PR #18, opened 2015-09-16
chore: add git hook for commit validation

Foxandxss:
All folders and filenames is just a proposal, I can rename it if needed.


wesleycho:
LGTM

PR #19, opened 2015-09-16
feat(modal): create initial implementation

wesleycho:
This is an initial implementation based off of https://github.com/angular/angular/blob/master/modules/angular2_material/src/components/dialog/dialog.ts (almost 1-1 port). This is still a work in progress, and needs tests too.

### DIFF
--- a/src/components/modal/modal.ts
+++ b/src/components/modal/modal.ts
@@ -1,0 +1,192 @@
+import {
+  bind,
+  forwardRef,
+  Component,
+  ComponentRef,
+  Directive,
+  DynamicComponentLoader,
+  ElementRef,
+  Host,
+  Injectable,
+  ResolvedBinding,
+  SkipSelf,
+  Injector,
+  View,
+  ViewEncapsulation
+} from 'angular2/angular2/core';
+
+import {ObservableWrapper, Promise, PromiseWrapper} from 'angular2/angular2/src/core/facade/async';
+import {isPresent, Type} from 'angular2/angular2/src/core/facade/lang';
+import {DOM} from 'angular2/angular2/src/core/dom/dom_adapter';
+import {MouseEvent, KeyboardEvent} from 'angular2/angular2/src/core/facade/browser';
+// TODO: pull key codes out from angular material or work with them to bring it into core?
+import {KeyCodes} from 'angular2/angular2_material/src/core/key_codes';
+
+export class UibModal {
+  constructor(private loader: DynamicComponentLoader) {}
+
+  open(type: Type, elementRef: ElementRef, options: UibModalConfig = null): Promise<UibModalRef> {
+    let config = isPresent(options) ? options : new UibModalConfig();
+    let modalRef = new UibModalRef();
+    let bindings = Injector.resolve([bind(UibModalRef).toValue(modalRef)]);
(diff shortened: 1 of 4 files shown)


src/components/modal/modal.ts
```diff
@@ -0,0 +1,196 @@
...
+  selector: 'Uib-backdrop',
+  host: {
+    '(click)': 'onClick()'
+  }
+})
+@View({template: '', encapsulation: ViewEncapsulation.None})
+class UibBackdrop {
+  modalRef: UibModalRef;
+
+  constructor(modalRef: UibModalRef, options: UibModalConfig) {
```
wesleycho: I need to investigate whether the adding of options here is valid.

src/components/modal/modal.ts
```diff
@@ -0,0 +1,196 @@
...
+  Injector,
+  View,
+  ViewEncapsulation
+} from 'angular2/core';
+
+import {ObservableWrapper, Promise, PromiseWrapper} from 'angular2/src/core/facade/async';
+import {isPresent, Type} from 'angular2/src/core/facade/lang';
+import {DOM} from 'angular2/src/core/dom/dom_adapter';
+import {MouseEvent, KeyboardEvent} from 'angular2/src/core/facade/browser';
+import {KeyCodes} from 'angular2_material/src/core/key_codes';
```
Foxandxss: All this lines, perhaps we need to change that to `angular2/angular2` or similar. Seems that the import "source" changes depending on what angular2 package the end user have in their project or something like that. Need to research on that.

src/components/modal/modal.ts
```diff
@@ -0,0 +1,196 @@
...
+
+@Injectable()
```
Foxandxss: That technically is not needed anymore.

src/components/modal/modal.ts
wesleycho: Oh right, that is obviously bunk - I need to fix that heh.

src/components/modal/modal.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  documentKeypress(event: KeyboardEvent) {
+    if (event.keyCode === KeyCodes.ESCAPE) {
+      this.modalRef.close();
+    }
+  }
+}
+
+@Directive({
+  selector: 'Uib-modal-content'
```
Foxandxss: bad uppercase.

src/components/modal/modal.ts
```diff
@@ -0,0 +1,196 @@
...
+import {DOM} from 'angular2/src/core/dom/dom_adapter';
+import {MouseEvent, KeyboardEvent} from 'angular2/src/core/facade/browser';
+import {KeyCodes} from 'angular2_material/src/core/key_codes';
+
+@Injectable()
+export class UibModal {
+  componentLoader: DynamicComponentLoader;
+
+  constructor(loader: DynamicComponentLoader) {
+    this.componentLoader = loader;
```
Foxandxss: Common convention I am seeing (but subjective I guess) is to put `private` on the constructor so we don't need line 29 and 26.

wesleycho:
Closing as I want to re-think this completely - I want to make the modal observable-based, which will allow for possibilities such as canceling modal opening during the animation.

Issue #4, opened 2015-09-15
Dropdown


BobbieBarker:
recreate BS3 drop down with ng2 component.

PR #23, opened 2015-09-26
feat(dropdown): create working dropdown

wesleycho:
- Create working dropdown component

Working Plunker [here](http://plnkr.co/edit/blhYNtxDtZxUFB4IQdsf?p=preview).

Implements #4

### DIFF
--- a/src/components/dropdown.ts
+++ b/src/components/dropdown.ts
@@ -1,0 +1,16 @@
+import {Component, View} from 'angular2/angular2';
+
+@Component({selector: 'ngb-dropdown', properties: ['open']})
+@View({
+  template: `<div class="dropdown" [class.open]="open">
+    <ng-content></ng-content>
+  </div>`
+})
+export class NgbDropdown {
+  private open: boolean;
+}
+
+@Component({selector: 'ngb-dropdown-menu', host: {class: 'dropdown-menu'}})
+@View({template: `<ng-content></ng-content>`})
+export class NgbDropdownMenu {
+}
--- a/src/core.ts
+++ b/src/core.ts
@@ -1,0 +1,1 @@
+export {NgbDropdown, NgbDropdownMenu} from './components/dropdown';
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,6 +6,7 @@
   },
   "files": [
     "typings/tsd.d.ts",
-    "src/core.ts"
+    "src/core.ts",
+    "src/components/dropdown.ts"
   ]
 }


src/components/dropdown.ts
```diff
@@ -0,0 +1,29 @@
...
+  </div>`
+})
+export class UibDropdown {
+  private open: boolean;
+}
+
+@Component({
+  selector: 'uib-dropdown-menu',
+  host: {
+    class: 'dropdown-menu'
```
Foxandxss: TS cries with this for me on the accordion. We will see when I merge the tooling.

src/components/dropdown.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'uib-dropdown',
+  properties: [
+    'open'
+  ]
+})
+@View({
+  directives: [NgClass],
+  template: `<div class="dropdown" [ng-class]="{open: open}">
```
Foxandxss: Did you try `host` in here? I think you had but you had issues.

src/components/dropdown.ts
```diff
@@ -0,0 +1,29 @@
+/// <reference path="../../typings/tsd.d.ts" />
```
Foxandxss: No bueno.

src/components/dropdown.ts
wesleycho: Pawel had a suggestion in the Angular repo when I mentioned not being able to use directives - I need to try
```js
properties: [
  '[class.open]: open'
]
```

wesleycho: Err, I had that suggestion wrong - [here](http://plnkr.co/edit/duj7JGoglJfX4b5oHK6g?p=preview) is a working implementation with changes though. This seems to have CSS ramifications though.

src/components/dropdown.ts
wesleycho: Seems to work fine

src/components/dropdown.ts
pkozlowski-opensource: I would definitively try `<div [class.open]="open">` here

src/components/dropdown.ts
```diff
@@ -0,0 +1,27 @@
+import {Component, View} from 'angular2/angular2';
+
+@Component({
+  selector: 'uib-dropdown',
```
pkozlowski-opensource: Did we decide in the prefix already? I was under the impression that we want to cut ties to angular-ui, no?

src/components/dropdown.ts
wesleycho: Already updated with that - see the current diff

src/components/dropdown.ts
Foxandxss: Open to ideas, but not any ng- related.

wesleycho: @Foxandxss was against the `ng-` suggestion - we should probably hash this out in the meeting shortly. I would prefer `ng-bs-` myself.

src/components/dropdown.ts
```diff
@@ -0,0 +1,27 @@
+import {Component, View} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-dropdown',
+  properties: [
+    'open'
+  ]
+})
+@View({
+  template: `<div class="dropdown" [class.open]="open">
```
Foxandxss: nitpicking, isn't better to put the template itself in a new line? @robwormald

Foxandxss: Like you said to rob the other day, it is easier for the eye :P

Foxandxss:
I give you the honors to push the very first angular 2 bootstrap.

PR #27, opened 2015-09-26
chore: add tooling

Foxandxss:
The never coming PR.

Basic workflow based on simple Gulp. It is able to generate ES6, ES5 and a final UMD build (thanks to webpack).

I deleted all the `reference` entries on the files. They are not needed since I feed it directly to `tsc`.

I can see the cases where an editor requires that `reference` to make proper highlight. I guess there should be tools for not requiring that. Webstorm directly reads from typings to offer proper highlighting.

If you run it now, stackedMap won't be transformed, it is broken, will fix it.

On the other hand, the ES6 gives a bunch of errors, nothing of our business and it should be fixed in .38.

I will continue working on the tooling time to time, to organize it better, move strings out and give it more features towards a future release, but for now it works perfectly to generate an UMD and use it directly on a plunker.

Ah, you can check a demo plunker with the end result:  http://plnkr.co/edit/UvD9er6I92aD3UfZqb5W?p=preview

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,0 +1,50 @@
+var gulp = require('gulp');
+var gutil = require('gulp-util');
+var ts = require('gulp-typescript');
+var del = require('del');
+var webpack = require('webpack');
+
+gulp.task('clean', function (cb) {
+  del('dist/', cb);
+});
+
+gulp.task('es5', function () {
+  var tsProject = ts.createProject('tsconfig.json', {
+    module: 'commonjs',
+    target: 'es5'
+  });
+
+  var tsResult = gulp.src(['typings/tsd.d.ts', 'src/**/*.ts'])
+    .pipe(ts(tsProject));
+
+  return tsResult.js.pipe(gulp.dest('dist/cjs'));
+});
+
+gulp.task('es6', function () {
+  var tsProject = ts.createProject('tsconfig.json');
+
+  var tsResult = gulp.src(['typings/tsd.d.ts', 'src/**/*.ts'])
+    .pipe(ts(tsProject));
+
+  return tsResult.js.pipe(gulp.dest('dist/es6'));
+});
+
+gulp.task('umd', ['es5', 'es6'], function (cb) {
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -0,0 +1,51 @@
+var gulp = require('gulp');
+var gutil = require('gulp-util');
+var ts = require('gulp-typescript');
+var del = require('del');
+var webpack = require('webpack');
+
+gulp.task('clean', function(cb) {
+  del('dist/');
```
wesleycho: don't you need to pass the `cb` to the second argument of `del`?

gulpfile.js
```diff
@@ -0,0 +1,51 @@
...
+});
+
+gulp.task('ES5', function() {
```
wesleycho: This is nitpicking, but I think lowercase would be better for the task name, i.e. `es5`

gulpfile.js
```diff
@@ -0,0 +1,51 @@
...
+    output: {
+      filename: 'dist/ng-bootstrap.js',
+      libraryTarget: 'umd'
+    },
+    externals: {
+      "angular2/angular2": "angular2/angular2"
+    }
+  }, function(err, stats) {
+    if (err) throw new gutil.PluginError('webpack', err);
+    gutil.log("[webpack]", stats.toString());
```
wesleycho: `cb` should probably be called after

gulpfile.js
Foxandxss: Definitely.

gulpfile.js
Foxandxss: Don't mind.

gulpfile.js
Foxandxss: I took this directly from official docs:

https://webpack.github.io/docs/usage-with-gulp.html

wesleycho: Official docs have callback being called after :P

wesleycho:
LGTM

PR #28, opened 2015-09-27
chore: cleanup initial tooling

Foxandxss:
When we started this project we hadn't a good idea of how to build it, but I think we have now.

So I propose to delete all those extra things we have now, because they add no real value to the project at this stage (Sorry Rob).

Since we start fresh, I would like to have the project under control and not having extra folders / configuration that does nothing for the time being.

So, no JSPM (we don't need that for a library), no examples folder (we haven't decided what to do with the demos, but is not going to be that at least). The tsconfig is not needed if we merge my new tooling.

So if you don't mind, I prefer to start as clean as possible and I promise that if any of this is needed in 1 month, I will code that again, but for now, it is adds no value.

There is still more tidy up to do, but I will create more PRs for that.


Foxandxss:
Some editors like Webstorm seems to also rely on `tsconfig.js` for highlight (when it should only worry about typings) so I am not sure if it is crying for me (in another unrelated project) because I am importing a library without `d.ts` or not.

If all editors have a different way of working with TS, we are going to have a problem :P

.gitignore
```diff
@@ -1,7 +1,3 @@
 /node_modules/
 /typings/
-/jspm_packages/
-/dist/
-npm-debug.log
-.sw*
-.idea
```
wesleycho: I think we should keep all these things in gitignore, minus maybe the `jspm_packages` one - `.sw*` is for Vim swap files, `.idea` is for IntelliJ, and `npm-debug.log` is for npm erroring out, all valid things to ignore.

tsconfig.json
```diff
@@ -1,18 +0,0 @@
-{
```
wesleycho: Can this file be safely removed? This is for the raw TS compiler I thought?

.gitignore
Foxandxss: I don't really mind, but you should create a global gitignore in your computer. If all of us uses 20 different editors / ide, we would need to add them all.

You can go to ~/.gitconfig and do:

```
[core]
	excludesfile = /Users/youruser/.gitignore_global
```

I have something like:

```
.idea
.DS_Store
npm-debug.log
playground
```

So that will get ignored to every project of mine.

tsconfig.json
Foxandxss: If we merge my tooling PR, this file wouldn't be needed. **BUT** it could be needed by some editor. I vote to delete it and then research if our editors needs it, if so, how and create a project one for everyone of have your own copy locally. Not sure yet.

.gitignore
wesleycho: Thing is, when we get more contributors, they will likely not have it stored in a global gitignore and will likely add these files.

tsconfig.json
wesleycho: Fine by me

.gitignore
Foxandxss: Okey, I think it is better to teach the users to have a clean environment but after all, not a big deal :)

wesleycho:
LGTM

Foxandxss:
I removed some reference things.

@robwormald Could this tsconfig.json be enough for Atom? I tested it and I get errors, highlight & shit with simply that.

wesleycho:
Some of the errors are on Angular 2's side from what I remember.

Foxandxss:
Oh yeah, I am mostly interested to have the tooling working perfectly for non webstorm users. Error fixing comes next.

PR #29, opened 2015-09-27
chore(stackedMap): export interface

Foxandxss:
If I am not mistaken, if you export something like a `class` (StackedMap in this case) and you use an interface that is not exported, typescript will cry. To fix that, you need to export the interface as well.


wesleycho:
LGTM - only question I have is whether we might want to think of renaming `keyVal` into something less generic, but we can always revisit this question later.

Foxandxss:
Definitely. We still need to use them properly, so we can do what we want without BCs yet :P

PR #30, opened 2015-09-29
add clang format

robwormald:
left the comments in about travis (just a straight copy pasta from ng2 docs)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,6 +3,8 @@ var gutil = require('gulp-util');
 var ts = require('gulp-typescript');
 var del = require('del');
 var webpack = require('webpack');
+var gulpFormat = require('gulp-clang-format');
+var clangFormat = require('clang-format');
 
 gulp.task('clean', function (cb) {
   del('dist/', cb);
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -46,5 +48,25 @@ gulp.task('umd', ['es5', 'es6'], function (cb) {
...
+
+gulp.task('enforce-format', function() {
+  return doCheckFormat().on('warning', function(e) {
+    console.log("ERROR: You forgot to run clang-format on your change.");
+    console.log("See https://github.com/angular/angular/blob/master/DEVELOPER.md#formatting");
+    process.exit(1);
+  });
+});
+
+gulp.task('build', ['clean', 'umd', 'check-format']);
```
Foxandxss: should this check before doing shit?

robwormald: the `check-format` task will currently *warn* you but not fail the build (and give you a hint on how to do the fix. the `enforce-format` task should be set up on CI when its ready so builds will fail if they are unformatted. note a couple of the files are already warning :)

robwormald: so to answer your question, i dont think the order matters much.

Foxandxss:
I merged it. Changed the url you point out in the `enforce-format` (It was even wrong in the angular repo, sent a PR).

Also moved the function that does the format to the bottom.

PR #31, opened 2015-09-29
chore: format files with clang

Foxandxss:
Alright, based on what @robwormald did with clang, I formatted our filed.

Not that I am too fond of the style, but I guess it is for the best.

### DIFF
--- a/src/services/multiMap.ts
+++ b/src/services/multiMap.ts
@@ -1,30 +1,15 @@
 export class UibMultiMap {
   _map: Object;
 
-  constructor() {
-    this._map = {};
-  }
+  constructor() { this._map = {}; }
 
-  entries() {
-    return Object.keys(this._map).map(key => {
-      return {
-        key: key,
-        value: this._map[key]
-      }
-    });
-  }
+  entries() { return Object.keys(this._map).map(key => {return {key: key, value: this._map[key]}}); }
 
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
Only slightly related to this PR but could we have 120 columns instead of 100? It kills me how the ng2 code is formatted / wrapped....

And yeh, +1 for landing this ASAP as it is pretty much impossible to change formatting as soon as we start to have PRs

Foxandxss:
SIGN ME IN!

No, seriously, I think that we the screens we have nowadays, 100 is too meh.

Foxandxss:
I applied it here.

src/services/multiMap.ts
```diff
@@ -1,30 +1,15 @@
...
 
-  entries() {
-    return Object.keys(this._map).map(key => {
-      return {
-        key: key,
-        value: this._map[key]
-      }
-    });
-  }
+  entries() { return Object.keys(this._map).map(key => {return {key: key, value: this._map[key]}}); }
```
wesleycho: It really makes this look a little harder to read :( .

Foxandxss: I agree.

wesleycho:
LGTM - I can live with my little gripe

PR #32, opened 2015-09-29
chore: clean tsconfig

Foxandxss:
Those things doesn't add value.

No need of exclude because that gets override by files and I am not even using that for compiling, just for IDE / Editors sake.

Also remove that extra d.ts, that will be fed to gulp directly.


wesleycho:
:+1:

PR #35, opened 2015-09-30
chore: tune tsconfig to prevent compilation on save

Foxandxss:
This is only a fix for Atom, but since Wes uses Atom, it is best to have it.

For the rest, this prevents `atom-typescript` to compile a `.ts` every time you save a file, you don't need that for this project.


wesleycho:
LGTM

PR #40, opened 2015-09-30
feat(accordion): create working accordion

Foxandxss:
There is a few things to notice, but I will add comments on the code directly.

Back in uibs@BS3 we used to toggle by clicking on the entire header, in BS4 it only toggles if you click on the text and not outside of it. Makes sense for me because the new accordion style (way uglier) doesn't make clear where is the separation between header and content.

[Plunker](http://plnkr.co/edit/XuH7mXBycODO7KYgNSbH?p=preview)

### DIFF
--- a/src/components/accordion.ts
+++ b/src/components/accordion.ts
@@ -1,0 +1,71 @@
+import {Component, View, OnDestroy} from 'angular2/angular2';
+
+@Component({selector: 'ngb-accordion, [ngb-accordion]', properties: ['onlyOneOpen: closeOthers']})
+@View({template: `<ng-content></ng-content>`})
+export class NgbAccordion {
+  private onlyOneOpen: boolean;
+  private groups: Array<NgbAccordionGroup> = [];
+
+  addGroup(group: NgbAccordionGroup): void { this.groups.push(group); }
+
+  closeOthers(openGroup): void {
+    if (!this.onlyOneOpen) {
+      return;
+    }
+
+    this.groups.forEach((group: NgbAccordionGroup) => {
+      if (group !== openGroup) {
+        group.isOpen = false;
+      }
+    });
+  }
+
+  removeGroup(group: NgbAccordionGroup): void {
+    const index = this.groups.indexOf(group);
+    if (index !== -1) {
+      this.groups.splice(index, 1);
+    }
+  }
+}
+
+@Component({selector: 'ngb-accordion-group, [ngb-accordion-group]', properties: ['heading', 'isOpen', 'isDisabled']})
+@View({
+  template: `
+    <div class="panel panel-default" [class.panel-open]="isOpen">
+      <div class="panel-heading">
+        <h4 class="panel-title">
+          <a href tabindex="0"><span [class.text-muted]="isDisabled" (click)="toggleOpen($event)">{{heading}}</span></a>
+        </h4>
+      </div>
+      <div class="panel-collapse" [hidden]="!isOpen">
+        <div class="panel-body">
+    	    <ng-content></ng-content>
+  	    </div>
+      </div>
+    </div>
+  `
+})
+export class NgbAccordionGroup implements OnDestroy {
+  private isDisabled: boolean;
+  private _isOpen: boolean = false;
+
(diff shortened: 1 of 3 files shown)


src/components/accordion.ts
```diff
@@ -0,0 +1,78 @@
+import {Component, View, NgClass, OnDestroy} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-accordion, [ngb-accordion]',
+  properties: ['onlyOneOpen: closeOthers'],
+  host: {'class': 'panel-group'}
```
Foxandxss: This makes the compiler cry, both here and in the dropdown.

I think that the syntax is correct in here, by having both key and value inside quotes. The dropdown is not using quotes for `class` and I think it should. Anyhow the `d.ts` seems to be broken. that or the doc in it is deprecated.

I say to ignore this, wait for .38 and see if any of the two gets fixed and change the other.

src/components/accordion.ts
```diff
@@ -0,0 +1,78 @@
...
+    const index = this.groups.indexOf(group);
+    if (index !== -1) {
+      this.groups.splice(index, 1);
+    }
+  }
+}
+
+@Component({selector: 'ngb-accordion-group, [ngb-accordion-group]', properties: ['heading', 'isOpen', 'isDisabled']})
+@View({
+  template: `
```
Foxandxss: In difference with the dropdown, I prefer to have multilines templates starting on a new line, easier to read. Thoughts?

src/components/accordion.ts
```diff
@@ -0,0 +1,78 @@
...
+    <div class="panel panel-default" [ng-class]="{'panel-open': isOpen}">
+      <div class="panel-heading">
+        <h4 class="panel-title">
+          <a href tabindex="0">
+            <span [ng-class]="{'text-muted': isDisabled}" (click)="toggleOpen($event)">{{heading}}</span>
```
pkozlowski-opensource: Could we drop the `ngClass` directive from here and use `[class.text-muted]="isDisabled"`?

Foxandxss: Probably, let me play with that.

pkozlowski-opensource:
So, when I was thinking about an accordion I was wondering it would be possible to do it as a collection of panel components where a panel can be used dependently of an accordion.

Foxandxss:
I updated the ng-class thing (also added a plunker, forgot).

About your suggestion, not sure what is the value. I remember our talk yesterday about having more reusable components, but I am not so sure what could we extract from here.

src/components/accordion.ts
wesleycho: I think it's actually unnecessary to even have that `panel-group` class in BS4 - it also is not useful in BS3 unless the element is a block element. Might be worth considering putting a wrapper div in the template with that class on it instead.

src/components/accordion.ts
wesleycho: Fine by me - probably is a little better when it comes to indentation.

src/components/accordion.ts
Foxandxss: I am seeing that too, perhaps I can remove it, no wrapper needed. That doesn't resolve it in the dropdown tho.

wesleycho:
Can you update the Plunker with the accessibility support as shown in the BS4 example [here](http://v4-alpha.getbootstrap.com/components/collapse/#accordion-example) - this should make it easier when it comes to documentation/demo site writing.

Foxandxss:
So I got this @pkozlowski-opensource http://plnkr.co/edit/eaGaEpds72KQlL0YrgRO?p=preview

That is with cards, but I am not sure we want a card component to be honest. Card has no functionality it is just *html*, so I don't see it.

wesleycho:
I took a stab at trying to implement cards as a component - I think if we go that route, it would end up opinionated due to the flexibility that Bootstrap provides in them: http://v4-alpha.getbootstrap.com/components/card/

Foxandxss:
I still think that our sole purpose with this library (and like we did on uibs) is to port functionality and not markup. That is why we didn't have a navbar component.

wesleycho:
This LGTM

Issue #14, opened 2015-09-15
Buttons


BobbieBarker:
Evaluate the need for an ng2 component for BS3 buttons.

pkozlowski-opensource:
Just to set a baseline, here are check-box buttons in vanilla ng2: http://plnkr.co/edit/pey0BwIo8dkKbViA02R0?p=preview 

Whatever we do it should be simpler / shorter to write as compared to this baseline.

pkozlowski-opensource:
Moving discussion over from #42 there were 2 proposal in plunkers:

* http://plnkr.co/edit/ZcNiSJR6rFfMAvXX9Q9X?p=preview from @wesleycho 
* http://plnkr.co/edit/5V4ZP8qThnM0xFyIIXFt?p=preview from @Foxandxss 

IMO both are going in a good directions, we just need to push this idea further. My biggest open question for now is the relation of those directives to the `ngModel` / `ngControl` - that is - do we want to plug into those to provide things like validation etc.

pkozlowski-opensource:
So I would like to go back to the topic of buttons and after experimenting a bit I'm not sure we want to have a directive for check-boxes. I'm saying this since the vanilla version looks like so:

```html
<div class="btn-group" data-toggle="buttons">
    <label class="btn btn-primary" [class.active]="foo">
         <input type="checkbox" [(ngModel)]="foo"> Checkbox 2
     </label>
</div>
``` 

while with a directive it would be:

```html
<div class="btn-group" data-toggle="buttons">
    <label class="btn btn-primary">
         <input type="checkbox" [(ngModel)]="foo" ngbButtonCheckbox> Checkbox 2
     </label>
</div>
``` 

so I'm not sure the directive pays for itself here.... Thoughts?

wesleycho:
I'm sort of in agreement - I feel like it's too simple.

icfantv:
agreed.  we should put something on our site though - or someone will say _something_.  maybe even just the template above.

pkozlowski-opensource:
Yes, agree with @icfantv - if we decide to skip a directive for a given piece of functionality because we think it is "not worth it" we should still show this on a demo page to say: "hey, look, this is actually easy with vanilla ng2, no need for a custom directives here"

Foxandxss:
And get endless issues because they don't manage to make the checkboxs to work.

maxokorokov:
I've been experimenting with radio buttons and for now I came up with this:

```html
[...]

PR #42, opened 2015-09-30
feat(buttons): create button directives

wesleycho:
- Create `NgbButtonCheckbox` and `NgbButtonRadio` directives

Working Plunker [here](http://plnkr.co/edit/xAOz2VWFvkG0dr2QVS5d?p=preview)

This is not quite there yet - the template needs to be applied for accessibility reasons, but unfortunately using a `Component` decorator causes click eventing to get thrown off.

This addresses #14.

### DIFF
--- a/src/components/buttons.ts
+++ b/src/components/buttons.ts
@@ -1,0 +1,38 @@
+import {Directive, View} from 'angular2/angular2';
+
+@Directive({
+  selector: '[ngb-button-checkbox]',
+  properties: [
+    'name: ngb-name', 'checked: ngb-checked', 'class: ngb-class'
+  ],
+  host: {
+    '[class]': 'class',
+    '[class.active]': 'checked'
+  }
+})
+@View({
+  template: `
+    <input type="checkbox" [checked]="checked" [name]="name" autocomplete="off">
+    <ng-content></ng-content>`
+})
+export class NgbButtonCheckbox {
+  private checked: boolean;
+}
+
+@Directive({
+  selector: '[ngb-button-radio]',
+  properties: [
+    'name: ngb-name', 'checked: ngb-checked', 'class: ngb-class'
+  ],
+  host: {
+    '[class]': 'class',
+    '[class.active]': 'checked'
+  }
+})
+@View({
+  template: `<input type="radio" [checked]="checked" [name]="name" autocomplete="off">
+    <ng-content></ng-content>`
+})
+export class NgbButtonRadio {
+  private checked: boolean;
+}
--- a/src/core.ts
+++ b/src/core.ts
@@ -1,1 +1,2 @@
+export {NgBButtonCheckbox, NgbButtonRadio} from './components/buttons';
 export {NgbDropdown, NgbDropdownMenu} from './components/dropdown';
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
   "files": [
     "typings/tsd.d.ts",
     "src/core.ts",
+    "src/components/buttons.ts",
     "src/components/dropdown.ts"
   ],
   "compileOnSave": false


Foxandxss:
The problem I see here is that this directive is completely different than the one on uibs and I prefer the style on the one in uibs.

There you can get any label / button with `ng-model` and stick the directive as an attribute.

Here we are forced to create a label and put lots of attributes on it.

In my vision, we should create a directive (not component) and not having a template. We don't need a template, we just need the element where we apply the directive to. The only real communication between both worlds (outside and directive) is the value you pass.

Now, to achieve that we need some kind of two-way databinding and that seems discouraged in ng2 community. I am not sure how can we achieve that. Perhaps ng-model.

pkozlowski-opensource:
Agree with @Foxandxss that I would expect this to be an attribute directive rather than sth that has a template. Actually the same goes for drop-down. I would expect those things to be `@Directive` and not have a template

Foxandxss:
This buttons directive needs to be redone, perhaps with the ideas I have on the ratings component (PR coming soon, tomorrow I expect).

The dropdown ideas, not sure if I follow. We could open an issue for that and check it out.

pkozlowski-opensource:
@Foxandxss hmm, I can't wait for your PR now, since I wouldn't expect rating and buttons to have anything in common :-)

I would expect that rating is quite similar to pagination (ex.: https://github.com/pkozlowski-opensource/ng2-bs/tree/master/src/pagination) so you got me curious :-)

Foxandxss:
The common thing is that both have inputs and outputs. In buttons, I give it a model and based on some radiobuttons I get the one checked as an output.

Rating you give it one default value (if any) and you get the selected rate output.

That is what I meant. Proper input/output management.

Something like....

```javascript
<label class="btn btn-primary" [(model)]="radioModel" ngb-btn-radio="'Left'">Left</label>
<label class="btn btn-primary" [(model)]="radioModel" ngb-btn-radio="'Middle'">Middle</label>
<label class="btn btn-primary" [(model)]="radioModel" ngb-btn-radio="'Right'">Right</label>
```

pkozlowski-opensource:
Oh, I see, it is about input / output. 

BTW, shouldn't we rather model the API so it looks like:

```
<div class="btn-group" role="group" aria-label="Basic example" [(radio-model)]="radioModel">
  <button type="button" class="btn btn-secondary" ngb-btn-radio="'Left'">Left</button>
  <button type="button" class="btn btn-secondary" ngb-btn-radio="'Middle'">Middle</button>
  <button type="button" class="btn btn-secondary" ngb-btn-radio="'Right'">Right</button>
</div>
```
Just thinking out loud here, didn't dig deep enough into it

Foxandxss:
No idea, but sounds good to me. I think that the important bit in here is more about getting the good visual result (like bootstrap does).

pkozlowski-opensource:
Right. Anyway, @wesleycho it seems like we explore more the possibility of having this as an attribute directive rather than a component

wesleycho:
That sounds good to me - I'll revisit this, been busy with work this week

wesleycho:
I created a new Plunker [here](http://plnkr.co/edit/ZcNiSJR6rFfMAvXX9Q9X?p=preview), but I'm not sure how much value there is in these directives - thoughts?

Foxandxss:
What do you think of [this](http://plnkr.co/edit/5V4ZP8qThnM0xFyIIXFt?p=preview)?

Haven't implemented the radio one yet because I have to rely on nativeElement and seems that that is not a good idea (that is why I used an extra internal state for checkbox).

pkozlowski-opensource:
I'm going to close this PR as based on the comments we want to go more in the attribute-based directives direction. Let's re-start the discussion in #14

PR #45, opened 2015-10-06
chore: update typings for angular2 .39

Foxandxss:
It doesn't fix the `d.ts` errors we have, but it is always better to work with latest typings and more because  it changes quite a bit since .37

### DIFF
--- a/tsd.json
+++ b/tsd.json
@@ -5,14 +5,11 @@
   "path": "typings",
   "bundle": "typings/tsd.d.ts",
   "installed": {
+    "es6-shim/es6-shim.d.ts": {
+      "commit": "a6732bf1c03fd285bb01641b3a5a0129147cdb5b"
+    },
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
I must say that the whole typing resolution is a mystery for me, but why can't we use typings distributed ing2 npm bundle?

Foxandxss:
@pkozlowski-opensource If I am not mistaken, that nice feature is not available yet. So for the time being we need to work with TSD.

wesleycho:
Fine by me - seems like this should be temporary anyhow.

wesleycho:
Although...this needs to be updated for .40

Foxandxss:
This is no relevant anymore since angular2 has the d.ts built in. Still that fails with ES6 and I don't know why.

Will keep researching.

PR #46, opened 2015-10-06
chore: fix umd for es5

Foxandxss:
This is what I meant on the talk.

Now our UMD build fails for plain ES5 users and this fixes it.

This tooling could potentially go away with other changes based on how we would need to do testing.

Still, for the time being, helps testing the library on a plunker for ES5 users.


wesleycho:
Looks fine to me

PR #47, opened 2015-10-09
feat(rating): add rating component

Foxandxss:
Stuff missing:

Custom titles.
Custom rating icons.

I will add that later. I prefer to work on other components so we can have more to show.

PS: Plunker http://plnkr.co/edit/2wtcZ5bmQk8xmvwWvFP6?p=preview

### DIFF
--- a/src/components/rating.ts
+++ b/src/components/rating.ts
@@ -1,0 +1,62 @@
+import {Component, View, NgFor, NgClass, EventEmitter, OnInit} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-rating',
+  inputs: ['max', 'rate', 'readonly'],
+  outputs: ['updateRating: rate', 'onHover', 'onLeave']
+})
+@View({
+  template: `
+    <span tabindex="0" (mouseleave)="reset()" aria-valuemin="0" [attr.aria-valuemax]="max" [attr.aria-valuenow]="rate">
+      <template ng-for #r [ng-for-of]="range" #index="index">
+        <span class="sr-only">({{ index < rate ? '*' : ' ' }})</span>
+        <i class="glyphicon" (mouseenter)="hover(index + 1)" (click)="update(index + 1)" [ng-class]="index < rate ? 'glyphicon-star' : 'glyphicon-star-empty'" [title]="r.title" [attr.aria-valuetext]="r.title"></i>
+      </template>
+    </span>
+  `,
+  directives: [NgFor, NgClass]
+})
+export class NgbRating implements OnInit {
+  private max: number = 10;
+  private oldRate: number;
+  private range: Array<any>;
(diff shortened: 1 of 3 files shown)


src/components/rating.ts
```diff
@@ -0,0 +1,62 @@
...
+  selector: 'ngb-rating',
+  inputs: ['max', 'rate', 'readonly'],
+  outputs: ['updateRating: rate', 'onHover', 'onLeave']
+})
+@View({
+  template: `
+    <span tabindex="0" (mouseleave)="reset()" aria-valuemin="0" [attr.aria-valuemax]="max" [attr.aria-valuenow]="rate">
+      <template ng-for #r [ng-for-of]="range" #index="index">
+        <span class="sr-only">({{ index < rate ? '*' : ' ' }})</span>
+        <i class="glyphicon" (mouseenter)="hover(index + 1)" (click)="update(index + 1)" [ng-class]="index < rate ? 'glyphicon-star' : 'glyphicon-star-empty'" [title]="r.title" [attr.aria-valuetext]="r.title"></i>
```
wesleycho: I would add the ability to bind to the `tabindex` here since there are no anchor tags.

wesleycho:
For the most part I'm fine with this, although I think it might be worth considering keyboard support, i.e. allowing the enter keypress to fire the update callback for accessibility reasons

Foxandxss:
I hear you Wes. I think that it is better for now to start rolling out directives and not having the perfect PR that does everything and move on.

Some kind of list of "What should this (merged) component still need to do."

pkozlowski-opensource:
I agree with @Foxandxss - let's get few components in first and let's keep them as simple as possible. We still don't have test infrastructure, so adding more functionality at this point is playing with fire :-)

@Foxandxss the directive LGTM. If I were you I would probably keep it even simpler at first and drop support for mouse hovering. But it is your call.

Foxandxss:
@pkozlowski-opensource Well, it is there yet, I don't think we should remove that feature and then add it again, right?

webteckie:
If you would follow a strict TDD approach then that's what would need to happen...remove/comment the code until tests exist!  But some of us cheat on that ( including me) and do TDD after☺  As long as a task is kept open to revisit adding tests it should be ok to leave it there for now, IMO.

pkozlowski-opensource:
@webteckie is it not about cheating, it is more that there is no good / easy way to setup testing infrastructure for ng2 libraries. We are working with the core team on this and yes, tests will be mandatory part of this library story.

@Foxandxss as I was mentioning, it is your call.

Foxandxss:
Yes, the problem is that the testing infrastructure is not there yet. On a
normal development, you would do testing, but this is an especial case.

We should move carefully and test everything ASAP.

PR #51, opened 2015-10-15
chore(dropdown): update to .42

Foxandxss:
I want to revisit this directive later, see if it can be done in a different way, but for the time being, I am going to update it to the latest version and recommendations.

[Plunker](http://plnkr.co/edit/8FNKrEURwSzqqtsK5Rxn?p=preview)

### DIFF
--- a/src/components/dropdown.ts
+++ b/src/components/dropdown.ts
@@ -1,16 +1,15 @@
-import {Component, View} from 'angular2/angular2';
+import {Component, Input} from 'angular2/angular2';
 
-@Component({selector: 'ngb-dropdown', properties: ['open']})
-@View({
+@Component({
+  selector: 'ngb-dropdown',
   template: `<div class="dropdown" [class.open]="open">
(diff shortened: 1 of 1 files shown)


src/components/dropdown.ts
```diff
@@ -1,16 +1,15 @@
...
-@View({
+@Component({
+  selector: 'ngb-dropdown',
   template: `<div class="dropdown" [class.open]="open">
     <ng-content></ng-content>
   </div>`
 })
 export class NgbDropdown {
-  private open: boolean;
+  @Input() private open: boolean;
```
wesleycho: Is this a new decorator? Where should I look this up, the source code?

Foxandxss: Yeah. Took me a while until I saw where it is defined in the source.

https://github.com/angular/angular/blob/master/modules/angular2/src/core/metadata/directives.ts#L977-L1025

wesleycho:
LGTM

PR #54, opened 2015-10-19
chore: simplify tooling

Foxandxss:
Now that angular2 bundles the d.ts within it, I removed the TSD support.

Algo @robwormald pointed out that it is better just to server CJS so I removed the ES6 part (which was giving me issues) but I am leaving the UMD for plunker purposes.

I moved to use `tsc` instead of `gulp-typescript` because for some reason, the gulp one didn't want to generate declaration files and well, it is overkill as well.


Foxandxss:
This will also help to develop components better, because we will be able to use the latest definition types in our projects.

wesleycho:
LGTM - keeps it simpler for now.

PR #56, opened 2015-10-24
chore: testing tooling

Foxandxss:
Alrighty, let me explain how this work.

Since testing as today is only possible using systemjs, I needed to extend the tooling to support that.

We needed two tsconfig, but we could only put one. Since that for editor supports we will need to list the files on the tsconfig, I put the testing tsconfig as the main one and "embed" the build one in the package.json.

So now, we have a `test` folder inside `src` where we can put our tests. (the convention is `foo_test.ts`).

As with source files, we need to add test files to `tsconfig.json`.

To make this work, you can run `npm run watch` in one tab and in another tab `npm test`.

There is no way yet to run the tests once nor generate a build after tests are green, but we don't care about atm. The important bit is that we can write tests now.

If you think that I should change some convention, feel free (but this can change in the future too).

### DIFF
--- a/karma-test-shim.js
+++ b/karma-test-shim.js
@@ -1,0 +1,64 @@
+// Tun on full stack traces in errors to help debugging
+Error.stackTraceLimit=Infinity;
+
+
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000;
+
+// // Cancel Karma's synchronous start,
+// // we will call `__karma__.start()` later, once all the specs are loaded.
+__karma__.loaded = function() {};
+
+
+System.config({
+  packages: {
+    'base/temp/components': {
+      defaultExtension: false,
+      format: 'register',
+      map: Object.keys(window.__karma__.files).
+            filter(onlyAppFiles).
+            reduce(function createPathRecords(pathsMapping, appPath) {
+              // creates local module name mapping to global path with karma's fingerprint in path, e.g.:
+              // './hero.service': '/base/temp/components/hero.service.js?f4523daf879cfb7310ef6242682ccf10b2041b3e'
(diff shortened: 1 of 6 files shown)


Foxandxss:
Btw, tested this with some real angular2 testing and works like charm. This is an adaptation of Julie Ralph work.

Foxandxss:
This generates a temp folder with the app & test inside, just for karma sake and not to dirt our work folders.

karma-test-shim.js
```diff
@@ -0,0 +1,64 @@
+// Tun on full stack traces in errors to help debugging
```
wesleycho: Does this shim come from anywhere in particular?

wesleycho:
Doesn't Angular 2 build in some tooling on top of jasmine? Is there a particular reason we're not using it?

karma-test-shim.js
Foxandxss: The thing is. Angular 2 is really picky with how tests run. You basically need to tell karma to stop. Run some "browser adapters" on angular 2 and then run the tests. That is how Angular does it on their repo.

https://github.com/juliemr/ng2-test-seed/blob/master/karma-test-shim.js

I modified it a bit with our needs. But basically what angular does in:

https://github.com/angular/angular/blob/master/test-main.js

Foxandxss:
Yes it does, it is all under `angular/testing` which is is built-in in the main package as today. Said in other words, it is already there :)

Foxandxss:
I removed the system.js generation as @robwormald suggestion. Still there is some dirt atm that I will fix along the way, but we need to start writing tests ASAP.

wesleycho:
This LGTM - nice work setting this up!

PR #57, opened 2015-10-27
chore(rating): add a first batch of tests

Foxandxss:
Just a couple of tests, I will progressively add more.

Feel free to ask if you want to learn more about testing.

Also, removed sanity tests, not needed anymore.

This won't work yet because I have something to fix on the tooling. I will research that now with @robwormald

### DIFF
--- a/karma-test-shim.js
+++ b/karma-test-shim.js
@@ -13,14 +13,14 @@ System.config({
   packages: {
     'base/temp/components': {
       defaultExtension: false,
-      format: 'register',
+      format: 'cjs',
       map: Object.keys(window.__karma__.files).
             filter(onlyAppFiles).
             reduce(function createPathRecords(pathsMapping, appPath) {
               // creates local module name mapping to global path with karma's fingerprint in path, e.g.:
               // './hero.service': '/base/temp/components/hero.service.js?f4523daf879cfb7310ef6242682ccf10b2041b3e'
               var moduleName = appPath.replace(/^\/base\/temp\/components\//, './').replace(/\.js$/, '');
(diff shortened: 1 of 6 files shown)


Foxandxss:
Fixed that issue as well.

Foxandxss:
Wait, broken.

Foxandxss:
I think we should merge #58 first.

src/test/rating_test.ts
```diff
@@ -0,0 +1,78 @@
...
+         expect(stars.length).toBe(10);
+       });
+     }));
+
+  it('should change the num of stars with `max`', injectAsync([TestComponentBuilder], (tcb) => {
+       const html = '<ngb-rating max="3"></ngb-rating>';
+
+       return tcb.overrideTemplate(TestComponent, html)
+           .createAsync(TestComponent)
+           .then((fixture) => {
```
wesleycho: How do you feel about `(fixture) =>` vs. `fixture =>`? This is a minor comment, all things considered.

wesleycho:
Should we namespace the tests in the filenames, i.e. `rating.spec.ts` or `rating.test.ts` instead of `rating_test.ts`?

src/test/rating_test.ts
Foxandxss: Personally I like the idea, but should we do that or stick to angular convention?

wesleycho: Is angular convention to always wrap it in parens?

Foxandxss: Yeah, they are doing that.

wesleycho: I'm fine with leaving that alone then - more of a code style comment than anything else.

wesleycho:
Awesome, LGTM

PR #58, opened 2015-10-27
chore: add travis support

Foxandxss:
I have no idea if this travis config would work, but I think it should.

I also configured travis already on their end.


wesleycho:
Don't forget to add the nice passing/failing icon in the README!

wesleycho:
Need to put `sudo: false` in the yml

PR #59, opened 2015-10-29
chore: update angular2 to 2.0.0-alpha.45

pkozlowski-opensource:


Foxandxss:
Nice. Saw the Changelog and I don't think we need to update our directives.
Will merge this when I get home

PR #62, opened 2015-10-30
chore: don't list individual files in tsconfig.json

pkozlowski-opensource:
We don't want to list each and individual file in tsconfig.json

`"moduleResolution": "node"` instructs TS to look for typings in node_modules folders so it can easily find angular2 typings.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,13 +7,6 @@
     "experimentalDecorators": true,
     "outDir": "temp/"
   },
-  "files": [
-    "src/core.ts",
-    "src/components/accordion.ts",
-    "src/components/dropdown.ts",
-    "src/components/rating.ts",
-    "src/test/rating.test.ts"
-  ],
   "compileOnSave": false,
   "exclude": [
     "node_modules"


Foxandxss:
We need the files array for IDE support like atom (with atom-typescript) and VSC (if I am not mistaken with that one).

About the moduleResolution, it is working perfectly and it finds the typings (at least for me).

What are the problems you're seeing?

pkozlowski-opensource:
I'm on the latest WebStorm and it doesn't find typings without `"moduleResolution": "node"`.

What is the TS version used by atom-typescript on your side? You need TS 1.6+ for this to work.
I'm also pretty positive that the proposed setup would work in VSC, need to try with the latest version, but this is a setup we want to push for and ng2 / TS team.

In short: listing all the files would be royal pain and we don't want to do this. Support for individual IDEs is important, but we need to make sure that those use the latest TS, as once again, this is the kind of setup we want to have in the end.

pkozlowski-opensource:
Just checked with the latest VSC and this setup looks alright. I'm not too familiar with VSC to say if there are use-cases that are not covered, but I can see all the typings etc.

Foxandxss:
I just played with it. Webstorm (which is the tool I use) I think it is kinda brokish (it was working correctly until recently).

With atom, works perfect the way it is now and with this PR it gets more clunky, but I can point to `atom-typescript` as the responsible.

I am the first one who wants to get rid of the files array, and seems that it is somehow working good without it.

the moduleResolution one is the one I don't see. `tsc` compiles correctly without it and Webstorm doesn't complain for me if I don't put it.

Foxandxss:
TL;DR; I am OK with no files (I appreciate not having to write down all the files we need), but I don't yet see the moduleResolution one. Without it works good for me in all editors.

pkozlowski-opensource:
What is the version of WebStorm you are on? In 11 angular2 typings are not seen without "moduleResolution": "node". Are you sure you are on TS 1.6+?

Foxandxss:
I am with latest 11 (the beta or RC that they released the other day).



Leaving files, it cries because it doesn't find the typings on the file list, but removing it, it suddenly works.

Also, typescript 1.6.2

Foxandxss:
Also, cmd clicking them sends me to the typing.

pkozlowski-opensource:
Hmm, after re-starting WS it seems to work without `moduleResolution`. Not sure what is going on, but I'm going to drop it for now.

PR #65, opened 2015-10-30
test(rating): test rate behavior

Foxandxss:
A bunch of more tests, which made me notice some missing features :+1:


wesleycho:
LGTM

PR #66, opened 2015-11-02
feat(pagination): add initial version of the pagination directive

pkozlowski-opensource:


src/components/pagination.ts
```diff
@@ -0,0 +1,81 @@
...
+})
+export class NgbPagination {
+  _collectionSize = 0;
+  _page = 0;
+  _pageSize = 10;
+  _pages: number[] = [];
+
+  @Input()
+  set page(newPage: number | string) {
+    this.selectPage(parseInt(`${newPage}`, 10));
```
Foxandxss: Any good reason of using `` ? Other than being cool :P

Foxandxss:
Could you please make a Plunker? That is all we have for "docs" right now and it is easy if we can see it working nicely (better to review).

src/components/pagination.ts
```diff
@@ -0,0 +1,81 @@
+import {Component, EventEmitter, Input, Output, NgFor} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-pagination',
+  directives: [NgFor],
+  template: `
+    <nav>
+      <ul class="pagination">
+        <li [class.disabled]="!hasPrevious()">
+            <a aria-hidden="true" (click)="selectPage(_page-1)">
```
Foxandxss: This should be aria-label, not hidden.

src/components/pagination.ts
pkozlowski-opensource: Open to alternatives, if you've got any suggestions. The basic thing here that we need to accept both stings and numbers and convert them to int.

src/components/pagination.ts
pkozlowski-opensource: Good catch, thnx! Fixing.

src/components/pagination.ts
Foxandxss: I just meant about something like `parseInt(newPage, 10)`. There is no need to do it that way.

pkozlowski-opensource:
Regarding plunks: can add it. But honestly I would much prefer spending time on actually making our demo page work so we don't have to do this manual operations :-)

pkozlowski-opensource:
OK, I've rebased this PR plus added some tests. Before I go any further with this I would love to have your input guys on this type of class-level testing (bypassing UI).

Foxandxss:
This tests are comparable to the old controller tests in ng1. They are nice indeed, but we need the other type of tests too, to test the view (aria support, and that we can see stuff on screen.

I will surely do this type of tests to my other components too.

src/components/pagination.ts
pkozlowski-opensource: Oh, I was _sure_ that doing `parseInt` would throw but after re-testing it this doesn't seem to be the case. Will change. Thnx!

src/pagination/pagination.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import {Component} from 'angular2/angular2';
+
+import {NgbPagination} from './pagination';
+
+describe('ngb-pagination', () => {
+
+  describe('component class', () => {
+
+      describe('component class', () => {
```
Foxandxss: repeated describe. Still, hold to see if we reach some convention.

src/components/pagination.ts
pkozlowski-opensource: Soooo, actually, removing conversion to string makes TS upset.... I'm going to leave it as-is for now, unless anyone has a better idea on how to tackle this.

src/pagination/pagination.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  beforeEachProviders
+} from 'angular2/testing';
+
+import {Component} from 'angular2/angular2';
+
+import {NgbPagination} from './pagination';
+
+describe('ngb-pagination', () => {
+
+  describe('component class', () => {
```
Foxandxss: use the new convention name.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+  private _page = 0;
+  private _pageSize = 10;
+  pages: number[] = [];
+
+  @Input()
+  set page(newPage: number | string) {
+    this._selectPage(parseInt(`${newPage}`, 10));
+  }
+
+  get page() : number | string {
```
Foxandxss: this is never going to return string, right?

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+    return this._page;
+  }
+
+  @Input()
+  set collectionSize(newSize: number | string) {
+    this._collectionSize = parseInt(`${newSize}`, 10);
+    this._updatePages();
+  }
+
+  get collectionSize() : number | string {
```
Foxandxss: same, never string.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+    return this._collectionSize;
+  }
+
+  @Input()
+  set pageSize(newSize: number | string) {
+    this._pageSize = parseInt(`${newSize}`, 10);
+    this._updatePages();
+  }
+
+  get pageSize() : number | string {
```
Foxandxss: :)

src/pagination/pagination.ts
pkozlowski-opensource: No, but TS complains if getter and setter don't agree on types

src/pagination/pagination.ts
pkozlowski-opensource: See https://github.com/ng-bootstrap/core/pull/66#discussion_r43863664

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+    return this._pageSize;
+  }
+
+  @Output() pageChange = new EventEmitter();
+
+  hasPrevious(): boolean { return this._page > 1; }
```
Foxandxss: I think it is better to use the getter now that we have it.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  hasNext(): boolean { return this._page < this.pages.length; }
```
Foxandxss: getter here too.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  private _selectPage(pageNumber: number): void {
```
Foxandxss: no underscore.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,93 @@
...
+    var prevPageNo = this._page;
+    this._page = Math.max(Math.min(pageNumber, this.pages.length), 1);
+
+    if (this._page != prevPageNo) {
+      this.pageChange.next(this._page);
+    }
+  }
+
+  private _updatePages(): void {
```
Foxandxss: no underscore.

src/pagination/pagination.ts
pkozlowski-opensource: I think it is fine the way it is.

Foxandxss: The problem is when getters / setters do special things (apart from simply return and assign), better to use them instead of the private value directly. If tomorrow you modify a getter to do a value conversion, you will need to change more code.

pkozlowski-opensource: But this is public, not private. And there is no setter.

Foxandxss: Ah no no, I mean `_page`. Change it to:

`hasPrevious(): boolean { return this.page > 1; }`

If an internal variable like `_page`, `_pageSize` has a set/get, it is better to always use them, because they could have extra logic that needs to be applied.

So if the `get page() { }` does extra logic in the future our code will be prepared to use that new logic. Apart, it is better to not having to code the _ all the time.

That is a classic convention on C# so I guess that Typescript will follow it too.

pkozlowski-opensource: Oh, right, I thought that you want me to change `hasPrevious()` to a getter! I see now, makes sense, thnx.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,87 @@
...
+  `
+})
+export class NgbPagination {
+  private _collectionSize = 0;
+  private _page = 0;
+  private _pageSize = 10;
+  pages: number[] = [];
+
+  @Input()
+  set page(newPage: number | string) {
```
wesleycho: Do we want to support strings here?

wesleycho:
This mostly LGTM, but I do have a question about supporting strings - the code becomes even nicer if we supported just numbers for page selection.

src/pagination/pagination.ts
Foxandxss: Yes, a `<foo val="10"></foo>`. `val` is a string in here.

wesleycho: That makes me sad :( .

pkozlowski-opensource:
@Foxandxss @wesleycho added tests, this is ready for the final review.

Foxandxss:
Sure, will review later in depth.

wesleycho:
Travis is saying no :)

src/pagination/pagination.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+function expectPages(nativeEl, pagesDef: string[]) {
+  var pages = nativeEl.querySelectorAll('li');
+
+  expect(pages.length).toEqual(pagesDef.length);
+
+  for (let i = 0; i < pagesDef.length; i++) {
+    let pageDef = pagesDef[i];
+    let classIndicator = pageDef.charAt(0);
+
+    if (classIndicator == '+') {
```
wesleycho: Should be `===`

src/pagination/pagination.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+      expect(pages[i]).toHaveCssClass('active');
+      expect(pages[i]).not.toHaveCssClass('disabled');
+      expect(pages[i].innerText.trim()).toEqual(pageDef.substr(1));
+    } else if (classIndicator == '-') {
```
wesleycho: Ditto

Foxandxss:
Yeah, always run "gulp" before commiting.

pkozlowski-opensource:
> Travis is saying no :)

It is saying "yes" now :-) It was an actual difference between Chrome that I use locally and FFox on TravisCI.

src/pagination/pagination.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+  beforeEach,
+  beforeEachProviders
+} from 'angular2/testing';
+
+import {Component} from 'angular2/angular2';
+
+import {NgbPagination} from './pagination';
+
+function getLink(nativeEl, idx) {
+  return nativeEl.querySelectorAll('li')[idx].querySelector('a');
```
wesleycho: Nitpick, but this could be written
```js
return nativeEl.querySelector(`li:nth-child(${idx}) a`);
```

src/pagination/pagination.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private _updatePages(): void {
+    // re-calculate new length of pages
+    var pageCount = Math.ceil(this._collectionSize / this._pageSize);
+
+    // fill-in model needed to render pages
+    this.pages.length = 0;
+    for (var i = 1; i <= pageCount; i++) {
+      this.pages.push(i);
```
wesleycho: I assume this is to maintain reference to the `this.pages` array? Otherwise, this could be condensed to
```js
this.pages = Array.from({ length: pageCount + 1 }, (v, k) => v + 1);
```

src/pagination/pagination.spec.ts
pkozlowski-opensource: Hmm, looks legit, but doesn't work in practice (tests start to fail, at least in Chrome). Didn't dig into it too much.

src/pagination/pagination.ts
pkozlowski-opensource: Yeh, we should be safer with keeping the same array reference.

pkozlowski-opensource:
Sooooo, anything else or is this one good to go?

src/pagination/pagination.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      paginationCmpt.collectionSize = 100;
+      paginationCmpt.pageSize = 20;
+      paginationCmpt.page = 2;
+
+      expect(paginationCmpt.page).toEqual(2);
+    });
+
+  });
+
+  describe('UI', () => {
```
Foxandxss: We decided to call this `UI logic`

src/pagination/pagination.ts
```diff
@@ -0,0 +1,88 @@
...
+    this._page = this._collectionSize ? Math.max(Math.min(pageNumber, this.pages.length), 1) : pageNumber;
+
+    if (this.page != prevPageNo) {
+      this.pageChange.next(this.page);
+    }
+  }
+
+  private _updatePages(): void {
+    // re-calculate new length of pages
+    var pageCount = Math.ceil(this._collectionSize / this._pageSize);
```
Foxandxss: getter in both there.

src/pagination/pagination.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    // fill-in model needed to render pages
+    this.pages.length = 0;
+    for (var i = 1; i <= pageCount; i++) {
+      this.pages.push(i);
+    }
+
+    // make sure that the selected page is within available pages range
+    this.selectPage(this._page);
```
Foxandxss: getter

Foxandxss:
After those nitpicks gets fixed, It looks good for me.

src/pagination/pagination.ts
pkozlowski-opensource: This won't work in TypeScrip:

> src/pagination/pagination.ts(77,31): error TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number' or an enum type.

The reason is that our getter is declared as `get collectionSize(): number | string { return this._collectionSize; }` and it is declared like this since TS requires same types on getters and setters

src/pagination/pagination.ts
pkozlowski-opensource: See https://github.com/ng-bootstrap/core/pull/66/files#r45610350

Foxandxss:
I can see what you say. Typing is problematic time to time.

PR #68, opened 2015-11-02
feat(all): make it easy to import all directives

pkozlowski-opensource:


Foxandxss:
Is there a real need of having to do it twice? one for import, one for export.

I was also modifying this file in a PR I am making. Basically yo do:

`export * from 'directive';`

which is what lot of people are doing.

pkozlowski-opensource:
@Foxandxss the problem I'm trying to solve is this:

```javascript
import {NGB_DIRECTIVES} from '...';

@Component({
   directives: [NGB_DIRECTIVES]
})
```

Not sure how `export * from 'directive'` would help with this...

Foxandxss:
I know what problem are you trying to solve, I just think that `core.ts` is now too verbose. For each new directive you need to import it and also export it.

I was just wondering if there is an easier way to achieve that.

The `export *` thing was something I was doing before, I reverted that in my PR.

pkozlowski-opensource:
> I just think that core.ts is now too verbose. I was just wondering if there is an easier way to achieve that.

I'm not aware of any other mechanism. But would love to see a less-verbose option if anyone is aware of such.

Foxandxss:
Seems like there is no other option. I was mistaken. I will rebase and merge this.

PR #76, opened 2015-11-03
chore: add basic CONTRIBUTING file

### LINKED ISSUE #72: Convention: private or not private
Should we define our component variable as private?

```
class Foo {
  foo;
  bar;
}
```

Or

```
class Foo {
  private foo;
  private bar;
}
```

This doesn't include private functions when we know that they are not mean to be used in public like helper functions.

### LINKED ISSUE #73: Convention: prepend underscore on variables
Should we prepend variables like:

```
class Foo {
  _foo;
  _bar;
}
```

The way to work with components is to create a setter for inputs, that way we can keep or component updated with changes, for example:

```
@input
set foo(value) {
  this._foo = value;
  // do something
}
```

So far so good, but you can use this like:

```html
<foo #foo></foo>
{{foo._foo}}
```

So if we decide to go with underscores, we will have that syntax, our we could create a getter.

### LINKED ISSUE #74: Convention: explicit or implicit typing
Something like:

`let foo = 0;`

is already a `number` variable, it is implicit due to initialization.

Should we rely on that or should we:

`let foo:number = 0;`

### LINKED ISSUE #75: Convention: Split testing for behavior and UI
**Final Decision:** Follow the snippet on this message.

A component can be tested in two ways (and we should do both).

You can test the behavior like we used to test directive's controllers. No UI involved. For example testing that setting a new value on a progressbar, triggers the logic to calculate the new completion percentage.

You can also test UI, that involves the component creation. Same as testing directives as we used to in ui-bootstrap. Following the previous example. We can test that changing the value, the progressbar width changed with the new value.

For this, I propose:

```
describe('componentname', () => {
  // some common code / matchers / initialization

  describe('business logic', () => {
  });

  describe('UI', () => {
  });
});
```

Foxandxss:
A bit basic, but there is no need of adding more at this stage of the game.

Closes #72 
Closes #73 
Closes #74 
Closes #75

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,12 @@
+## You want to contribute some code?
+
+We are always looking for the quality contributions and will be happy to accept your Pull Requests as long as those adhere to some basic rules:
+
+* Please make sure that your contribution fits well in the project's context:
+  * we are aiming at rebuilding bootstrap components in pure Angular 2, without any dependencies on any external JavaScript library;
+  * the only dependency should be bootstrap CSS and its markup structure
+* Please assure that you are submitting quality code, specifically make sure that:
+  * your component has accompanying tests and all the tests are passing
+  * your PR doesn't break the build; check the Travis-CI build status after opening a PR and push corrective commits if anything goes wrong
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
LGTM

PR #78, opened 2015-11-04
chore(tooling): move everything to gulp

Foxandxss:
Not the simple PR you want to see.

This is a workflow made with the right base to make it group up when we need some extra tasks.

I split it in 3 separate categories. The important bits are:

Instead of having a insanely big task like we did on ui-bootstrap that would do everything, I split it in two tasks:

* `gulp` - This is the task you want to run before commit, to make sure that everything is correct (formatting and tests).
* `gulp build` - Does all the previous things, but also builds the project.

For testing + watch, you need two tabs (no way to simplify that I think):

One with `gulp watch` and the other, `gulp test` or `karma start` or `npm test`. I left the `npm test` for convenience.

Travis will run the  `gulp` one. It doesn't need to build anything, that only adds time.

Cleaned the package.json, and moved the git hook to a `install` hook. That is a better way of managing it.

After this gets merged, I will update the docs to explain how to work with this tooling.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,8 +1,27 @@
 var gulp = require('gulp');
+var ts = require('gulp-typescript');
 var gutil = require('gulp-util');
+var sourcemaps = require('gulp-sourcemaps');
+var ddescribeIit = require('gulp-ddescribe-iit');
+var karmaServer = require('karma').Server;
 var webpack = require('webpack');
-var gulpFormat = require('gulp-clang-format');
+var del = require('del');
+var merge = require('merge2');
 var clangFormat = require('clang-format');
+var gulpFormat = require('gulp-clang-format');
+var runSequence = require('run-sequence');
+
+// Transpiling & Building
+
+var buildProject = ts.createProject('tsconfig.json', {declaration: true});
+
+gulp.task('clean:build', function() { return del('dist/'); });
+
+gulp.task('cjs', function() {
+  var tsResult = gulp.src(['src/**/*.ts', '!src/**/*.spec.ts']).pipe(ts(buildProject));
+
+  return merge([tsResult.dts.pipe(gulp.dest('dist/cjs')), tsResult.js.pipe(gulp.dest('dist/cjs'))]);
+});
 
 gulp.task('umd', function(cb) {
   webpack(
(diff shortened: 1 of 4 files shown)


Foxandxss:
Tslint will come later, seems that it needs more configuring and fidling and that is for another PR.

Foxandxss:
Oh, clang was complaining about formats now, will update that in master and rebase (to not dirt this PR).

pkozlowski-opensource:
@Foxandxss I'm not big fan of splitting gulpfile over multiple files as:
* it makes, IMO, harder to see what is in build
* usually leads to config / code duplication

The other thing I'm not totally sold on is `gulp-load-plugins` - personally I find it harder to see what is imported / used.

Having said the above I can perfectly live with the proposed setup, was just spelling out my personal preferences. 

Didn't spend too much time reviewing every details of the individual tasks - we can always update / fix those as we go.

Foxandxss:
I split them because a 500+ gulpfile is hard to grow up. I don't think I
have duplication at all. Said that, I don't mind putting it back to one
file.

On the other hand, I have mixed feeling with load plugins, so that go away.

About the task, yeah, they seems fine for now, and this is enough
bikeshedding for now.

wesleycho:
I'm personally fine with the split, but not hugely opinionated now - maybe better to keep it in a singular file until it gets more complex though for simplicity.

Foxandxss:
Updated, please re-check.

gulpfile.js
```diff
@@ -34,11 +84,15 @@ gulp.task('enforce-format', function() {
   });
 });
 
-gulp.task('build', ['umd', 'check-format']);
-gulp.task('default', ['build']);
-
-// formatting
 function doCheckFormat() {
-  return gulp.src(['gulpfile.js', 'karma-test-shim.js', 'src/**/*.ts'])
+  return gulp.src(['gulpfile.js', 'tasks/**/*.js', 'karma-test-shim.js', 'src/**/*.ts'])
```
pkozlowski-opensource: we don't have `'tasks/**/*.js'` any more, right?

gulpfile.js
```diff
@@ -1,8 +1,27 @@
...
+gulp.task('clean:build', function() { return del('dist/'); });
+
+gulp.task('cjs', function() {
+  var tsResult = gulp.src(['src/**/*.ts', '!src/**/*.spec.ts']).pipe(ts(buildProject));
+
+  return merge([tsResult.dts.pipe(gulp.dest('dist/cjs')), tsResult.js.pipe(gulp.dest('dist/cjs'))]);
+});
 
 gulp.task('umd', function(cb) {
   webpack(
```
pkozlowski-opensource: could we import (`require`) `webpack` here? More generally - if a given plugin is only used in one tasks, could we import it in this task only? This would speed up gulp start.

gulpfile.js
```diff
@@ -1,8 +1,27 @@
...
+var runSequence = require('run-sequence');
+
+// Transpiling & Building
+
+var buildProject = ts.createProject('tsconfig.json', {declaration: true});
+
+gulp.task('clean:build', function() { return del('dist/'); });
+
+gulp.task('cjs', function() {
+  var tsResult = gulp.src(['src/**/*.ts', '!src/**/*.spec.ts']).pipe(ts(buildProject));
```
pkozlowski-opensource: We start to have `'src/**/*.ts'` and other globs duplicated in few places. Could we pull it to some kind of config object? Ex.: https://github.com/pkozlowski-opensource/ng2-play/blob/master/gulpfile.js#L3-L5

pkozlowski-opensource:
Left few comments, LGTM otherwise.

PR #79, opened 2015-11-04
feat(progressbar): add initial progressbar

Foxandxss:
It is the barebones just yet, but there is something important I want to discuss in here and I am waiting for the tooling to make the development a bit easier.

The problem I have in here is that we use it like:

```
<ngb-progressbar class="progress-striped" value="22"></ngb-progressbar>
```

The problem is that it was going to generate a whole progressbar inside, copy the `progress-striped` class and apply it inside it. Works good but then we will have two elements carrying the same class, and that can backfire pretty easily.

I hacked it a bit so I delete the classes on the parent but I am not loving the approach.

Check this [plunker](http://plnkr.co/edit/9Vts7W6wNtWSjRRhhazX?p=preview).

Thoughts?

### DIFF
--- a/src/progressbar/progressbar.spec.ts
+++ b/src/progressbar/progressbar.spec.ts
@@ -1,0 +1,116 @@
+import {iit, it, ddescribe, describe, expect, injectAsync, TestComponentBuilder} from 'angular2/testing';
+
+import {Component} from 'angular2/angular2';
+
+import {NgbProgressbar} from "./progressbar";
+
+function getBarWidth(nativeEl): string {
+  return nativeEl.querySelector('.progress-bar').style.width;
+}
+
+function getProgressbar(nativeEl: Element): Element {
+  return nativeEl.querySelector('progress');
+}
+
+describe('ng-progressbar', () => {
+
+  describe('business logic', () => {
+    let progressCmp: NgbProgressbar;
+
+    beforeEach(() => { progressCmp = new NgbProgressbar(); });
+
+    it('should calculate the percentage (default max size)', () => {
+      progressCmp.value = 50;
+      expect(progressCmp.calculatePercentage()).toBe(50);
+
+      progressCmp.value = 25;
+      expect(progressCmp.calculatePercentage()).toBe(25);
+    });
+
+    it('should calculate the percentage (custom max size)', () => {
+      progressCmp.max = 150;
+
+      progressCmp.value = 75;
+      expect(progressCmp.calculatePercentage()).toBe(50);
+
+      progressCmp.value = 30;
+      expect(progressCmp.calculatePercentage()).toBe(20);
+    });
+  });
+
+  describe('UI logic', () => {
+    it('accepts a value and respond to value changes', injectAsync([TestComponentBuilder], (tcb) => {
+         const html = '<ngb-progressbar [value]="value"></ngb-progressbar>';
+
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           expect(getBarWidth(fixture.debugElement.nativeElement)).toBe('10%');
+
+           fixture.debugElement.componentInstance.value = 30;
+           fixture.detectChanges();
+           expect(getBarWidth(fixture.debugElement.nativeElement)).toBe('30%');
+         });
+       }));
+
+    it('accepts a max value and respond to max changes', injectAsync([TestComponentBuilder], (tcb) => {
+         const html = '<ngb-progressbar [value]="value" [max]="max"></ngb-progressbar>';
+
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           expect(getBarWidth(fixture.debugElement.nativeElement)).toBe('20%');
+
+           fixture.debugElement.componentInstance.max = 200;
+           fixture.detectChanges();
+           expect(getBarWidth(fixture.debugElement.nativeElement)).toBe('5%');
+         });
+       }));
+
+    it('accepts a custom type', injectAsync([TestComponentBuilder], (tcb) => {
+         const html = '<ngb-progressbar [value]="value" [type]="type"></ngb-progressbar>';
+
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           expect(getProgressbar(fixture.debugElement.nativeElement)).toHaveCssClass('progress-warning');
+
+           fixture.debugElement.componentInstance.type = 'info';
+           fixture.detectChanges();
+           expect(getProgressbar(fixture.debugElement.nativeElement)).toHaveCssClass('progress-info');
+         });
+       }));
+
+    it('accepts striped as boolean attr', injectAsync([TestComponentBuilder], (tcb) => {
+         const html = '<ngb-progressbar [value]="value" striped></ngb-progressbar>';
+
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           expect(getProgressbar(fixture.debugElement.nativeElement)).toHaveCssClass('progress-striped');
+         });
+       }));
+
+    it('accepts striped as normal attr', injectAsync([TestComponentBuilder], (tcb) => {
+         const html = '<ngb-progressbar [value]="value" [striped]="striped"></ngb-progressbar>';
+
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           expect(getProgressbar(fixture.debugElement.nativeElement)).toHaveCssClass('progress-striped');
+
+           fixture.debugElement.componentInstance.striped = false;
+           fixture.detectChanges();
+           expect(getProgressbar(fixture.debugElement.nativeElement)).not.toHaveCssClass('progress-striped');
+         });
+       }));
+  });
+});
+
+@Component({selector: 'test-cmp', directives: [NgbProgressbar], template: ''})
+class TestComponent {
+  value = 10;
+  max = 50;
+  striped = true;
+  type = 'warning';
+}
--- a/src/progressbar/progressbar.ts
+++ b/src/progressbar/progressbar.ts
@@ -1,0 +1,34 @@
+import {Component, Input} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-progressbar',
+  template: `
+    <progress class="progress {{striped && 'progress-striped'}} {{type && 'progress-' + type}}" [value]="value" [max]="max">
+      <div class="progress">
+        <span class="progress-bar" [style.width.%]="calculatePercentage()"><ng-content></ng-content></span>
+      </div>
+    </progress>
+  `
+})
+export class NgbProgressbar {
+  private _striped: boolean;
+
+  @Input() max = 100;
+
+  @Input()
+  set striped(value: boolean | string) {
+    if (value === '') {  // boolean usage
+      value = true;
+    }
+
+    this._striped = !!value;
+  }
+
+  get striped(): boolean | string { return this._striped; }
+
+  @Input() type: string;
+
+  @Input() value: number;
+
+  calculatePercentage() { return 100 * this.value / this.max; }
+}
--- a/src/core.ts
+++ b/src/core.ts
@@ -2,13 +2,23 @@ import {NgbAccordion, NgbAccordionPanel} from './accordion/accordion';
 import {NgbCollapse} from './collapse/collapse';
 import {NgbDropdown, NgbDropdownToggle} from './dropdown/dropdown';
 import {NgbPagination} from './pagination/pagination';
+import {NgbProgressbar} from './progressbar/progressbar';
 import {NgbRating} from './rating/rating';
 
 export {NgbAccordion, NgbAccordionPanel} from './accordion/accordion';
 export {NgbCollapse} from './collapse/collapse';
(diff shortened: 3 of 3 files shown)


src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,36 @@
+import {Component, Input, ElementRef} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-progressbar',
+  template: `
+    <progress class="progress {{extraClasses}}  {{type && 'progress-' + type}}" [value]="value" [max]="max">
+      <div class="progress">
```
Foxandxss: I should perhaps add the classes in here for IE9, not sure and not liking it.

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.calculatePorcentage();
+  }
+
+  get value() { return this._value; }
+
+  @Input() type: string;
+
+  @Input() max = 100;
+
+  constructor(private element: ElementRef) {
```
pkozlowski-opensource: We can do everything without injecting `ElementRef`:
* you can get classes set on an element by doing a setter for `class`, see: https://github.com/angular/angular/blob/master/modules/angular2/src/core/directives/ng_class.ts
* this will also make sure that the class is not propagated to the host element so you won't need to clear it by hand

In short: I don't think we need / should be injecting `ElementRef` here

Foxandxss: That would be fantastic. I'm not at home now, so I will check that later

pkozlowski-opensource:
OK, so apart from the `ElementRef` comment I think that we need to think a bit more about an API. IMO there are 2 aspects here:
* types of bindings
* abstraction level

Let's start with types. So, we can't prevent people from using this component as `<ngb-progressbar value="22"></ngb-progressbar>` or `<ngb-progressbar value="{{22}}"></ngb-progressbar>`. In this case we are going to get `value` as string and I'm not sure this is handled properly. Same applies to `min` and `max`.

Regarding abstractions on the API level: personally I would expect to have a higher level api than class names. I would love to work with sth like:

`<ngb-progressbar value="22" type="danger" striped>` 

or 

`<ngb-progressbar value="22" type="danger" striped="true">` 

I fully appreciate that it might be seen as too restrictive but it is less typing and I'm able to express what I _want_ instead of how it is done. 

Once again, I'm not saying that we should do just this - just expressing my personal preference here.

Foxandxss:
About the striped as an attribute, I had the same idea as you, as an easy
way of avoiding the class issue, but then I thought that better to learn a
proper way. Still I think that we better create a striped argument for this
concrete use case.

About the foo="{{val}}". I can see that, but why would a user do that?
Still, I see your point

pkozlowski-opensource:
> About the foo="{{val}}". I can see that, but why would a user do that?

Hopefully people wouldn't do it  too often, but this is equivalent to `<ngb-progressbar value="22">` => `<ngb-progressbar [value]="'' + 22">`

src/progressbar/progressbar.ts
Foxandxss: So what ng-class does is, gets `class` as an input (so you get all the classes on the parent), but to delete it from the parent, it is toggling them. So yes, it is manually deleting the classes on the host element and needed a ElementRef and Renderer.

The idea of a striped attribute is something I had in mind too, but the problem is that it will work for this use case, but this topic will come back for sure. Still, not sure if there is a clean solution (not even in ng1).

Foxandxss: Here is what it would like by doing ng-class way: http://plnkr.co/edit/bTKqnK5MgGXWY0iTteFX?p=preview

More code but no nativeElement usage (AKA it can work in  a web worker).

Foxandxss:
About that issue, since our directive will get the string / number and multiply it by 100, it will do a conversion in the flight.

src/progressbar/progressbar.ts
pkozlowski-opensource: @Foxandxss could you put this new version in a PR. There are few things that can / should be changed but it is super-hard to comment on the code in a plunk.

Foxandxss:
Ok @pkozlowski-opensource I updated this with those changes.

I have been playing a bit with the idea yesterday and I could do something like:

```
@HostBinding() get class { return aEmptyString; }
```

More or less, to empty the host classes after something happens.

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,50 @@
...
+      </div>
+    </progress>
+  `
+})
+export class NgbProgressbar {
+  _value: number;
+  _class: string;
+  min = 0;
+
+  @Input() set class
```
pkozlowski-opensource: Could we use sth else than "class" here? I know it "works" but is a pain in a neck for older browsers.

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,50 @@
...
+  (newClasses: string) {
+    this._class = newClasses;
+    const initialClasses = newClasses && newClasses.split(' ');
+    initialClasses.forEach(className => this.toggleClass(className));
```
pkozlowski-opensource: I don't understand what we are doing here... Could you leave a comment as to why this logic is needed?

src/progressbar/progressbar.ts
Foxandxss: Definitely, not only for older browsers but clang is doing weird things to it.

Still, we have to decide if we want this behavior, perhaps use @HostBindings or even go the `striped` attribute (so we won't need to do all this class thing for this component).

pkozlowski-opensource:
@Foxandxss it looks better but I'm still not sold on the CSS-based API, let's see what others got to say.

Also I don't thin we need to use renderer / element ref, but I can't be sure since I don't understand why do we have this logic at all...

src/progressbar/progressbar.ts
Foxandxss: Sure, that is for toggling out (remove) the classes in the host element. That is what ng-class does too. The classes are still on host until you remove them, and I remove them in the same way as ng-class, with a toggleClass method that will modify the host.

pkozlowski-opensource: Bah, I really don't see why we need it... If someone removed classes from host, they are gone already, no?

If you want to copy classes from host to a progress element just do this, as an entire string. Having said this I _really_ don't think we should be basing our API css classes. IMO we should have our own attributes and map those to the corresponding bootstrap css

Foxandxss: No one will remove the classes. You will end with:

```
<ng-progressbar class="foo bar baz" value="20">
  <progress class="progress foo bar baz" ...
```

You need the remove the host one to not having them twice.

Said that, for this use case, let's go with attributes and remove all of this.

pkozlowski-opensource: > Said that, for this use case, let's go with attributes and remove all of this.

:+1:

src/progressbar/progressbar.ts
wesleycho: cssClass perhaps?

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,50 @@
+import {Component, Input, ElementRef, Renderer} from 'angular2/angular2';
+
+@Component({
+  selector: 'ngb-progressbar',
+  template: `
+    <progress class="progress {{class}} {{type && 'progress-' + type}}" [value]="value" [max]="max">
+      <div class="progress">
+        <span class="progress-bar" [style.width.%]="calculatePorcentage()"><ng-content></ng-content></span>
```
wesleycho: Should be `Percentage`

Foxandxss:
Code part is done I think (basic implementation). Time for more tests. Aria stuff is needed, for another PR.

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,49 @@
...
+        <span class="progress-bar" [style.width.%]="calculatePercentage()"><ng-content></ng-content></span>
+      </div>
+    </progress>
+  `
+})
+export class NgbProgressbar {
+  private _max = 100;
+  private _striped = false;
+  private _value: number;
+  min = 0;
```
pkozlowski-opensource: Shouldn't `min` be an `@Input`?

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  @Input()
+  set max(value: number | string) {
+    this._max = parseInt(`${value}`, 10);
+    this.calculatePercentage();
```
pkozlowski-opensource: no need to call `calculatePercentage()` here - it is going to be called from a template

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  get striped(): boolean | string { return this._striped; }
+
+  @Input() type: string;
+
+  @Input()
+  set value(value: number) {
+    this._value = value;
+    this.calculatePercentage();
```
pkozlowski-opensource: no need to call `calculatePercentage()` here - it is going to be called from a template

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  get value(): number { return this._value; }
+
+  private calculatePercentage() { return 100 * this.value / (this._max - this.min); }
```
pkozlowski-opensource: It shouldn't be `private` since we are calling it from a template

pkozlowski-opensource: How about making it a getter and naming `valueAsPercent` or `perecentValue`?

src/progressbar/progressbar.spec.ts
```diff
@@ -0,0 +1,27 @@
+import {iit, it, ddescribe, describe, expect, injectAsync, TestComponentBuilder} from 'angular2/testing';
+
+import {Component} from 'angular2/angular2';
+
+import {NgbProgressbar} from "./progressbar";
+
+describe('ng-progressbar', () => {
+  describe('UI logic', () => {
```
pkozlowski-opensource: Could we add a bunch of tests for "business logic" - in the case of this directive we only care about percent calculation. I guess we should test a happy path and few corner cases

src/progressbar/progressbar.ts
Foxandxss: We don't have it back at ui-bootstrap. Never made sense to us, but that can be reconsidered.

src/progressbar/progressbar.ts
Foxandxss: private thing is a leftover (we, I mean I, changed conventions lots of times lately :P)

Not sure about the getter. Will think.

src/progressbar/progressbar.ts
pkozlowski-opensource: OK, but in this case there is no point of having this variable at all - you can just delete it.

src/progressbar/progressbar.spec.ts
Foxandxss: Didn't touch tests yet, and I was thinking about business logic to test calculation yes.

Foxandxss:
I am not entirely sure how the dirty checking works in here (I guess that pretty much the same as angular) so I never realized that `calculatePercentage` was getting called. Makes sense.

Foxandxss:
I think this is good to go. Review it please.

src/progressbar/progressbar.spec.ts
```diff
@@ -0,0 +1,116 @@
+import {iit, it, ddescribe, describe, expect, injectAsync, TestComponentBuilder} from 'angular2/testing';
+
+import {Component} from 'angular2/angular2';
+
+import {NgbProgressbar} from "./progressbar";
+
+function getBarWidth(nativeEl): number {
```
pkozlowski-opensource: `: Element`?

Foxandxss: That is something to research. `Element` doesn't have `style` on it so it fails. That being said, it returns a string, not a number.

src/progressbar/progressbar.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  return nativeEl.querySelector('.progress-bar').style.width;
+}
+
+function getProgressbar(nativeEl: Element): Element {
+  return nativeEl.querySelector('progress');
+}
+
+describe('ng-progressbar', () => {
+
+  describe('business logic', () => {
```
pkozlowski-opensource: How do we want to handle corner cases (ex. max <0, value > max etc,)?

Foxandxss: There, but it needs extra logic to be done and better if we start with a smaller PR with an initial implementation and then we move to add more features. That is my next one.

src/progressbar/progressbar.ts
```diff
@@ -0,0 +1,34 @@
...
+  template: `
+    <progress class="progress {{striped && 'progress-striped'}} {{type && 'progress-' + type}}" [value]="value" [max]="max">
+      <div class="progress">
+        <span class="progress-bar" [style.width.%]="calculatePercentage()"><ng-content></ng-content></span>
+      </div>
+    </progress>
+  `
+})
+export class NgbProgressbar {
+  private _striped: boolean;
```
pkozlowski-opensource: What about animated stripes (http://v4-alpha.getbootstrap.com/components/progress/#animated-stripes). This seems to be a pure CSS thing in BS4

pkozlowski-opensource:
This LGTM, generally speaking. We can get this one in and do additional stuff in subsequent PRs:
- aria
- animated
- corner cases in % calculation

src/progressbar/progressbar.ts
Foxandxss: Yeah, that needs more researching outside this PR because it doesn't work for me out of the box (in fact, I don't see them animated on that page).

So I will open a PR when I figure that out.

Foxandxss:
Sure. I will tackle that myself.

wesleycho:
This LGTM too

PR #81, opened 2015-11-07
chore(dropdown): switch menu to directive

Foxandxss:
A `<ng-content>` template makes sense if you have something else on the template. If you remove the template entirely, the inner content will appear nonetheless.

Not sure about the chore thing, but I think that the first changelog won't be autogenerated (lot of experimentation that would result in a weird changelog.

[Plunker](http://plnkr.co/edit/0GZqyybzjOuc88Gw31p3?p=preview)

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -1,4 +1,4 @@
-import {Component, Input} from 'angular2/angular2';
+import {Component, Directive, Input} from 'angular2/angular2';
 
 @Component({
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
> Not sure about the chore thing,

`refactor` is the way to go in cases like this

Foxandxss:
Definitely. Forgot about it.

wesleycho:
LGTM

PR #82, opened 2015-11-08
feat(dropdown): change API to be more natural

wesleycho:
- Remove unnecessary `dropdownMenu` component
- Change dropdown to directive
- Add tests for toggle behavior


Foxandxss:
Same things I said with accordion, and this is where the `hasClass` global matcher could shine. We could probably merge it as is (once you do the describe thing for ui logic), and then create a file for external matchers and apply them. I can do that.

I would call this PR a refactor and not a feature.

wesleycho:
Alright, will merge then and change the commit message.

Foxandxss:
Fix the tests conventions and then we will see if it can go merge.

PR #84, opened 2015-11-08
chore(tooling): clean temp folder before any major action

Foxandxss:
This fixes the issue where switching branches would result on "stale" temp folder.


wesleycho:
LGTM

PR #85, opened 2015-11-09
feat(carousel): create initial implementation

### LINKED ISSUE #12: Carousel
Recreate BS3 Carousel with ng2 Component.

wesleycho:
- Create carousel and slide implementations

This is a WIP atm - currently having problems hooking into the lifecycle event in the test, any comments/suggestions are welcome.

[Plunk](http://plnkr.co/edit/AaJYbYHPxng58vsX5pcm?p=preview) of WIP (buggy atm)

This should close #12.

### DIFF
--- a/src/carousel/carousel.spec.ts
+++ b/src/carousel/carousel.spec.ts
@@ -1,0 +1,97 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  injectAsync,
+  TestComponentBuilder,
+  beforeEachProviders
+} from 'angular2/testing';
+
+import {Component, NgFor} from 'angular2/angular2';
+
+import {NgbCarousel, NgbSlide} from './carousel';
+
+function getSlides(compiled) {
+  return compiled.querySelectorAll('.carousel-item');
+}
+
+function hasClass(element: Element, str: string): Boolean {
+  return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.className)
+}
+
+describe('ngb-carousel', () => {
+  describe('UI logic', () => {
+    let html: string;
+
+    beforeEach(() => {
+      html = `
+      <ngb-carousel [(active)]="activeSlide">
+        <ngb-slide>
+          <div class="text-content">{{slides[0].content}}</div>
+        </ngb-slide>
+        <ngb-slide>
+          <div class="text-content">{{slides[1].content}}</div>
+        </ngb-slide>
+        <ngb-slide>
+          <div class="text-content">{{slides[2].content}}</div>
+        </ngb-slide>
+      </ngb-carousel>
+    `;
+    });
+
+    it('should set the first slide as active when no active', injectAsync([TestComponentBuilder], (tcb) => {
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           const compiled = fixture.debugElement.nativeElement;
+
+           let slides = getSlides(compiled);
+
+           expect(hasClass(slides[0], 'active')).toBe(true);
+           expect(hasClass(slides[1], 'active')).toBe(false);
+           expect(hasClass(slides[2], 'active')).toBe(false);
+         });
+       }));
+
+    it('should change slides appropriately', injectAsync([TestComponentBuilder], (tcb) => {
+         return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+
+           const tc = fixture.debugElement.componentInstance;
+           const compiled = fixture.debugElement.nativeElement;
+
+           let slides = getSlides(compiled);
+
+           tc.activeSlide = 1;
+           fixture.detectChanges();
+
+           expect(hasClass(slides[0], 'active')).toBe(false);
+           expect(hasClass(slides[1], 'active')).toBe(true);
+           expect(hasClass(slides[2], 'active')).toBe(false);
+
+           tc.activeSlide = 0;
+           fixture.detectChanges();
+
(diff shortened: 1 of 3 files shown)


Foxandxss:
Read the slack conversation. You don't really need to use `onChange` in here and I think you shouldn't because that overrides the whole change mechanism so you need to do everything by hand. (if I got the hook idea correctly)

What you need to do is:

```
private _active = false;

@Input()
set active(value) {
  this._active = value;
  // Do something with the new value
}

get active() {
  return this._active;
}
```

Try with that.

src/carousel/carousel.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+describe('ngb-carousel', () => {
+  describe('UI logic', () => {
+    let html: string;
+
+    beforeEach(() => {
+      html = `
+      <ngb-carousel>
+        <ngb-slide [(active)]="slides[0].active">
```
pkozlowski-opensource: IMO we should re-think the API we've got - `active` on each slide is not great since in theory it allows multiple active slides. 

How about an "active index" or similar on the `ngb-carousel` level?

src/carousel/carousel.ts
```diff
@@ -0,0 +1,113 @@
...
+      <ng-content></ng-content>
+    </div>
+  `
+})
+export class NgbSlide implements OnDestroy {
+  @Input() private active: boolean = false;
+  @Output() private activeChange: EventEmitter = new EventEmitter();
+  private _isActive: boolean = false;
+
+  constructor(private carousel: NgbCarousel) {
```
pkozlowski-opensource: I think that with Angular2 we cam do it other way around - that is - have the carousel to keep a list of children

Foxandxss: And it is not what he is doing in here? Adding new NgbSlides to the NgbCarousel list.

pkozlowski-opensource: This is my point: I don't think we need to add it: it is enough that parent queries for children and maintains an index of an active one. This would mean that we don't need any intermediate data structures.

Foxandxss: So you say that a children shouldn't need to register with the parent anymore? That the parent is smart enough to know how many children are and do the job?

That sounds good but no idea in how to do it.

pkozlowski-opensource: > So you say that a children shouldn't need to register with the parent anymore? That the parent is smart enough to know how many children are and do the job?

Exactly

> That sounds good but no idea in how to do it.

See https://github.com/angular/angular/blob/master/modules/angular2/test/core/linker/query_integration_spec.ts

Foxandxss: What should I see in there? I am doing some demos, query Query, ViewChildren... and I am not reaching anything yet.

wesleycho: Does this mean for the accordion, there is a better way to get the children from the parent accordion?

pkozlowski-opensource: > Does this mean for the accordion, there is a better way to get the children from the parent accordion?

Didn't look into accordion impl yet, but yes, potentially yes. In short, with ng2 we've got other patters / mechanism in our toolbox.

Foxandxss: Will fix accordion as soon as I discover how this work.

Foxandxss: Alright, I was dumb enough, I will send a PR for accordion later.

wesleycho:
Updated code to where the Plunker is in a working state - tests still fail though.

src/carousel/carousel.ts
```diff
@@ -0,0 +1,81 @@
...
+export class NgbCarousel {
+  @Input() private active: number;
+  @Output() private activeChange: EventEmitter = new EventEmitter();
+  @ViewChildren(NgbSlide) private slidesQuery: QueryList<NgbSlide>;
+  private slides: NgbSlide[] = [];
+  private activeIndex: number;
+
+  constructor(@Query(NgbSlide) query: QueryList<NgbSlide>) { this.slidesQuery = query; }
+
+  afterContentInit() {
```
wesleycho: I want to change this to instead observe the changes in the query (maybe initialize call in constructor?

pkozlowski-opensource:
@wesleycho I had some more time to look into the proposed implementation and it goes in the right direction, so let's land it sooner than later. However I would like to propose few changes:
* let's keep a reference to an active slide, instead of an index;
* let's use slides as attributes on a template element;

Why those changes? The most important argument is an API where it is a carousel who knows which slide is active, instead of each slide tracking its own "active" state. This, IMO, creates less confusing API and avoid corner cases where multiple sides have the "active" slide set to "true" (I know that we can check and correct those situation but I would say it is better when those situations are disallowed "by design").

Here is a plunker with what I'm proposing: http://plnkr.co/edit/fL78MvzQY5L7nXja4MJ5?p=preview

Note that this implementation requires an equivalent of the `NgbInsert` directive. Such a directive will be part of the core (see https://github.com/angular/angular/issues/7615). I hope to land it in core in the coming days but we could host our impl till then.

Open to discuss this more!

wesleycho:
This makes a lot of sense for now - I'll refresh this PR sometime in the next week and refactor it to a single component.

I did a similar refactor with UI Bootstrap to have the state tracked at the carousel level because of weird code slide level active state causes, making it hard to reason about.

pkozlowski-opensource:
FYI, I've just created PR https://github.com/angular/angular/pull/8021 to add a `NgInsert` directive. When it lands to the core our carousel directive will look like so: http://plnkr.co/edit/0SYcvPewTbQIjlL4Zyv2?p=preview

Foxandxss:
So this like an ng-include for <template>?

pkozlowski-opensource:
@Foxandxss yes, more or less. It should support components as well in the future, see discussion in the PR

pkozlowski-opensource:
Suppressed by #179

Issue #49, opened 2015-10-13
Accordion extra features


Foxandxss:
The accordion is missing:

- [ ] All the accessibility
- [ ] Be able to use HTML on the header.
- [x] Swap hidden for collapse.
- [x] Use ContentChildren instead of Query.
- [ ] Remove circular deps
- [x] See if it is feasible to create a generic panel from the groups.
- [ ] Change API of `isOpen`.

wesleycho:
We could do a named ng-content element with `select` usage, i.e.
`<ng-content select="ngb-accordion-heading"></ng-content>`

Then it would be used like
```html
<ngb-accordion>
  <ngb-accordion-group>
    <ngb-accordion-header>Foo</ngb-accordion-header>
    <ngb-accordion-content>Bar</ngb-accordion-content>
  </ngb-accordion-group>
</ngb-accordion>
```

Foxandxss:
Yup, that was already in my mental todo list. But before I want a 100% (or the closest we can) test coverage for what we have already. Before accordion grows too much and we lose the track of what we need to test yet.

pkozlowski-opensource:
Closing in favor of more focused issues I've just opened.

azarus:
Any chance we could have it any soon?

PR #86, opened 2015-11-09
refactor(accordion): apply conventions and querylist

Foxandxss:
Seeing the diff seems like I rewrote the entire thing. Not true, but please re-review it.

I applied all the conventions we talked about (I updated the wiki, check it please).

I simplified the accordion to be a `Directive` and I moved all the logic to use `QueryList`.

The `QueryList` thing is awesome, I was able to simplify the code a bit more by removing all the "removeGroups" logic, that is not needed anymore.

For the tests, since we are not doing business logic (at least atm), I removed the inner describe (don't kill me Wes).

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -1,40 +1,12 @@
-import {Component, Input, OnDestroy} from 'angular2/angular2';
-
-@Component({selector: 'ngb-accordion, [ngb-accordion]', template: `<ng-content></ng-content>`})
-export class NgbAccordion {
-  @Input('closeOthers') private onlyOneOpen: boolean;
-  private groups: Array<NgbAccordionGroup> = [];
-
-  addGroup(group: NgbAccordionGroup): void { this.groups.push(group); }
-
-  closeOthers(openGroup): void {
-    if (!this.onlyOneOpen) {
-      return;
-    }
-
-    this.groups.forEach((group: NgbAccordionGroup) => {
-      if (group !== openGroup) {
-        group.isOpen = false;
-      }
-    });
-  }
-
-  removeGroup(group: NgbAccordionGroup): void {
-    const index = this.groups.indexOf(group);
-    if (index !== -1) {
-      this.groups.splice(index, 1);
-    }
-  }
-}
+import {Component, Directive, forwardRef, Inject, Input, Query, QueryList} from 'angular2/angular2';
 
 @Component({
   selector: 'ngb-accordion-group, [ngb-accordion-group]',
-  inputs: ['heading', 'isOpen', 'isDisabled'],
   template: `
     <div class="panel panel-default" [class.panel-open]="isOpen">
       <div class="panel-heading">
         <h4 class="panel-title">
-          <a href tabindex="0"><span [class.text-muted]="isDisabled" (click)="toggleOpen($event)">{{heading}}</span></a>
+          <a href tabindex="0"><span [class.text-muted]="isDisabled" (click)="toggleOpen($event)">{{title}}</span></a>
         </h4>
       </div>
       <div class="panel-collapse" [hidden]="!isOpen">
@@ -45,27 +17,46 @@ export class NgbAccordion {
     </div>
   `
 })
-export class NgbAccordionGroup implements OnDestroy {
-  private isDisabled: boolean;
-  private _isOpen: boolean = false;
+export class NgbAccordionGroup {
+  private _isOpen = false;
+  @Input() isDisabled: boolean;
+  @Input() title: string;
 
-  constructor(private accordion: NgbAccordion) { this.accordion.addGroup(this); }
+  @Input()
+  set isOpen(value: boolean) {
+    this._isOpen = value;
+    if (value) {
+      this.accordion.closeOthers(this);
+    }
+  }
 
-  toggleOpen(event) {
+  get isOpen(): boolean { return this._isOpen; }
+
+  constructor(@Inject(forwardRef(() => NgbAccordion)) private accordion: NgbAccordion) {}
+
+  toggleOpen(event): void {
     event.preventDefault();
     if (!this.isDisabled) {
       this.isOpen = !this.isOpen;
     }
   }
+}
 
-  onDestroy(): void { this.accordion.removeGroup(this); }
+@Directive({selector: 'ngb-accordion, [ngb-accordion]'})
+export class NgbAccordion {
+  @Input('closeOthers') onlyOneOpen: boolean;
 
-  public get isOpen(): boolean { return this._isOpen; }
+  constructor(@Query(NgbAccordionGroup) public groups: QueryList<NgbAccordionGroup>) {}
 
-  public set isOpen(value: boolean) {
-    this._isOpen = value;
-    if (value) {
-      this.accordion.closeOthers(this);
+  closeOthers(openGroup: NgbAccordionGroup): void {
+    if (!this.onlyOneOpen) {
+      return;
     }
+
+    this.groups.toArray().forEach((group: NgbAccordionGroup) => {
+      if (group !== openGroup) {
+        group.isOpen = false;
+      }
+    });
   }
 }
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -22,208 +22,206 @@ function hasClass(element: Element, str: string): Boolean {
   return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.className)
 }
 
-function hasHeading(element: Element, str: string): Boolean {
+function hasTitle(element: Element, str: string): Boolean {
   return element.textContent === str;
 }
 
 describe('ngb-accordion', () => {
-  describe('UI logic', () => {
-    let html: string;
+  let html: string;
 
-    beforeEach(() => {
-      html = `
+  beforeEach(() => {
+    html = `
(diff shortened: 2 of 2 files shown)


wesleycho:
A little trick for reviewing that makes it easier: https://github.com/ng-bootstrap/core/pull/86/files?w=1

src/accordion/accordion.ts
```diff
@@ -45,27 +17,48 @@ export class NgbAccordion {
...
+    if (value) {
+      this.accordion.closeOthers(this);
+    }
+  }
 
-  constructor(private accordion: NgbAccordion) { this.accordion.addGroup(this); }
+  get isOpen(): boolean { return this._isOpen; }
 
-  toggleOpen(event) {
+  constructor(@Inject(forwardRef(() => NgbAccordion)) accordion: NgbAccordion) { this.accordion = accordion; }
```
wesleycho: Do `constructor(@Inject(forwardRef(() => NgbAccordion)) private accordion: NgbAccordion) {}`

src/accordion/accordion.ts
```diff
@@ -45,27 +17,48 @@ export class NgbAccordion {
...
+}
 
-  onDestroy(): void { this.accordion.removeGroup(this); }
+@Directive({selector: 'ngb-accordion, [ngb-accordion]'})
+export class NgbAccordion {
+  @Input('closeOthers') onlyOneOpen: boolean;
+  groups: QueryList<NgbAccordionGroup>;
 
-  public get isOpen(): boolean { return this._isOpen; }
+  constructor(@Query(NgbAccordionGroup) groups: QueryList<NgbAccordionGroup>) { this.groups = groups; }
```
wesleycho: Could just do
```js
constructor(@Query(NgbAccordionGroup) public groups: QueryList<NgbAccordionGroup>) {}
```
I believe, and remove the defining of groups above

wesleycho:
I'm fine with this PR as is - my comments are stylistic, may be not we want if we decide we do not want to take advantage of TypeScript's automatic publishing on the class instance.

src/accordion/accordion.ts
pkozlowski-opensource: `@Query` is going away (https://github.com/angular/angular/issues/4919), we should be using `@ContentChildren` instead (please note that this is a field annotation, not a constructor argument annotation)

src/accordion/accordion.ts
```diff
@@ -45,27 +17,48 @@ export class NgbAccordion {
...
+  private _isOpen = false;
+  private accordion: NgbAccordion;
+  @Input() isDisabled: boolean;
+  @Input() heading: string;
+
+  @Input()
+  set isOpen(value: boolean) {
+    this._isOpen = value;
+    if (value) {
+      this.accordion.closeOthers(this);
```
pkozlowski-opensource: Soooo, this is a circular dependency (accordion -> group and group -> accordion). Ideally we should remove this circle, but I'm not sure how atm... Let me fiddle with different ideas....

pkozlowski-opensource:
So, I really like how the implementation got simplified! There are few detailed comments from @wesleycho and me, but more importantly I would suggest that we do some more thinking on couple of items:

* API - from what I remember a common problem with the our accordion / tabset implementation for ng1 was how we manage `isOpen` (it is more pronounced for tabset, but applies here as well). So currently we've got `isOpen`  on each panel which is pain in a neck if we want to mark a group as opened pragmatically. Essentially we need to create model entry for each and every panel, which is pain. How about exploring ideas of having an open index/ indexes on the accordion level?
* reausability / composition: the way the directive is coded today is that we have an accordion with panels. IMO ideally we should be having a panel / collapse directive that is totally independent from accordion and then accordion that just manages a group of panels. Not sure if it is easy / possible, but I would explore this avenue.
* circular deps between directives: related to the above - ideally we should just have accordion -> panel dependency.

I know that I'm raising topics that might sound like turning things upside down, but IMO we need to explore all the options (in terms of impl and API) before committing to anything.

src/accordion/accordion.ts
```diff
@@ -45,27 +17,48 @@ export class NgbAccordion {
...
+  toggleOpen(event): void {
     event.preventDefault();
     if (!this.isDisabled) {
       this.isOpen = !this.isOpen;
     }
   }
+}
 
-  onDestroy(): void { this.accordion.removeGroup(this); }
+@Directive({selector: 'ngb-accordion, [ngb-accordion]'})
```
pkozlowski-opensource: Shouldn't it be a component with its own template? AFAIK we still need some markup, from the bs4 docs: `<div id="accordion" role="tablist" aria-multiselectable="true">`

src/accordion/accordion.ts
```diff
@@ -45,27 +17,48 @@ export class NgbAccordion {
     </div>
   `
 })
-export class NgbAccordionGroup implements OnDestroy {
-  private isDisabled: boolean;
-  private _isOpen: boolean = false;
+export class NgbAccordionGroup {
```
pkozlowski-opensource: `NgbAccordionGroup` name sounds off. Bootstrap is using "Collapsible Group _Item_" as a name. My suggestion is to stick to `NgbPanel` and make a panel an independent component (not tied to accordion).

src/accordion/accordion.ts
```diff
@@ -45,27 +17,48 @@ export class NgbAccordion {
...
+  private _isOpen = false;
+  private accordion: NgbAccordion;
+  @Input() isDisabled: boolean;
+  @Input() heading: string;
```
pkozlowski-opensource: I think that we should use the `title` here as this is what users can change here. `heading` should be reserved for changing the entire heading (which we don't have yet)

src/accordion/accordion.ts
Foxandxss: Nice, I tried a few permutations but definitely not that one.

pkozlowski-opensource:
One more thing: bs4 is using a combination of `collapse in` classes to close a panel instead of `[hidden]="!isOpen"`

Foxandxss:
It is Pawel, it is. But we decided that a collapse directive wasn't making sense until we have animations. A collapse directive right now is simply a wrapping over `[hidden]`. So when we get animations, I will create that collapse and refactor this, until there, just extra code to maintain.

pkozlowski-opensource:
> But we decided that a collapse directive wasn't making sense until we have animations. 

IMO it still makes sense as it provides toggling functionality with aria support. And more importantly it would allow us to re-use it in the accordion implementation.

pkozlowski-opensource:
Another item: we can't have empty `href` as it is going to trigger page reloads.

Foxandxss:
> IMO it still makes sense as it provides toggling functionality with aria support. And more importantly it would allow us to re-use it in the accordion implementation.

Sure, but that is outside the scope of this PR. We can create a collapse and then refactor this later. Let's not convert this PR in a monster of changes because it will get out of hand quickly. I don't mind writing 20 little PRs to fix it than having a 200 messages discussion in here.

> Another item: we can't have empty href as it is going to trigger page reloads.

What do you suggest? I am not getting reloads btw ( I preventDefault in my code)

pkozlowski-opensource:
>  Let's not convert this PR in a monster of changes because it will get out of hand quickly. I don't mind writing 20 little PRs to fix it than having a 200 messages discussion in here.

OK, let's leave out:
* circular deps
* reusable panel
* API change (selected index on accordion)
* empty href
to separate PRs - it will be easier to discuss indeed. 

Could address other comments and submit so we can merge and start discussions on other topics?

Foxandxss:
Sure mate.

Foxandxss:
Alright mates.

I Changed `heading` for `title`.
Used the constructor shortcuts as Wes suggested.
Instead of using a component with template I used `host` to put the role and aria (ends with a much better HTML than having yet another element).

I wasn't able to do the ContentChildren and it is due the circular dep that Pawel mentioned.

The problem is that ContentChildren takes a while to populate and every time a new group is instantiated, it is calling `closeOthers` in the accordion which is going to fail because `this.groups` is empty yet so `toArray()` fails. (uhhh, I figured out a good way to fix this while explaining, for a new pull request).

So, verify this, merge this, I will add a few extra bullets on #49 and then move on to create another PRs.

src/accordion/accordion.ts
```diff
@@ -45,27 +17,46 @@ export class NgbAccordion {
...
+  toggleOpen(event): void {
     event.preventDefault();
     if (!this.isDisabled) {
       this.isOpen = !this.isOpen;
     }
   }
+}
 
-  onDestroy(): void { this.accordion.removeGroup(this); }
+@Directive({selector: 'ngb-accordion, [ngb-accordion]', host: {'role': 'tablist', 'aria-multiselectable': 'true'}})
```
pkozlowski-opensource: `'aria-multiselectable': 'true'` shouldn't be hard-coded, I guess you've meant `[attr.aria-multiselectable]: !onlyOneOpen` or sth. BTW, maybe we should rename the var name to `multiselectable`?

pkozlowski-opensource:
Left one more comment for `'aria-multiselectable': 'true'`.

Otherwise it LGTM - let's get it in and tackle subsequent items for this directive.

Foxandxss:
I removed the accessibility thing entirely Pawel. It is something I have in #49 as a TODO and it is not simply multiselectable, it also needs expanded and more stuff and well, accessibility is another PR entirely.

pkozlowski-opensource:
OK, let's merge it and tackle remaining items one by one.

PR #92, opened 2015-11-12
refactor(accordion): remove attribute selector

Foxandxss:
Better to leave it as a element, it doesn't make sense as an attribute.


pkozlowski-opensource:
LGTM if Travis is green

PR #94, opened 2015-11-12
chore(docs): update readme file

icfantv:
* add new sections using UIBS project as guide
* add stubs for sections with no content
* update doc from meeting minutes

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,6 @@ This project is being created by the [ui-bootstrap team](https://github.com/angu
     - [Code of Conduct](#code-of-conduct)
     - [Supported browsers](#supported-browsers)
 - [Contributing to the project](#contributing-to-the-project)
-- [Development, meeting minutes, roadmap and more.](#development-meeting-minutes-roadmap-and-more)
 
 # Demo
 
@@ -45,7 +44,3 @@ See [this](https://github.com/angular/angular/blob/master/README.md) for up-to-d
 ## Contributing to the project
 
 We are always looking for the quality contributions! Please check the [CONTRIBUTING.md](CONTRIBUTING.md) doc for contribution guidelines.
-
-## Development, meeting minutes, roadmap and more.
-
-We support TypeScript and ES5 using the CJS module system.
--- a/README.md
+++ b/README.md
@@ -2,12 +2,113 @@
 
 [![Build Status](https://travis-ci.org/ng-bootstrap/core.svg?branch=master)](https://travis-ci.org/ng-bootstrap/core)
 
-This library is a work in progress. Please check out our list of [issues](https://github.com/ng-bootstrap/core/issues) to see all the things we are implementing. Feel free to make any comment there ;)
+Welcome to the Angular 2 version of the []Angular UI Bootstrap](https://github.com/angular-ui/bootstrap) library.
+This library is being built from scratch in TypeScript using the Bootstrap 4 CSS framework.
 
-As for demos, we have a list of Plunkers with our components:
+As with Bootstrap 4 and Angular 2, this library is a work in progress. Please check out our list of [issues](https://github.com/ng-bootstrap/core/issues)
+to see all the things we are implementing. Feel free to make comments there.
+
+
+This project is being created by the [ui-bootstrap team](https://github.com/angular-ui/bootstrap).
+
+### Quick links
+- [Demo](#demo)
+- [Installation](#installation)
+    - [NPM](#install-with-npm)
+    - [Bower](#install-with-bower)
+    - [NuGet](#install-with-nuget)
+    - [Custom](#custom-build)
+    - [Manual](#manual-download)
+- [Support](#support)
+    - [FAQ](#faq)
+    - [Code of Conduct](#code-of-conduct)
+    - [PREFIX MIGRATION GUIDE](#prefix-migration-guide)
+    - [Supported browsers](#supported-browsers)
+    - [Need help?](#need-help)
+    - [Found a bug?](#found-a-bug)
+- [Contributing to the project](#contributing-to-the-project)
+- [Development, meeting minutes, roadmap and more.](#development-meeting-minutes-roadmap-and-more)
+
+# Demo
 
 * [Accordion](http://plnkr.co/edit/XuH7mXBycODO7KYgNSbH)
 * [Dropdown](http://plnkr.co/edit/8FNKrEURwSzqqtsK5Rxn?p=preview)
 * [Rating](http://plnkr.co/edit/2wtcZ5bmQk8xmvwWvFP6?p=preview)
 
-This project is being created by the [ui-bootstrap team](https://github.com/angular-ui/bootstrap).
+# Installation
+
+TBD
+
+#### Install with NPM
+
+TBD
+
+#### Install with Bower
+
+TBD
+
+#### Install with NuGet
+
+TBD
+
+#### Custom build
+
+TBD
+
+#### Manual download
+
+TBD
+
+### Adding dependency to your project
+
+TBD
+
+# Support
+
+## FAQ
+
+TBD
+
+## Code of Conduct
+
+Take a moment to read or [Code of Conduct](CODE_OF_CONDUCT.md)
+
+## Supported browsers
+
+As Bootstrap 4 and Angular 2 are still in active development this list of supported browsers is subject to
+change. It is our intention to support the same browsers and versions supported by both Bootstrap 4 and
+Angular 2, which ever is the _most_ restrictive.
+
+As of this writing, Angular 2 browser support is unknown, but Bootstrap's are:
+
+* Chrome (stable and canary channel)
+* Firefox
+* IE 9+ (but see [Bootstrap 4's notes](http://v4-alpha.getbootstrap.com/getting-started/browsers-devices/#internet-explorer-9) on IE9)
+* Opera
+* Safari
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,12 +2,113 @@
...
-This project is being created by the [ui-bootstrap team](https://github.com/angular-ui/bootstrap).
+# Installation
+
+TBD
+
+#### Install with NPM
+
+TBD
+
+#### Install with Bower
```
Foxandxss: No bower.

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+TBD
+
+#### Install with NuGet
```
Foxandxss: Hopefully not this either.

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+TBD
+
+#### Custom build
```
Foxandxss: No custom builds.

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+TBD
+
+#### Manual download
+
+TBD
+
+### Adding dependency to your project
```
Foxandxss: Not this either.

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+As of this writing, Angular 2 browser support is unknown, but Bootstrap's are:
+
+* Chrome (stable and canary channel)
+* Firefox
+* IE 9+ (but see [Bootstrap 4's notes](http://v4-alpha.getbootstrap.com/getting-started/browsers-devices/#internet-explorer-9) on IE9)
+* Opera
+* Safari
+
+While modern mobile browsers should work without problems, we have made certain design decisions surrounding
```
Foxandxss: Please not this.

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+# Support
+
+## FAQ
+
+TBD
+
+## Code of Conduct
+
+Take a moment to read or [Code of Conduct](CODE_OF_CONDUCT.md)
```
Foxandxss: our

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+## Supported browsers
+
+As Bootstrap 4 and Angular 2 are still in active development this list of supported browsers is subject to
+change. It is our intention to support the same browsers and versions supported by both Bootstrap 4 and
+Angular 2, which ever is the _most_ restrictive.
+
+As of this writing, Angular 2 browser support is unknown, but Bootstrap's are:
```
Foxandxss: They are known.

pkozlowski-opensource:
@icfantv I'm not big fan of leaving placeholders for things we _might_ add in the future. We can add content when we know what to add but I'm not sure we want to create placeholders for those.

Foxandxss:
I totally agree with you Pawel. The original idea was to just add that we want to support X browsers and bootstrap 4. Nothing else, no placeholders.

We discussed whether or not to add stuff like "We are not going to support tooltips and popovers" and my response was no, because we have no idea how is this going to evolve, and you can be surprised.

I vote to just add the support browsers and bootstrap version bits and even so, it can change in the future I guess. For the rest of things, we will add more sections to the readme when they are needed.

icfantv:
@pkozlowski-opensource, funny you say that.  I was telling @Foxandxss that we should add the tooltip/popover bit on modals to set expectations and it never occurred to me that adding placeholders could, even indirectly, set expectations.  I agree with you both there and will remove them.

I'd like another set of eyes on that bit for supporting tooltips/popovers on modals.  I think we either support it or we don't.  We've already made the decision in the UIBS project to **NOT** support them so I don't know that it makes sense to do so in this project.  But we should discuss.  @wesleycho, can you please also chime in?  Thanks.

pkozlowski-opensource:
> I'd like another set of eyes on that bit for supporting tooltips/popovers on modals. I think we either support it or we don't. We've already made the decision in the UIBS project to NOT support them so I don't know that it makes sense to do so in this project.

Are those supported in the original bootstrap?

README.md
wesleycho: They are not official yet - I think this might need some chiming in from @pkozlowski-opensource

Foxandxss:
The problem Adam is that Angular 2 with webcomponents is a box of surprises. What in the past wasn't possible, now it is possible.

Not saying that tooltip/popover on mobiles (you mean mobiles and not modals, right?) is now a possibility, but we shouldn't close doors when we don't have a popover/tooltip yet.

README.md
wesleycho: This whole passage is unnecessary - remove it.

README.md
Foxandxss: Well, their README has some kind of browser support :P

wesleycho:
I was not aware that we did not support tooltips and popovers in modals in UI Bootstrap. Was there a particular technical limitation there?

README.md
pkozlowski-opensource: They _are_ known and _officially_ - things were announced during AngularConnect (keynote). You can see a list of browsers to be supported when you go to the main angular repo (SauceLabs badge)

icfantv:
Gah!  Sorry.  I meant "mobile" not modal.  Will fix.

icfantv:
@pkozlowski-opensource, see above comment.  Sorry for the confusion.  @Foxandxss, remember, ours was a UX decision, not an architecture one.  That is, we made it because we thought it was poor UX to have tooltips and popovers on mobile devices not because we couldn't do it from a code perspective.

README.md
```diff
@@ -2,12 +2,113 @@
...
+
+#### Install with NuGet
+
+TBD
+
+#### Custom build
+
+TBD
+
+#### Manual download
```
wesleycho: Do we want this section too? The only reason I can think of to support this is for Plunkers and such, but we should strongly adopt npm as the main mechanism for consuming the library.

pkozlowski-opensource:
@icfantv ok. So, I wouldn't worry about anything mobile at this point - it is way too early as we don't know how we are going to code tooltips / popovers nor what mobile support we are going to have from ng2.

icfantv:
@pkozlowski-opensource.  ok.  I'll take it out.  I do feel like we should have a conversation eventually about cross-project consistency though.  Changes in architecture and what we can support is one thing, but design decisions made around UX (e.g., only supporting one dropdown open at a time), is another and shouldn't change just because the architecture does.  Right?

wesleycho:
It's too early to mention anything specific IMO.

Foxandxss:
As I said yesterday. We shouldn't get anything from ui-bootstrap for granted. We should question **EVERYTHING** we did in there and do it better now.

README.md
icfantv: Sorry, I was going by our meeting notes from early October which is why I missed it.

icfantv:
Ok fellas.  Rev 2 is posted for your perusal.  I should note that several hours of therapy were necessary after the last review.

README.md
```diff
@@ -1,13 +1,51 @@
...
+* Firefox (40+)
+* IE (9+) (but see [Bootstrap 4's notes](http://v4-alpha.getbootstrap.com/getting-started/browsers-devices/#internet-explorer-9) on IE9)
+* Edge (20+)
+* Safari (7+)
+
+## Contributing to the project
+
+We are always looking for the quality contributions! Please check the [CONTRIBUTING.md](CONTRIBUTING.md) doc for contribution guidelines.
+
+## Development, meeting minutes, roadmap and more.
```
Foxandxss: I don't think this adds any value.

icfantv: Sorry, the comments in the diff aren't clear.  Which line/section are you referring to?

Foxandxss: The Developmeng, meeting minutes section.

icfantv: Ok.  Will remove.

README.md
```diff
@@ -1,13 +1,46 @@
...
 
-As for demos, we have a list of Plunkers with our components:
+As with Bootstrap 4 and Angular 2, this library is a work in progress. Please check out our list of
+[issues](https://github.com/ng-bootstrap/core/issues) to see all the things we are implementing.
+Feel free to make comments there.
+
+This project is being created by the [ui-bootstrap team](https://github.com/angular-ui/bootstrap).
+
+# Quick links
+- [Demo](#demo)
```
Foxandxss: space in here.

icfantv: sorry, where?   what space?

Foxandxss: Between #quick links and the first entry of the TOC

README.md
```diff
@@ -1,13 +1,46 @@
...
-This project is being created by the [ui-bootstrap team](https://github.com/angular-ui/bootstrap).
+# Support
+
+## Code of Conduct
+
+Please take a moment and read our [Code of Conduct](CODE_OF_CONDUCT.md)
+
+## Supported browsers
+
+We support the same browsers and versions supported by both Bootstrap 4 and Angular 2, which ever is _more_ restrictive.
```
Foxandxss: whichever

README.md
```diff
@@ -1,13 +1,47 @@
-### NG Bootstrap - [Angular 2](http://angular.io/) directives specific to [Bootstrap](http://getbootstrap.com)
+# NG Bootstrap - [Angular 2](http://angular.io/) directives specific to [Bootstrap](http://getbootstrap.com)
```
sirajc: For Bootstrap4, we should link to http://v4-alpha.getbootstrap.com/

Issue #11, opened 2015-09-15
Collapse


BobbieBarker:
Recreate BS3 Collapse with ng2 component

PR #96, opened 2015-11-15
feat(collapse): add initial collapse

sirajc:
Initial implementation of collapse. Please advise for changes required. Will add a plunker for this.
Usage `<div [ngb-collapse]="collapsed">` where `collapsed` is a boolean on container component

### DIFF
--- a/src/collapse/collapse.spec.ts
+++ b/src/collapse/collapse.spec.ts
@@ -1,0 +1,118 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  injectAsync,
+  TestComponentBuilder,
+  beforeEachProviders
+} from 'angular2/testing';
+
+import {Component} from 'angular2/angular2';
+
+import {NgbCollapse} from './collapse';
+
+function getCollapsibleContent(element: Element): Element {
+  return element.querySelector('.collapse');
+}
+
+function getButton(element) {
+  return element.querySelector('button');
+}
+
+
+function hasClass(element: Element, str: string): Boolean {
+  return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.className);
+}
+
+function hasStyle(element: Element, str: string): Boolean {
+  return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.getAttribute('style'));
+}
+
+describe('ngb-collapse', () => {
+  let html: string;
+
+  beforeEach(() => {
+    html = `
+      <button type="button" (click)="collapsed = !collapsed">Toggle collapse</button>
+      <div [ngb-collapse]="collapsed">
+        <div class="card">
+          <div class="card-block">
+            Some content
+          </div>
+        </div>
+      </div>
+    `;
+  });
+
+  it('should have content open and aria-expanded true', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         const compiled = fixture.debugElement.nativeElement;
+
+         let content = getCollapsibleContent(compiled);
+
+         expect(hasClass(content, 'in')).toBe(true);
+         expect(content.getAttribute('aria-expanded') === "true").toBe(true);
+       });
+     }));
+
+  it('should have content closed  and aria-expanded false', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         const tc = fixture.debugElement.componentInstance;
+         tc.collapsed = true;
+         fixture.detectChanges();
+
+         const compiled = fixture.debugElement.nativeElement;
+
+         let content = getCollapsibleContent(compiled);
+
+         expect(hasClass(content, 'in')).toBe(false);
+         expect(content.getAttribute('aria-expanded') === "false").toBe(true);
+       });
+     }));
+
+  it('should have height style of 0px', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         const tc = fixture.debugElement.componentInstance;
+         tc.collapsed = true;
+         fixture.detectChanges();
+
+         const compiled = fixture.debugElement.nativeElement;
+
+         let content = getCollapsibleContent(compiled);
+
+         expect(hasStyle(content, 'height: 0px;')).toBe(true);
+       });
+     }));
+
+  it('should toggle collapsed content on click', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+         const compiled = fixture.debugElement.nativeElement;
+         let content = getCollapsibleContent(compiled);
+         expect(hasClass(content, 'in')).toBe(true);
+
+         let buttonEl = getButton(compiled);
+
+         buttonEl.click();
+         fixture.detectChanges();
+         expect(hasClass(content, 'in')).toBe(false);
+
+         buttonEl.click();
+         fixture.detectChanges();
+         expect(hasClass(content, 'in')).toBe(true);
+
+       });
+     }));
+
+
+});
+
+@Component({selector: 'test-cmp', directives: [NgbCollapse], template: ''})
+class TestComponent {
+  collapsed = false;
+}
--- a/src/collapse/collapse.spec.ts
+++ b/src/collapse/collapse.spec.ts
@@ -26,10 +26,6 @@ function hasClass(element: Element, str: string): Boolean {
   return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.className);
 }
 
-function hasStyle(element: Element, str: string): Boolean {
-  return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.getAttribute('style'));
-}
-
 describe('ngb-collapse', () => {
   let html: string;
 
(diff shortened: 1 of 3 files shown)


Foxandxss:
Thank you for the PR, but we decided not to create a collapse yet because there is no animations. The only gain is having a aria attribute.

On the other hand, we are trying to focus on a few components at once so this doesn't get out of hand quickly.

Still, high quality PR, I give you that.

Will keep it on hold until we decide what to do with collapse.

Thanks.

Foxandxss:
I suggest that you create an issue (or participate on the issues that are already created) and inform that you want to work on something. That way you don't waste your time writing PRs that could not be merged and also, you don't step on other person work. One of our team members was already assigned to do this collapse (when the time was right) and well, you got here first.

sirajc:
I added the plunker for this http://plnkr.co/edit/SW68aAIMP6iwVtNUZZis?p=preview 
I have requested for Slack login to start with discussion, will update/create issues henceforth to pick up something.
Also, I wanted to get my feet wet with this ecosystem, before picking up sth complex, and collapse was easy piece.
Thanks

Foxandxss:
Still, everything I said applies. Ask before doing anything to not step in other works. Collapse was already assigned to someone in our team.

sirajc:
:+1:

pkozlowski-opensource:
> but we decided not to create a collapse yet because there is no animations.

@Foxandxss IMO this is still open to debate. I would _love_ to have collapse, even without animations.

Foxandxss:
@pkozlowski-opensource Wes and I decided not to, but it is open to debate. I still don't see the value on it, it does nothing for us just yet.

pkozlowski-opensource:
> Wes and I decided not to, but it is open to debate. 

If it is open to debate, I disagree with this decision. 

> I still don't see the value on it, it does nothing for us just yet

I'm afraid that I've got a different opinion here:
* aria support
* semantic HTML
* can be re-used in accordion
* we _will_ have animations

The most important argument, IMO, is that we could compose accordion from smaller pieces (collapse). I see it as a _good thing_ when we build higher-order directives from smaller building blocks.

Foxandxss:
Alright. I normally would say no to create a directive just for the sake of adding one aria attribute. [hidden] still does the work and is semantically pretty.

But since we are going to need a collapse yes or yes, I see no point on discussing that.

I am going to review this PR, but still I ask again to @sirajc to inform on what he want to do before he step on someones work.

src/core.ts
```diff
@@ -1,8 +1,10 @@
 import {NgbAccordion, NgbAccordionGroup} from './accordion/accordion';
+import {NgbCollapse} from './collapse/collapse';
 import {NgbDropdown} from './dropdown/dropdown';
 import {NgbRating} from './rating/rating';
 
 export {NgbAccordion, NgbAccordionGroup} from './accordion/accordion';
+export {NgbCollapse} from './collapse/collapse';
 export {NgbDropdown} from './dropdown/dropdown';
 export {NgbRating} from './rating/rating';
```
Foxandxss: Also add it to the NGB_DIRECTIVES

src/collapse/collapse.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+         const compiled = fixture.debugElement.nativeElement;
+
+         let content = getCollapsibleContent(compiled);
+
+         expect(hasClass(content, 'in')).toBe(true);
+         expect(content.getAttribute('aria-expanded') === "true").toBe(true);
+       });
+     }));
+
+  it('should have content closed  and aria-expanded false', injectAsync([TestComponentBuilder], (tcb) => {
```
Foxandxss: extra space in there.

src/collapse/collapse.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+         expect(hasClass(content, 'in')).toBe(false);
+
+         buttonEl.click();
+         fixture.detectChanges();
+         expect(hasClass(content, 'in')).toBe(true);
+
+       });
+     }));
+
+
```
Foxandxss: remove this two empty lines

src/collapse/collapse.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+         let buttonEl = getButton(compiled);
+
+         buttonEl.click();
+         fixture.detectChanges();
+         expect(hasClass(content, 'in')).toBe(false);
+
+         buttonEl.click();
+         fixture.detectChanges();
+         expect(hasClass(content, 'in')).toBe(true);
+
```
Foxandxss: remove this empty line

Foxandxss:
About the directive itself, take a peek into the conventions in the wiki, you have a few things to change in there.

sirajc:
@Foxandxss, I agree that I need to be careful before starting any work, won't jump in directly henceforth. I just saw the label PR's plz on issue #11 and picked it up, didn't notice that it was assinged to BobbieBarker, but after you mentioned. Apologies
Also, Will incorporate the review comments as suggested.

Foxandxss:
Don't worry mate. In that case is not your fault.

Still, we have lots of internal talks, meetings and we take some decisions. We are trying not to rush to get as much components as possible right now because Angular 2 is not beta yet and we would like to do things as better as possible. Angular 2 comes with lot of changes and we are still trying to learn.

So in short, we are trying not to pull more components for the time being, and try to do make right the ones we have and then the others are pretty much the same.

That being said, I can see how the collapse can be useful (in the long term) so fix those issues and I will merge it.

sirajc:
Done with suggested changes, @Foxandxss please review, have also updated the plnkr.

PS: This is my first PR in a public repo, would like to see this merged and keep learning from you all. thnx

src/collapse/collapse.ts
```diff
@@ -0,0 +1,28 @@
+import {Directive, Input} from 'angular2/angular2';
+
+@Directive({
+  selector: '[ngb-collapse]',
+  host: {'class': 'collapse', '[class.in]': '_open', '[attr.aria-expanded]': '_open', '[style.height.px]': '_height'}
+})
+export class NgbCollapse {
+  private _open: boolean;
+  private _height: number;
```
Foxandxss: Since this goes into the template, it shouldn't be private nor have a underscore.

Foxandxss:
Sure, I will keep checking it a bit later and when it is ready, I will merge it, don't worry.

src/collapse/collapse.ts
sirajc: BS4 adds height only when content is collapsed, and so I have added it accordingly, hence this is private

Foxandxss: The problem is that in angular 2 you can do stuff like:

```
<div [ngb-collapse]="foo" #collapse>
...
</div>
{{ collapse.height }}
```

That being said and after I explained this to you, perhaps `height` makes more sense as private. I will leave that to discussion of my fellow team members. For now, leave it as is.

src/collapse/collapse.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  @Input()
+  set ngbCollapse(value: boolean) {
+    this._open = !value;
+    this._setHeight()
+  }
+
+  get ngbCollapse(): boolean { return this._open; }
+
+  private _setHeight() {
```
Foxandxss: this returns "void"

src/collapse/collapse.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (this._open === true) {
```
Foxandxss: There is no need for the `=== true`. Also, you can use a ternary in here.

Foxandxss:
Apart from what I noted. I would rename `open` for `collapsed`. It doesn't make too much sense to talk about collapsing and having the internal variable as `open`.

That being said, I would do something like:

```typescript
@Input('ngbCollapse')
set collapsed(value: boolean) { ... }
```

also rename the get and use it in the template.

On the other hand, you need an output for the `collapsed` variable.

src/collapse/collapse.ts
```diff
@@ -0,0 +1,28 @@
+import {Directive, Input} from 'angular2/angular2';
+
+@Directive({
+  selector: '[ngb-collapse]',
```
pkozlowski-opensource: Hmm, I wonder why we would prefer attribute vs. element here. Somehow I was assuming:

```html
<ngb-collapse [collapsed]="condition">
   Content to collapse goes here...
</ngb-collapse>
```

but honestly I'm not sure _why_ I would lean towards this setup. Thoughts?
```

File: src/collapse/collapse.ts [hunk 9]
```diff
@@ -0,0 +1,28 @@
...
+  host: {'class': 'collapse', '[class.in]': '_open', '[attr.aria-expanded]': '_open', '[style.height.px]': '_height'}
```
pkozlowski-opensource: Personally I would drop any code that deals with `height` for now, as we are not calculating / making any use of it for now.

src/collapse/collapse.ts
Foxandxss: Because we can make everything collapsible without a wrapper element

src/collapse/collapse.ts
Foxandxss: I am in.

sirajc:
> On the other hand, you need an output for the collapsed variable.

Not sure what is expected here, details please

Foxandxss:
Sure, you need an `@Output` that is a EventEmitter. Check the rating directive.

pkozlowski-opensource:
@Foxandxss I'm not sure output is that much useful in this particular case. Could we keep it simple for now?

Foxandxss:
If you want any other component to react to collapse changes, it is useful.

pkozlowski-opensource:
> If you want any other component to react to collapse changes, it is useful.

I'm not saying it is not useful. I would prefer that we leave this out for now and come back to this when integrating collapse into accordion.

Foxandxss:
That is exactly my next step, integrate this into the accordion.

Foxandxss:
Ok @sirajc. Sorry for this back and forward. The remaining things to do in here:

* Remove the `height` stuff for now, it is not adding value.
* Use `collapsed` directive on the `host` part (you should always use a getter or setter if available instead of using the inner variable directly.
* Update tests accordingly.

I will merge it tomorrow.

sirajc:
> Remove the height stuff for now, it is not adding value.

Agreed that height will be helpful when animations come into play, can remove now. But after removing this no logic remains inside setter, we strip down this class to 
```
export class NgbCollapse {
  @Input('ngbCollapse') collapsed: boolean;
}
```
or should I keep setter/getter as we will adding stuff later.

Also, I tried adding output `@Output() collapse: EventEmitter<boolean> = new EventEmitter();`
and in setter `this.collapse.next(value);` it works locally, can add to this PR or later

@Foxandxss , @pkozlowski-opensource let me know your thoughts

Foxandxss:
Yeah, no need of set/get for now. The output, we will add it when it gets needed, don't worry.

sirajc:
> this back and forward

@Foxandxss I am open to discussion and changes. Design decisions are more critical then hurrying up on the feature

Foxandxss:
Yes, don't worry. We are going to make this simple, merge it and move on to other stuff. Collapse will get updated when animation arrives, we don't know when is that going to happen yet.

Foxandxss:
Okay, I tuned the tests a bit and I merged it. Thanks. And remember, post an issue / write on existing issues if you want to tackle another thing.

PR #97, opened 2015-11-17
Ddropdown tests

pkozlowski-opensource:
Some tests changes before I start adding more functionality / tests.

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -26,9 +26,7 @@ describe('ngb-dropdown', () => {
          fixture.detectChanges();
          const compiled = fixture.debugElement.nativeElement;
 
-         let dropdownEl = getDropdownEl(compiled);
-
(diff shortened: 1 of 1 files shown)


Foxandxss:
Where does that `toHaveCssClass` comes from? Is that angular 2 doing? Never saw it in pure jasmine.

On the other hand, don't you have the .git hook for the commits? (Weird PR title in there :P)

On the other hand, good for me.

Foxandxss:
Oh, it definitely is. That is good. Poor @wesleycho was fighting with some regex and we had that.

pkozlowski-opensource:
>Where does that toHaveCssClass comes from? Is that angular 2 doing? Never saw it in pure jasmine.

https://github.com/angular/angular/blob/master/modules/angular2/src/testing/matchers.ts

> On the other hand, don't you have the .git hook for the commits? (Weird PR title in there :P)

This is just PR title since there are 2 commits in this PR. Individual commit messages are correct (as far as I can tell).

PR #98, opened 2015-11-17
feat(dropdown): support dropdown and dropups

pkozlowski-opensource:
Here is a proposal to drop the automatic addition of the `dropdown` so we can support dropups as well: http://v4-alpha.getbootstrap.com/components/button-dropdown/#dropup-variation

An alternative is to introduce a switch attribute which boils down to:

```html
<ngb-dropdown class="dropdown">.../<ngb-dropdown> 
```

vs.

```html
<ngb-dropdown direction="down">.../<ngb-dropdown> 
```

Thoughts?

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -1,6 +1,6 @@
 import {Directive, Input} from 'angular2/angular2';
 
-@Directive({selector: 'ngb-dropdown', exportAs: 'ngbDropdown', host: {'class': 'dropdown', '[class.open]': 'open'}})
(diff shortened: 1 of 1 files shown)


Foxandxss:
That kind of stuff doesn't need a position service?

pkozlowski-opensource:
> That kind of stuff doesn't need a position service?

Doesn't seem to be the case in BS4. It might need a position service for more involved use-cases, but seem to work out of the box for down / up variants.

wesleycho:
Dropdown is a little simpler since the dropdown/dropup takes the entire width & doesn't have corner positioning.

We probably want an auto position setting down the road though.

wesleycho:
This sounds good to me.

PR #103, opened 2015-11-17
feat(dropdown): allow toggling dropdown from outside

pkozlowski-opensource:
It covers a common use case where people want to toggle things programatically. In ng1 one would create an associated service to manage dropdowns / tooltips etc. With ng2 we've got much better story as we can simply expose a component instance and allow people to write code to toggle things at will.

Check the test for more details.

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -63,6 +63,27 @@ describe('ngb-dropdown', () => {
          expect(dropdownEl).not.toHaveCssClass('open');
        });
      }));
+
+  it('should allow toggling dropdown from outside', injectAsync([TestComponentBuilder], (tcb) => {
+       const html = `
+      <button (click)="drop.open = !drop.open">Toggle</button>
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.spec.ts
```diff
@@ -63,6 +63,27 @@ describe('ngb-dropdown', () => {
...
+  it('should allow toggling dropdown from outside', injectAsync([TestComponentBuilder], (tcb) => {
+       const html = `
+      <button (click)="drop.open = !drop.open">Toggle</button>
+      <ngb-dropdown #drop="ngbDropdown"></ngb-dropdown>`;
+
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+         const compiled = fixture.debugElement.nativeElement;
+         let dropdownEl = getDropdownEl(compiled);
+         let buttonEl = compiled.querySelector('button');
```
Foxandxss: this should be `const` and perhaps we should have a convention discussion around this too :)

Foxandxss:
LGTM (apart from the const thing)

PR #104, opened 2015-11-17
refactor(accordion): use ngb-collapse

Foxandxss:
Realized also that tests in here needs work, but that is a separate PR.


pkozlowski-opensource:
LGTM

Issue #105, opened 2015-11-17
External helpers


Foxandxss:
So I am in the need of having a external file for helpers. @wesleycho created a `hasClass` which is useful for example on `rating` where I am using `classList` (which is not IE9 compatible).

The thing is... how?

If I create it for example in `misc/test-lib/functions.ts` I have to do in `rating.spec.ts`:

```typescript
import {hasClass} from '../../misc/test-lib/functions
```

IDE agrees on that, but after it gets transpiled into `.js`, it is not going to work anymore (the CJS will try to require a .ts file).

So... transpile it? At first I thought about transpiling all of that into `temp/`, but the problem is that the paths will change from the original `spec.ts` file and the one that gets into `temp`.

So if I have to do:

```typescript
import {hasClass} from '../../misc/test-lib/functions
```

In the `temp` folder I would need to require it from `../../functions.js` or `../../test-lib/functions.js`. In both options, path changes from original to transpiled.

I could leave helpers in plain javascript and tune karma to pick them. By coincidence, the path won't change in this case, because both original test files and transpiled one are two folder deep so it will always be `../../misc/test-lib/functions.js` ...

Problem with that is that typescript is gonna need definition files or perhaps simply complain...

What to do... What to do... Thoughts?

Foxandxss:
Yet another task / watch for helpers that will transpile it in the same exact folder could work, but is that pretty?

wesleycho:
IMO it is better to just do the `../../` - by changing the file path, it adds confusion & isn't as clear what is going on.

pkozlowski-opensource:
We've got https://github.com/ng-bootstrap/core/blob/master/src/util/util.ts for now so let's stick helpers in there for now. Will see how things evolve.

PR #106, opened 2015-11-17
test(accordion): simplify the panel-open logic

Foxandxss:
I think that now it is a bit more legible. We are doing that with the `rating` already and sounds good to me.

I would like to extract `hasClass` to a helper, but that is for #105

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -18,7 +18,15 @@ function getPanels(element: Element): Element[] {
   return Array.from(element.querySelectorAll('ngb-accordion-group .panel'));
 }
 
-function hasTitle(element: Element, str: string): Boolean {
+function getOpenState(panels: Element[]): boolean[] {
+  return panels.map((panel) => { return hasClass(panel, 'panel-open'); });
+}
+
+function hasClass(element: Element, str: string): boolean {
+  return new RegExp(`(^|\\s)${str}(\\s|$)`).test(element.className)
+}
+
+function hasTitle(element: Element, str: string): boolean {
   return element.textContent === str;
 }
 
@@ -55,7 +63,7 @@ describe('ngb-accordion', () => {
 
          let panels = getPanels(compiled);
 
-         panels.forEach((panel) => expect(panel).not.toHaveCssClass('panel-open'));
+         expect(getOpenState(panels)).toEqual([false, false, false]);
        });
      }));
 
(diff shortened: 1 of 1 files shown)


wesleycho:
Question - I thought ng2 is providing a nice matcher for the has class stuff?

src/accordion/accordion.spec.ts
```diff
@@ -18,7 +18,15 @@ function getPanels(element: Element): Element[] {
   return Array.from(element.querySelectorAll('ngb-accordion-group .panel'));
 }
 
-function hasTitle(element: Element, str: string): Boolean {
+function getOpenState(panels: Element[]): boolean[] {
+  return panels.map((panel) => { return hasClass(panel, 'panel-open'); });
```
wesleycho: Any reason this couldn't be done like
```js
  return panels.map((panel) => { return panel.toHaveCssClass('panel-open'); });
```

Foxandxss: Yes. `toHaveCssClass` is a matcher (AKA usable with `expect`) not a method on `Element`.

Foxandxss:
It does wes, but I don't think `expect` returns something so I cannot map the result of X expect into a boolean array. So I had to use `hasClass`.

wesleycho:
Only thing I wonder here is whether it might be better to do the expectation in the helper function, i.e.
```js
function checkPanelsClasses(panels, classes) {
  panels.forEach((panel, idx) => expect(hasClass(panel, classes[idx])));
}
```
But I am fine with this as is.

Foxandxss:
Im still fine as it is.

The point about readability that Pawel was worried about, I don't see it.

If you have:

```
expect(panels[0]).toHaveCssClass('panel-open');
expect(panels[1]).toHaveCssClass('panel-open');
expect(panels[2]).not.toHaveCssClass('panel-open');
```

Your eyes are going to focus on:



so mentally you map that to [true, true, false].

I think it is better to simply read:

```
expect(getOpenState(panels)).toEqual([true, true, false]);
```

No need to focus on the existence or not of a `.not.`. You can simply read... `true, true, false' and easily see what you mean.

> Only thing I wonder here is whether it might be better to do the expectation in the helper function, i.e.

I prefer to keep the `expect` on the test, it makes it more clear than doing a:

```
// test stuff
checkPanelsClasses(panels, [true, true, false]);
```

That is clear, but doesn't tell you if that is doing expectations or something else.

PR #107, opened 2015-11-18
chore(build): make test and tdd single gulp commands

### LINKED ISSUE #99: Testing workflows
I'm very much used to 2 testing workflows where I've got 2 separate commands (gulp or whatever):
* `gulp tdd` - do whatever compilation / preparation, start Karma in watch mode. Re-compile code /tests on each change and re-run Karma tests
* `gulp test` - do whatever compilation / preparation, start Karma in single-run mode and exit

After quick look at the current gulp file I don't see commands corresponding to those 2 workflows.

How do you guys work on tests? What is your workflow? Do we need to add / change anything in our build?

pkozlowski-opensource:
This is far from perfect but at least gives us `test` and `tdd` commands that work as described in #99 and are single commands that can be run in a single terminal.

Closes #99

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -53,10 +53,12 @@ gulp.task('build-tests', function() {
   return tsResult.js.pipe(sourcemaps.write('.')).pipe(gulp.dest('temp'));
 });
 
+gulp.task('clean-build-tests', function(done) { runSequence('clean:tests', 'build-tests', done); });
+
 gulp.task(
     'ddescribe-iit', function() { return gulp.src(PATHS.specs).pipe(ddescribeIit({allowDisabledTests: false})); });
 
-gulp.task('test', function(done) {
+gulp.task('test', ['clean-build-tests'], function(done) {
   var karmaServer = require('karma').Server;
   var travis = process.env.TRAVIS;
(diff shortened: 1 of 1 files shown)


Foxandxss:
It certainly works. I will check it better latter because I gotta run.

So the secret was to put a watch inside the task. That is clever.

I don't think I have any issue at all, so you can merge it.

I don't think I have any problem with it

gulpfile.js
```diff
@@ -70,7 +72,25 @@ gulp.task('test', function(done) {
   new karmaServer(config, done).start();
 });
 
-gulp.task('watch', ['clean:tests', 'build-tests'], function() { gulp.watch(PATHS.src, ['build-tests']); });
+gulp.task('tdd', ['clean-build-tests'], function(done) {
+  var karmaServer = require('karma').Server;
+  var travis = process.env.TRAVIS;
+
+  var config = {configFile: __dirname + '/karma.conf.js', singleRun: false, autoWatch: true};
```
Foxandxss: I don't think this option parameters are needed. They exist on the karma conf already. (singleRun, autoWatch).

PR #108, opened 2015-11-18
refactor(accordion): rename NgbAccordionGroup to NgbAccordionPanel

pkozlowski-opensource:
Group is not a great name since we are not grouping anything here. The NgbAccordionPanel is close to the actual css class names used (and funnily enough we were already referring to those things as "panels" in tests)

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -12,10 +12,10 @@ import {
 
 import {Component} from 'angular2/angular2';
 
-import {NgbAccordion, NgbAccordionGroup} from './accordion';
+import {NgbAccordion, NgbAccordionPanel} from './accordion';
 
 function getPanels(element: Element): Element[] {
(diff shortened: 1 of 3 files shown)


Foxandxss:
I think that both names are fine. You're actuality grouping stuff.
Different group of options.

pkozlowski-opensource:
> I think that both names are fine. You're actuality grouping stuff.

You group things in an accordion not in panel. If you go to http://v4-alpha.getbootstrap.com/components/collapse/#accordion-example you will see that BS folks refer to panels as "Collapsible Group Item #1". So an accordion would be a group of collapsibles while a panel is would an _item_ in this group. 

IMO "panel", "collapsible", "accordion item" or similar variants are acceptable names but group to denote an item doesn't work for me, I'm afraid. Happy to hear a different interpretation, though.

pkozlowski-opensource:
Landed as c47ed85d972d64df43a2004ea0937f45466a02fe

PR #110, opened 2015-11-19
fix(dropdown): use attribute as a selector

### LINKED ISSUE #109: Dropdown is not behaving correctly
Yay, first real issue.

The dropdown as is as we have it now (even with pawel changes on toggle) is not working right on the CSS.

Check this [plunker](http://plnkr.co/edit/X5nOU0W2F9QPW2IX913R?p=preview)

First two works good, the third one doesnt.

There seems to be a dropdown and button dropdown. First two are button dropdown (they need a `btn-group`) and the third one which is the raw one (the one we support directly) has the popup weirdly.

pkozlowski-opensource:
Fixes #109


Foxandxss:
LGTM

PR #117, opened 2015-11-24
refactor(pagination): use value as the setter param

Foxandxss:
I changed this on the wiki a couple of weeks ago since it is what everybody is doing (including angular IIRC) and well, it is was always `value` in all languages.


pkozlowski-opensource:
LGTM

PR #119, opened 2015-11-24
feat(progressbar): manage values below 0 and bigger than max

Foxandxss:
This is the solution I had in mind.

If the value is less than 0, put it at 0, if more than 0, set it to max. If max changes, change value.

There are stuff to discuss about `max` typing, so hold this until meeting.

### DIFF
--- a/src/progressbar/progressbar.ts
+++ b/src/progressbar/progressbar.ts
@@ -11,9 +11,17 @@ import {Component, Input} from 'angular2/angular2';
   `
 })
 export class NgbProgressbar {
+  private _max = 100;
   private _striped: boolean;
+  private _value: number;
 
-  @Input() max = 100;
+  @Input()
+  set max(value: number) {
+    this._max = value;
+    this.value = Math.min(this.value, this._max);
+  }
+
+  get max(): number { return this._max; }
 
   @Input()
   set striped(value: boolean | string) {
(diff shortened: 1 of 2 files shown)


src/progressbar/progressbar.ts
```diff
@@ -28,7 +38,17 @@ export class NgbProgressbar {
 
   @Input() type: string;
 
-  @Input() value: number;
+  @Input()
+  set value(value: number) {
+    if (value < 0) {
```
pkozlowski-opensource: You can do this in one-liner, similar to what is done in pagination: https://github.com/ng-bootstrap/core/blob/9c6f488ae75a22519c38299d369a616f22203b77/src/pagination/pagination.ts#L68

src/progressbar/progressbar.ts
```diff
@@ -11,9 +11,19 @@ import {Component, Input} from 'angular2/angular2';
...
   private _striped: boolean;
+  private _value: number;
 
-  @Input() max = 100;
+  @Input()
+  set max(value: number) {
+    this._max = value;
+    if (value < this.value) {
+      this.value = value;
+    }
```
pkozlowski-opensource: Rewrite as `this.value = Math.min(this.value, this._max)`?

Foxandxss:
Yeah, updated that. Those days you're blank :P

PR #120, opened 2015-11-25
test(progressbar): expose problem with order of setters invocation

pkozlowski-opensource:
I'm not sure if I was clear during yesterday's meeting about the issue we can see here in this test but the basic problem with setters influencing state is that we might end up with different state depending on the order of setters invocation. And we can't control this order!

Having the problem clear we can have a basic rule here: we can't calculate / update state in setters! All our directives (pagination, progress and rating) currently suffer from the problem exposed here.

Sending a PR to make everyone aware of the issue.

### DIFF
--- a/src/progressbar/progressbar.ts
+++ b/src/progressbar/progressbar.ts
@@ -1,4 +1,4 @@
-import {Component, Input} from 'angular2/angular2';
+import {Component, Input, OnChanges} from 'angular2/angular2';
 
 @Component({
   selector: 'ngb-progressbar',
@@ -10,18 +10,11 @@ import {Component, Input} from 'angular2/angular2';
     </progress>
   `
 })
-export class NgbProgressbar {
-  private _max = 100;
+export class NgbProgressbar implements OnChanges {
   private _striped: boolean;
-  private _value: number;
 
-  @Input()
-  set max(value: number) {
-    this._max = value;
-    this.value = Math.min(this.value, this._max);
-  }
-
-  get max(): number { return this._max; }
+  @Input() max = 100;
+  @Input() value: number;
 
   @Input()
   set striped(value: boolean | string) {
@@ -36,12 +29,7 @@ export class NgbProgressbar {
 
   @Input() type: string;
 
-  @Input()
-  set value(value: number) {
-    this._value = Math.max(Math.min(value, this.max), 0);
-  }
-
-  get value(): number { return this._value; }
+  onChanges() { this.value = Math.max(Math.min(this.value, this.max), 0); }
 
   calculatePercentage() { return 100 * this.value / this.max; }
 }
--- a/src/progressbar/progressbar.spec.ts
+++ b/src/progressbar/progressbar.spec.ts
@@ -21,9 +21,11 @@ describe('ng-progressbar', () => {
 
     it('should calculate the percentage (default max size)', () => {
       progressCmp.value = 50;
+      progressCmp.onChanges();
       expect(progressCmp.calculatePercentage()).toBe(50);
 
       progressCmp.value = 25;
+      progressCmp.onChanges();
       expect(progressCmp.calculatePercentage()).toBe(25);
     });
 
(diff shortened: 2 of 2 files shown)


pkozlowski-opensource:
cc @Foxandxss @robwormald @wesleycho

wesleycho:
Travis is complaining about needing to run clang-format :P

Foxandxss:
I can see what you mean and I thought about that too. I am not sure what can we do in here. Even outside the setters issue.

In uibs we watch both fields but we don't update `value` based on `max`, we do calculations afterwards. So perhaps we can do the same in here. Not update `value` based on `max` nor `value` changes. Then when we re-run the calculation, we check if the % is more than 100% or lower. and then update `value` perhaps.

Do you have another idea?

wesleycho:
I see what happened now - one has to be judicious about when one uses a getter/setter I think because of this.

pkozlowski-opensource:
@Foxandxss see the second commit with a solution proposal
@wesleycho it has proper formatting :-)

Foxandxss:
That makes me want @robwormald solution. We could potentially combine both subscription and make the update in there.

pkozlowski-opensource:
> That makes me want @robwormald solution. We could potentially combine both subscription and make the update in there.

Possibly. But I can't think of any other solution with the current framework. IMO we should fix this stuff, regardless of what lands in the framework in the future, as currently things are pretty much broken.

But yeh, we need to discuss this with the core team as once again this is pretty common scenario and I would expect that people keep bumping into this.

src/progressbar/progressbar.ts
```diff
@@ -38,10 +37,12 @@ export class NgbProgressbar {
 
   @Input()
   set value(value: number) {
```
Foxandxss: you don't need setters anymore (nor getters). Both max and value.

Foxandxss:
I like this, but I would call this a refactor, no test.

Foxandxss:
> Possibly. But I can't think of any other solution with the current framework. IMO we should fix this stuff, regardless of what lands in the framework in the future, as currently things are pretty much broken.

> But yeh, we need to discuss this with the core team as once again this is pretty common scenario and I would expect that people keep bumping into this.

Yeah, I wasn't saying of doing it today, just to propose better sugar to the team.

src/progressbar/progressbar.ts
pkozlowski-opensource: I think we still need them for max as currently we don't handle max-as-string properly

wesleycho:
It sounds like they're already working on a solution though, so maybe sync up with Rob/Victor based on what Rob mentioned about it and chime in when the proposal gets brought up for review for getting into Angular proper?

pkozlowski-opensource:
> I like this, but I would call this a refactor, no test.

There are 2 commits in here, I believe that both got proper commit messages.

src/progressbar/progressbar.ts
Foxandxss: "Yes". You could parseInt on max to convert it to int. With the current implementation, a `50 / "100"` works, but perhaps not elegant.

pkozlowski-opensource:
> It sounds like they're already working on a solution though, so maybe sync up with Rob/Victor based on what Rob mentioned about it and chime in when the proposal gets brought up for review for getting into Angular proper?

I wouldn't expect it to happen before beta so few weeks from now at best. I wouldn't hold my breath waiting to quick solution from the framework.

Foxandxss:
> There are 2 commits in here, I believe that both got proper commit messages.

That is cheating, you should squash :P

pkozlowski-opensource:
Squashed

Foxandxss:
So, it only needs the `max` setter to parse int or if not, remove it.

pkozlowski-opensource:
> So, it only needs the max setter to parse int or if not, remove it.

@Foxandxss OK, removed get / set for now. We will take care about type conversions + tests in a subsequent PR

robwormald:
fyi you can just add mine in, but it wont solve the ordering problem :)

Foxandxss:
Couldn't you merge both streams or something?

PR #122, opened 2015-11-25
refactor(rating): organize code inside component

Foxandxss:
I updated the [conventions](https://github.com/ng-bootstrap/core/wiki/Code-conventions) to explain how to organize the code. I leave this as an example. If you like it as is, I merge and update the rest, if not, we can discuss it a bit and make changes.

Also changed the way we make arrays to do `number[]` instead of `Array<number>`.


wesleycho:
I'm fine with this.

pkozlowski-opensource:
LGTM

PR #123, opened 2015-11-25
refactor(pagination): add typings to tests

Foxandxss:
Fiddling a bit with `lib.d.ts` I figured out a better way of do typing on our elements. I think this is more idiomatic and works really good.

Also renamed `paginationCmpt` to `paginationCmp` to match the other business logic tests we have and also our test components. Also, less letters :P

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -15,15 +15,7 @@ import {Component} from 'angular2/angular2';
 
 import {NgbPagination} from './pagination';
 
-function getLink(nativeEl, idx) {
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
I'm fine with typings, less fan of the variable rename. Maybe we should name it `ngbPagination`?

Foxandxss:
Or perhaps `component` for all of them?

wesleycho:
Just `ngbPagination` or `pagination` probably is better - they are instances of a class, and not a component in particular in those cases.

Foxandxss:
I renamed it to `pagination`.

pkozlowski-opensource:
LGTM

PR #125, opened 2015-11-27
refactor(accordion): organize code inside component

Foxandxss:
Another more.

If there is a constructor, always the first method (after inputs / outputs).

Input goes in alphabetical order, doesn't matter if they have setters or not. Hopefully we will get rid of setters anytime soon.

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -22,6 +22,7 @@ import {NgbCollapse} from '../collapse/collapse';
 })
 export class NgbAccordionPanel {
   private _isOpen = false;
+
   @Input() isDisabled: boolean;
   @Input() title: string;
 


pkozlowski-opensource:
I must say that I find it harder to read (that is, field declaration mixed with methods).
For now it is -1 from me, but open to be convinced

Foxandxss:
So, should we do:

```typescript
class Foo {

  @Input() a;
  @Input() h;

  @Input()
  set b(...) { ... }

  get b() { ... }

  @Input()
  set r(...) { ... }

  get r() { ... }
```

AKA, all simple inputs first and then the setter / getters one. What about Outputs? before Input setters or after?

wesleycho:
I think I agree with Pawel - having the simple declarations all together is easier to read since it doesn't require an extra line in between to separate it from the getters/setters.

Foxandxss:
Updated it.

wesleycho:
Now it looks like an extra line is added - this LGTM, although you should just merge it in if it's this trivial IMO.

Foxandxss:
Yeah, I merge trivial for sure, but never realized that. I moved code around, made the PR, moved code around again and pushed back, but never looked into the diff.

PR #126, opened 2015-11-27
refactor(accordion): add typings to tests

Foxandxss:
Added typing, changed `Boolean` to `boolean` (recommendation from TS team) and simplified some initializations.

I need to revise this tests, add more to cover better what we have, but the accordion is broken atm anyway.


pkozlowski-opensource:
LGTM

wesleycho:
LGTM

PR #128, opened 2015-11-28
refactor(collapse): add typing and simplify a bit

Foxandxss:
Also added `collapseEl` because we are using that pattern already and I like it.


pkozlowski-opensource:
LGTM

wesleycho:
LGTM

PR #131, opened 2015-12-08
chore: add tooling for demo

Foxandxss:
Alrighty, this is what I was working on the past few days.

With this, we get complete automation for the demo site.

The idea is that we can do anything we want in `/demo` and by doing `$ gulp deploy-demo` it would "compile" the demo into `/demo/dist` and the content of that goes to `gh-pages` automagically.

Everything (webpack conf, deps) sits in root. The reason for that is that you cannot run certain commands from the root that would involve the children alone. So this a choice between webpack.conf + package.json for the demo but all demo related stuff should sit in the demo folder with a gulpfile. That is not good.

If we add more packages for the demo, they need to sit on the project's package.json, but that is not a big deal.

I am more wary about the automatic push to gh-pages. It is the automation we needed in ui-bootstrap but is quite limiting. I don't have versioned docs figured out (not needed yet) but with automation is a bit harder.

About the webpack config. I decided to create a separate bundle with angular 2 and dependencies and a bundle for the app itself (which will contain also ngbs).

For css I haven't decided anything yet. Bootstrap 4 is not bundled yet in npm (I guess) so I could put a <link> tag or download the .css directly into a folder in the demo. Also we need to decide if scss or less (not sure what bs4 uses, scss IIRC).

And of course, figure out how to lay a demo.

When this get merged, we can create a dummy page with BS4 as a welcoming page.

I created this demo in [here](https://github.com/Foxandxss/ngbs-demo-playground) and you can see the result in [here](http://foxandxss.github.io/ngbs-demo-playground/)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,12 @@
     "gulp": "^3.9.0",
     "gulp-clang-format": "^1.0.21",
     "gulp-ddescribe-iit": "^1.3.0",
+    "gulp-gh-pages": "^0.5.4",
+    "gulp-shell": "^0.5.1",
     "gulp-sourcemaps": "^1.6.0",
     "gulp-typescript": "^2.9.2",
     "gulp-util": "^3.0.6",
+    "html-webpack-plugin": "^1.7.0",
     "jasmine": "^2.3.2",
     "jasmine-core": "^2.3.4",
     "karma": "^0.13.14",
@@ -36,9 +39,13 @@
     "karma-jasmine": "^0.3.6",
     "karma-sourcemap-loader": "^0.3.6",
     "merge2": "^0.3.6",
+    "reflect-metadata": "^0.1.2",
     "run-sequence": "^1.1.4",
     "systemjs": "^0.19.5",
+    "ts-loader": "^0.7.2",
     "typescript": "^1.6.2",
-    "webpack": "^1.12.2"
+    "webpack": "^1.12.2",
+    "webpack-dev-server": "^1.14.0",
+    "zone.js": "^0.5.8"
   }
 }
--- a/webpack.demo.js
+++ b/webpack.demo.js
@@ -1,0 +1,54 @@
+var webpack = require('webpack');
+var HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: {
+    'angular2': [
+      'zone.js',
+      'reflect-metadata',
+      'angular2/angular2'
+    ],
+    'app': [
+      './demo/src/demo.ts'
+    ]
+  },
+  output: {
+    path: __dirname + '/demo/dist',
+    filename: '[name].bundle.js',
+    chunkFilename: '[name].bundle.js'
+  },
+  resolve: {
+    extensions: ['', '.ts', '.js']
+  },
+  devtool: 'eval',
+  module: {
+    loaders: [
+      {
+        test: /\.ts$/,
+        loader: 'ts-loader'
+      }
+    ]
+  },
+  plugins: [
+    new webpack.optimize.OccurrenceOrderPlugin(),
+    new webpack.optimize.DedupePlugin(),
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'angular2',
+      minChunks: Infinity,
+      filename: 'angular2.bundle.js'
+    }),
+    new HtmlWebpackPlugin({
+      template: './demo/src/index.html',
+      inject: 'body'
+    })
+  ],
+  devServer: {
+    contentBase: './demo/dist',
+    stats: {
+      modules: false,
+      cached: false,
+      colors: true,
+      chunk: false
+    }
+  }
+};
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 /typings/
 /dist/
 /temp/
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -18,16 +18,20 @@
   },
   "homepage": "https://github.com/ng-bootstrap/core#readme",
   "devDependencies": {
-    "angular2": "^2.0.0-alpha.46",
+    "@reactivex/rxjs": "5.0.0-alpha.7",
```
wesleycho: Not sure it matters, but the package is now provided as rxjs: https://www.npmjs.com/package/rxjs

Foxandxss: Yeah, but I thought it was only for the new version and not for the old ones.

I am waiting for angular .49 to fix an issue that prevents this library from upgrade and then updates rxjs too.

wesleycho: Ah, gotcha

wesleycho:
This looks fine for me - I like this approach, keeps it separate from the package distribution folder.

webpack.demo.js
```diff
@@ -0,0 +1,54 @@
+var webpack = require('webpack');
+var HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: {
+    'angular2': [
+      'zone.js',
+      'reflect-metadata',
```
pkozlowski-opensource: We now have https://code.angularjs.org/2.0.0-alpha.48/external-dependencies.js bundle that contains all the external dependencies (zone and reflect-metadata for now) so we can use it (and by this I mean just stick a good old script tag) instead of having it done through webpack

pkozlowski-opensource: What about es6-shim? I guess we should include it in the demo so it works in all the supported browsers. Here again I would just stick it in a script tag

pkozlowski-opensource:
Happy to see progress on this item! Left 2 minor comments, will review more carefully tomorrow.

webpack.demo.js
Foxandxss: So the .48 thing. We are stuck on .46 just yet so there is nothing to be done as today.

About putting extra script tags. I don't like that. That is why we have webpack at the end of the day so we can bundle and minify the deps. Unbundled for dev, bundled for prod. If we stick to script tags, we will have always dev ones or prod ones.

At the end you get the same behavior in both ways but I prefer to define my deps in webpack (lets use the tool we have in hand, right?).

That being said, I have no idea how to make that work with .48 external dependencies and webpack, but I will investigate for sure.

TL;DR I am still fine with this way and I will update to .49 when it gets out.

pkozlowski-opensource: > About putting extra script tags. I don't like that. 

This is not an argument :-) That is why we have webpack at the end of the day so we can bundle and minify the deps. 

> Unbundled for dev, bundled for prod. If we stick to script tags, we will have always dev ones or prod ones.

Not sure I understand this one...

Putting external dependencies in script tags has 2 advantages:
* enables effective caching in browsers
* avoid unnecessary work on WebPack side

Foxandxss: I mean, the point of having webpack is to use webpack. I like to have a simple index.html and let webpack inject all the needed script tags in there.

And about bundled stuff. If we put script tags with development versions of zone.js and reflect-metadata. When I run the uglify, those two dependencies won't be uglified (or I will need to do that in a separate step).

> enables effective caching in browsers

Why is that? I mean, what is the difference if we are not using cache bust?

> avoid unnecessary work on WebPack side

That is not an argument. It is unnecessary work on webpack or unnecessary work on the template side.

wesleycho: I'm not sure how much the caching will be an issue - in practice, the releases won't be *that* frequent, will they? Caching will take over in the browser after the first time they hit the new page...caching external dependencies separately via script tag has the advantage of displaying a good practice for end users, but as far as this project goes, it doesn't really matter too much IMO.

The avoiding unnecessary webpack work is a valid point though - decreasing build tool runtime has productivity ramifications.

I'm indifferent about choosing one approach over the other, so I'm fine with either as long as the one with the better arguments win out.

Foxandxss: There is an issue in here. I was going to do it as Pawel said, but that would require the final demo to have a node_modules with the dependencies or something. I mean if I put script tags like:

```
<script src="/node_modules/angular2/node_modules/reflect-metadata/Reflect.js"></script>
<script src="/node_modules/angular2/node_modules/zone.js/dist/zone.js"></script>
```

I need that path to exist on gh-pages. If I use the `external-dependencies` thingy, I would need also to have a node_modules with angular in it for the gh-pages.

The idea is to serve just the `index.html` and 1-2 bundle files with minified dependencies. Not a node_modules with stuff in it.

Perhaps I am doing it wrong.

Foxandxss:
Ok @pkozlowski-opensource I rebased this.

So, we cannot use script tags to deploy the demos so we are stuck to webpack like we have now or we can do `import 'zone.js'` (and the same for reflect-metadata) in `demo.ts`.

We can use the new polyfill but webpack will complain for feeding it with a bundle, it prefers the raw deps.

Foxandxss:
@pkozlowski-opensource we need to get a solution for this so we can create some dummy page (not that angular is beta)

Foxandxss:
continue at #144

PR #133, opened 2015-12-09
refactor: re-organize angular2 imports

pkozlowski-opensource:
Changes are due to https://github.com/angular/angular/issues/5710

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -10,7 +10,7 @@ import {
   beforeEachProviders
 } from 'angular2/testing';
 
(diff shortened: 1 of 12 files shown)


Foxandxss:
I see, I tried that by recommendation of @robwormald but saw the NgFor failing.

Is there now a list of what can you import from where?

and LGTM.

pkozlowski-opensource:
> Is there now a list of what can you import from where?

The best way is to check top-level files in https://github.com/angular/angular/tree/master/modules/angular2, see what is available and what gets exported (once again, knowing that angular2/angular2 goes away)

PR #134, opened 2015-12-09
refactor(tests): drop debugElement when possible

Foxandxss:
More sugar. There is more with injectAsync, need to research that too.


pkozlowski-opensource:
LGTM

PR #135, opened 2015-12-09
refactor(tests): move to inject from injectAsync

Foxandxss:
Bit of sugar added by Julie.

Not that it helps with clang tho ;(

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -5,7 +5,6 @@ import {
   describe,
   expect,
   inject,
-  injectAsync,
   TestComponentBuilder,
   beforeEachProviders
 } from 'angular2/testing';
@@ -47,8 +46,8 @@ describe('ngb-accordion', () => {
     `;
   });
 
-  it('should have no open panel', injectAsync([TestComponentBuilder], (tcb) => {
-       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+  it('should have no open panel', inject([TestComponentBuilder], (tcb) => {
+       tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
          fixture.detectChanges();
 
          const panels = getPanels(fixture.nativeElement);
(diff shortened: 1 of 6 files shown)


pkozlowski-opensource:
I didn't follow the whole story too close but my very limited understanding is that now test can pass if we don't return a promise thus masking a failure:
* https://github.com/angular/angular/pull/5721
* https://github.com/angular/angular/issues/5515

Once again, I don't know all the details but we need to check that tests don't pass when they should fail. It might be unnecessary noise but I prefer to be safe than sorry.

Foxandxss:
Oh, I wasn't aware of that. I am not reading angular issues / PRs.

juliemr:
Details:

It is indeed possible for tests to fail if you don't return a promise. We've reintroduced `injectAsync` to attempt to give the developer more mental clues to guard against this. When it's supported by browsers we'll be able to catch those errors, but for now we're still recommending `injectAsync`.

For the record, you can always shoot yourself in the foot by accidentally using `inject` :(

```js
it('...', inject([], () => {
  service.doAsyncThing().then(() => {
    expect(2).toEqual(5); 
  });  // BAD: This will pass
}));

it('...', injectAsync([], () => {
 service.doAsyncThing().then(() => {
    expect(2).toEqual(5); 
  });  // GOOD: This will fail, since it will yell at you because you forgot to return a promise
}));

it('...', injectAsync([], () => {
  return service.doAsyncThing().then(() => {
    expect(2).toEqual(5); 
  });  // GOOD: This will fail, as it should
}));
```

icfantv:
All that work @Foxandxss...sorry mate.  :-)

wesleycho:
Any reason why we wouldn't just do `injectAsync(...)` and then inside do `return tcb.overrideTemplate()...` or whatever method used? It sounds like that would fix it.

Foxandxss:
That is what we are doing now. I made this PR seeing @juliemr work on removing injectAsync, but there are other problems, so well, stuff happens :P

wesleycho:
Ah I see, I missed the `return` deletion

wesleycho:
Closing based on discussion

PR #136, opened 2015-12-10
chore: update to angular2 .52

Foxandxss:
This PR will make me cry all day and probably the rest of the week also. Worst update ever.

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -1,12 +1,12 @@
 import {Directive, Input, HostListener} from 'angular2/core';
 
-@Directive({selector: '[ngb-dropdown]', exportAs: 'ngbDropdown', host: {'[class.open]': 'open'}})
+@Directive({selector: '[ngbDropdown]', exportAs: 'ngbDropdown', host: {'[class.open]': 'open'}})
 export class NgbDropdown {
   @Input() open = false;
 }
 
 @Directive({
-  selector: '[ngb-dropdown-toggle]',
+  selector: '[ngbDropdownToggle]',
   host: {'class': 'dropdown-toggle', 'aria-haspopup': 'true', '[attr.aria-expanded]': '_dropdown.open'}
 })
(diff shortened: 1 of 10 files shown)


src/dropdown/dropdown.ts
```diff
@@ -1,12 +1,12 @@
 import {Directive, Input, HostListener} from 'angular2/core';
 
-@Directive({selector: '[ngb-dropdown]', exportAs: 'ngbDropdown', host: {'[class.open]': 'open'}})
+@Directive({selector: '[ngbDropdown]', exportAs: 'ngbDropdown', host: {'[class.open]': 'open'}})
```
pkozlowski-opensource: do we need / want to change those?

Foxandxss: Do I want? No, I don't want any of this, this is all BS. Do we need to do it? Yes.

pkozlowski-opensource: Can't we leave a selector as `ngb-dropdown`? I'm not saying we should, just didn't test so don't know if technically possible...

Foxandxss: We can't.

wesleycho: That was a breaking change as far as I understand - it makes me sad, as it is more natural to understand than this camelCase :( .

Foxandxss: Completely agree. But we have to move on and merge this.

pkozlowski-opensource: I thought that we can still use `ngb-dropdown` in a selector....

Foxandxss: dash-stuff for element selectors, attributes ones are camelCase.

pkozlowski-opensource: I would like to check this, since it sounds wrong to me...

Foxandxss: please do, you're the one with connections :P

PR #141, opened 2015-12-16
chore: add compiler settings to Travis

wesleycho:
I noticed [this message](https://travis-ci.org/angular-ui/bootstrap/builds/97249326) in UI Bootstrap's Travis, likely related to the set of devDependencies used - I ported those settings here since it seems like this might be relevant to us in the future.


pkozlowski-opensource:
LGTM

PR #143, opened 2015-12-17
feat(pager): add pager component

pkozlowski-opensource:
This still needs tests, but I'm opening an issue as soon as possible since this one is quite different as compared to ng1 version one.

I also need to push further work on adding some sugar for boolean / number attributes as it is pain in the neck to deal with them.

### DIFF
--- a/src/pager/pager.ts
+++ b/src/pager/pager.ts
@@ -1,0 +1,43 @@
+import {Component, ChangeDetectionStrategy, OnChanges, Input, Output, EventEmitter} from 'angular2/core';
+
+@Component({
+  selector: 'ngb-pager',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <nav>
+      <ul class="pager">
+        <li [class.disabled]="!hasPrev()"><a (click)="prev()">Previous</a></li>
+        <li [class.disabled]="!hasNext()"><a (click)="next()">Next</a></li>
+      </ul>
+    </nav>
+    `
+})
+export class NgbPager implements OnChanges {
+  private _currentPage = 0;  // internal state
+  @Input() noOfPages = 0;
+  @Input() page = 0;
+  @Output() pageChange = new EventEmitter();
+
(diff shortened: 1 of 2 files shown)


wesleycho:
One thing I'm wondering about, a pager is somewhat similar to pagination - what drives the difference, and how do we handle components with shared features?

src/pager/pager.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (this.hasNext()) {
+      this.pageChange.emit(++this._currentPage)
+    }
+  }
+
+  hasPrev(): boolean { return this._currentPage > 1; }
+
+  hasNext(): boolean { return this._currentPage < this.noOfPages; }
+
+  ngOnChanges(): void { this._currentPage = Math.max(Math.min(this.page, this.noOfPages), 1) }
```
wesleycho: One thing I noticed is that the pagination `page` is allowed to be a number or string - do we have to account for this too with a getter/setter as well since page is an input?

pkozlowski-opensource:
> One thing I'm wondering about, a pager is somewhat similar to pagination - what drives the difference, 

There are similarities but there are more differences, IMO. Most notably I'm not sure we want to have the same inputs (collection size, items per page and page no). To be discussed.

> and how do we handle components with shared features?

This is a good question! I would expect that "normal" refactoring practices (extract common code to a utility function and / or base class etc.) will get us covered. But this comes at the price of coupling components.

src/pager/pager.ts
pkozlowski-opensource: Yeh, we need to do it consistently... Still not sure what is the best way... See my other comment about shared utils / base classes.

pkozlowski-opensource:
This should be good to review / go. @wesleycho do you want to chat more about pagination / pager relation?

wesleycho:
I'm fine with this

Foxandxss:
Sounds good but I created some conventions, for example, reorder methods to be in alphabetical order.

PR #144, opened 2015-12-17
chore: add demo toolings

Foxandxss:
Broke the other PR (by deleting my fork due a problem).

I tried the <script> tag solution. It requires lot of scripting to move the file around. Much better to simply import it and use it.

That is the less ugly solution I could find.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 /typings/
 /dist/
 /temp/
+/demo/dist/
 npm-debug.log
+.publish
 .sw*
 .idea
--- a/demo/index.html
+++ b/demo/index.html
@@ -1,0 +1,11 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>ng-bootstrap</title>
+</head>
+<body>
+  <demo-app></demo-app>
+  <script src="lib/angular2-polyfills.js"></script>
+</body>
+</html>
--- a/demo/src/demo.ts
+++ b/demo/src/demo.ts
@@ -1,0 +1,13 @@
+import {Component} from 'angular2/core';
+import {bootstrap} from 'angular2/platform/browser';
+
+@Component({
+  selector: 'demo-app',
+  template: `
+    <div>Hello World</div>
+  `
+})
+class Demo {
+}
+
+bootstrap(Demo);
--- a/webpack.demo.js
+++ b/webpack.demo.js
@@ -1,0 +1,39 @@
+var webpack = require('webpack');
+var HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: {
+    'app': './demo/src/demo.ts'
+  },
+  output: {
+    path: __dirname + '/demo/dist',
+    filename: 'bundle.js'
+  },
+  resolve: {
+    extensions: ['', '.ts', '.js']
+  },
+  devtool: 'eval',
+  module: {
+    loaders: [
+      {
+        test: /\.ts$/,
+        loader: 'ts-loader'
+      }
+    ]
+  },
+  plugins: [
+    new HtmlWebpackPlugin({
+      template: './demo/index.html',
+      inject: 'body'
+    })
+  ],
+  devServer: {
+    contentBase: './demo/dist',
+    stats: {
+      modules: false,
+      cached: false,
+      colors: true,
+      chunk: false
+    }
+  }
+};
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -3,13 +3,17 @@ var ts = require('gulp-typescript');
 var gutil = require('gulp-util');
 var sourcemaps = require('gulp-sourcemaps');
 var ddescribeIit = require('gulp-ddescribe-iit');
+var shell = require('gulp-shell');
+var ghPages = require('gulp-gh-pages');
 var del = require('del');
 var merge = require('merge2');
 var clangFormat = require('clang-format');
 var gulpFormat = require('gulp-clang-format');
 var runSequence = require('run-sequence');
+var webpack = require('webpack');
+var webpackDemoConfig = require('./webpack.demo.js');
 
-var PATHS = {src: 'src/**/*.ts', specs: 'src/**/*.spec.ts'};
+var PATHS = {src: 'src/**/*.ts', specs: 'src/**/*.spec.ts', demo: 'demo/**/*.ts', demoDist: 'demo/dist/**/*'};
 
 // Transpiling & Building
 
@@ -24,7 +28,6 @@ gulp.task('cjs', function() {
 });
 
 gulp.task('umd', function(cb) {
-  var webpack = require('webpack');
   webpack(
       {
         entry: './dist/cjs/core.js',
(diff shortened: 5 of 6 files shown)


demo/src/vendor.ts
```diff
@@ -0,0 +1,6 @@
+import 'angular2/bundles/angular2-polyfills';
+
+import 'angular2/platform/browser';
+import 'angular2/core';
+import 'angular2/http';
+import 'angular2/router';
```
wesleycho: I don't think the http and router imports are necessary at this stage.

pkozlowski-opensource: +1 - let's only import what we use

demo/src/vendor.ts
```diff
@@ -0,0 +1,6 @@
+import 'angular2/bundles/angular2-polyfills';
```
pkozlowski-opensource: I thought that we had a discussion about it :-) I need to fix it in all the other seed repos.

webpack.demo.js
```diff
@@ -0,0 +1,49 @@
+var webpack = require('webpack');
+var HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: {
+    'app': './demo/src/demo.ts',
+    'vendor': './demo/src/vendor.ts'
```
pkozlowski-opensource: I really wouldn't bother about splitting it into vendor right now.

webpack.demo.js
```diff
@@ -0,0 +1,49 @@
...
+        test: /\.ts$/,
+        loader: 'ts-loader'
+      }
+    ],
+    noParse: [ /angular2\/bundles\/.+/ ]
+  },
+  plugins: [
+    new webpack.optimize.OccurrenceOrderPlugin(),
+    new webpack.optimize.DedupePlugin(),
+    new webpack.optimize.CommonsChunkPlugin({
```
pkozlowski-opensource: So as soon as you use commons chunk plugin all the benefits of excluding vendor go out of the window. My vote: let's not bother and keep webpack's config simple

pkozlowski-opensource: We can always make it more complex if / when needed.

demo/src/vendor.ts
Foxandxss: The thing is that angular imports them anyway and it ends on the app bundle. By importing them by hand, they end on the vendor bundle.

TL;DR I can remove those two but will be on the bundle anyway. I will need that for the demo tho.

demo/src/vendor.ts
Foxandxss: <script> tag is -1 for me.

webpack.demo.js
Foxandxss: I don't mind that. We can require `vendor.ts` on `demo.ts`.

Foxandxss:
I deleted the extra vendor bundle. Now you have to require `vendor.ts` on the `demo.ts`.

demo/src/vendor.ts
pkozlowski-opensource: Soooo, we _really_ don't want to do this, see https://github.com/angular/angular/issues/5980

Foxandxss: It is not failing for me at all.

pkozlowski-opensource: It just means that you don't see issue with your current use case. But there are problems. If I can't convince you maybe Misko will? https://github.com/angular/angular/issues/5304#issuecomment-158133227

Foxandxss: It is not about trust. It was always working and everybody was doing it too without problems.

I will change it and write some scripts for deployment.

pkozlowski-opensource: >  It was always working and everybody was doing it too without problems.

It is more that you didn't have use-cases that would expose problems. This doesn't mean that those don't exist.

Foxandxss: agree on that

Foxandxss:
Review please.

demo/index.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>ng-bootstrap</title>
+</head>
+<body>
+  <demo></demo>
```
pkozlowski-opensource: use `-` in component selectors, ex.: `demo-app`

.gitignore
```diff
@@ -2,6 +2,8 @@
 /typings/
 /dist/
 /temp/
+/demo/dist/
```
pkozlowski-opensource: maybe let's use `dist/demo`?

webpack.demo.js
```diff
@@ -0,0 +1,44 @@
+var webpack = require('webpack');
+var HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: {
+    'app': [
+      'angular2/core',
```
pkozlowski-opensource: why do we need `angular2/core` as an entry here?

demo/src/demo.ts
```diff
@@ -0,0 +1,13 @@
+import {Component} from 'angular2/core';
+import {bootstrap} from 'angular2/bootstrap';
```
pkozlowski-opensource: import from `angular2/platform/browser` - `angular2/bootstrap` is going away

demo/index.html
```diff
@@ -0,0 +1,11 @@
...
+  <script src="lib/angular2-polyfills.js"></script>
```
pkozlowski-opensource: do we want to add `es-shim` as well?

.gitignore
Foxandxss: Dist folder is something we have to talk about yet. We are not really sure in how are we going to deploy it.

On the other hand, the demo deploy is just a temporary folder. I could even create a task to clean it after success push.

So the idea is, when you need to update the gh-pages, you run a command, `/demo/dist` gets generated and then pushed to gh-pages. After that you don't need and if we decide to put root's `/dist` on NPM, I wouldn't certainly want the demo dist in there.

So it is a -1 for me.

webpack.demo.js
```diff
@@ -0,0 +1,44 @@
...
+      './demo/src/demo.ts'
+    ]
+  },
+  output: {
+    path: __dirname + '/demo/dist',
+    filename: '[name].bundle.js',
+    chunkFilename: '[name].bundle.js'
```
pkozlowski-opensource: Do we still need `chunkFilename`?

webpack.demo.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  devtool: 'eval',
+  module: {
+    loaders: [
+      {
+        test: /\.ts$/,
+        loader: 'ts-loader'
+      }
+    ],
+    noParse: [ /angular2\/bundles\/.+/ ]
```
pkozlowski-opensource: I don't think we need this one any more

pkozlowski-opensource:
Left some comments. Good stuff!

demo/index.html
Foxandxss: sure

webpack.demo.js
Foxandxss: seems like we don't.

demo/src/demo.ts
Foxandxss: Good to know.

demo/index.html
Foxandxss: I don't have an answer for that tbh.

Foxandxss: Well, we can add it when it gets needed, not for now I think.

webpack.demo.js
Foxandxss: I can simplify that.

webpack.demo.js
Foxandxss: Indeed.

Foxandxss:
I addressed all of them (except the dist thing).

pkozlowski-opensource:
LGTM!

PR #147, opened 2015-12-30
refactor(pagination): use ChangeDetectionStrategy.OnPush

pkozlowski-opensource:


pkozlowski-opensource:
With this modification of the `ChangeDetectionStrategy` method used in a template (`hasPrevious()`, `hasNext()` etc.) are going to be called only once in response to input changes or events emitted (instead of being called systematically on each and every turn of the change detection).

Foxandxss:
Didn't we try this unsuccessfully in the past? Well I am happy it works now.

pkozlowski-opensource:
> Didn't we try this unsuccessfully in the past?

Don't remember exactly, but I think I wasn't aware about the fact that that `ChangeDetectionStrategy.OnPush` will trigger change detection on _both_ input changes _and_ output changes (firing events).

wesleycho:
This looks fine to me

PR #148, opened 2016-01-04
refactor(accordion): use @ContentChildren instead of @Query

pkozlowski-opensource:
@Query is going away: https://github.com/angular/angular/issues/4919

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -1,4 +1,4 @@
-import {Component, Directive, forwardRef, Inject, Input, Query, QueryList} from 'angular2/core';
+import {Component, Directive, forwardRef, Inject, Input, ContentChildren, QueryList} from 'angular2/core';
 
 import {NgbCollapse} from '../collapse/collapse';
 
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
BTW: we can't remove child->parent relationship (and thus circular deps...) till there is a solution similar to https://github.com/angular/angular/issues/5467

Foxandxss:
So I never did this because they are both broken (accordion in master and your PR is equally broken).

We need those changes @robwormald proposed to listen for children stuff.

pkozlowski-opensource:
> So I never did this because they are both broken (accordion in master and your PR is equally broken).

Did you try this stuff with Beta.0? I see it working on my end, will put together a plunker if needed.

Foxandxss:
try plunker try. There is some race conditions that happens in some places.

pkozlowski-opensource:
@Foxandxss here is a plunker with a version from this PR: https://plnkr.co/edit/7GarmofzfeG01L8RtEgU?p=preview

> There is some race conditions that happens in some places.

Do you have more info? Do you reproduce those in the plunker I've linked?

pkozlowski-opensource:
So, I've spent a couple days trying to dig deeper into this issue / general design of accordion-like components. As a result I will be sharing a doc with the angular team shortly with proposals on how to change the core of the framework. At the same time I believe that I've found an acceptable solution with the beta.7 version of the framework. It requires some refactoring to the existing code / design on our end, but I believe that those refactorings go in the right direction.

Closing this PR, will send another one shortly.

PR #149, opened 2016-01-05
feat: add input conversion utils

pkozlowski-opensource:
Adding few utils so we can consistently deal with type conversion for `@Input` values. IMO we need to have such utils in order to:
* have once central place where we handle conversion logic (for example: shell `foo="{{false}}"` evaluate to `true` or `false`?)
* have consistent user experience across all directives

What worries me is that is is creating coupling between components (that is - we are going to have harder time moving each component to its own dedicated repository) so we might want to extract those utils into a separate NPM package at some point.

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -1,0 +1,11 @@
+export function toBoolean(value: any): boolean {
+  return value === '' ? true : !!value;
+}
+
+export function toInteger(value: any): number {
+  return parseInt(`${value}`, 10);
+}
+
+export function getValueInRange(value: number, max: number, min = 0): number {
+  return Math.max(Math.min(value, max), min);
+}
--- a/src/pagination/pagination.ts
+++ b/src/pagination/pagination.ts
@@ -1,5 +1,6 @@
 import {Component, EventEmitter, Input, Output, OnChanges, ChangeDetectionStrategy} from 'angular2/core';
 import {NgFor} from 'angular2/common';
+import {getValueInRange, toInteger} from '../util/util';
 
 @Component({
   selector: 'ngb-pagination',
@@ -44,14 +45,14 @@ export class NgbPagination implements OnChanges {
 
   @Input()
   set collectionSize(value: number | string) {
-    this._collectionSize = parseInt(`${value}`, 10);
+    this._collectionSize = toInteger(value);
   }
 
   get collectionSize(): number | string { return this._collectionSize; }
 
   @Input()
   set pageSize(value: number | string) {
-    this._pageSize = parseInt(`${value}`, 10);
+    this._pageSize = toInteger(value);
   }
 
   get pageSize(): number | string { return this._pageSize; }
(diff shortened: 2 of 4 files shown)


src/util/util.ts
```diff
@@ -0,0 +1,11 @@
+export function toBoolean(value: any): boolean {
+  return value === '' ? true : !!value;
```
wesleycho: The semantics of this is a bit worrisome - one typically thinks of an empty string as a falsy value, so the function name `toBoolean` is a bit misleading.

pkozlowski-opensource: Ha! This is _exactly_ the discussion I wanted to trigger with this PR :-) Also why `"false"` doesn't get converted to `false`? We might rename this function to better indicate that this is for inputs, but the most important part for me is to agree on the semantic

wesleycho: Maybe `toInputBoolean` would work?

pkozlowski-opensource:
So, to agree on the semantic of those `toBoolean` conversions, let's list "all" the cases we can have and the resulting Boolean value:

* `false` => `false`
* `true` => `true`
* `""` (empty string) => `true`
* `"true"` => `true`
* `"false"` => `true`
* `"foo"` => `true`
* `0` => `false`
* `1` => `true`

IMO the only "controversial" ones are:
*  `""` which corresponds to `<div foo>` - it makes sense to have it as "true" for "Boolean" inputs
* `"false"` which could correspond to `<div foo="{{false}}">` - I'm on the fence with this one

icfantv:
Do we REALLY want to make the empty string truthy?  I suspect we'd hear about that from folks.  I understand why we would treat it as truthy for an input and if someone marks the field as required, the validator kicks in anyway.

If we do this, we should certainly document the heck out of it.  __WITH BIG, BOLD LETTERS__ :-)

wesleycho:
Is there a better way to check for whether an attribute is a boolean in ng2? This seems like an important use case...or we could just say we don't support `foo=""`.

pkozlowski-opensource:
> Is there a better way to check for whether an attribute is a boolean in ng2?

Nope, I'm not aware if any mechanism. We can add one if we can come up with a consistent story

> or we could just say we don't support foo="".

This is more about `<div foo>`, or using more concrete example: `<ngb-progress striped animated>` which is nicer as compared to `<ngb-progress [striped]="true" [animated]="true">`

pkozlowski-opensource:
FYI this is what Ionic2 is using in its "is boolean input" checks:

```javascript
get isActive() {
    return (this._isActive === undefined || this._isActive === true || this._isActive === 'true');
}
```

https://github.com/driftyco/ionic2/blob/f74606dc1be56d34f134d7b9feb790b78d5362dc/ionic/components/icon/icon.ts#L119

wesleycho:
My concern is mainly that `foo=""` and `foo` would both evaluate to `true` - is that the interpretation we want to support? I'm fine if that's the decision made, but I want us to be absolutely sure here.

Otherwise, this is fine to me.

pkozlowski-opensource:
> My concern is mainly that foo="" and foo would both evaluate to true - is that the interpretation we want to support?

I don't have strong opinion here but yeh, leaning towards having `foo` as evaluating to "true" when evaluating in the context of Boolean attributes. 

At the same time I don't want to "agonize" over a discussion on this one. IMO a _far_ more important decision is this one: do we want to have this type of helpers shared by all components. If we answer YES than let's take _any_ decision: it will be so much easier to change a helper in one place if we change our mind in the future.

wesleycho:
I'm fine with this getting in as is.

PR #150, opened 2016-01-11
feat(datepicker): initial implementation

Foxandxss:
This is just the simplest implementation ever.  Not to merge yet.

I just want to test the waters, get feedback and give it some extra love.

So the idea is to extract all the logic into a separate Engine so we can release (in a different repo) a MomentEngine and people can implement their own if needed.

I created a `bindings` file that every people will need to add to their bootstrap:

```typescript
bootstrap(App, [NGB_BINDING])
```

I guess we will have more stuff to add to that with other components.

That engine has just one method (for now) that simply generate an array of days.

I put the styling inline because I have no idea in how to make it on a separate file and make it work with TSC and the UMD bundle.

And of course, the whole design is just a prototype (but I would like to keep it simple, our uibs one is a space shuttle with that many buttons).

http://plnkr.co/edit/y8YSQrLl39ECLXRxaqNJ?p=preview

I also realized that my design is coupled with the glyphicons.css I use on my playground.

### DIFF
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -1,0 +1,53 @@
+import {Component, Input, OnInit} from 'angular2/core';
+import {DatePipe} from 'angular2/common';
+
+import {NgbDatepickerEngine} from './datepickerEngine';
+
+@Component({
+  selector: 'ngb-datepicker',
+  template: `
+    <div class="card">
+      <div class="card-header">
+        {{date | date:'MMMM yyyy'}}
+      </div>
+      <div class="card-block">
+        <div *ngFor="#day of days" class="a-day">
+          <span>{{day}}</span>
+        <div>
+      </div>
+    </div>
+  `,
+  styles: [
+    `
+    .card {
+        width: 285px;
+        height: 245px;
+    }
+    .card-header {
+      text-align: center;
+    }
+    .a-day {
+        display: inline-block;
+        text-align: right;
+        width: 42px;
+        height: 25px;
+    }
+    .a-day span {
+        margin-right: 4px;
+        margin-top: 4px;
+    }
+  `
+  ]
+})
+export class NgbDatepicker implements OnInit {
+  days: number[] = [];
+
+  @Input() date: Date;
+  @Input() startOnMonday: boolean | string;
+
+  isMondayFirst(): boolean { return this.startOnMonday === '' ? true : !!this.startOnMonday; }
+
+  constructor(private engine: NgbDatepickerEngine) {}
+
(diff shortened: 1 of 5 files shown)


Foxandxss:
Another thing to mention is bindings or providers or ... ?

Foxandxss:
Also, wasn't sure where to put that bindings file nor how to export it, I accept suggestions.

wesleycho:
To be clear on this, the DatepickerEngine will be provided upon bootstrap of the application by the user, and that is how it will be injected into the component?

Foxandxss:
Yes, all users will need to inject the bindings on bootstrap. Now we will have 1 with this Engine but we can potentially have 5-10 (for upcoming components).

So the idea to inject it in the bootstrap is that it makes the engine available everywhere to inject (including our datepicker).

So if we have bootstrap -> A -> B -> Datepicker (And A component using a B component that uses the datepicker).Datepicker will ask B if he defined a engine, nope, A, nope, on bootstrap? Yes.

The nicety in here is that you can do in B (or A):

```
providers: [provide(DatepickerEngine, {useClass: MyCustomDatepickerEngine})]
```

Now when the datepicker ask B if it has a engine, well, it does so the datepicker is injected with that one instead.

So having the interface as a contract, people can create their own engines and use them where they like to.

wesleycho:
So I guess the thing we really need to get right is the engine then - it's a little different than I was originally thinking of for an implementation, but could be fine.

I need to think about this some.

src/datepicker/datepicker.ts
```diff
@@ -0,0 +1,53 @@
...
+})
+export class NgbDatepicker implements OnInit {
+  days: number[] = [];
+
+  @Input() date: Date;
+  @Input() startOnMonday: boolean | string;
+
+  isMondayFirst(): boolean { return this.startOnMonday === '' ? true : !!this.startOnMonday; }
+
+  constructor(private engine: NgbDatepickerEngine) {}
```
pkozlowski-opensource: Couldn't we mark it with `@Optional()` and instantiate engine "by hand" if not provided? It would make things _slightly_ easier on people ("nothing to do" by default).

Foxandxss: Open to suggestions.

Foxandxss: We need to establish a solid "way to go" for other devs / libraries.

I think that tending to offer a providers / binding array to inject at bootstrap is what the other devs are doing. But we need to care and think twice what we want to share and not merge the first idea without thinking :)

pkozlowski-opensource:
I think that this work is stale to the point that there is no reason to keep this PR open. Let's close it to keep the queue clean and re-open when this PR is ready.

PR #154, opened 2016-01-19
feat(alert): add alert component

pkozlowski-opensource:
This is more or less an equivalent of our angular-ui/bootstrap implementation and I expect this proposal to trigger some discussion so... let it begin :-)

### DIFF
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -1,0 +1,23 @@
+import {Component, Input, Output, EventEmitter, ChangeDetectionStrategy} from 'angular2/core';
+
+@Component({
+  selector: 'ngb-alert',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `
+    <div [class]="'alert alert-' + type" role="alert">
+      <template [ngIf]="dismissible">
+        <button type="button" class="close" aria-label="Close" (click)="closeHandler()">
+              <span aria-hidden="true">&times;</span>
+        </button>
+      </template>
+      <ng-content></ng-content>
+    </div>
+    `
+})
+export class NgbAlert {
+  @Input() dismissible = true;
+  @Input() type = 'warning';
+  @Output() close = new EventEmitter();
+
+  closeHandler() { this.close.emit(null); }
+}
--- a/src/alert/alert.spec.ts
+++ b/src/alert/alert.spec.ts
@@ -1,0 +1,61 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  injectAsync,
+  TestComponentBuilder,
+} from 'angular2/testing';
+
+import {Component} from 'angular2/core';
+
+import {NgbAlert} from './alert';
+
+function getAlertElement(element: HTMLElement): HTMLDivElement {
+  return <HTMLDivElement>element.querySelector('.alert');
+}
+
+function getCloseButton(element: HTMLElement): HTMLButtonElement {
+  return <HTMLButtonElement>element.querySelector('button');
+}
+
+describe('ngb-alert', () => {
+
+  it('should have type warning and by dismissible by default', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, '<ngb-alert>Watch out!</ngb-alert>')
+           .createAsync(TestComponent)
+           .then((fixture) => {
+             fixture.detectChanges();
+
+             const alertEl = getAlertElement(fixture.nativeElement);
+
+             expect(alertEl).toHaveCssClass('alert-warning');
+             expect(alertEl.getAttribute('role')).toEqual('alert');
+             expect(getCloseButton(alertEl)).toBeTruthy();
+           });
+     }));
+
+  it('should allow specifying alert type', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, '<ngb-alert type="success">Cool!</ngb-alert>')
+           .createAsync(TestComponent)
+           .then((fixture) => {
+             fixture.detectChanges();
+             expect(getAlertElement(fixture.nativeElement)).toHaveCssClass('alert-success');
+           });
+     }));
+
+  it('should render close button only if dismissible', injectAsync([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, '<ngb-alert [dismissible]="false">Don\'t close!</ngb-alert>')
+           .createAsync(TestComponent)
+           .then((fixture) => {
+             fixture.detectChanges();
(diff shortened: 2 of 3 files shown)


pkozlowski-opensource:
So, there are couple of usability issues with this component, as we've learned from angular-ui/bootstrap. The main issue is that the "close" button doesn't do anything useful by default (that is - it doesn't actually close alert). 

Unfortunately with the current Angular2 version there is no way of expressing "hey, component, go and destroy yourself" so the only way of proceeding would be to wrap everything in a `ngIf`:

```
@Component({
  selector: 'ngb-alert',
  template: `
    <template [ngIf]="dismissed">
    <div [class]="'alert alert-' + (type || 'warning')" role="alert">
      <template [ngIf]="dismissible">
        <button type="button" class="close" aria-label="Close" (click)="closeHandler()">
              <span aria-hidden="true">&times;</span>
        </button>
      </template>
      <ng-content></ng-content>
    </div>
    </template>
    `,
  directives: [NgIf]
})
export class NgbAlert {
  dismissed = false;
  @Input() dismissible = true;
  @Input() type;
  @Output() close = new EventEmitter();

  closeHandler() { dismissed = true; this.close.emit(null); }
}
```

but the downside of this is that:
* people can't capture / prevent closing of alerts
* the actual `<ngb-alert>` element would stay in the DOM after "closing" and this might screw up peoples' CSS styles

pkozlowski-opensource:
Idea of an alternative API:

```html
<template ngb-alert (close)="onAlertClose($event)">
   Alert's content
</template>
```
and in `onClose` people could call `$event.preventDefault()` to prevent an alert from closing.

By having alerts on `<template>` elements we could easily "destroy" them without leaving any leftowers in the DOM.

Thoughts?

wesleycho:
I don't quite like the original alert design in UI Bootstrap - it's a bit awkward.

Linking for reference the [Bootstrap 4 version](http://v4-alpha.getbootstrap.com/components/alerts/).

CSS-wise, this is a pretty straightforward component, so there is no reason to necessarily care about the pure CSS version. This leaves the dismissable version - there are two use cases for these sort of messages. There is the toastr type messages that pop on top of the whole UI, and there is the type that just show up in the container directly. These use cases would be best captured by utilizing the DCL I think, which also gives the ability to remove them & dynamically manage them.

WDYT about this approach?

Foxandxss:
Alert and toasts perhaps have different goals, but worth a new discussion imo.

pkozlowski-opensource:
So, my [previous idea](https://github.com/ng-bootstrap/core/pull/154#issuecomment-202896646) doesn't sound like an implementable one :-/ (or at least I don't know how to implement it...).

Regarding @wesleycho remarks:

 > I don't quite like the original alert design in UI Bootstrap - it's a bit awkward

I think we all agree here :-) The goal is to come up with sth better and this is what we are struggling with here....

> CSS-wise, this is a pretty straightforward component, so there is no reason to necessarily care about the pure CSS version.

Well, I still prefer to write `<ngb-alert type="warning">Watch out!</ngb-alert>`  as compared to dumping the whole HTML + CSS, remembering to add all the aria stuff, making sure that animations work etc. So still there is some value in it.

> This leaves the dismissable version

Yes, this is the crux. And I wouldn't drag toastr into the conversation - IMO toastr could build on top of "dumb" alerts.

pkozlowski-opensource:
To sum up: as much as I don't like the design of the current ui-bootstrap alert it is only one I know how to implement :-/ To cover the admissible part we need one of:
* changes in the framework ("hey component, destroy yourself" or make [other API](https://github.com/ng-bootstrap/core/pull/154#issuecomment-202896646) implementable)
* new idea
* someone convincing that says that we are trying to solve an issue that doesn't exist

Otherwise we are stuck with the ui-bootstrap API / impl :-(

pkozlowski-opensource:
Soooo, I had a chat with Misko yesterday on it (based on this doc: https://docs.google.com/document/d/1tMEhoAHhi-BZgHlbOMdxa5fu-Ev9PVM78SgC5uiGySk/edit?usp=sharing).

The tl;dr; is that:
- there will be mechanism to do `<template ngb-alert>` (it is not implementable yet but you can have a look at the gist toward the end of the linked doc);
- we still going to need a "wrapper" component as in this PR.

pkozlowski-opensource:
My suggestion based on the above: land what we've got in this PR and improve close support as soon as we've got sufficient mechanism in the framework. @Foxandxss @wesleycho WDYT?

wesleycho:
I'm fine with that.

Foxandxss:
You were the people who didn't want to add some features that needed to be changed later. I am fine with the idea tho.

wesleycho:
Well, we need to move faster (and I need to spend more time with ng-bootstrap), so I can bend.

pkozlowski-opensource:
@Foxandxss for the record - we will still need the exact implementation from this PR even if support in Angular gets better. So no code will be removed afterwards.

Foxandxss:
Yeah, you are right.

pkozlowski-opensource:
Any other comments or is this one good to go?

src/alert/alert.ts
```diff
@@ -0,0 +1,25 @@
+import {Component, Input, Output, EventEmitter, ChangeDetectionStrategy} from 'angular2/core';
+import {NgIf} from 'angular2/common';
```
Foxandxss: out

src/alert/alert.ts
```diff
@@ -0,0 +1,25 @@
...
+    <div [class]="'alert alert-' + (type || 'warning')" role="alert">
+      <template [ngIf]="dismissible">
+        <button type="button" class="close" aria-label="Close" (click)="closeHandler()">
+              <span aria-hidden="true">&times;</span>
+        </button>
+      </template>
+      <ng-content></ng-content>
+    </div>
+    `,
+  directives: [NgIf]
```
Foxandxss: out

src/alert/alert.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+export class NgbAlert {
+  @Input() dismissible = true;
+  @Input() type;
+  @Output() close = new EventEmitter();
+
+  closeHandler() { this.close.emit(null); }
```
Foxandxss: is null needed?

src/alert/alert.ts
```diff
@@ -0,0 +1,25 @@
...
+        </button>
+      </template>
+      <ng-content></ng-content>
+    </div>
+    `,
+  directives: [NgIf]
+})
+export class NgbAlert {
+  @Input() dismissible = true;
+  @Input() type;
```
Foxandxss: could we have type in here?

Foxandxss:
Apart from what I said, perhaps some extra tests, but that can be done later (I want to revisit tests and do a few)

PR #157, opened 2016-02-12
chore: update to angular 2 beta 6

Foxandxss:
After trying beta 2, beta 3, beta 5, beta 6 seems to not have any potential issue.

This fails on the typings for jasmine but since this PR needs to be merged as is and the other is more opinion based, I am going to create another PR for that.


wesleycho:
LGTM

PR #158, opened 2016-02-12
chore: add typings

### LINKED ISSUE #156: Were to add typings?
Since beta 2/3 jasmine typings are not included so we need to manually load it.

Official recommendation is to use typings, but the problem is that we need to add a /// reference tag thingy in a file that is loaded in our test, AKA a source file or test file.

We just need to add it once, but where? Perhaps `core.ts`? Were we add a reference to the whole `main.d.ts` generated by typings?

Another solution is `tsconfig.json` but I don't think we can do it without having a files array and listing all files there is a real pain.

I could hijack something with gulp but then the IDE won't see it and will raise errors.

I will implement the idea that we get after I can update to a new beta.

Foxandxss:
Don't merge this until #157 is merged so I can try it for reals.

Not that we have lot of typings for this project but jasmine is not bundled anymore and we need to use it.

Fixes #156

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,7 +13,13 @@ var runSequence = require('run-sequence');
 var webpack = require('webpack');
 var webpackDemoConfig = require('./webpack.demo.js');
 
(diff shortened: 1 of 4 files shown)


Foxandxss:
I rebased this. It needed also es6-shim typings.

We need this yes or yes :P

wesleycho:
LGTM

PR #161, opened 2016-03-16
style(all): re-format with clang 1.0.36

pkozlowski-opensource:
Looks like clang 1.0.36 was released and it has impact on our code base.
Not sure how I feel about all the changes but hey, clang sets the rules here...

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -18,7 +18,7 @@ var PATHS = {src: 'src/**/*.ts', specs: 'src/**/*.spec.ts', demo: 'demo/**/*.ts'
 function webpackCallBack(taskName, gulpDone) {
   return function(err, stats) {
     if (err) throw new gutil.PluginError(taskName, err);
-    gutil.log("[" + taskName + "]", stats.toString());
+    gutil.log('[' + taskName + ']', stats.toString());
     gulpDone();
   }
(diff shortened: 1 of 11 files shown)


pkozlowski-opensource:
If clang keeps on doing those kind of things we might consider pinning a version we are using...

Foxandxss:
I am completely against this change. One quote for HTML is horrible.

pkozlowski-opensource:
> I am completely against this change. One quote for HTML is horrible.

Not to mention that it breaks our code.... The only alternative I can see is to pin to a previous version of clang... Sight...

icfantv:
I think it's a good idea, in general, to pin versions until we have a chance to test the changed library in a controlled upgrade.  Otherwise, bugs significantly more subtle than a breaking build can creep in and we won't know it until someone reports it.  This has happened to us at work.

pkozlowski-opensource:
> I think it's a good idea, in general, to pin versions until we have a chance to test the changed library in a controlled upgrade.

In this case we might consider shrink-wrapping our package.json. It is pain n a neck, but at least deals with the problem we are seeing here :-)

icfantv:
I can't tell if you're agreeing with me or not.  :-)

PR #167, opened 2016-03-30
refactor(accordion): use reusable panels

pkozlowski-opensource:


src/accordion/accordion.spec.ts
```diff
@@ -123,41 +121,53 @@ describe('ngb-accordion', () => {
        });
      }));
 
-  it('should only open one at a time', injectAsync([TestComponentBuilder], (tcb) => {
-       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
-         const tc = fixture.componentInstance;
-         tc.closeOthers = true;
-         fixture.detectChanges();
+  xit('should only open one at a time', injectAsync([TestComponentBuilder], (tcb) => {
```
Foxandxss: no xit.

wesleycho:
Is there a particular reason for the removal of collapse?

Foxandxss:
I think he wants the ngbPanel to be the collapse (they do the same thing) but I think this PR is half baked or lacks proper explanation :P

pkozlowski-opensource:
Here is some more context about the proposed refactor: if you look at the HTML structure of bs4 accordion, you will notice that tis is just a bunch of things called "panels". A panel can have a header + title, collapsible content and (optionally) a footer. We haven't had "panel" as a stand-alone directive but this PR introduces one. The reasoning here is that we should be building higher-level components from the lower-level building blocks.

Now, a panel can be collapsible and we might have it as a separate directive as well - this is one of the things I wanted to debate.

src/accordion/accordion.spec.ts
pkozlowski-opensource: yeh, for whatever reason I didn't mange to make this test behave as expected.... On it right now...

pkozlowski-opensource:
@wesleycho @Foxandxss this should be ready for review, I'm done with all the code changes and test. Hopefully my previous explanations give a bit more context but please let me know if you need more info.

wesleycho:
I guess I'm fine with this if this is the way we want to go - I am worried about panels potentially becoming a mega component though.

pkozlowski-opensource:
> I am worried about panels potentially becoming a mega component though.

I'm totally cool with extracting the `collapse` directive as we used to have. But I would like to be able to use panels stand-alone as well.

Foxandxss:
Why remove isOpen from the panels?

And I think I like our collapse, we could need it for other stuff.

pkozlowski-opensource:
> Why remove isOpen from the panels?

It was rename to `open` but otherwise still part of a panel: https://github.com/ng-bootstrap/core/pull/167/files#diff-b50f479a3cf0e488ce11076b561c5993R50

> And I think I like our collapse, we could need it for other stuff.

OK, will add it back. Out of curiosity - what is the "other stuff" you've got on your mind?

Foxandxss:
The diffs annoy me. Yeah, I see it now.

Other stuff... Well, isn't there any other component that could be collapsed?

pkozlowski-opensource:
@Foxandxss `NgbCollapse` is back. Anything else or is this one good to go?

Foxandxss:
Yeah, merge it. I wanted to work with this a bit to add more tests or something (I have a few things in my TODO).

PR #170, opened 2016-04-04
TEST, DO NOT MERGE

pkozlowski-opensource:
Opening to trigger tests as those are not passing locally on my end, testing in TravisCI


pkozlowski-opensource:
Oh, I see what is going on - we've got tests passing in FFox but not in Chrome! (at least not in 50.0.2661.18 on Linux). Will investigate, although from my first checks I'm not sure how things can work in FFox...

PR #172, opened 2016-04-07
test(pagination): make tests pass on Chrome

pkozlowski-opensource:
Don't ask me why but without this change pagination tests are failing in Chrome (perfectly fine in FFox, thus not visible on TravisCI)

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -41,6 +41,7 @@ describe('ngb-dropdown', () => {
        });
      }));
 
+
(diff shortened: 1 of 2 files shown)


Foxandxss:
I'll try later when I get home, it is really weird.

wesleycho:
Works for me, although sounds like a potential ng2 bug here.

Foxandxss:
It works for me perfect Pawel, even in chrome (locally). Are you telling me that running "gulp" locally fails for you?

pkozlowski-opensource:
I've just updated Chrome on my Ubuntu box and tests are passing now :-/ Must have been particular Chrome version messing up things. Closing for now.

PR #174, opened 2016-04-15
feat(NgbDialog): add the NgbDialog helper service

pkozlowski-opensource:
The NgbDialog service can be used to dynamically load
components next to the body element. This is an internal
service to be used by modals, popovers, tooltips etc.

### DIFF
--- a/src/dialog/dialog.ts
+++ b/src/dialog/dialog.ts
@@ -1,0 +1,40 @@
+import {Injectable, ComponentRef, DynamicComponentLoader, Injector, Component} from 'angular2/core';
+
+const CONTAINER_TAG_NAME = 'ngb-dialog-container';
+
+@Component({selector: CONTAINER_TAG_NAME, template: ''})
+class NgbDialogContainer {
+}
+
+export class NgbDialogOpenResult {
+  constructor(public componentRefs: ComponentRef[]) {}
+
+  disposeAll() {
+    this.componentRefs.forEach((component) => { component.dispose(); });
+  }
+}
+
+@Injectable()
+export class NgbDialog {
+  private _container: Promise<ComponentRef>;
+
+  constructor(private _dcl: DynamicComponentLoader, private _injector: Injector) {
+    var el = document.querySelector(CONTAINER_TAG_NAME);
+
+    if (!el) {
+      document.body.appendChild(document.createElement(CONTAINER_TAG_NAME));
+    }
+
+    this._container = this._dcl.loadAsRoot(NgbDialogContainer, CONTAINER_TAG_NAME, _injector);
+  }
+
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
To make it easier for everyone to review modal implementation I've decided to split it into smaller chunks that should be manageable.

The first PR is a supporting service that allows us to load / render components into a container close to a body. It is needed for Modal but also necessary for appending tooltips / popovers / other dialogs to `body`.

src/dialog/dialog.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Component({selector: CONTAINER_TAG_NAME, template: ''})
+class NgbDialogContainer {
+}
+
+export class NgbDialogOpenResult {
+  constructor(public componentRefs: ComponentRef[]) {}
+
+  disposeAll() {
+    this.componentRefs.forEach((component) => { component.dispose(); });
```
wesleycho: Do we need to set `this.componentRefs = []` after or do some sort of gc?

Foxandxss:
I don't understand half of the injections done there but I get the idea.

wesleycho:
Question - how does the backdrop get removed?

pkozlowski-opensource:
> I don't understand half of the injections done there but I get the idea.

@Foxandxss feel free to ask any specific questions if this would help clarify things.

> Question - how does the backdrop get removed?

@wesleycho it will be removed by the `Modal` service (PR tomorrow). Think about the component from this PR as about a "dumb" component that just needs to display sth - it doesn't know nor care who creates / destroys it.

src/dialog/dialog.ts
pkozlowski-opensource: `NgbDialogOpenResult` itself shouldn't be (too) long-living, so not sure it matters. Also I don't know how "expensive" `ComponentRef` is...

In short: I don't know, but adding it wouldn't hurt.

src/dialog/dialog.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Injectable()
+export class NgbDialog {
+  private _container: Promise<ComponentRef>;
+
+  constructor(private _dcl: DynamicComponentLoader, private _injector: Injector) {
+    var el = document.querySelector(CONTAINER_TAG_NAME);
+
+    if (!el) {
+      document.body.appendChild(document.createElement(CONTAINER_TAG_NAME));
```
wesleycho: Should we switch to use the new Renderer service for the DOM appending?

pkozlowski-opensource:
`DynamicComponentLoader` is depreciated and now I know how to do it in an easier way so closing this PR. Will have an alternative for review.

PR #175, opened 2016-04-15
feat(NgbModalBackdrop): add NgbModalBackdrop needed for Modal support

pkozlowski-opensource:
This one should be fairly easy to review :-)

### DIFF
--- a/src/modal/modal_backdrop.spec.ts
+++ b/src/modal/modal_backdrop.spec.ts
@@ -1,0 +1,26 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  injectAsync,
+  TestComponentBuilder,
+  beforeEachProviders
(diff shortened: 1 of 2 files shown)


wesleycho:
LGTM for an initial start - we'll probably want to implement backdrop clicking logic later for dismissing the modal, but that can come later.

pkozlowski-opensource:
> Oh, I see the backdrop logic is in the modal window

Exactly - if I understand things correctly the HTML structure is different in B4.

Foxandxss:
Do we need a component for something without template?

pkozlowski-opensource:
> Do we need a component for something without template?

We can't dynamically load directives, so pretty much yes.... I'm sure that there are alternative designs so I'm open to other suggestions.

Foxandxss:
Didn't know about that limitation. Good for me then.

PR #176, opened 2016-04-18
feat(NgbModalWindow): add NgbModalWindow component

pkozlowski-opensource:
NgbModalWindow is needed for modals support (Modal service).

### DIFF
--- a/src/modal/modal_window.ts
+++ b/src/modal/modal_window.ts
@@ -1,0 +1,57 @@
+import {
+  Component,
+  Output,
+  EventEmitter,
+  ElementRef,
+  DynamicComponentLoader,
+  Input,
+  HostListener,
+  ComponentRef
+} from "angular2/core";
+import {ModalDismissReasons} from './modal_dismiss_reasons';
+
+@Component({
+  selector: 'ngb-modal-window',
+  host: {"class": "modal", "role": "dialog", "tabindex": "-1", "style": "display: block;"},
+  template: `
+        <div [class]="'modal-dialog' + (size ? ' modal-' + size : '')">
+            <div class="modal-content" (click)="stopPropagation($event)">
+                <content-sibiling #content></content-sibiling>                          
+            </div>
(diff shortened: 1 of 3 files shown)


src/modal/modal_window.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (this.keyboard) {
+      this.dismiss(ModalDismissReasons.ESC);
+    }
+  }
+
+  close(result: any) { this.closeEvent.emit(result); }
+
+  dismiss(reason) { this.dismissEvent.emit(reason); }
+
+  stopPropagation($event) { $event.stopPropagation(); }
```
wesleycho: Can we type the event, or is that a TS limitation?

src/modal/modal_window.ts
```diff
@@ -0,0 +1,46 @@
...
+  @Input() keyboard = true;
+  @Input() size: string;
+
+  @Output('close') closeEvent = new EventEmitter();
+  @Output('dismiss') dismissEvent = new EventEmitter();
+
+  constructor(private _elRef: ElementRef, private _dcl: DynamicComponentLoader) {}
+
+  @HostListener('click')
+  backdropClick() {
```
wesleycho: I'm a bit confused by this - on click, it will run this function? Seems to be a bit of confusing naming here.

pkozlowski-opensource: Which name is confusing? `backdropClick`? Can you think of an alternative proposal? `backdropClickHandler`?

Foxandxss: It is fine for me.

src/modal/modal_window.ts
Foxandxss: IIRC you can type it as MouseEvent or something like that.

pkozlowski-opensource: Done

wesleycho:
LGTM

PR #177, opened 2016-04-19
chore: swap tests to webpack

Foxandxss:
So since I created all the tooling, I learnt a lot of new cool stuff so I am going to open a few PRs over the week.

First one, swap testing to karma.

Pros:

* All testing is done in memory, that means:
* No temp folder, so no more having to worry about stale code there.
* No need to have a transpilation in background for tdd mode.
* Sourcemaps? (not sure if we had that before)

* coverage (not here yet)
* proper linting (not here yet)

Cons:

* Our source needs to be webpack compatible, not an issue right now tho. Will see what the future bring us, but no problem as today.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -58,8 +58,6 @@ gulp.task('umd', function(cb) {
 
 // Testing
 
-var testProject = ts.createProject('tsconfig.json');
-
 function startKarmaServer(isTddMode, done) {
   var karmaServer = require('karma').Server;
   var travis = process.env.TRAVIS;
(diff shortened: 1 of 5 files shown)


pkozlowski-opensource:
I must say that it looks great after the first, short look. Will take it for a test-drive tomorrow. Thnx!

icfantv:
This is great stuff.  I thought we were going to switch away from gulp?  Am I not remembering correctly?

Foxandxss:
You are not remembering correctly hah.

We are not moving away, it is handy for libraries.

gulpfile.js
```diff
@@ -74,28 +72,16 @@ function startKarmaServer(isTddMode, done) {
...
-  return tsResult.js.pipe(sourcemaps.write('.')).pipe(gulp.dest('temp'));
-});
-
-gulp.task('clean:build-tests', function(done) { runSequence('clean:tests', 'build:tests', done); });
-
 gulp.task(
     'ddescribe-iit', function() { return gulp.src(PATHS.specs).pipe(ddescribeIit({allowDisabledTests: false})); });
 
-gulp.task('test', ['clean:build-tests'], function(done) { startKarmaServer(false, done); });
+gulp.task('test', function(done) { startKarmaServer(false, done); });
```
wesleycho: Not sure if it matters, but you could fat arrow this

wesleycho:
If it works, it's fine with me - it passes Travis with no issue.

gulpfile.js
Foxandxss: I could, but I am thinking in removing lot of this stuff, just don't tell Pawel.

pkozlowski-opensource:
So, I've just taken it for a test drive and I see 100% slowdown for running tests (both `test` and on each re-run with `tdd`). Here are the results from my machine:

Before:

```
[16:52:28] Finished 'test' after 2.68 s

real	0m5.816s
user	0m4.902s
sys	0m0.452s
```

With WebPack:

```
[16:51:02] Finished 'test' after 8.1 s

real	0m9.045s
user	0m8.973s
sys	0m0.707s
```

So it is ~5s slower on each file change. This is not much of a problem for `test` but super-annoying for `tdd`

pkozlowski-opensource:
@Foxandxss I hope there there is some WebPAck magic to speed things up, otherwise ~5s for `tdd` is noticable. I see the value in this change but I also see value in the smooth workflow.

wesleycho:
Probably the fact that Webpack/ts-loader is just slow.

icfantv:
@Foxandxss, what do you want to do with this PR?

Foxandxss:
For now, fire.

PR #179, opened 2016-04-21
feat(carousel): add NgbCarousel component

pkozlowski-opensource:
This is ready for review now!

### DIFF
--- a/src/carousel/carousel.spec.ts
+++ b/src/carousel/carousel.spec.ts
@@ -1,0 +1,319 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  fakeAsync,
+  discardPeriodicTasks,
+  tick,
+  beforeEach,
+  beforeEachProviders
+} from '@angular/core/testing';
+import {TestComponentBuilder} from '@angular/compiler/testing';
+
+import {By} from '@angular/platform-browser';
+import {Component} from '@angular/core';
+
+import {NgbCarousel, NgbSlide} from './carousel';
+
+function expectActiveSlides(nativeEl: HTMLDivElement, active: boolean[]) {
+  const slideElms = nativeEl.querySelectorAll('.carousel-item');
+  const indicatorElms = nativeEl.querySelectorAll('ol.carousel-indicators > li');
+
+  expect(slideElms.length).toBe(active.length);
+  expect(indicatorElms.length).toBe(active.length);
+
+  for (let i = 0; i < active.length; i++) {
+    if (active[i]) {
+      expect(slideElms[i]).toHaveCssClass('active');
+      expect(indicatorElms[i]).toHaveCssClass('active');
+    } else {
+      expect(slideElms[i]).not.toHaveCssClass('active');
+      expect(indicatorElms[i]).not.toHaveCssClass('active');
+    }
+  }
+}
+
+describe('ngb-carousel', () => {
+
+  it('should render slides and navigation indicators', fakeAsync(inject([TestComponentBuilder], (tcb) => {
+       const html = `
+      <ngb-carousel>
+        <template ngbSlide>foo</template>
+        <template ngbSlide>bar</template>
+      </ngb-carousel>
+    `;
+
+       tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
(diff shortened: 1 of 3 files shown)


src/carousel/carousel.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      expect(slideElms[i]).toHaveCssClass('active');
+      expect(indicatorElms[i]).toHaveCssClass('active');
+    } else {
+      expect(slideElms[i]).not.toHaveCssClass('active');
+      expect(indicatorElms[i]).not.toHaveCssClass('active');
+    }
+  }
+}
+
+describe('ngb-pagination', () => {
```
wesleycho: I think you meant `ngb-carousel` here :)

pkozlowski-opensource: Indeed, thnx!

pkozlowski-opensource:
I think that the implementation in this PR is the best I can do with my current state of knowledge. It is missing few things:
* keyboard navigation + tests for it
* tests for mouse enter / leave
* tests for set / reset interval

If anyone could give me a hand with those it would be much appreciated. If not I will research more next week.

Also I can't fully decide between 2 forms of the API:

```html
<ngb-carousel>
        <template ngb-slide>foo</template>
        <template ngb-slide>bar</template>
</ngb-carousel>
```

vs.

```html
<ngb-carousel>
        <ngb-slide>foo</ngb-slide>
        <ngb-slide>bar</ngb-slide>
</ngb-carousel>
```

I'm leaning more towards usage of `<template>` elements as we will have to use them in accordion / tabs to properly handle titles. But I want to hear your input!

wesleycho:
The template elements probably gives a little more flexibility for styling, although Bootstrap 4 is using better selectors compared to its predecessors (mostly class-based, as opposed to stuff like direct child element selectors found in areas of Bootstrap 3).

The downside of the template elements is that they aren't as nice to read in a template, which may be ok.

I don't have a huge opinion on this, if the template elements are more flexible, then I am fine with their usage.

src/carousel/carousel.ts
```diff
@@ -0,0 +1,118 @@
...
+  QueryList,
+  Input,
+  OnDestroy,
+  ViewRef,
+  ViewContainerRef,
+  AfterContentChecked,
+  OnInit
+} from 'angular2/core';
+
+// TODO: remove when https://github.com/angular/angular/commit/f4e6994634a9cf6c2c20aa8c2268233e2667edf8 is released
```
wesleycho: This TODO comment can be removed I think

wesleycho: Oh, ignore comment

wesleycho: This looks like it was released - should it be removed?

Foxandxss: Yes.

pkozlowski-opensource:
I've rebased to bring all the changes from the latest ng2 version but I still don't have tests running. I should figure it out tomorrow though, as I'm working with the ng team this week.

pkozlowski-opensource:
Proper testing solution is blocked on https://github.com/angular/angular/issues/8616

I can probably hack around it with all the help I got from Julie. Will see how bad the work-around is and if it is not too bad we will add it. If not we need to wait the fix.

pkozlowski-opensource:
https://github.com/angular/angular/issues/8616 got fixed in master so we should be unblocked soon. We just need to wait for rc.2 being released. As soon it is done I'm going to rebase this one and see how it goes.

src/carousel/carousel.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.select(prevIdx < 0 ? (this.wrap ? prevIdx + this._slides.length : prevIdx + 1) : prevIdx);
+    this._restartTimer();
+  }
+
+  next() {
+    this._cycleToNext();
+    this._restartTimer();
+  }
+
+  pause() { clearInterval(this._slideChangeInterval); }
```
wesleycho: Not a huge deal I think, but perhaps it'd be a good thing to null out the cleared interval so it can be marked for gc

pkozlowski-opensource: > to null out the cleared interval so it can be marked for gc

we could definitively do this but what we get from `setInterval` is a primitive (number) so I'm not sure there is much to GC here...

PR #180, opened 2016-04-21
chore: update travis for node 5

Foxandxss:
With frameworks and libraries moving to give only node 5 support and npm3, I prefer to bump travis as well to avoid surprises.

Also I think that all those extra options are not needed. Let say what travis has to say about this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,6 @@
 language: node_js
 node_js:
-  - "4.2.1"
-env:
-  - CXX=g++-4.8
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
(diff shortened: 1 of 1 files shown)


wesleycho:
Works for me, but the extra options should go back
```
Starting with io.js 3 and Node.js 4, building native extensions requires C++11-compatible compiler, which seems unavailable on this VM. Please read https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements.
```

Foxandxss:
It is not compiling node but getting a binary.

pkozlowski-opensource:
LGTM

PR #181, opened 2016-04-25
feat(tabset): add tabset component

### LINKED ISSUE #7: Tab
Recreate BS3 Tab with ng2 component

pkozlowski-opensource:
Closes #7

### DIFF
--- a/src/tabset/tabset.ts
+++ b/src/tabset/tabset.ts
@@ -1,0 +1,58 @@
+import {
+  Component,
+  Input,
+  ContentChildren,
+  QueryList,
+  Directive,
+  TemplateRef,
+  ContentChild,
+  AfterContentChecked
+} from 'angular2/core';
+
+@Directive({selector: 'template[ngb-tab-title]'})
+export class NgbTabTitle {
+  constructor(public templateRef: TemplateRef) {}
+}
+
+@Directive({selector: 'template[ngb-tab-content]'})
+export class NgbTabContent {
+  constructor(public templateRef: TemplateRef) {}
+}
+
+@Directive({selector: 'ngb-tab'})
+export class NgbTab {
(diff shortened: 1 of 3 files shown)


src/tabset/tabset.ts
```diff
@@ -0,0 +1,80 @@
...
+  @ContentChild(NgbTabTitle) titleTpl: NgbTabTitle;
+}
+
+@Component({
+  selector: 'ngb-tabset',
+  template: `
+    <ul class="nav nav-tabs" role="tablist">
+      <li class="nav-item" *ngFor="#tab of tabs; #i=index">
+        <a class="nav-link" [class.active]="i === activeIdx" (click)="select(i)">
+          {{tab.title}}<template [ngbTemplateOutlet]="tab.titleTpl?.templateRef"></template>
```
Foxandxss: Wouldn't this allow both title input and titleTpl to exist both at once?

Foxandxss:
My only complain is that you are using `1` and `0` to test the status of the active tab and we are using `true` and `false` already in other tests and I personally prefer the latter.

src/tabset/tabset.ts
pkozlowski-opensource: Yup, this is "by design", see: https://github.com/ng-bootstrap/core/pull/181/files#diff-f4f98e2db742c4772ab831ae9adb2dacR69

Having said this I don't mind either way so if you / anyone else prefers other logic, please let me know.

pkozlowski-opensource:
Changing 0,1 to false,true.

Foxandxss:
Works for me.

pkozlowski-opensource:
Rebased on top of master / beta.16 and using built-in `ngTemplateOutlet` now. Please review.

PR #182, opened 2016-04-25
refactor(alert): drop template element in favour of *ngIf

pkozlowski-opensource:


pkozlowski-opensource:
Makes a slightly smaller template so less bytes to send over a wire.

Foxandxss:
LGTM

PR #183, opened 2016-04-26
chore: update angular2 to 2.0.0-beta.16

pkozlowski-opensource:
I had to remove modal bits as due to changes in DCL we need a new thinking / design; this will take few days and we might need to wait for a new DCL (next release?)


pkozlowski-opensource:
This update also revealed bugs in the core that we can work-arround but should track and remove work-arrounds at some point:
* https://github.com/angular/angular/issues/8242

PR #184, opened 2016-04-28
chore: add linting

Foxandxss:
I added linting to the project.

I was going to add codelyzer (a set of rules for angular 2) but that breaks the entire project, so I leave that out for now.

I also disabled the unused var because we have all those `iit` and stuff that are handy while developing and having to add the import and remove it all times is painy.


Foxandxss:
Although, we can remove those `iit`, `ddescribe` etc and use `fit` and `fdescribe` that gives the same result and don't need the import. Thoughts?

wesleycho:
I think this is the reason for the imports: https://github.com/angular/angular/blob/master/modules/angular2/src/testing/testing.ts#L75

Foxandxss:
Nah, I just meant the ones we don't actually use.

PR #185, opened 2016-04-29
feat(scrollspy): add scrollspy

wesleycho:
- Add an initial scrollspy mechanism

### DIFF
--- a/src/scrollspy/scrollspy.spec.ts
+++ b/src/scrollspy/scrollspy.spec.ts
@@ -1,0 +1,95 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  async,
+  beforeEach,
+  beforeEachProviders
+} from '@angular/core/testing';
+
+import {TestComponentBuilder} from '@angular/compiler/testing';
+
+import {Component, provide} from '@angular/core';
+
+import {NgbScrollSpier, NgbScrollSpy, NgbScrollTarget} from './scrollspy';
+
+function expectToBeActive(componentElement, idx) {
+  let anchors = componentElement.querySelectorAll('a');
+  Array.from(anchors).forEach((anchor, aIdx) => {
+    if (aIdx === idx) {
+      expect(anchor).toHaveCssClass('active');
+    } else {
+      expect(anchor).not.toHaveCssClass('active');
+    }
+  });
+}
+
+describe('ngb-scroll-spy', () => {
+  let html: string;
+
+  beforeEach(() => {
+    html = `
+      <div>
+        <navbar>
+          <nav>
+            <ul>
+              <li><a href [ngbScrollSpy]="selectors[0]">Anchor 1</a></li>
+              <li><a href [ngbScrollSpy]="selectors[1]">Anchor 2</a></li>
+              <li><a href [ngbScrollSpy]="selectors[2]">Anchor 3</a></li>
+            </ul>
+          </nav>
+        </navbar>
+        <section ngbScrollTarget>
+          <div id="foo" style="height: 500px; width: 100px"></div>
+          <div id="bar" style="height: 500px; width: 100px"></div>
+          <div id="baz" style="height: 500px; width: 100px"></div>
+        </section>
+      </div>
+    `;
+  });
+
+  beforeEachProviders(() => [provide(NgbScrollSpier, {useClass: NgbScrollSpier})]);
+
+  it('should have no active class', async(inject([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         let anchors = fixture.nativeElement.querySelectorAll('a');
+         Array.from(anchors).forEach((anchor: Element) => { expect(anchor.classList.contains('active')).toBe(false); });
+       });
+     })));
+
+  it('should set the first anchor class as active', async(inject([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         fixture.nativeElement.querySelector('[ngbScrollTarget]').scrollTop = 10;
+         let evt = new Event('scroll');
+         fixture.nativeElement.dispatchEvent(evt);
+         fixture.detectChanges();
+
+         expectToBeActive(fixture.nativeElement, 0);
+       });
+     })));
+
+  it('should set the second anchor class as active', async(inject([TestComponentBuilder], (tcb) => {
+       return tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         fixture.nativeElement.querySelector('[ngbScrollTarget]').scrollTop = 501;
+         let evt = new Event('scroll');
+         fixture.nativeElement.dispatchEvent(evt);
+         fixture.detectChanges();
+
+         expectToBeActive(fixture.nativeElement, 1);
+       });
+     })));
+});
+
+@Component({selector: 'test-cmp', directives: [NgbScrollSpy, NgbScrollTarget], template: ''})
+class TestComponent {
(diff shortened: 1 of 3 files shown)


src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,42 @@
+import {Directive, ElementRef, Injectable, Input, OnDestroy, OnInit} from 'angular2/core';
+
+@Directive({selector: '[ngb-scroll-spy]', host: {'[class.active]': 'active'}, exportAs: 'ngbScrollSpy'})
```
Foxandxss: pascalCase for the selector.

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,42 @@
...
+  @Input() ngbScrollSpy: string;
+
+  constructor(private scrollSpier: NgbScrollSpier) {}
+
+  ngOnDestroy(): void { this.scrollSpier.deregister(this); }
+
+  ngOnInit(): void { this.scrollSpier.register(this); }
+}
+
+@Directive({selector: '[ngb-scroll-target]', exportAs: 'ngbScrollTarget'})
```
Foxandxss: ditto

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    spy.active = target.offsetTop <= element.scrollTop && target.offsetTop + target.offsetHeight >= element.scrollTop;
+  }
+
+  ngOnInit(): void { this.elementRef.nativeElement.addEventListener('scroll', (e) => this.checkSpies(e)); }
+}
+
+@Injectable()
+export class NgbScrollSpier {
+  public spies: Set<NgbScrollSpy> = <Set<NgbScrollSpy>>new Set();
```
Foxandxss: public is still a default, so I am not sure if it is really needed. Also, you don't need to do `Set<NgbScrollSpy>` twice, or at least, I hope so.

wesleycho: You do, because TypeScript types `new Set()` as `Set<{}>` :(

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    spy.active = target.offsetTop <= element.scrollTop && target.offsetTop + target.offsetHeight >= element.scrollTop;
+  }
+
+  ngOnInit(): void { this.elementRef.nativeElement.addEventListener('scroll', (e) => this.checkSpies(e)); }
+}
+
+@Injectable()
+export class NgbScrollSpier {
+  spies: Set<NgbScrollSpy> = new Set<NgbScrollSpy>();
```
Foxandxss: Since we initialize it, we don't need the type.

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,44 @@
...
+        spy.elementRef.nativeElement, 'active',
+        target.offsetTop <= element.scrollTop && target.offsetTop + target.offsetHeight >= element.scrollTop);
+  }
+
+  ngOnInit(): void { this.renderer.listen(this.elementRef.nativeElement, 'scroll', (e) => this.checkSpies(e)); }
+}
+
+@Injectable()
+export class NgbScrollSpier {
+  spies = new Set<NgbScrollSpy>();
```
wesleycho: It would be nice to use a WeakSet for this, but I assume we should stay away from that ES6 feature until we're supporting Edge and up.

wesleycho:
I've no clue why the tests are failing - the stack trace is completely useless. As far as I've been able to determine, the error is triggered once `fixture.detectChanges()` is fired.

wesleycho:
So I updated the PR to the latest on `master`, and now the error has shifted. It seems to be triggered by the act of compiling the components.

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  checkSpy(spy: NgbScrollSpy, evt: Event) {
+    let element = this.elementRef.nativeElement;
+    let target = element.querySelector(spy.selector);
+
+    if (!target) {
+      return;
+    }
+
+    this.renderer.setElementClass(
```
pkozlowski-opensource: Shouldn't we have a method like 'setActive(...)' on the `NgbScrollSpy` instead? Just an idea...

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@Directive({selector: '[ngbScrollSpy]'})
+export class NgbScrollSpy implements OnDestroy, OnInit {
+  active: boolean;
+  selector: string;
+
+  constructor(public elementRef: ElementRef, private scrollSpier: NgbScrollSpier) {}
+
+  @Input()
+  set ngbScrollSpy(selector: string) {
```
pkozlowski-opensource: Why are we using a getter / setter here instead of `@Input('ngbScrollSpy') selector: string;`?

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,51 @@
+import {Directive, ElementRef, Injectable, Input, OnDestroy, OnInit, Renderer} from '@angular/core';
+
+@Directive({selector: '[ngbScrollSpy]'})
+export class NgbScrollSpy implements OnDestroy, OnInit {
+  active: boolean;
```
pkozlowski-opensource: Is `active` used anywhere? Doesn't look like so...

wesleycho:
Updated as per comments, but those irksome tests are still failing - I'll have to look into this deeper some.

src/scrollspy/scrollspy.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+          <div id="bar" style="height: 500px; width: 100px"></div>
+          <div id="baz" style="height: 500px; width: 100px"></div>
+        </section>
+      </div>
+    `;
+  });
+
+  beforeEachProviders(() => [provide(NgbScrollSpier, {useClass: NgbScrollSpier})]);
+
+  it('should have no active class', async(inject([TestComponentBuilder], (tcb) => {
```
pkozlowski-opensource: Test name should say under which conditions there should be no active class. The way it reads now (`'should have no active class'`) makes me think that it should never have an active class. Are you trying to say something like `'should have no active class when no elements spied on are visible'`?

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,48 @@
...
+    let target = element.querySelector(spy.selector);
+
+    if (!target) {
+      return;
+    }
+
+    spy.setActive(target);
+  }
+
+  ngOnInit(): void { this.renderer.listen(this.elementRef.nativeElement, 'scroll', () => this.checkSpies()); }
```
pkozlowski-opensource: Maybe we should think of some de-bouncing here? Otherwise we run pretty heavy operation here (reading positional information causing re-flows +  querySelector on each scroll).

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,48 @@
...
+
+@Directive({selector: '[ngbScrollTarget]'})
+export class NgbScrollTarget implements OnInit {
+  constructor(private elementRef: ElementRef, private renderer: Renderer, private scrollSpier: NgbScrollSpier) {}
+
+  checkSpies() { this.scrollSpier.spies.forEach(spy => this.checkSpy(spy)); }
+
+  checkSpy(spy: NgbScrollSpy) {
+    let element = this.elementRef.nativeElement;
+    let target = element.querySelector(spy.selector);
```
pkozlowski-opensource: Maybe we should explore different API where we don't need to use `querySelector`? How about making each spy target a directive?

pkozlowski-opensource:
@wesleycho added more comments to the PR - some of them probably don't make sense as I'm not super familiar with those scroll spies. Maybe we could jump on a hangout to discuss various API / impl options?

pkozlowski-opensource:
OK so this one doesn't merge cleanly any more and IMO we need some more conceptual work on it so going to close this PR for now. @wesleycho let me know if / when you want to chat about next steps.

PR #187, opened 2016-05-09
chore: update demo project

Foxandxss:
Having a different repo for the docs was a dead end. Webpack will get crazy when trying to import stuff from a different project.

I updated what we got using a simplified version of my starter. No lint, no test. Will add those when the time is right, but we need something to show like for yesterday.

As soon as I merge this, I will make 2-3 demos so @icfantv can play with the CSS


icfantv:
That really stinks.  So we have to have everything under core?  What happens when we break out components?  Is that going to cause the same issue?

Foxandxss:
We will cross that bridge when that happens.

Perhaps when we push this repo to npm, we can install it into the demo project and work with it, but as today, I don't find a clean way.

webpack.demo.js
```diff
@@ -1,39 +1,269 @@
...
+  }
+
+  /**
+   * PostCSS
+   * Reference: https://github.com/postcss/autoprefixer-core
+   * Add vendor prefixes to your css
+   */
+  config.postcss = [
+    autoprefixer({
+      browsers: ['last 2 version']
```
wesleycho: Isn't this config default with autoprefixer?

webpack.demo.js
```diff
@@ -1,39 +1,269 @@
...
+    historyApiFallback: true,
+    stats: 'minimal' // none (or false), errors-only, minimal, normal (or true) and verbose
+  };
+
+  return config;
+}();
+
+// Helper functions
+function root(args) {
+  args = Array.prototype.slice.call(arguments, 0);
```
wesleycho: No need for this, you can do `function root(...args) { ... }` - remember, webpack runs in Node, you have ES6!

webpack.demo.js
```diff
@@ -1,39 +1,269 @@
...
+  return path.join.apply(path, [__dirname].concat(args));
+}
+
+function rootNode(args) {
+  args = Array.prototype.slice.call(arguments, 0);
```
wesleycho: Ditto

webpack.demo.js
```diff
@@ -1,39 +1,269 @@
...
+    stats: 'minimal' // none (or false), errors-only, minimal, normal (or true) and verbose
+  };
+
+  return config;
+}();
+
+// Helper functions
+function root(args) {
+  args = Array.prototype.slice.call(arguments, 0);
+  return path.join.apply(path, [__dirname].concat(args));
```
wesleycho: You can do `[__dirname].push(...args)` here - this is the spread operator

webpack.demo.js
```diff
@@ -1,39 +1,269 @@
...
+}
+
+function rootNode(args) {
+  args = Array.prototype.slice.call(arguments, 0);
+  return root.apply(path, ['node_modules'].concat(args));
```
wesleycho: `['node_modules'].push(...args)`

icfantv:
@Foxandxss just lemme know when you're ready and I'll get started

wesleycho:
After discussion, this LGTM

icfantv:
I'm ok with this.

PR #188, opened 2016-05-09
docs: accordion demo

Foxandxss:
Werk in progress.

I need to tidy up the code and stuff like that (the templates is another subject).

This setup will allow multiple demos per component (to show different features).

I tried to extract as much as possible in the json so in the future we can make this more generic.

Also notice how I imported the accordion in the demo (@ng-bootstrap/accordion) which is how we are going to use it in prod, so better to show in the demo that way. This alias won't work twice, so I will need to make changes on the core itself so each component is exported. We will probably need something like that for the scoped packages. I will think about this.

Probably a few things are horrible, but I am asleeeep.

### DIFF
--- a/demo/src/app/app.component.ts
+++ b/demo/src/app/app.component.ts
@@ -5,6 +5,8 @@ import { HomeComponent } from './home';
 import { AccordionComponent } from './components/accordion';
 import { SideNavComponent } from './shared';
 
+import 'prismjs/themes/prism-okaidia.css';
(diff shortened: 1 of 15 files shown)


Foxandxss:
Ah, I managed the alias thing so every package can be imported in that way for the docs without the need of touching the codebase.

icfantv:
For the path thing we saw Sunday?  Nice.

wesleycho:
LGTM

icfantv:
LGTM.

PR #189, opened 2016-05-09
refactor(pagination): bring back ChangeDetectionStrategy.OnPush

pkozlowski-opensource:
https://github.com/angular/angular/issues/8242 was fixed so we can go back to using `ChangeDetectionStrategy.OnPush`


Foxandxss:
wops, hadn't enough coffee and thought that the option to not add merge commits was on but was the contrary.

PR #190, opened 2016-05-10
feat(accordion): add aria-labelledby attributes

pkozlowski-opensource:
This is a little PR that adds `aria-labelledby` support and - more importantly - should serve as an example  for generating "unique" ids.


wesleycho:
LGTM

PR #192, opened 2016-05-12
feat(modal): bring back modal backdrop and window

pkozlowski-opensource:
This PR brings back modal's window and backdrop components previously removed in 1f34c8781dca3d371e4543917834b4846fb6a292. The good news is that we don't need _that_ much gymnastic to load modal's content. By using a technique demonstrated [here](https://github.com/ng-bootstrap/core/issues/2#issuecomment-218851066) we can use regular `<ng-content>` and insert content synchronously.

### DIFF
--- a/src/modal/modal_window.spec.ts
+++ b/src/modal/modal_window.spec.ts
@@ -1,0 +1,94 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  async,
+  beforeEach,
+  beforeEachProviders
+} from '@angular/core/testing';
+import {TestComponentBuilder} from '@angular/compiler/testing';
+
+import {NgbModalWindow} from './modal_window';
+import {ModalDismissReasons} from "./modal_dismiss_reasons";
+
+describe('ngb-modal-dialog', () => {
+
+  describe('basic rendering functionality', () => {
+
+    it('should render default modal window', async(inject([TestComponentBuilder], (tcb) => {
+         tcb.createAsync(NgbModalWindow).then((fixture) => {
+           fixture.detectChanges();
+
+           var modalEl: Element = fixture.nativeElement;
+           var dialogEl: Element = fixture.nativeElement.querySelector('.modal-dialog');
+
+           expect(modalEl).toHaveCssClass('modal');
+           expect(dialogEl).toHaveCssClass('modal-dialog');
+         });
+       })));
+
+    it('should render default modal window with a specified size', async(inject([TestComponentBuilder], (tcb) => {
+         tcb.createAsync(NgbModalWindow).then((fixture) => {
+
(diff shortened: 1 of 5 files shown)


icfantv:
This looks really good.  Nice work Pawel.  I really like that we can emit and trap [custom] named events.  That's really cool.

src/modal/modal_window.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @HostListener('keyup.esc')
+  escKey(): void {
+    if (this.keyboard) {
+      this.dismiss(ModalDismissReasons.ESC);
+    }
+  }
+
+  close(result: any): void { this.closeEvent.emit(result); }
```
wesleycho: Don't need `: any` here

src/modal/modal_window.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+describe('ngb-modal-dialog', () => {
+
+  describe('basic rendering functionality', () => {
+
+    it('should render default modal window', async(inject([TestComponentBuilder], (tcb) => {
+         tcb.createAsync(NgbModalWindow).then((fixture) => {
+           fixture.detectChanges();
+
+           var modalEl: Element = fixture.nativeElement;
+           var dialogEl: Element = fixture.nativeElement.querySelector('.modal-dialog');
```
wesleycho: Would it make more sense to do `modalEl.querySelector('.modal-dialog')`?

src/modal/modal_window.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+        fixture.componentInstance.backdrop = false;
+        fixture.detectChanges();
+
+        fixture.componentInstance.dismissEvent.subscribe(($event) => {
+          expect($event).toBe(ModalDismissReasons.BACKDROP_CLICK);
+          done(new Error('Should not trigger dismiss event'));
+        });
+
+        fixture.nativeElement.querySelector('.modal-dialog').click();
+        setTimeout(done, 200);
```
wesleycho: Is this `setTimeout` necessary? Might be better to use a variable that is toggled in the subscription then check if it has been called. I guess the caveat might be whether the subscriber is fired async (I'm an observable newbie still)

src/modal/modal_window.spec.ts
pkozlowski-opensource: I would like to keep those 2 vars independent.

src/modal/modal_window.spec.ts
pkozlowski-opensource: Yeh, precisely, in rc.1 events are async. I think it changes in rc.2 so we might be able to get rid of this setTimeout then.

pkozlowski-opensource:
Thnx for the review @wesleycho - I've addressed / commented on your comments.

PR #193, opened 2016-05-12
refactor(tabset): use proper camelCase directives

Foxandxss:
Hey @pkozlowski-opensource correct me if I am wrong, but shouldn't we code tabset like this?


pkozlowski-opensource:
You are probably right. What do we have in the style guide for this?

Foxandxss:
"Do Use lower camel case for naming the selectors of our directives."

We normally put all attribute selectors as camelCase, but this one is a bit of an advanced case.

Still, I see stuff like `<template ngSwitchWhen...>`. It is not the saaame selector a the end, but still camelCase.

PR #194, opened 2016-06-02
Popover / tooltip windows

pkozlowski-opensource:
Those changes add wrapper components for tooltips / popovers. I'm still struggling with dynamic loading (see part of the conversation on the topic in https://github.com/angular/angular/issues/8941) but we will need those wrappers anyway. 

Hopefully we will start connecting all those elements together soonish :-)

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -1,0 +1,16 @@
+import {Component, Input, ChangeDetectionStrategy} from '@angular/core';
+
+@Component({
+  selector: 'ngb-popover-window',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {'[class]': '"popover " + placement', 'role': 'tooltip'},
+  template: `
+    <div class="popover-arrow"></div>
+    <h3 class="popover-title">{{title}}</h3>
+    <div class="popover-content"><ng-content></ng-content></div>
+    `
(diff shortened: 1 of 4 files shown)


src/popover/popover.ts
```diff
@@ -0,0 +1,16 @@
+import {Component, Input, ChangeDetectionStrategy} from '@angular/core';
+
+@Component({
+  selector: 'ngb-popover-window',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {'[class]': '"popover " + placement', 'role': 'tooltip'},
```
Foxandxss: Wouldn't this override all classes used in the component?

pkozlowski-opensource: huh? This is just a popover decorator, a component that we would be opening ourselves (user's content go into ng-content) so I'm not sure what you mean...

Foxandxss: Oh right, I meant that if you do `<ngb-popover-window class="foo bar baz"...` those 3 classes would be removed to add the popover-placement one.

pkozlowski-opensource: Correct. It is just that we are never going to do this in practice.

Foxandxss: 👍

PR #198, opened 2016-06-10
feat(triggers): introduce triggers parsing utility

pkozlowski-opensource:
Needed for tooltips / popovers as described in:
https://github.com/ng-bootstrap/core/issues/8#issuecomment-225168469

### DIFF
--- a/src/util/triggers.spec.ts
+++ b/src/util/triggers.spec.ts
@@ -1,0 +1,100 @@
+import {
+  iit,
+  it,
+  ddescribe,
+  describe,
+  expect,
+  inject,
+  injectAsync,
+  beforeEach,
+  beforeEachProviders
+} from '@angular/core/testing';
+
+import {parseTriggers} from './triggers';
+
+describe('triggers', () => {
+
+  describe('parseTriggers', () => {
+
+    it('should parse single trigger', () => {
+      var t = parseTriggers('foo');
+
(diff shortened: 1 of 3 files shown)


src/util/triggers.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    it('should throw when more than one manual trigger detected', () => {
+      expect(() => {
+        parseTriggers('manual click manual');
+      }).toThrow('Triggers parse error: only one manual trigger is allowed');
+    });
+
+    it('should throw when manual trigger is mixed with other triggers', () => {
+      expect(() => {
+        parseTriggers('click manual');
+      }).toThrow('Triggers parse error: manual trigger can\'t be mixed with other triggers');
```
wesleycho: Could use a template string here to avoid escaping

wesleycho:
Probably should be a test to test it handling strings with newline characters appropriately as well.

pkozlowski-opensource:
> Probably should be a test to test it handling strings with newline characters appropriately as well.

I don't think ng2 allows new lines in attribute values (HTML5 spec probably doesn't allow them either).

wesleycho:
It does - I actually implemented that to fix an issue reported against lack of support against that.

pkozlowski-opensource:
@wesleycho you are right! Will add a test for newline characters case. Thnx!

PR #216, opened 2016-06-15
feat(radio): add radio buttons

maxokorokov:
Adding radio buttons component

```html
<div class="btn-group" data-toggle="buttons" [(ngModel)]="model" ngb-radio-group>
    <label class="btn btn-primary">
        <input type="radio" name="rad2" [value]="one" ngb-radio>One
    </label>
    <label class="btn btn-primary">
        <input type="radio" name="rad2" [value]="two" ngb-radio>Two
    </label>
</div>
```

Behaviour:
- single model
- can use arbitrary values, not only strings
- when a model is updated => proper radio gets selected
- when a selected radio value is updated => model is updated
- when a selected radio is hidden => model sets to null

Here is the updated plunker:
http://plnkr.co/edit/hHNhl8lBgZrMdOTY8jDj?p=preview

### DIFF
--- a/src/radio/radio.ts
+++ b/src/radio/radio.ts
@@ -1,0 +1,118 @@
+import {Directive, forwardRef, Provider, Host, Optional, Input, Renderer, ElementRef, OnDestroy} from '@angular/core';
+import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/common';
+
+const NGB_RADIO_VALUE_ACCESSOR =
+    new Provider(NG_VALUE_ACCESSOR, {useExisting: forwardRef(() => NgbRadioGroup), multi: true});
+
+@Directive({selector: '[ngb-radio-group][ngModel]', bindings: [NGB_RADIO_VALUE_ACCESSOR]})
+export class NgbRadioGroup implements ControlValueAccessor {
+  private _radios: Set<NgbRadio> = new Set<NgbRadio>();
+  private _selectedRadio: NgbRadio = null;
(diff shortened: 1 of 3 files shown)


src/radio/radio.ts
```diff
@@ -0,0 +1,116 @@
...
+      this.group.unregister(this);
+    }
+  }
+
+  onChange() {
+    if (this.group) {
+      this.group.onRadioChange(this);
+    }
+  }
+}
```
pkozlowski-opensource: could we add sth like `export const NGB_BUTTON_DIRECTIVES = [NgbRadio, NgbRadioLabel, NgbRadioGroup]` so people can add all those directives in one go?

Also  https://github.com/ng-bootstrap/core/blob/master/src/core.ts needs updating.

pkozlowski-opensource:
LGTM after quick scan, just left one minor comment.

@wesleycho @Foxandxss could you guys have a look as well?

wesleycho:
The only thing I'm wondering is in the tests, should the inputs have the same name? Currently they are missing it, but generally inputs in a set of radio inputs should have the same input name.

maxokorokov:
Nice catches, amended:
- core.ts update
- NGB_RADIO_DIRECTIVES export
- name attribute for tests
- double --> single quotes in typescript
- some comments for tests

pkozlowski-opensource:
Thnx @maxokorokov, merging! Great stuff, very high quality for the very first PR. Keep them coming!

PR #218, opened 2016-06-15
style(all): add types for all @Input members

pkozlowski-opensource:
This is not for API doc extraction.


Foxandxss:
It is not or it is?

pkozlowski-opensource:
Should be "This is needed for API doc extraction.", will amend

PR #225, opened 2016-06-16
docs: refactor structure a bit

Foxandxss:
A bit of a refactor, is not much, but I want to stop bikeshedding on that for a bit.

Now we can add effort into real docs.


icfantv:
Very nice.  LGTM.

PR #227, opened 2016-06-17
chore(docs): update get-doc

icfantv:
* update doc gen script to iterate over components and generate the component
  documentation JSON to the appropriate component's demo directory.

### DIFF
--- a/misc/get-doc.js
+++ b/misc/get-doc.js
@@ -1,6 +1,43 @@
-var doc = require('./api-doc');
+const fs = require('fs');
+const doc = require('./api-doc');
+const excluded = new Set([
+  'util'
+]);
+
+function isDirectory(path) {
+  return fs.lstatSync(path).isDirectory();
+}
 
 function getFileNames(directiveName) {
   return [`./src/${directiveName}/${directiveName}.ts`];
 }
-console.log(JSON.stringify(doc(getFileNames('tabset'))));
+
+function fileExists(file) {
+  let result = false;
+  try {
+    fs.accessSync(file);
+    result = true;
+  }
+  catch (e) {
+    console.error(`the specified file does not exist: ${file}`)
+  }
+
+  return result;
+}
+
+function getComponentMap() {
+  const componentMap = new Map();
+  const items = fs.readdirSync('./src');
+  items.forEach(component => {
+    if (!excluded.has(component) && isDirectory(`./src/${component}`)) {
+      const names = getFileNames(component);
+      if (names.every(name => fileExists(name))) {
+        componentMap.set(component, doc(names));
+      }
+    }
+  });
+
+  return componentMap;
+}
+
+module.exports = getComponentMap;


pkozlowski-opensource:
As commented on slack, I don't think it is the best approach - we should rather generate JSON files in memory, without creating temp files. Also I would scan fs for components instead of hard-coding a list of them.

I can help / we can pair on wiring doc generation with this script.

masaanli:
Do you guys have an public slack channel?

pkozlowski-opensource:
> Do you guys have an public slack channel?

Nope, sorry, not at the moment :-/ But you can always reach out to use here on GH

icfantv:
@pkozlowski-opensource, I went the scan route originally, but we have that `utils` directory in there along with our components.  If we intend for those to be public and supported then scanning is fine.  Thoughts?

pkozlowski-opensource:
> I went the scan route originally, but we have that utils directory in there along with our components

Exclude `utils` for now. Still I think it is better to have one special case (exclude `utlis`) as compared to listing all directives by hand.

icfantv:
Agreed.  It certainly scales better.

icfantv:
@pkozlowski-opensource updated to be in memory and to filter out utils.  right now the `get-doc.js` script holds the component map - is that ok or do we want to pull it out and store it elsewhere?

misc/get-doc.js
```diff
@@ -1,6 +1,38 @@
...
+  return result;
+}
+
+const componentMap = new Map();
+const items = fs.readdirSync('./src');
+items.forEach(component => {
+  if (!excluded.has(component) && isDirectory(`./src/${component}`)) {
+    const names = getFileNames(component);
+    if (names.every(name => fileExists(name))) {
+      componentMap.set(component, JSON.stringify(doc(names)));
```
pkozlowski-opensource: Why are you calling `JSON.stringify` here? Isn't it going to make your life harder when generating API docs?

pkozlowski-opensource:
>  right now the get-doc.js script holds the component map - is that ok or do we want to pull it out and store it elsewhere?

Personally I wouldn't store it anywhere - would just have a function call that returns data. But it is your call.

misc/get-doc.js
icfantv: wasn't thinking.  will remove.

icfantv:
noted.

icfantv:
updated.

misc/get-doc.js
```diff
@@ -1,6 +1,43 @@
...
+    result = true;
+  }
+  catch (e) {
+    console.error(`the specified file does not exist: ${file}`)
+  }
+
+  return result;
+}
+
+function getComponentMap() {
```
pkozlowski-opensource: @icfantv I think that we should simply use an existing globbing library, ex.: https://www.npmjs.com/package/glob

pkozlowski-opensource:
@icfantv I think that we can have it reduced to 2-liner by using existing npm libraries. See an alternative implementation in #236

PR #231, opened 2016-06-20
test(panel): replace use of deprecated injectAsync

pkozlowski-opensource:


Foxandxss:
Thought I did that in the past. Or it was somewhere else.

PR #239, opened 2016-06-20
docs: use new doc generator

Foxandxss:
This is my proposed solution. A file at `demo/src` called `docs.json` that contains all the doc generated so from the demo's Angular, I can generate the doc.

I gitignored the file and I force a re-creation of the file always, so we don't end with stale information.

WDYT?


demo/src/app/components/accordion/accordion.component.ts
```diff
@@ -3,17 +3,20 @@ import { Component } from '@angular/core';
...
 
 @Component({
   selector: 'ngbd-accordion',
   template: require('./accordion.component.html'),
   directives: [AccordionBasicComponent, ComponentDocsComponent, ExampleBoxComponent]
 })
 export class AccordionComponent {
-  accordionDocs = accordionDocs;
-  panelDocs = panelDocs;
+  accordionDocs = docs['NgbAccordion'];
```
wesleycho: Does `docs.NgbAccordion` not work?

icfantv: it should - i don't think any of our keys are non-standard.  but i admittedly have no preference here.

gulpfile.js
```diff
@@ -19,7 +19,8 @@ var PATHS = {
   specs: 'src/**/*.spec.ts',
   demo: 'demo/**/*.ts',
   demoDist: 'demo/dist/**/*',
-  typings: 'typings/browser.d.ts'
+  typings: 'typings/browser.d.ts',
+  demoDocsJson: 'demo/src/docs.json'
```
icfantv: this is ok for now - we should either .gitignore this or move to a directory that's ignored so git doesn't keep pestering us that it's untracked.

Foxandxss: Yes, it is ignored by now.

pkozlowski-opensource: Could we move it to another folder, something that is specifically not under 'src'? Like `demo/temp` or `demo/generated`? This way people won't be tempted to generate + edit + change `.gitignore`.

pkozlowski-opensource:
So my biggest grip with this approach is that I don't know how we are going to handle formatting of API description (ex. I would to highlight sth or just indicate `varName`). Shell we just use HTML directly in JSdocs? Markdown is sooooo convenient for those little things...

pkozlowski-opensource:
At the same time let's not die over it - it is better to have sth working now - we can always change later. I'm going to pull it tomorrow and see how it works in practice with "real" content. Will report back.

demo/src/app/components/accordion/accordion.component.ts
```diff
@@ -3,17 +3,20 @@ import { Component } from '@angular/core';
...
 export class AccordionComponent {
-  accordionDocs = accordionDocs;
-  panelDocs = panelDocs;
+  accordionDocs = docs.NgbAccordion;
+  panelDocs = docs.NgbPanel;
   basicHtmlContent = basicHtmlContent;
   basicTsContent = basicTsContent;
+
+  constructor() {
+    console.log(this.accordionDocs);
```
pkozlowski-opensource: remove

demo/src/app/components/accordion/accordion.component.html
```diff
@@ -1,8 +1,8 @@
-<ngbd-component-docs [name]="accordionDocs.name"
+<ngbd-component-docs [name]="accordionDocs.selector"
```
pkozlowski-opensource: We should have 2 separate things:
* name (class name) => people need to list it in `[directives]`
* selector

Foxandxss: We have no "name" right now on the output file.

pkozlowski-opensource: We do have `className`. Oh, while we are at it -> we should probably also use `fileName` to:
* let people know how to import this thing
* have a link to sources on GH

Foxandxss: I don't think we want to recommend people to import concrete components by `fileName`.

Better to output scoped packages.

Issue #214, opened 2016-06-13
Tabset: support disabled tabs


pkozlowski-opensource:
Similar to https://github.com/ng-bootstrap/core/blob/faf49d53739bf1ddb21e60d05cc4f3a2a6518ee1/src/accordion/accordion.ts#L36

PR #240, opened 2016-06-20
feat(tabset): add disable option to tabset

sendilkumarn:
#214

### DIFF
--- a/src/tabset/tabset.spec.ts
+++ b/src/tabset/tabset.spec.ts
@@ -149,6 +149,25 @@ describe('ngb-tabset', () => {
        });
      })));
 
+  it('should not change active tab on disabled tab title click', async(inject([TestComponentBuilder], (tcb) => {
+       const html = `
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
It is a good beginning, but I think that disabled tabs are getting particular CSS classes, no?

sendilkumarn:
disabled class to list item ?

pkozlowski-opensource:
`<a class="nav-link disabled">Disabled</a>`, see example here: http://v4-alpha.getbootstrap.com/components/navs/#tabs

Obviously we also need a test that shows that a tab gets proper CSS classes in response to changes in the disabled state.

sendilkumarn:
is it okay to add that in all methods.. wherever `expectactivetabs` is present?

sendilkumarn:
@pkozlowski-opensource  let me know if anything else to be added here

Issue #208, opened 2016-06-13
Pager: align to sides


pkozlowski-opensource:
BS4 spec: http://v4-alpha.getbootstrap.com/components/pagination/#aligned-links

We should introduce a new `@Input` called `allignToSides` (or whatever other name that works). When this attribute is present we should add `pager-prev` and `pager-next` classes on the `<li>` element as shown in the docs.

PR #241, opened 2016-06-20
feat(pager): add align to sides

sendilkumarn:
#208

### DIFF
--- a/src/pager/pager.spec.ts
+++ b/src/pager/pager.spec.ts
@@ -36,6 +36,20 @@ function getLink(nativeEl: HTMLElement, idx: number): HTMLAnchorElement {
   return <HTMLAnchorElement>nativeEl.querySelectorAll('li')[idx].querySelector('a');
 }
 
+function expectAlignClasses(nativeEl: HTMLElement, active: boolean[]) {
+  const pagerCtrls = nativeEl.querySelectorAll('li');
+
+  for (let i = 0; i < active.length; i++) {
+    if (active[i]) {
+      expect(pagerCtrls[0]).toHaveCssClass('pager-prev');
+      expect(pagerCtrls[1]).toHaveCssClass('pager-next');
+    } else {
+      expect(pagerCtrls[0]).not.toHaveCssClass('pager-prev');
+      expect(pagerCtrls[1]).not.toHaveCssClass('pager-next');
+    }
+  }
+}
+
 describe('ngb-pagination', () => {
 
   describe('business logic', () => {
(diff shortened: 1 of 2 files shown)


src/pager/pager.spec.ts
```diff
@@ -131,6 +145,23 @@ describe('ngb-pagination', () => {
...
+         tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+           expectPager(fixture.nativeElement, [true, true]);
+           expectAlignClasses(fixture.nativeElement, [true, true]);
+
+           fixture.componentInstance.noOfPages = 5;
+           fixture.componentInstance.page = 5;
+           fixture.detectChanges();
+           expectPager(fixture.nativeElement, [true, false]);
+           expectAlignClasses(fixture.nativeElement, [true, true]);
```
pkozlowski-opensource: I don't think there is much point in having a separate `expectAlignClasses` method. Just check the presence of CSS classes directly here in this test. Or just create a simpler helper method.

src/pager/pager.spec.ts
```diff
@@ -131,6 +145,23 @@ describe('ngb-pagination', () => {
            expectPager(fixture.nativeElement, [true, true]);
          });
        })));
+
+    it('should have pager-prev and pager-next classes set', async(inject([TestComponentBuilder], (tcb) => {
+         const html = '<ngb-pager [noOfPages]="noOfPages" page="5" [alignLinks]=true></ngb-pager>';
```
pkozlowski-opensource: `[alignLinks]=true>` => `[alignLinks]="true">`

src/pager/pager.spec.ts
sendilkumarn: noted. Thought of using it elsewhere. Okay will change it 👍

src/pager/pager.spec.ts
```diff
@@ -131,6 +145,23 @@ describe('ngb-pagination', () => {
...
+
+         tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+           expectPager(fixture.nativeElement, [true, true]);
```
pkozlowski-opensource: We don't need to test for this, already covered in other tests.

src/pager/pager.spec.ts
```diff
@@ -131,6 +145,23 @@ describe('ngb-pagination', () => {
...
+           expectAlignClasses(fixture.nativeElement, [true, true]);
+
+           fixture.componentInstance.noOfPages = 5;
+           fixture.componentInstance.page = 5;
+           fixture.detectChanges();
+           expectPager(fixture.nativeElement, [true, false]);
```
pkozlowski-opensource: This section of the above 4 lines is not needed, as already covered in other tests. What we should have instead is a change in the `alignLinks` binding to see that CSS classes are added / removed.

sendilkumarn: toggle align?  Think adding a function to handle alignLink is not needed right?

pkozlowski-opensource: What I meant is that: `<ngb-pager [noOfPages]="noOfPages" page="5" [alignLinks]="shouldAlign">` and then in this test we should flip it between true / false and verify that the corresponding CSS classes are added / removed.

pkozlowski-opensource:
Merged, thnx! In the future please squash commits into one and add `Fixes #xyz` to indicate which issue is being fixed by the commit.

PR #243, opened 2016-06-21
fix(radio): changes after rc2 migration

### LINKED ISSUE #221: Buttons: 2 tests failing after update to angular 2.0.0-rc.2
Needs investigation. cc: @maxokorokov

maxokorokov:
Changes:
* Removal of deprecated Provider
* Tests simplification
* Removal of non-unidirectional data flow issues (radio modifying radio-group)
* Simplifying behaviour to match the select-option component from angular

Initial version of the radio button had issues with the one-directional data flow for change detection that worked for some reason with rc1 (radio updated the radio group ng-model after change detection was already done for the group). So now radio behaviour is simplified and matches select-option directives from the angular framework:

* when a selected radio is hidden → ng-model still holds radio value
* when a selected radio value is updated → ng-model still holds the old value

Closes #221

### DIFF
--- a/src/radio/radio.spec.ts
+++ b/src/radio/radio.spec.ts
@@ -1,4 +1,4 @@
-import {it, iit, expect, inject, async, tick, fakeAsync, describe, ddescribe} from '@angular/core/testing';
+import {it, iit, expect, inject, async, describe, ddescribe} from '@angular/core/testing';
 import {TestComponentBuilder} from '@angular/compiler/testing';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
Could you please have a look at the lint failures? Seems like there are few formatting / style issues. Thnx @maxokorokov !

maxokorokov:
@pkozlowski-opensource thanks, haven't noticed that linting was introduced. Amended

PR #245, opened 2016-06-21
refactor(radio): minor style changes

pkozlowski-opensource:
@maxokorokov few changes that I've spotted only now... Could you PTAL?

### DIFF
--- a/src/radio/radio.spec.ts
+++ b/src/radio/radio.spec.ts
@@ -23,9 +23,9 @@ function getInput(nativeEl: HTMLElement, idx: number): HTMLInputElement {
   return <HTMLInputElement>nativeEl.querySelectorAll('input')[idx];
 }
 
-describe('ng-radio-group', () => {
+describe('ngbRadioGroup', () => {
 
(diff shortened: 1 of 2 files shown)


icfantv:
LTGM.  I'd love an explanation of what `@Host` was.

pkozlowski-opensource:
> I'd love an explanation of what @Host was.

When looking for things to inject stop traversing injectors at the component (host view) boundary.

maxokorokov:
Yeah, sure it looks good. I knew I was gonna forget something.
So why would you need to continue checking beyond the component level?

pkozlowski-opensource:
> So why would you need to continue checking beyond the component level?

I could get back with the same question: "Why would you stop at a component level?" :-)
And since I like less code than more code...

icfantv:
Cool.  got it.

maxokorokov:
Yeah, I wrote the message above and thought that it looks like a useless "optimisation" in this case

pkozlowski-opensource:
> useless "optimisation" in this case

Queries are auto-generated at compilation time so we are not getting any runtime optimization from restricting search scope (at least this is the theory :-))

Issue #232, opened 2016-06-20
Carousel: add keyboard navigation + tests


pkozlowski-opensource:
It is about the `keyboard` option from http://v4-alpha.getbootstrap.com/components/carousel/

Check https://github.com/ng-bootstrap/core/blob/7ae57e3ee5737e12e3a53e534bc98acfff391663/src/dropdown/dropdown.spec.ts#L112-L132 for an example of testing with keyboard events.

sendilkumarn:
Is this enough or am I missing anything...? 78e0a9ca35deca6cdede0d63325d8ba3062a8528

sendilkumarn:
Adding a host event would be enough right? 

Something like this 78e0a9c

pkozlowski-opensource:
> Adding a host event would be enough right?

Did you test it outside of unit tests, in a real app? Things from your commit should be enough but we need to make sure that the carousel can be focused to actually receive key events (tabIndex?)

pkozlowski-opensource:
Landed via 01b0542c213f9719312e90db9bebd7a8a1256731

PR #251, opened 2016-06-21
feat(carousel): fix to carousel keyboard navigation

sendilkumarn:
#232 

@pkozlowski-opensource  Thanks 👍


pkozlowski-opensource:
Two more things:
* in BS 4 the keyboard navigation is an option behind a flag: http://v4-alpha.getbootstrap.com/components/carousel/#options (true by default). I wonder if we should support this option as well
* please amend the commit message to include `Fixes #232`

src/carousel/carousel.ts
```diff
@@ -53,6 +61,11 @@ export class NgbCarousel implements AfterContentChecked,
   @Input() interval: number = 5000;
   @Input() wrap: boolean = true;
 
+  @HostBinding('tabIndex')
```
pkozlowski-opensource: We don't need `@HostBinding` for constants.... we could achieve the same effect in `host: {tabIndex: 0}`. Having said this I'm not sure that `0` is the best value. Maybe we should `-1` instead? Or ask users to provide the value?

I must say that I don't have enough knowledge on tab indexes to answer this question myself...

sendilkumarn: As per docs,
-1 won't allow you to have sequential focus navigation
0 will make browser takes care of order in sequential  focus navigation
And 1 or any other positive value for enforcing the order 

I think this is why we use tabindex and also for an element to be focusable we need to set tab index focus flag.

Yes we can achieve it with host.

sendilkumarn:
I thinking supporting it better we can add in that also just like wrap.

I will change the text content.

pkozlowski-opensource:
@sendilkumarn seems like something went wrong during the rebase - the second commit is completely broken. You should have only one commit with only your changes.

sendilkumarn:
Merged with master and then rebased. I will close this open another if okay.

pkozlowski-opensource:
> Merged with master and then rebased. I will close this open another if okay.

Whatever works for you

PR #252, opened 2016-06-22
chore: update to Angular 2 RC 3

Foxandxss:
With new router for the demo.


demo/src/main.ts
```diff
@@ -15,6 +16,6 @@ if (process.env.ENV === 'build') {
 
 bootstrap(AppComponent, [
     // These are dependencies of our App
-    ...ROUTER_PROVIDERS
+    ...APP_ROUTER_PROVIDERS
```
pkozlowski-opensource: I don't think we need destructing here, providers will be flattened.

pkozlowski-opensource:
LGTM with one minor comment that can be safely ignored.

demo/src/main.ts
Foxandxss: Yeah, probably not, remnants.

pkozlowski-opensource:
Great stuff @Foxandxss !

PR #254, opened 2016-06-23
docs(alert): add API docs

pkozlowski-opensource:
Would be great if one of the native speakers could check this one. Thnx!

### DIFF
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -1,5 +1,8 @@
 import {Component, Input, Output, EventEmitter, ChangeDetectionStrategy} from '@angular/core';
 
+/**
+ * Alerts can be used to provide feedback messages.
+ */
 @Component({
   selector: 'ngb-alert',
   changeDetection: ChangeDetectionStrategy.OnPush,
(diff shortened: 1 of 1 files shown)


src/alert/alert.ts
```diff
@@ -13,8 +16,19 @@ import {Component, Input, Output, EventEmitter, ChangeDetectionStrategy} from '@
     `
 })
 export class NgbAlert {
+  /**
+   * A flag indicating if a given alert can be dismissed (closed) by a user. If this is flag is set a close button (in a
```
Foxandxss: I would say "If this flag is set, a close button..." There are two things to fix there, an extra "is" and the comma.

src/alert/alert.ts
```diff
@@ -13,8 +16,19 @@ import {Component, Input, Output, EventEmitter, ChangeDetectionStrategy} from '@
...
+   * form of a cross) will be displayed.
+   */
   @Input() dismissible: boolean = true;
+  /**
+   * Presentation type (CSS class) for an alert. Bootstrap 4 recognizes following types: "success", "info", "warning"
```
Foxandxss: I would resume the first sentence as "Alert type (CSS class)"

Foxandxss:
Looks good to me.

Issue #202, opened 2016-06-10
docs: add component demos


icfantv:
The following components require demos:
- [ ] Accordion (cleaned up?)
- [ ] Alert
- [x] Buttons
- [x] Carousel
- [x] Collapse
- [x] Dropdown
- [x] Pager
- [x] Pagination
- [x] Popover
- [x] Progressbar
- [x] Rating
- [x] Tabset
- [x] Tooltip

pkozlowski-opensource:
At this point we've got most of the demos in place or PRs opened for the remaining ones (accordion). Some of the existing demos are a bit "light" but we can grow them as we go. I will open separate issues for parts where I believe we can have more / better demos.

In any case I think that we can consider this issue as done.

PR #259, opened 2016-06-23
demo(alert): first crack at alert demo

icfantv:
First crack at alert demo.

Addresses alert portion of #202

### DIFF
--- a/demo/src/app/components/alert/alert.component.ts
+++ b/demo/src/app/components/alert/alert.component.ts
@@ -1,12 +1,30 @@
 import {Component} from '@angular/core';
-import {NgbdApiDocs} from '../shared/api-docs';
+import {ExampleBoxComponent, NgbdApiDocs} from '../shared';
+import {AlertBasicComponent, basicHtmlContent, basicTsContent,
+        AlertCloseableComponent, closeableHtmlContent, closeableTsContent,
+        AlertCustomComponent, customHtmlContent, customTsContent} from './demos';
 
 @Component({
   selector: 'ngbd-alert',
   template: `
     <ngbd-api-docs directive="NgbAlert"></ngbd-api-docs>        
+    <ngbd-example-box demoTitle="Basic Alert" [htmlSnippet]="basicHtmlContent" [tsSnippet]="basicTsContent">
+      <ngbd-alert-basic></ngbd-alert-basic>
+    </ngbd-example-box>
+    <ngbd-example-box demoTitle="Custom Alert" [htmlSnippet]="customHtmlContent" [tsSnippet]="customTsContent">
+      <ngbd-alert-custom></ngbd-alert-custom>
+    </ngbd-example-box>
+    <ngbd-example-box demoTitle="Closeable Alert" [htmlSnippet]="closeableHtmlContent" [tsSnippet]="closeableTsContent">
+      <ngbd-alert-closeable></ngbd-alert-closeable>
+    </ngbd-example-box>
   `,
-  directives: [NgbdApiDocs]
+  directives: [AlertBasicComponent, AlertCloseableComponent, AlertCustomComponent, NgbdApiDocs, ExampleBoxComponent]
 })
 export class NgbdAlert {
+  basicHtmlContent = basicHtmlContent;
+  basicTsContent = basicTsContent;
+  closeableHtmlContent = closeableHtmlContent;
+  closeableTsContent = closeableTsContent;
+  customHtmlContent = customHtmlContent;
+  customTsContent = customTsContent;
 }
--- a/demo/src/app/components/alert/demos/basic/basic.alert.ts
+++ b/demo/src/app/components/alert/demos/basic/basic.alert.ts
@@ -1,0 +1,10 @@
+import { Component } from '@angular/core';
+
+import { NgbAlert } from '@ng-bootstrap/alert';
+
+@Component({
+  selector: 'ngbd-alert-basic',
+  template: require('./basic.alert.html'),
+  directives: [NgbAlert]
+})
+export class AlertBasicComponent {}
--- a/demo/src/app/components/alert/demos/closeable/closeable.alert.ts
+++ b/demo/src/app/components/alert/demos/closeable/closeable.alert.ts
@@ -1,0 +1,52 @@
+import { Input, Component } from '@angular/core';
+
+import { NgbAlert } from '@ng-bootstrap/alert';
+
+@Component({
+  selector: 'ngbd-alert-closeable',
+  template: require('./closeable.alert.html'),
+  directives: [NgbAlert]
+})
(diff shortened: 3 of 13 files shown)


icfantv:
This was my first real foray into ng2 coding so if I've done something heinous, please be gentle.  :-)

demo/src/app/components/alert/demos/closeable/closeable.alert.ts
```diff
@@ -0,0 +1,52 @@
...
+    }, {
+      id: 3,
+      type: 'warning',
+      message: 'This is a warning alert',
+    }, {
+      id: 4,
+      type: 'danger',
+      message: 'This is a danger alert',
+    });
+    this.backup = this.alerts.map((alert: IAlert) => Object.assign({}, alert))
```
icfantv: is there a better way to deep clone an array?

demo/src/app/components/alert/demos/closeable/closeable.alert.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public closeAlert(id: number) {
+    const index: number = this.alerts.findIndex((alert: IAlert) => alert.id === id);
+    this.alerts.splice(index, 1);
+  }
+
+  public reset() {
+    this.alerts = this.backup.map((alert: IAlert) => Object.assign({}, alert))
```
icfantv: ditto

demo/src/app/components/alert/alert.component.ts
```diff
@@ -1,12 +1,24 @@
...
+    </ngbd-example-box>
+    <ngbd-example-box demoTitle="Closeable demo" [htmlSnippet]="closeableHtmlContent" [tsSnippet]="closeableTsContent">
+      <ngbd-alert-closeable></ngbd-alert-closeable>
+    </ngbd-example-box>
   `,
-  directives: [NgbdApiDocs]
+  directives: [AlertBasicComponent, AlertCloseableComponent, NgbdApiDocs, ExampleBoxComponent]
 })
 export class NgbdAlert {
+  basicHtmlContent = basicHtmlContent;
```
icfantv: I really don't like this paradigm as it doesn't scale with multiple demos.  I _think_ this has more to do with how we are doing exports, but we'll still need the names to be uniqe...I think...based on how we've got the wrapper demo constructed.

demo/src/app/components/alert/alert.component.ts
```diff
@@ -1,12 +1,24 @@
 import {Component} from '@angular/core';
-import {NgbdApiDocs} from '../shared/api-docs';
+import {ExampleBoxComponent, NgbdApiDocs} from '../shared';
+import {AlertBasicComponent, basicHtmlContent, basicTsContent,
+        AlertCloseableComponent, closeableHtmlContent, closeableTsContent} from './demos';
```
icfantv: see note below re exports.

demo/src/app/components/alert/demos/basic/basic.alert.ts
```diff
@@ -0,0 +1,17 @@
+import { Component } from '@angular/core';
+
+import { NgbAlert } from '@ng-bootstrap/alert';
```
pkozlowski-opensource: @icfantv did this import work for you? I'm not sure how it could work since we didn't publish anything to npm...

pkozlowski-opensource:
@icfantv I was trying to test this PR but I don't know how `@ng-bootstrap/alert` imports could work... and they failing on my end... Are there are parts of WebPack config you've forgotten to commit? Or there is sth else going on?

icfantv:
@pkozlowski-opensource gah!  Omg, yes.  There's a line in the webpack file you need to add and I didn't commit that file because I'd changed the port from 8080 to 9090 because we run our work web server on 8080.  If you want to test, look in the webpack config file for where we map @ng-bootstrap for accordion and do the same for alert.  I'll jump on a computer and update the PR.

Foxandxss:
Try and if you can, I do it for you. I needed to rethink the idea to be able to add more than one alias.

gulpfile.js
```diff
@@ -158,8 +158,7 @@ gulp.task('build', function(done) {
   runSequence('lint', 'enforce-format', 'ddescribe-iit', 'test', 'clean:build', 'cjs', 'umd', done);
 });
 
-gulp.task('deploy-demo', function(done) {
-  runSequence('clean:demo', 'build:demo', 'demo-push', 'clean:demo-cache', done);
-});
+gulp.task(
```
icfantv: this change was because the clang formatter.  i didn't even touch this file...

pkozlowski-opensource:
LGTM generally speaking. But having looked at the existing demos here my ideas / remarks:

* the first basic demo should be minimal. Like really minimal, showing the simplest possible usage. Then we can have more involved cases in subsequent demos. Applying this principle to the alert demos would mean that we start with just one alert with default options (so remove custom colors)
* HTML tag should be present first and selected by default as HTML is the first thing people are going to copy & paste

Let me know if you want to do the above changes or if you want me to land this PR and do changes.

I agree that the current model of passing snippets is not ideal. As soon as we land this PR I will try to propose a slightly simpler model (unless @Foxandxss has ideas / wants to take a crack at it)

icfantv:
@pkozlowski-opensource I like the idea of having the simplest and most basic use case first with more advanced uses later.  This dovetails nicely into my desire to somewhat mimic AM's demos in that more, smaller ones are better and more easily manageable than one large one.

The other thing I want to be sure of is to have 100% coverage of our API calls across all the demos.  In UIBS, we have directive properties that are documented but we don't show the user an example that uses them.  I realize this may seem silly for some really basic/simple properties, but I think this is a good idea regardless.  One of the biggest complaints ng1 users had (as reported by Brad) was the lack of good examples and I think this ties in to that.

As far as which tab is selected first, I don't have a preference and am happy to switch.

I'll change up the alerts demo to incorporate the changes.

icfantv:
AM puts the HTML tab first, so I've changed my lack of having a preference to having one.  We should do the same.

Foxandxss:
Let's put this PR in hold so tomorrow we can hangouts and discuss a few things.

icfantv:
@Foxandxss, @pkozlowski-opensource: updated per meeting discussion.  Please disregard spacing between demo components and demo vs. API documentation - just trying to nail down the contents of the alert demo.

demo/src/app/components/alert/demos/basic/basic.alert.ts
sendilkumarn: @pkozlowski-opensource  It will work. adding this in webpack.demo is enough I guess.

pkozlowski-opensource:
@icfantv this looks good! I will send a follow-up PR with some other proposed changes.

icfantv:
Woo hoo!

Issue #222, opened 2016-06-16
chore: API extraction script missing features


pkozlowski-opensource:
* [x] - extract input types from setters
* [x] - extract info about _public_ members that are: not input / outputs, not lifecycle hooks. There should be a flag to hide a given member from a doc as a "template private"

pkozlowski-opensource:
Implementation note: we need a way of distinguishing between things that are "template private" (methods to be used in a template only) and "public" (ones that can be invoked by people having reference to an exported directive instance).

Foxandxss:
Technically the stuff behind `private` keyword are the private things.

maxokorokov:
Another use case to (possibly) support: default value in a private variable + accessor

```javascript
class Pagination {
  private _page = 0;
  
  /**
  * Comment goes here, but what's the default value?...
  */
  @Input() set page(value: number | string) { this.page = toInteger(value); }
}
```

pkozlowski-opensource:
We also need start extracting APIs from regular classes / interfaces as it will be needed for:
* services (ex. `NgbModal`)
* configuration classes
* other parts of the API we are exposing to our users

pkozlowski-opensource:
I think that we are done here, generally speaking. There are still some improvements that we could do, I will open small, more focused issues.

PR #262, opened 2016-06-24
fix(tabset): use user-supplied tab indexes

### LINKED ISSUE #215: Tabset: use user-defined ids / indexes instead of array indexes
See: http://angular-ui.github.io/bootstrap/#/tabs

pkozlowski-opensource:
Fixes #215

### DIFF
--- a/src/tabset/tabset.ts
+++ b/src/tabset/tabset.ts
@@ -9,6 +9,8 @@ import {
   AfterContentChecked
 } from '@angular/core';
 
+let nextId = 0;
+
 @Directive({selector: 'template[ngbTabTitle]'})
 export class NgbTabTitle {
   constructor(public templateRef: TemplateRef<any>) {}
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
I didn't get much feedback on #215 so I assume no one is against the change :-)

src/tabset/tabset.ts
```diff
@@ -19,27 +21,43 @@ export class NgbTabContent {
...
 
+/**
+ * A directive representing an individual tab.
+ */
 @Directive({selector: 'ngb-tab'})
 export class NgbTab {
+  /**
+   * Unique tab identifier. Must be unique for the entire document for proper accessibility support.
+   */
+  @Input() id: string = `ngb-tab-${nextId++}`;
```
icfantv: this is why typescript rocks.  we don't have to say what the data type is anymore.

src/tabset/tabset.ts
```diff
@@ -19,27 +21,43 @@ export class NgbTabContent {
...
+ * A component that makes it easy to create tabbed interface.
+ */
 @Component({
   selector: 'ngb-tabset',
   template: `
     <ul class="nav nav-tabs" role="tablist">
-      <li class="nav-item" *ngFor="let tab of tabs; let i=index">
-        <a class="nav-link" [class.active]="i === activeIdx" [class.disabled]="tab.disabled" (click)="select(i)">
+      <li class="nav-item" *ngFor="let tab of tabs">
+        <a [id]="tab.id" class="nav-link" [class.active]="tab.id === activeId" [class.disabled]="tab.disabled" (click)="select(tab.id)">
```
icfantv: in the UIBS version we pass the event object to the click handler so that the user can cancel.  i don't think we need it for the initial release, but know that someone will ask for it as some point.

src/tabset/tabset.ts
```diff
@@ -48,17 +66,26 @@ export class NgbTab {
...
-  @Input() activeIdx: number = 0;
+  /**
+   * An identifier of a tab that should be selected (active).
+   */
+  @Input() activeId: string;
 
-  select(tabIdx: number) {
-    if (!this.tabs.toArray()[tabIdx].disabled) {
-      this.activeIdx = tabIdx;
+  select(tabIdx: string) {
```
icfantv: see comment above about passing event so that it can be cancelled.

icfantv:
LGTM - please see note about `select` event.

src/tabset/tabset.ts
pkozlowski-opensource: we actually need it, at least now, for API docs documentation (that is, when extracting API docs I don't know how to infer types). Will update #222

src/tabset/tabset.ts
pkozlowski-opensource: We've got a separate issue for this: #267

PR #265, opened 2016-06-24
chore: update to typings 1.x

Foxandxss:
Changes:

New typings with its breaking changes.
I removed the postinstall. A library **shouldn't** have it because that will force all consumers to have typings installed globally and that is a no-no, so from now one, manual install for us (this doesn't involve anything for the end user).

Please delete your typings folder and install them again.


pkozlowski-opensource:
I don't know much about this stuff but looks good to me on pronciple

PR #271, opened 2016-06-24
feat(exports): Fixes #219 Adding exports grouping directives

### LINKED ISSUE #219: Add exports grouping all directives when there is more than one cooperating direcctvie
ex.:

```
const TABSET_DIRECTIVES = [NgbTabset, NgbTab, NgbTabTitle, NgbTabContent];
```

so people can simply do `directives: [TABSET_DIRECTIVES]` in their components.

This needs to be done for all the existing directives.

sendilkumarn:
Exporting Accordion / Tabset / Dropdown components

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -77,3 +77,5 @@ export class NgbAccordion implements AfterContentChecked {
     }
   }
 }
+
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
Thnx, this is a good start. On top of this we need:
* do this for all directives (even if there is one in a given file)
* https://github.com/ng-bootstrap/core/blob/master/src/core.ts needs to be modified accordingly

PR #280, opened 2016-06-28
chore(config): config/build tweaks

icfantv:
* update typescript to latest
* update `tsconfig.json` to exclude `node_modules`


Foxandxss:
IIRC node_modules is excluded automatically but I am not 200% sure. We don't exclude it on angular.io.

PR #282, opened 2016-06-28
feat(carousel): defining id instead of running sequence

### LINKED ISSUE #263: Carousel: use user-defined ids / indexes instead of array indexes
Same issue as #215 and the same solution as in #262

sendilkumarn:
fixes #263

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -10,8 +10,11 @@ import {
   OnInit
 } from '@angular/core';
 
+let nextId = 0;
+
 @Directive({selector: 'template[ngbSlide]'})
 export class NgbSlide {
+  @Input() id: string = `ngb-slide-${nextId++}`;
   constructor(public tplRef: TemplateRef<any>) {}
 }
 
@@ -29,10 +32,10 @@ export class NgbSlide {
   },
   template: `
     <ol class="carousel-indicators">
-      <li *ngFor="let slide of _slides; let slideIdx = index" [class.active]="slideIdx === _activeIdx" (click)="select(slideIdx)"></li>
+      <li *ngFor="let slide of _slides" [id]="slide.id" [class.active]="slide.id === activeId" (click)="select(slide.id)"></li>
     </ol>
     <div class="carousel-inner" role="listbox">
-      <div *ngFor="let slide of _slides; let slideIdx = index" class="carousel-item" [class.active]="slideIdx === _activeIdx">
+      <div *ngFor="let slide of _slides" class="carousel-item" [class.active]="slide.id === activeId">
         <template [ngTemplateOutlet]="slide.tplRef"></template>
       </div>
     </div>
@@ -49,33 +52,32 @@ export class NgbSlide {
 export class NgbCarousel implements AfterContentChecked,
     OnDestroy, OnInit {
   @ContentChildren(NgbSlide) private _slides: QueryList<NgbSlide>;
-  private _activeIdx = 0;
   private _slideChangeInterval;
-
-  @Input()
-  set activeIdx(newActiveIdx: number) {
-    this.select(newActiveIdx || 0);
-    this._restartTimer();
-  };
 
   @Input() interval: number = 5000;
   @Input() wrap: boolean = true;
   @Input() keyboard: boolean = true;
+  @Input() activeId: string;
 
   ngAfterContentChecked() {
-    // auto-correct _activeIdx that might have been set incorrectly as input
-    this._activeIdx = Math.min(Math.max(0, this._activeIdx), this._slides.length - 1);
+    let activeSlide = this._getSlideById(this.activeId);
+    this.activeId = activeSlide ? activeSlide.id : (this._slides.length ? this._slides.first.id : null);
   }
 
   ngOnInit() { this.cycle(); }
 
   ngOnDestroy() { clearInterval(this._slideChangeInterval); }
 
-  select(slideIdx: number) { this._activeIdx = slideIdx; }
+  select(slideIdx: string) {
+    let selectedSlide = this._getSlideById(slideIdx);
+    if (selectedSlide) {
+      this.activeId = selectedSlide.id;
+    }
+  }
 
   prev() {
-    const prevIdx = this._activeIdx - 1;
-    this.select(prevIdx < 0 ? (this.wrap ? prevIdx + this._slides.length : prevIdx + 1) : prevIdx);
+    let selectedId: string = this._getPrevSlide(this.activeId);
+    this.select(selectedId);
     this._restartTimer();
   }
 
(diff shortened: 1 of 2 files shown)


src/carousel/carousel.ts
```diff
@@ -103,14 +105,38 @@ export class NgbCarousel implements AfterContentChecked,
...
+    let selectedId: string = this._getOtherSlide(this.activeId, true);
+    this.select(selectedId);
   }
 
   private _restartTimer() {
     this.pause();
     this.cycle();
   }
+
+  private _getSlideById(slideIdx: string): any {
```
pkozlowski-opensource: `:NgbSlide` instead of `:any`

src/carousel/carousel.ts
```diff
@@ -49,33 +52,32 @@ export class NgbSlide {
...
+    this.activeId = activeSlide ? activeSlide.id : (this._slides.length ? this._slides.first.id : null);
   }
 
   ngOnInit() { this.cycle(); }
 
   ngOnDestroy() { clearInterval(this._slideChangeInterval); }
 
-  select(slideIdx: number) { this._activeIdx = slideIdx; }
+  select(slideIdx: string) {
+    let selectedSlide: NgbSlide = this._getSlideById(slideIdx);
```
pkozlowski-opensource: no need for `: NgbSlide` here, it will be inferred.

src/carousel/carousel.ts
```diff
@@ -103,14 +105,38 @@ export class NgbCarousel implements AfterContentChecked,
...
     this.pause();
     this.cycle();
   }
+
+  private _getSlideById(slideIdx: string): any {
+    let slideWithId: NgbSlide[] = this._slides.filter(slide => slide.id === slideIdx);
+    return slideWithId.length ? slideWithId[0] : null;
+  }
+
+  private _getOtherSlide(id: string, next: boolean): string {
```
pkozlowski-opensource: This methods looks quite complex. Could you think of writing it differently so it is easier to read / reason about? Maybe split it into 2 methods (`_getNextSlide` / `_getPrevSlide`)?

src/carousel/carousel.ts
```diff
@@ -49,33 +52,32 @@ export class NgbSlide {
...
-
   @Input() interval: number = 5000;
   @Input() wrap: boolean = true;
   @Input() keyboard: boolean = true;
+  @Input() activeId: string;
 
   ngAfterContentChecked() {
-    // auto-correct _activeIdx that might have been set incorrectly as input
-    this._activeIdx = Math.min(Math.max(0, this._activeIdx), this._slides.length - 1);
+    let activeSlide: NgbSlide = this._getSlideById(this.activeId);
```
wesleycho: I believe the `: NgbSlide` here is unnecessary (should be inferred)

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
...
     this.cycle();
   }
+
+  private _getSlideById(slideIdx: string): NgbSlide {
+    let slideWithId: NgbSlide[] = this._slides.filter(slide => slide.id === slideIdx);
+    return slideWithId.length ? slideWithId[0] : null;
+  }
+
+  private _getNextSlide(id: string): string {
+    let otherSlideId: string = id;
```
wesleycho: Maybe rename to `nextSlideId`

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
...
+        let lastSlide: boolean = (idx === (slideArr.length - 1));
+        otherSlideId =
+            lastSlide ? (this.wrap ? slideArr[0].id : slideArr[slideArr.length - 1].id) : slideArr[idx + 1].id;
+      }
+    });
+    return otherSlideId;
+  }
+
+  private _getPrevSlide(id: string): string {
+    let otherSlideId: string = id;
```
wesleycho: Maybe rename to `prevSlideId` for readability

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
...
+    let slideArr: any[] = this._slides.toArray();
```
wesleycho: I think this can be typed as `: NgbSlide[]`

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
   }
 
   private _cycleToNext() {
-    const nextIdx = this._activeIdx + 1;
-    this.select(nextIdx >= this._slides.length ? (this.wrap ? 0 : nextIdx - 1) : nextIdx);
+    let selectedId: string = this._getNextSlide(this.activeId);
```
wesleycho: This typing can be removed too as inferred

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
...
     this.cycle();
   }
+
+  private _getSlideById(slideIdx: string): NgbSlide {
+    let slideWithId: NgbSlide[] = this._slides.filter(slide => slide.id === slideIdx);
+    return slideWithId.length ? slideWithId[0] : null;
+  }
+
+  private _getNextSlide(id: string): string {
+    let nextSlideId: string = id;
```
wesleycho: No need for the typing here, it's obvious from the arguments

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
...
+        let lastSlide: boolean = (idx === (slideArr.length - 1));
+        nextSlideId =
+            lastSlide ? (this.wrap ? slideArr[0].id : slideArr[slideArr.length - 1].id) : slideArr[idx + 1].id;
+      }
+    });
+    return nextSlideId;
+  }
+
+  private _getPrevSlide(id: string): string {
+    let prevSlideId: string = id;
```
wesleycho: Ditto about the tying

src/carousel/carousel.ts
```diff
@@ -103,14 +105,46 @@ export class NgbCarousel implements AfterContentChecked,
...
   }
+
+  private _getSlideById(slideIdx: string): NgbSlide {
+    let slideWithId: NgbSlide[] = this._slides.filter(slide => slide.id === slideIdx);
+    return slideWithId.length ? slideWithId[0] : null;
+  }
+
+  private _getNextSlide(id: string): string {
+    let nextSlideId: string = id;
+    let slideArr: NgbSlide[] = this._slides.toArray();
```
wesleycho: I think the `: NgbSlide[]` may not be necessary here and below - can you try removing and seeing if it works?

sendilkumarn: isn't it similar to line no: 118. I think it is needed here.

wesleycho: Are they needed? `_slides` is typed as `QueryList<NgbSlide>`, and if one looks at the definition of QueryList, it looks like it should work (see https://github.com/angular/angular/blob/master/modules/%40angular/core/src/linker/query_list.ts#L82 and https://github.com/angular/angular/blob/master/modules/%40angular/core/src/linker/query_list.ts#L56-L58). I could be missing something though.

PR #286, opened 2016-06-28
feat(tabs): export the component and expose navigation methods

### LINKED ISSUE #268: Tabs: export the component and expose navigation methods
To make programmatic interaction with a tabset easy we should export this component (`exportAs`) so a minimum the `select(id)` method is exposed. As a bonus we should add methods like `prev()` and `next()` to move between tabs.

Tests are essential.

ExFlo:
Closes #268

### DIFF
--- a/src/tabset/tabset.spec.ts
+++ b/src/tabset/tabset.spec.ts
@@ -54,6 +54,10 @@ function expectTabs(nativeEl: HTMLElement, active: boolean[], disabled?: boolean
   }
 }
 
+function getButton(nativeEl: HTMLElement) {
+  return nativeEl.querySelectorAll('button');
+}
+
 describe('ngb-tabset', () => {
 
   it('should render tabs and select first tab as active by default', async(inject([TestComponentBuilder], (tcb) => {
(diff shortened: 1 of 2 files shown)


src/tabset/tabset.spec.ts
```diff
@@ -225,6 +229,55 @@ describe('ngb-tabset', () => {
          expectTabs(fixture.nativeElement, [true], [true]);
        });
      })));
+
+    it('should change active tab by using a button click', async(inject([TestComponentBuilder], (tcb) => {
```
pkozlowski-opensource: I would rather say "should change active tab by calling select on an exported directive instance" since using buttons is just one way of invoking the `select` method.

src/tabset/tabset.spec.ts
```diff
@@ -225,6 +229,55 @@ describe('ngb-tabset', () => {
...
+            expectTabs(fixture.nativeElement, [false, true]);
+
+            // Click on a button to select the first tab
+            (<HTMLAnchorElement>button[0]).click();
+            fixture.detectChanges();
+            expectTabs(fixture.nativeElement, [true, false]);
+        });
+    })));
+
+    it('should not change active tab by using a button click in case of disable tab', async(inject([TestComponentBuilder], (tcb) => {
```
pkozlowski-opensource: same as above - remove reference to "button" in a test name.

pkozlowski-opensource:
LGTM. Just left 2 minor comments regarding tests naming. 
It needs rebasing.

PR #289, opened 2016-06-28
feat(positioning): add positioning service

RobJacobs:
Closes #289

Need feedback/help with::

1.  Is the name 'Positioning' acceptable?
2.  Does the Positioning class need to be marked as @Injectable?
3.  I'm getting conflicting clang-format and tslint errors on spaces around curly braces, tslint wants:

` func(): number { return 0; }`

clang-format wants:

` func(): number{return 0;}`

to get the build to complete locally, I had to remove the tslint rules.
4.  Rather than create an ELementPosition interface, should I use the ClientRect interface instead?  It has 2 properties (right and bottom) that we don't currently return, but we could easily enough. 

I still need to write the test, but I wanted to make sure I'm going down the right path, thanks.

### DIFF
--- a/src/util/positioning.spec.ts
+++ b/src/util/positioning.spec.ts
@@ -1,0 +1,263 @@
+import {it, describe, expect, beforeEach, afterEach} from '@angular/core/testing';
+
+import {Positioning} from './positioning';
+
+describe('Positioning', () => {
+  const positioning = new Positioning();
+  const documentMargin = document.documentElement.style.margin;
+  const bodyMargin = document.body.style.margin;
+  const bodyHeight = document.body.style.height;
+  const bodyWidth = document.body.style.width;
+
+  function createElement(height: number, width: number, marginTop: number, marginLeft: number): HTMLElement {
+    let element = document.createElement('div');
+    element.style.display = 'inline-block';
+    element.style.height = height + 'px';
+    element.style.width = width + 'px';
+    element.style.marginTop = marginTop + 'px';
+    element.style.marginLeft = marginLeft + 'px';
+
+    return element;
+  }
+
+  let cleanUp = false;
+  let element = createElement(200, 300, 100, 150);
+
+  document.documentElement.style.margin = '0';
+  document.body.style.margin = '0';
+  document.body.style.height = '2000px';
+  document.body.style.width = '2000px';
+
+  beforeEach(() => {
+    if (!cleanUp) {
+      document.body.appendChild(element);
+    }
+  });
+
+  afterEach(() => {
+    if (!cleanUp) {
+      document.body.removeChild(element);
+    }
+  });
+
+  it('should calculate the element offset', () => {
+    let position = positioning.offset(element);
+
+    expect(position.height).toBe(200);
+    expect(position.width).toBe(300);
+    expect(position.top).toBe(100);
+    expect(position.bottom).toBe(300);
+    expect(position.left).toBe(150);
+    expect(position.right).toBe(450);
+  });
+
+  it('should calculate the element offset when scrolled', () => {
+    document.documentElement.scrollTop = 1000;
+    document.documentElement.scrollLeft = 1000;
+
+    let position = positioning.offset(element);
+
+    expect(position.top).toBe(100);
+    expect(position.bottom).toBe(300);
+    expect(position.left).toBe(150);
+    expect(position.right).toBe(450);
+
+    document.documentElement.scrollTop = 0;
+    document.documentElement.scrollLeft = 0;
+  });
+
+  it('should calculate the element position', () => {
+    let position = positioning.position(element);
+
+    expect(position.height).toBe(200);
+    expect(position.width).toBe(300);
+    expect(position.top).toBe(100);
+    expect(position.bottom).toBe(300);
+    expect(position.left).toBe(150);
+    expect(position.right).toBe(450);
+  });
+
+  it('should calculate the element position when scrolled', () => {
+    document.documentElement.scrollTop = 1000;
+    document.documentElement.scrollLeft = 1000;
+
+    let position = positioning.position(element);
+
+    expect(position.top).toBe(100);
+    expect(position.bottom).toBe(300);
+    expect(position.left).toBe(150);
+    expect(position.right).toBe(450);
+
+    document.documentElement.scrollTop = 0;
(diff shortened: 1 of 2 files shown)


src/util/positioning.ts
```diff
@@ -0,0 +1,102 @@
...
+    let parentOffset: ElementPosition = {width: 0, height: 0, top: 0, left: 0};
+    let offsetParentEl: HTMLElement = this.parentOffsetElement(element);
+
+    if (offsetParentEl !== document.documentElement) {
+      parentOffset = this.offset(offsetParentEl);
+      parentOffset.top += offsetParentEl.clientTop - offsetParentEl.scrollTop;
+      parentOffset.left += offsetParentEl.clientLeft - offsetParentEl.scrollLeft;
+    }
+
+    return <ElementPosition>{
```
icfantv: don't need the cast here.

src/util/positioning.ts
```diff
@@ -0,0 +1,102 @@
...
+      width: elBCR.width || element.offsetWidth,
+      height: elBCR.height || element.offsetHeight,
+      top: elOffset.top - parentOffset.top,
+      left: elOffset.left - parentOffset.left
+    };
+  }
+
+  offset(element: HTMLElement): ElementPosition {
+    let elBcr: ClientRect = element.getBoundingClientRect();
+    return <ElementPosition>{
```
icfantv: ditto

src/util/positioning.ts
RobJacobs: It adds type safety to the object I'm returning, or is that not important?

icfantv: the compiler will check that if the function declares a return type, that the returned type matches what's declared.

src/util/positioning.ts
```diff
@@ -0,0 +1,102 @@
+import {ElementRef} from '@angular/core';
+
+export interface ElementPosition {
```
icfantv: should be named `IElementPosition` since it is an interface

src/util/positioning.ts
```diff
@@ -0,0 +1,102 @@
+import {ElementRef} from '@angular/core';
```
pkozlowski-opensource: Doesn't look like we are using `ElementRef` anywhere (and probably shouldn't be) so this import can be removed.

pkozlowski-opensource:
@RobJacobs I will review in grater details tomorrow but I was wondering if there is any way to write meaningful tests for this service?

RobJacobs:
@pkozlowski-opensource I'm going to look at how material is testing their stuff and see what we can do.

src/util/positioning.ts
```diff
@@ -0,0 +1,100 @@
...
+      offsetParentEl = offsetParentEl.offsetParent as HTMLElement;
+    }
+    return offsetParentEl || document.documentElement;
+  }
+
+  position(element: HTMLElement): ElementPosition {
+    let elOffset: ElementPosition = this.offset(element);
+    let elBCR: ClientRect = element.getBoundingClientRect();
+    let parentOffset: ElementPosition = {width: 0, height: 0, top: 0, left: 0};
+    let offsetParentEl: HTMLElement = this.parentOffsetElement(element);
```
wesleycho: The typings in these several of lines are unnecessary I think

wesleycho: Well, excepting `parentOffset` - that is probably ok.

src/util/positioning.ts
```diff
@@ -0,0 +1,100 @@
...
+    };
+  }
+
+  positionElements(hostElement: HTMLElement, targetElement: HTMLElement, placement: string, appendToBody: boolean):
+      ElementPosition {
+    let hostElPosition: ElementPosition = appendToBody ? this.offset(hostElement) : this.position(hostElement);
+    let shiftWidth: any = {
+      left(): number{return hostElPosition.left;},
+      center(): number{return hostElPosition.left + hostElPosition.width / 2 - targetElement.offsetWidth / 2;},
+      right(): number{return hostElPosition.left + hostElPosition.width;}
```
wesleycho: How come these have to be functions? Is this to readjust after the element has been rendered?

src/util/positioning.ts
```diff
@@ -0,0 +1,100 @@
...
+    };
+    let shiftHeight: any = {
+      top(): number{return hostElPosition.top;},
+      center(): number{return hostElPosition.top + hostElPosition.height / 2 - targetElement.offsetHeight / 2;},
+      bottom(): number{return hostElPosition.top + hostElPosition.height;}
```
wesleycho: Same question as above

src/util/positioning.ts
```diff
@@ -0,0 +1,100 @@
...
+    };
+    let targetElBCR: ClientRect = targetElement.getBoundingClientRect();
```
wesleycho: Typing can probably be removed as inferred.

pkozlowski-opensource:
@RobJacobs as a very minimum we should have tests for calculations (those could be decoupled from the DOM - that is we could write tests for the part that does calculations after dimensions of all the involved elements is known).

RobJacobs:
Updated based on suggestions.  I switched from creating an ElementPosition interface to using the Angular ClientRect interface and now return right/bottom values.  That will put us in a better position to support smart positioning down the road.  I'm going to work on tests today.

RobJacobs:
@pkozlowski-opensource Tests added, did a little more refactoring.  I think this is good to go, let me know if you see anything that needs work.

src/util/positioning.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+  const bodyHeight = document.body.style.height;
+  const bodyWidth = document.body.style.width;
+
+  function createElement(height: number, width: number, marginTop: number, marginLeft: number): HTMLElement {
+    let element = document.createElement('div');
+    element.style.display = 'inline-block';
+    element.style.height = height + 'px';
+    element.style.width = width + 'px';
+    element.style.marginTop = marginTop + 'px';
+    element.style.marginLeft = marginLeft + 'px';
```
wesleycho: Could use template strings for these 4 lines

pkozlowski-opensource:
@RobJacobs thnx, I've merged it to unblock the situation. We know that we need to improve it but I think it is good as a starting point. Expect to hear from me more as soon as I start impl of tooltips / popovers (just after alpha.0 is out).

PR #290, opened 2016-06-29
chore(demo): add prismjs as dependency

sirajc:
We have an `import 'prismjs/themes/prism-okaidia.css';` statement in _app.component.ts_ but we don't have prismjs as dependency due to which `gulp demo-server` fails

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "merge2": "^1.0.1",
     "node-sass": "^3.7.0",
     "postcss-loader": "^0.9.1",
+    "prismjs": "^1.5.1",
     "prismjs-loader": "0.0.3",
     "raw-loader": "^0.5.1",
     "reflect-metadata": "0.1.2",


icfantv:
@Foxandxss, please see [my comment here](https://github.com/ng-bootstrap/core/issues/278#issuecomment-229238992) and the rest of the thread before we consider merging this.  Cheers.

pkozlowski-opensource:
On the first sight it looks like a correct fix to me. It is true that it works without an explicit dependency on prism on npm 3.x, but this is just "by accident". It works since npm 3.x de-dupes dependencies by default so it will move prism higher up n the folders hierarchy.

IMO we should have explicit dependency if we require something in our code.

@sirajc what is your npm version?
@icfantv I don't think it is Windows-specific, it is probably npm-version specific (or at least this is my theory for now).

sirajc:
@pkozlowski-opensource My npm version is `2.4.17` as per what is mentioned in https://github.com/ng-bootstrap/core/blob/master/DEVELOPER.md#prerequisite-software
I agree with your theory on npm 3.x

Foxandxss:
Problem I have with this PR is that the loader depend on 1.4.1 and this is 1.5.1 and I don't know if there are problems with that.

As I mentioned in a different issue. You need npm 3.x, the scoped packages of angular are problematic with 2.x anyway.

pkozlowski-opensource:
@Foxandxss but in this case I think that npm is going to do "the right thing" - that is, put prism used by loader under its node_modules.

I agree that this is borderline but I doesn't sound "right" that we are importing things and we don't have dependency on it. Once again, for me it works "by accident".

pkozlowski-opensource:
I'm going to merge it for now as there is no harm in it and "logically" it makes sense. Having said this @sirajc you should definitively update to npm 3 (and probably update your node version).

sirajc:
Thanks, will update my local npm to v3 
we might want to update this doc as well to reflect this, I was using 2.x just because of this document
https://github.com/ng-bootstrap/core/blob/master/DEVELOPER.md#prerequisite-software

Issue #264, opened 2016-06-24
docs: add API docs


pkozlowski-opensource:
The following components API docs:
- [x] Accordion
- [x] Alert
- [x] Buttons
- [x] Carousel
- [x] Collapse
- [x] Dropdown
- [x] Pager
- [x] Pagination
- [x] Progressbar
- [x] Rating
- [x] Tabset

sirajc:
I can pick up docs for collapse.
Will have two examples, one Basic and other with default close
Something similar to this old plunker
http://plnkr.co/edit/SW68aAIMP6iwVtNUZZis?p=preview

@pkozlowski-opensource Let me know if I should proceed

icfantv:
@sirajc i would say go ahead and please submit a PR and we'll do it iteratively.  we had a meeting yesterday and the demos should be structured as the alert demo:  most simple/basic first, the more advanced as appropriate.

to be clear, we don't want 10 demos, but we want to showcase the different possibilities as applicable.  additionally, we want 100% API coverage in the demos, meaning, across all the demos, we are using (demoing) every public API property/component.

sirajc:
Thanks @icfantv , I am following every bit of alert demo structure. There is nothing much in Collapse so 2 demos is sufficient one basic and one showing the usage of default value. 
Will raise a PR and then we can discuss on it further, I am open to discussions and consume feedback so no worries on number of iterations

maxokorokov:
@pkozlowski-opensource any ideas on how to handle default values for accessors to private variables? I mean this:

```javascript
class Pagination {
  private _page = 0;
  
  /**
  * Comment goes here, but what's the default value?...
  */
  @Input() set page(value: number | string) { ... }
}
```

Can't find simpler solution than adding it directly, even though the value is in two places in this case...

```javascript
  /**
  * Comment goes here, but what's the default value?...
  * @default 0
  */
```

pkozlowski-opensource:
@maxokorokov hmmm, we don't support it right now with the default API extraction in place. Could you please add this use-case to #222?

In your particular case `0` is an implementation detail (somehow) so not showing default value won't do much harm. But yeh, we need to support this....

wesleycho:
Going to close this, this looks done minus pending PRs

PR #291, opened 2016-06-29
docs(demo): add demo component for collapse

sirajc:
Adding Demo Components for collapse as part of #264

### DIFF
--- a/demo/src/app/components/collapse/demos/basic/basic.collapse.ts
+++ b/demo/src/app/components/collapse/demos/basic/basic.collapse.ts
@@ -1,0 +1,12 @@
+import { Component } from '@angular/core';
+
+import { NgbCollapse } from '@ng-bootstrap/collapse';
+
+@Component({
+  selector: 'ngbd-collapse-basic',
+  template: require('./basic.collapse.html'),
+  directives: [NgbCollapse]
+})
+export class CollapseBasicComponent {
+  private isCollapsed = false;
+}
--- a/demo/src/app/components/collapse/collapse.component.html
+++ b/demo/src/app/components/collapse/collapse.component.html
@@ -1,0 +1,6 @@
+<ngbd-content-wrapper component="Collapse">
+  <ngbd-api-docs directive="NgbCollapse"></ngbd-api-docs>
+  <ngbd-example-box demoTitle="Demo" [htmlSnippet]="basicHtmlContent" [tsSnippet]="basicTsContent">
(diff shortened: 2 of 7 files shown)


demo/src/app/components/collapse/demos/default-close/default.close.collapse.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { NgbCollapse } from '@ng-bootstrap/collapse';
+
+@Component({
+  selector: 'ngbd-collapse-default-close',
+  template: require('./default.close.collapse.html'),
+  directives: [NgbCollapse]
+})
+export class CollapseDefaultCloseComponent {
+  isCollapsed: boolean = true; // Keep content collapsed by default
```
icfantv: should be `private`, don't need data type as it's inferred from the initial value.

demo/src/app/components/collapse/demos/basic/basic.collapse.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { NgbCollapse } from '@ng-bootstrap/collapse';
+
+@Component({
+  selector: 'ngbd-collapse-basic',
+  template: require('./basic.collapse.html'),
+  directives: [NgbCollapse]
+})
+export class CollapseBasicComponent {
+  isCollapsed: boolean = false;
```
icfantv: don't need the data type here as it's inferred.

icfantv:
@sirajc, a couple notes for multiple files:
1. please ensure there's a newline at the end of every file
2. please match the existing naming convention for files and directories (e.g. `default.close.collapse.ts` ==> `default.collapse.ts` and `default-close` ==> `default`)

icfantv:
Unless the rest of the team has anything else to add, other than my notes, this LGTM.

Foxandxss:
My only concerns is what we talked already, 20 demos for one component. Do we really need a demo where by default is open and another where by default is closed? Is too much imo.

pkozlowski-opensource:
Agree with @Foxandxss - in case of this simple directive just one demo is enough (we can stick with the default open).

@sirajc could you please make the changes corresponding to the comments so we can merge it? Thnx!

sirajc:
Ok, will update PR as per comments provided by you all

sirajc:
Updated
* Rebased with master
* Type inference
* private isCollapsed
* Removed default close
* New line in all files at end

icfantv:
LGTM

PR #294, opened 2016-06-29
docs(progressbar): add api docs

pkozlowski-opensource:


src/progressbar/progressbar.ts
```diff
@@ -14,10 +17,30 @@ import {getValueInRange, toBoolean} from '../util/util';
   `
 })
 export class NgbProgressbar {
+  /**
+   * Maximal value to be displayed in the progressbar.
+   */
   @Input() max: number = 100;
+
+  /**
+   * A flag indicating if a progress bar should be animated when the value changes. Takes effect only for browsers
```
wesleycho: Is this note needed about CSS3 animations? I feel like all the browsers we're supporting support them.

wesleycho:
Other than my question, this LGTM

src/progressbar/progressbar.ts
pkozlowski-opensource: Well, there is still IE9 so I'm going to leave this one in place for now.

PR #299, opened 2016-06-29
docs(rating): add API docs

pkozlowski-opensource:


src/rating/rating.ts
```diff
@@ -17,12 +20,37 @@ export class NgbRating implements OnInit {
...
   @Input() max: number = 10;
+
+  /**
+   * Current rating.
+   */
   @Input() rate: number;
-  @Input() readonly: boolean;
 
+  /**
+   * A flag indicating if rating could be read only (true) or updated (false).
```
Foxandxss: I find this a bit verbose. I don't think we need to explain what true or false means in this case. I mean, bit obvious from the name and the type of the input (boolean).

src/rating/rating.ts
```diff
@@ -17,12 +20,37 @@ export class NgbRating implements OnInit {
...
+   */
+  @Input() readonly = false;
+
+  /**
+   * An event fired when a user is hovering over a given rating.
+   * Event's payload equals to the rating being hoovered over.
```
Foxandxss: hoovered, that is a typo.

src/rating/rating.ts
```diff
@@ -17,12 +20,37 @@ export class NgbRating implements OnInit {
...
+   */
   @Output() hover = new EventEmitter<number>();
+
+  /**
+   * An event fired when a user stops hovering over a given rating.
+   * Event's payload equals to the rating of the last item being hoovered over.
```
Foxandxss: ditto

src/rating/rating.ts
pkozlowski-opensource: Could propose an alternative wording?

Foxandxss: What about...

"A flag indicating whether the rating is read only"

Probably @icfantv knows more english than we do.

icfantv: i'm ok w/ @Foxandxss' proposal.

wesleycho: Or how about `A flag indicating if rating can be updated`. I think this is actually a little better at communicating the intent

PR #300, opened 2016-06-29
demo: use hash location strategy

### LINKED ISSUE #293: demo site: switch router to hash location strategy
Since we are going to deploy to gh-pages where we can't have server side redirections we need to stick with the hash location strategy. It should be a very simple fix, essentially the same as in https://github.com/angular/angular2-seed/blob/452d6daad7c024860d9c84947a812ce4cbcb1910/src/main.browser.ts#L14

Foxandxss:
Fixes #293


pkozlowski-opensource:
LGTM. Will merge as soon as CI is green :-)

PR #301, opened 2016-06-29
demo: properly build for dist

Foxandxss:
Now it builds a proper build for dist which work on gh-pages.

We may need to adjust a parameter depending on where we will deploy, but we will cross that bridge eventually.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -147,7 +147,7 @@ gulp.task(
 
 gulp.task(
     'build:demo', ['clean:demo', 'generate-docs'],
-    shell.task(['webpack --config webpack.demo.js --progress --profile --bail']));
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -147,7 +147,7 @@ gulp.task(
 
 gulp.task(
     'build:demo', ['clean:demo', 'generate-docs'],
-    shell.task(['webpack --config webpack.demo.js --progress --profile --bail']));
+    shell.task(['MODE=build webpack --config webpack.demo.js --progress --profile --bail']));
```
pkozlowski-opensource: Is the `MODE` flag used anywhere?

Foxandxss: Yes, at webpack config.

PR #307, opened 2016-06-29
docs(accordion): add api docs

wesleycho:
- Add api docs for the accordion and panel components

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -33,9 +33,26 @@ let nextId = 0;
   directives: [NgbCollapse]
 })
 export class NgbPanel {
+  /**
+   *  A flag determining whether the panel is disabled or not.
+   *  When disabled, the panel cannot be toggled.
+   */
   @Input() disabled = false;
+
+  /**
+   *  An optional id for the panel. The id should be unique.
+   *  If not provided, it will be auto-generated.
+   */
   @Input() id = `ngb-panel-${nextId++}`;
+
+  /**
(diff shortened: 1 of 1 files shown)


src/accordion/accordion.ts
```diff
@@ -33,9 +33,25 @@ let nextId = 0;
   directives: [NgbCollapse]
 })
 export class NgbPanel {
+  /**
+   *  A flag determining whether the panel is disabled or not.
+   *  When disabled, the panel cannot be toggled.
```
pkozlowski-opensource: cannot be toggled -> cannot be selected?

src/accordion/accordion.ts
```diff
@@ -33,9 +33,25 @@ let nextId = 0;
...
+   */
   @Input() disabled = false;
+
+  /**
+   *  An optional id for the panel. The id should be unique.
```
pkozlowski-opensource: I would add a note: "If not provided by a user it will be auto-generated." or sth along those lines.

pkozlowski-opensource:
LGTM, just left 2 remarks.

src/accordion/accordion.ts
wesleycho: If it is open and later becomes disabled, then one will not be able to close the panel. This language was intentional.

pkozlowski-opensource: You are right. Don't know why but I had tabset on my mind while I was writing this....

wesleycho:
Merging this as per comments

PR #308, opened 2016-06-29
docs(carousel): add api docs

wesleycho:
- Add documentation for carousel api

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -54,9 +54,24 @@ export class NgbCarousel implements AfterContentChecked,
   @ContentChildren(NgbSlide) private _slides: QueryList<NgbSlide>;
   private _slideChangeInterval;
 
+  /**
+   *  Amount of time in milliseconds before next slide is shown.
+   */
   @Input() interval = 5000;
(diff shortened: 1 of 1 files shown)


src/carousel/carousel.ts
```diff
@@ -54,9 +54,24 @@ export class NgbCarousel implements AfterContentChecked,
   @ContentChildren(NgbSlide) private _slides: QueryList<NgbSlide>;
   private _slideChangeInterval;
 
+  /**
+   *  Amount of time before next slide is shown.
```
icfantv: in _ms_?

src/carousel/carousel.ts
```diff
@@ -54,9 +54,24 @@ export class NgbCarousel implements AfterContentChecked,
...
+   */
   @Input() interval = 5000;
+
+  /**
+   *  Whether user can navigate between the first and last slide.
```
icfantv: Whether slide navigation can wrap from the last slide to the first slide.  (or something close)

What you wrote is normal carousel behavior - navigating between first and last slide.  :-)

PR #312, opened 2016-06-30
fix(demo): add cursor pointer for code icon

sirajc:
Fixes: Pointer stays as arrow when we hover over code icon in Example wrapper

### DIFF
--- a/demo/src/app/components/shared/example-box/example-box.component.html
+++ b/demo/src/app/components/shared/example-box/example-box.component.html
@@ -3,7 +3,7 @@
     <div class="card-header">
       {{ demoTitle }}
       <div class="pull-xs-right">
-        <img src="img/code-icon.png" (click)="showCode = !showCode"/>
+        <img role="button" src="img/code-icon.png" (click)="showCode = !showCode"/>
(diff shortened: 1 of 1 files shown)


icfantv:
@pkozlowski-opensource, @Foxandxss: I'm reading [this document](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role) and it's sounding like we should make this an `<input type="image" role="button"/>`.  The latter attribute will give us both ARIA compliance and the pointer cursor because BS4 has a `[role="button"]` selector.  Thoughts?

wesleycho:
Wouldn't wrapping it in an anchor tag and using that as the click target accomplish the purpose as well?

sirajc:
@wesleycho Wrapping it in anchor shows cursor, but then we need to add `href="#"` and hovering shows the address. Also we need to add a style on a, because it renders `underlined` on hover
Since this is not a link and actually an action, IMO this should be a button

@icfantv just adding `role="button"` on `img` is also working great, Have updated PR
@wesleycho your thougts?

wesleycho:
No, you don't need to add `href="#"`, just `href`.

PR #313, opened 2016-06-30
chore(demo): add routerLinkActive on sidenav links

sirajc:
Highligh active link in SideNav, See preview below
image

### DIFF
--- a/demo/src/style/app.scss
+++ b/demo/src/style/app.scss
@@ -65,11 +65,6 @@ p.github-buttons > iframe {
   display: block
 }
 
-.bd-toc-item .active {
-  margin-top: 1rem;
-  margin-bottom: 1rem
-}
-
 .bd-toc-item a.active {
   margin-top: inherit;
   margin-bottom: inherit;
@@ -101,7 +96,7 @@ p.github-buttons > iframe {
 .bd-sidebar .nav > .active:focus > a,
 .bd-sidebar .nav > .active:hover > a,
 .bd-sidebar .nav > .active > a {
-  font-weight: 500;
+  font-weight: 700;
   color: #373a3c;
   background-color: transparent
 }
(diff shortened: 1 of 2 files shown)


sirajc:
For active links increased the font-weight and used existing color `#373a3c` from _app.scss_

demo/src/style/app.scss
```diff
@@ -89,6 +89,11 @@ p.github-buttons > iframe {
   cursor: pointer;
 }
 
+.bd-sidebar .nav > li > a.active{
+  font-weight: 700;
```
icfantv: how does the UI look w/ `bold` instead of a number?

icfantv:
@sirajc, i just realized there's CSS already there for this.  IIRC, the `active` class just needs to be added to the `li` tag around the anchor.  [According to this](https://github.com/angular/angular/blob/master/modules/%40angular/router/src/directives/router_link_active.ts) it should just be a matter of moving the `routerLinkActive` to the `li` tag and removing the `[]`.

demo/src/style/app.scss
```diff
@@ -89,6 +89,11 @@ p.github-buttons > iframe {
   cursor: pointer;
 }
 
+.bd-sidebar .nav > li > a.active{
```
wesleycho: Add a space before `{`

icfantv: that's not his code - that was mine originally.  also i neglected to put newlines on some compound css defs.

icfantv: gah, nevermind.  i missed that change.  i still need to put the newlines in...

sirajc:
@icfantv Added `routerLinkActive` on `li` and removed CSS rule added by me
This is how it looks now
image

It added a good amount of margin, Also do we need to make it bold?

icfantv:
Ooof, that looks horrible.  Please go ahead and:
1. remove the `.bd-toc-item .active` selector as this behavior is not wanted and isn't used elsewhere.
2. change the `font-weight` from 500 to 700 for the `.bd-sidebar .nav > .active > a` group of selectors.

Thanks.

sirajc:
@icfantv This now looks like First screenshot in this PR (darker and bold without margins)

icfantv:
Sweet!  If you're happy with it then LGTM.

PR #316, opened 2016-06-30
refactor(progressbar): add defaults values so they show up in API docs

pkozlowski-opensource:


src/progressbar/progressbar.ts
```diff
@@ -41,7 +41,7 @@ export class NgbProgressbar {
   /**
    * Current value to be displayed in the progressbar. Should be smaller or equal to "max" value.
    */
-  @Input() value: number;
+  @Input() value: number = 0;
```
Foxandxss: wasn't the type not necessary anymore? (I can see the other two as valid)

pkozlowski-opensource: you are right, good catch. Thnx!

PR #320, opened 2016-06-30
feat(pagination): change pagination display size (sm/lg)

### LINKED ISSUE #314: Pagination: support "size" input
We should introduce the `size` `@Input` that could take the following values: `sm` or `lg` (empty by default). When this input is set to something else than empty we should add `size-sm` or `size-lg` class. This is very similar to what we do for a modal window here: https://github.com/ng-bootstrap/core/blob/fb4d331443f48321826d09a2292d3bd68ae21168/src/modal/modal_window.ts#L15

Ref: http://v4-alpha.getbootstrap.com/components/pagination/#sizing

maxokorokov:
Adding input 'size' to pagination component to support bootstrap 'pagination-lg' and 'pagination-sm' styles

Closes #314

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -45,6 +45,10 @@ function getLink(nativeEl: HTMLElement, idx: number): HTMLAnchorElement {
   return <HTMLAnchorElement>nativeEl.querySelectorAll('li')[idx].querySelector('a');
 }
 
+function getList(nativeEl: HTMLElement) {
+  return <HTMLUListElement>nativeEl.querySelector('ul');
+}
+
 function normalizeText(txt: string): string {
   return txt.trim().replace(/\s+/g, ' ');
 }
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.spec.ts
```diff
@@ -243,6 +247,40 @@ describe('ngb-pagination', () => {
...
+           fixture.detectChanges();
+           expect(classList.length).toBe(2);
+           expect(classList).toContain('pagination');
+           expect(classList).toContain('pagination-lg');
+
+           // removing large size
+           fixture.componentInstance.size = '';
+           fixture.detectChanges();
+           expect(classList.length).toBe(1);
+           expect(classList).toContain('pagination');
```
pkozlowski-opensource: Please add `expect(classList).not.toContain('pagination-lg');`

maxokorokov: Hmm, you don't like checking length to be 1?

pkozlowski-opensource:
LGTM part from the remark about tests.

maxokorokov:
Thanks for the explanation, amended

pkozlowski-opensource:
LGTM. Will merge as soon as TravisCI is green.

PR #325, opened 2016-06-30
fix(buttons): add  data-toggle='buttons' to radio group

pkozlowski-opensource:
Spotted when working on buttons demo

cc.: @maxokorokov


maxokorokov:
Yeah, I was using the markup from the [bs4 documentation](http://v4-alpha.getbootstrap.com/components/buttons/#checkbox-and-radio-buttons) as a base, but this is good in case you forget it. Might also add the 'btn-group' style as well, though I don't know where to put the boundary on what we should auto-add.

pkozlowski-opensource:
> Might also add the 'btn-group' style as well, though I don't know where to put the boundary on what we should auto-add. There is also the autocomplete=off in their examples on the input

Yeh, we could add those as well.  `btn-group` would be an easy fix. I will update the PR.

pkozlowski-opensource:
@maxokorokov updated, PTAL

maxokorokov:
Looks good!

PR #327, opened 2016-06-30
fix: add empty hrefs for styling

### LINKED ISSUE #142: Dealing with styling of "a" tags (with or without hrefs?)
A very common markup in Angular version of bootstrap directives goes like this: `<a (click)="prev()">Previous</a>`

Pure bootstrap version would have an "empty" href (`<a (click)="prev()" href="#">Previous</a>`) to provide appropriate styling (cursor) but this is not needed in ng-version and can interfere with routing.

The question: how do we deal with this situation? Few ideas:
- pass `$event` and call `preventDefault()` on it; this is what we do in the accordion, for example: https://github.com/ng-bootstrap/core/blob/1d912de06578aca1eaea9ad69942978090da43ff/src/accordion/accordion.ts#L42
- remove `href` altogether and add cursor styling through CSS (this would be quite easy with ng2 since we've got styles encapsulation)
- try to convince bootstrap folks to "fix it" <- not even sure it makes sense

Now, I must admit that I don't know what is the impact of removing `href` on accessibility and keyboard navigation, so the encapsulated CSS path might be a non-option in the end... 

Thoughts?

pkozlowski-opensource:
Just tested on the accordion example and it looks like empty `href`s are not causing any negative consequences and assure proper styling. We will have to keep an eye on this item, but let's try those empty `href`s.

Fixes #142

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -24,7 +24,7 @@ let nextId = 0;
     <div class="panel panel-default" [class.panel-open]="open">
       <div class="panel-heading" role="tab" [id]="id">
         <h4 class="panel-title">
-          <a tabindex="0" (click)="toggleOpen($event)"><span [class.text-muted]="disabled">{{title}}</span></a>
+          <a tabindex="0" (click)="toggleOpen($event)" href><span [class.text-muted]="disabled">{{title}}</span></a>
         </h4>
       </div>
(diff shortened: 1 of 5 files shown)


pkozlowski-opensource:
Hmm, actually it fails tests with:

> Some of your tests did a full page reload!

So not good. Closing for now.

PR #329, opened 2016-06-30
feat(angular): upgrade to rc4

wesleycho:
- Upgrade angular to rc4


demo/src/main.ts
```diff
@@ -1,6 +1,6 @@
 import { enableProdMode } from '@angular/core';
 import { bootstrap } from '@angular/platform-browser-dynamic';
-import { ELEMENT_PROBE_PROVIDERS } from '@angular/platform-browser';
+import { ELEMENT_PROBE_PROVIDERS } from '@angular/platform-browser/src/dom/debug/ng_probe';
```
pkozlowski-opensource: This sounds bad... Why do we need `ELEMENT_PROBE_PROVIDERS`?

wesleycho: No clue - @Foxandxss, do you know the reason? Removing the logic around here doesn't seem to break anything.

Foxandxss: That is a good question, I carried that over from my starter. Fire with it.

Foxandxss:
I merged this by  hand.

PR #332, opened 2016-07-01
feat(pagination): support for 'maxSize', 'ellipses' and 'rotate'

### LINKED ISSUE #210: Pagination: support "max-size", "rotate" and "ellipses"
Same as in http://angular-ui.github.io/bootstrap/#/pagination

maxokorokov:
Implements inputs in question for the pagination component + minor cleanup (reorder methods alphabetically)

Some implementation details:
- ellipsis link is not clickable
- ellipses are stored in the 'pages' array as -1
- ellipses always come together with boundary page numbers (1... or ... n)
- new '_pageCount' field to store the actual last page number

Plunker to play with: http://plnkr.co/edit/UIeOOHCgfyxXIkS94ClD?p=preview

Closes #210

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -267,6 +267,132 @@ describe('ngb-pagination', () => {
            expect(classList).toContain('pagination-123');
          });
        })));
+
+    it('should render and respond to maxSize change correctly', async(inject([TestComponentBuilder], (tcb) => {
+         const html =
+             '<ngb-pagination collectionSize="70" [page]="page" [maxSize]="maxSize" [ellipses]="ellipses"></ngb-pagination>';
+
+         tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['-« Previous', '+1', '2', '3', '4', '5', '6', '7', '» Next']);
+
+           // disabling ellipsis
+           fixture.componentInstance.ellipses = false;
+
+           // limiting to 3 page numbers
+           fixture.componentInstance.maxSize = 3;
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['-« Previous', '+1', '2', '3', '» Next']);
+
+           fixture.componentInstance.page = 3;
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['« Previous', '1', '2', '+3', '» Next']);
+
+           fixture.componentInstance.page = 4;
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['« Previous', '+4', '5', '6', '» Next']);
+
+           fixture.componentInstance.page = 7;
+           fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.ts
```diff
@@ -110,13 +136,24 @@ export class NgbPagination implements OnChanges {
   @Output() pageChange = new EventEmitter();
 
   /**
+   *  If true current page will always be displayed in the middle.
```
wesleycho: Maybe reword as `Whether to rotate pages with more pages than maxSize. Current page will be in middle.`

src/pagination/pagination.ts
```diff
@@ -125,22 +162,101 @@ export class NgbPagination implements OnChanges {
...
+      end = this._page + rightOffset;
+    }
+
+    return [start, end];
+  }
+
+  /**
+   * Paginates page numbers based on maxSize items per page
+   */
+  private _applyPagination(): [number, number] {
```
wesleycho: I don't think this needs to be documented with jsdoc. Maybe change to `//`?

wesleycho:
This mostly LGTM. I just have a question - do we want to use jsdoc to document private methods? Maybe we need a way of excluding comments from documentation with some sort of hook.

pkozlowski-opensource:
> I just have a question - do we want to use jsdoc to document private methods? Maybe we need a way of excluding comments from documentation with some sort of hook.

We are never going to include private methods in the generated API docs so how we document those is totally up to us. That is - we are not going expose this to directive users.

wesleycho:
Ah, didn't realize that was already built in - disregard that comment then.

src/pagination/pagination.spec.ts
```diff
@@ -267,6 +267,132 @@ describe('ngb-pagination', () => {
...
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['« Previous', '1', '2', '+3', '» Next']);
+
+           fixture.componentInstance.page = 4;
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['« Previous', '+4', '5', '6', '» Next']);
+
+           fixture.componentInstance.page = 7;
+           fixture.detectChanges();
+           expectPages(fixture.nativeElement, ['« Previous', '+7', '-» Next']);
```
pkozlowski-opensource: This doesn't look right. Shouldn't we get `['« Previous', '5', '6', '+7', '-» Next']` in this case?

pkozlowski-opensource:
This also mostly LGTM, although one test is suspicious - left a comment. @maxokorokov - could you PTAL?

maxokorokov:
@pkozlowski-opensource The test is ok for me as there is no rotation, so page numbers a paginated with maxSize=3. The third page will have only number 7. Your example is true for the following test when the rotation is on.

src/pagination/pagination.ts
maxokorokov: @wesleycho agreed, so it will be in line with other docs

maxokorokov:
Rebased and changed `rotate` doc to `Whether to rotate pages when maxSize > number of pages. Current page will be in the middle`

PR #333, opened 2016-07-01
feat: add npmignore

wesleycho:
- Add basic npmignore file


pkozlowski-opensource:
@wesleycho we are not going to release root of this repo, but rather a specific sub-folder (dist, most probably). So let's wait on this one.

wesleycho:
Ahh gotcha - this will be easy to recreate when we do do it, I'll close this.

PR #335, opened 2016-07-01
feat(tooltip): initial version of the tooltip directive

pkozlowski-opensource:

PR #337, opened 2016-07-02
feat(popover): add popover directive

### LINKED ISSUE #9: Popover
Recreate BS3 Popover with ng2 component.

pkozlowski-opensource:
This is essentially the same as  #335 but for popovers. As you can see this is mostly copy & paste but I want to land this one first and then factor out common code with the tests in place.

Closes #9


wesleycho:
LGTM

PR #341, opened 2016-07-02
chore(demo): update logo and css

icfantv:
* fix css formatting
* another take on new bootstrap logo and favicon


demo/src/style/app.scss
```diff
@@ -121,34 +121,33 @@ div.api-doc-component > h2 > a:hover {
...
+  0 -5px 0 #fff,
+  0 -6px 0 #fff;
+  box-shadow: 0 1px #fff, 0 3px #0275d8;
+  word-spacing: 0.5em;
+  font-size: 2.5em;
   border-bottom: 2px solid #fff;
   display: table;
 }
 
 div.api-doc-component > h2:hover {
```
wesleycho: This selector could be nested under the above as

```scss
div.api-doc-component > h2 {
  ...
  &:hover {
    ...
  }
}
```

wesleycho:
I like the logo, but I think @Foxandxss said he didn't like the red being used.

One thing that would be good for our doc styles is to pull out the colors into semantic variables, that way we can theme it easily.

demo/src/style/app.scss
icfantv: i'd no idea you could do something like that.

icfantv:
@pkozlowski-opensource updated per comments on slack.

PR #343, opened 2016-07-04
docs(demo): add carousel demo

pkozlowski-opensource:


pkozlowski-opensource:
This PR adds demo for the carousel component. More importantly, though, it:
* put in place rudimentary scripts to generate demo skeleton (see `demo-gen.js`). Since I couldn't find any satisfactory way of reducing code duplication I've written a script to generate skeletons. If people are OK with this approach I will document and make this scrip "usable"
* proposes change to the demo structure - it removes a barrel for each demo and normalizes `demos` barrel. This significantly reduces amount of code needed for each demo component (and `demos/index.ts` can be auto-generated - check the script).

pkozlowski-opensource:
@icfantv @Foxandxss PTAL

misc/demo-gen.js
```diff
@@ -0,0 +1,70 @@
...
+
+function genDemosIndex(component) {
+  const base = `demo/src/app/components/${component}/demos`;
+  const path = `${base}/*/`;
+  const demoNames = glob.sync(path, {}).map((dir) => {
+      const dirNoEndingSlash = dir.substr(0, dir.length - 1);
+      return dirNoEndingSlash.substr(dirNoEndingSlash.lastIndexOf('/') + 1);
+    }).sort();
+
+  const demoImports =demoNames.map((demo) => {
```
Foxandxss: Nit: Space after demo. In fact, are you sure this passed clang?

Foxandxss:
I like it.

misc/demo-gen.js
pkozlowski-opensource: I don't think we've got clang for scripts in misc. Let me check if I can activate it (and if it is not messing up templates).

icfantv:
@pkozlowski-opensource, damn dude.  This is slick.  Really slick.  I was thinking about how to do the auto generation and it never occurred to me to build the TS components from scratch.  Nice work.

PR #348, opened 2016-07-04
docs(demo): add tooltip basic demo

pkozlowski-opensource:


demo/src/app/components/tooltip/demos/basic/tooltip-basic.html
```diff
@@ -0,0 +1,12 @@
+<button type="button" class="btn btn-secondary" placement="top" ngbTooltip="Tooltip on top">
```
icfantv: do we need `role="button"` on all these?

icfantv:
@pkozlowski-opensource this also looks awesome.  Please see my comment re showing the full possible range of triggers (say, separate demo w/ dropdown to let the user select?) so that we have 100% API coverage.

Also, we should document (eventually) the valid list of events supported by the `triggers` property.

demo/src/app/components/tooltip/demos/basic/tooltip-basic.html
pkozlowski-opensource: I don't think we need `role="button"` on native `<button>` elements, ex.: http://stackoverflow.com/questions/26934719/do-i-need-role-button-on-a-button

pkozlowski-opensource:
@icfantv yup, this is just very basic demo. We should have more. Let's discuss during the today's meeting what we need to add.

sendilkumarn:
@pkozlowski-opensource Is it possible to join the conversation?

pkozlowski-opensource:
@sendilkumarn sorry, we don't have any public dev channel yet... Will discuss this with the entire team to see how we can make things more "in the open".

PR #349, opened 2016-07-04
docs(demo): add popovers basic demo

pkozlowski-opensource:


demo/src/app/components/popover/demos/basic/popover-basic.html
```diff
@@ -0,0 +1,19 @@
+<button type="button" class="btn btn-secondary" placement="top"
```
icfantv: Do we want `role="button"` on these for i11y?

icfantv:
@pkozlowski-opensource this looks awesome.  Maybe not for alpha, but as I'd indicated earlier, I'd like 100% API coverage so I'd like a demo of all the triggers, be that in this PR or a subsequent one.

PR #354, opened 2016-07-05
feat(tabset): support tab 'change' event

### LINKED ISSUE #267: Tabs: support "select" output
Tabset should have the `select` output (event) which should be fired before a new tab is selected. Event's payload should have:
* id of a tab to be selected
* id of the actual tab
* `preventDefault()` method to abort tab change.

Ref: http://v4-alpha.getbootstrap.com/components/navs/#events (but IMO we should implement only `show`)

maxokorokov:
Adds the 'change' output, that is fired right before tab selection. Payload contains current tab id, next tab id and preventDefault() function to cancel tab selection.

Closes #267

### DIFF
--- a/src/tabset/tabset.ts
+++ b/src/tabset/tabset.ts
@@ -6,7 +6,9 @@ import {
   Directive,
   TemplateRef,
   ContentChild,
-  AfterContentChecked
+  AfterContentChecked,
+  Output,
+  EventEmitter
 } from '@angular/core';
 
 let nextId = 0;
@@ -50,6 +52,15 @@ export class NgbTab {
 }
 
 /**
+ * The payload of the tab change event
+ */
+export interface NgbTabChangeEvent {
+  activeId: String;
+  nextId: String;
+  preventDefault();
+}
+
+/**
  * A component that makes it easy to create tabbed interface.
  */
 @Component({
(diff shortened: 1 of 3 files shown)


src/tabset/tabset.ts
```diff
@@ -6,8 +6,10 @@ import {
   Directive,
   TemplateRef,
   ContentChild,
-  AfterContentChecked
+  AfterContentChecked,
+  Output
 } from '@angular/core';
+import {EventEmitter} from '@angular/common/src/facade/async';
```
pkozlowski-opensource: This should be imported from `@angular/core` as it is re-exported from there. Any path that has `src` is considered "private" and shouldn't be used.

src/tabset/tabset.ts
```diff
@@ -83,10 +94,22 @@ export class NgbTabset implements AfterContentChecked {
...
+   * A tab change event fired right before the tab selection happens
+   */
+  @Output() change = new EventEmitter<NgbTabChangeEvent>();
+
   select(tabIdx: string) {
     let selectedTab = this._getTabById(tabIdx);
-    if (selectedTab && !selectedTab.disabled) {
-      this.activeId = selectedTab.id;
+    if (selectedTab && !selectedTab.disabled && this.activeId !== selectedTab.id) {
+      let preventSelection = false;
```
pkozlowski-opensource: In the DOM this is named `defaultPrevented` and IMO it would be good to use the same name: https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented

src/tabset/tabset.ts
maxokorokov: Should be careful with WebStorm autocompletion :/

pkozlowski-opensource:
LGTM, there are just 2 minor comments that need resolving. Good stuff :+1:

src/tabset/tabset.ts
maxokorokov: Ok, so throw away the function and leave just the `defaultPrevented` flag?

pkozlowski-opensource: No, no. Just rename `preventSelection` to `defaultPrevented`. The "public API" for users should still be `preventDefault()`

maxokorokov: Yeah, got that already, thanks. Will just rename the variable

maxokorokov:
Amended, fixed two comments and also added `NgbTabChangeEvent` to exports in `index.js`

Issue #344, opened 2016-07-04
Carousel: setInterval not cleared up correctly


pkozlowski-opensource:
Try clicking on prev / next buttons and you will see that `clearInterval` is not cleaned up properly (you can see slides cycling faster / randomly).

I can see few places where the behavior is fishy: 
* we need to re-think clicking on next/prev with mouseover / mouseleave - those seems to conflict atm

pkozlowski-opensource:
Actually 69565f3 fixes it....

PR #362, opened 2016-07-05
refactor(carousel): make it clear when timers are started / stopped

pkozlowski-opensource:
Previously it wasn't clear when timers are started / canceled. This refactring hopefully makes it more obvious and opens doors to easy fixes for issues like #344


icfantv:
LGTM.

PR #365, opened 2016-07-05
fix(dropdown): added ngbdropdown to have dropdown class always

### LINKED ISSUE #361: Dropdown: dropdown-menu is not properly aligned with dropdown-toggle button
screen shot 2016-07-05 at 8 33 19 pm

sendilkumarn:

 - [X] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/core/blob/master/CONTRIBUTING.md) guide.

fixes #361

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -19,6 +19,7 @@ describe('ngb-dropdown', () => {
          fixture.detectChanges();
          const compiled = fixture.nativeElement;
 
+         expect(getDropdownEl(compiled)).toHaveCssClass('dropdown');
          expect(getDropdownEl(compiled)).not.toHaveCssClass('open');
        });
      })));
@@ -30,6 +31,7 @@ describe('ngb-dropdown', () => {
          fixture.detectChanges();
          const compiled = fixture.nativeElement;
 
+         expect(getDropdownEl(compiled)).toHaveCssClass('dropdown');
          expect(getDropdownEl(compiled)).toHaveCssClass('open');
        });
      })));
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -6,7 +6,12 @@ import {Directive, Input, Output, HostListener, EventEmitter} from '@angular/cor
 @Directive({
   selector: '[ngbDropdown]',
   exportAs: 'ngbDropdown',
-  host: {'[class.open]': '_open', '(keyup.esc)': '_closeFromOutside()', '(document:click)': '_closeFromOutside()'}
+  host: {
+    'class': 'dropdown',
+    '[class.open]': '_open',
+    '(keyup.esc)': '_closeFromOutside()',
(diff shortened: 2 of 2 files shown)


pkozlowski-opensource:
This is a good fix but we need to add a check for this class in one of the unit tests.

wesleycho:
Do we want tests for all these classes on host? Wouldn't this theoretically mean we need to do this for the other components too?

pkozlowski-opensource:
> Do we want tests for all these classes on host? Wouldn't this theoretically mean we need to do this for the other components too?

We do have tests for other cases already!

wesleycho:
Ah, was not aware - fair enough then

sendilkumarn:
@pkozlowski-opensource I checked, since there was no check for dropdowntoggle class. I kinda omitted it. Again I think it should be there. I will update

sendilkumarn:
@pkozlowski-opensource ping

icfantv:
@sendilkumarn, please be patient.  we are all busy and Pawel will answer when he has a moment to do so.  he typically bikes home from work around now.

also, you forgot to run the clang formatter so the build failed.

sendilkumarn:
@icfantv just pinged to inform that the issue is fixed. 

point noted :)

src/dropdown/dropdown.spec.ts
```diff
@@ -274,6 +274,28 @@ describe('ngb-dropdown-toggle', () => {
          expect(dropdownEl).toHaveCssClass('open');
        });
      })));
+  it('should have css class dropdown on div and dropdown-toggle on button',
```
pkozlowski-opensource: This test should be moved to the ` describe('ngb-dropdown'` section and made much simpler (for example we don't need to have `ngbDropdown` inside at all)

pkozlowski-opensource: What I was saying is that we can test the functionality in question with the following markup only: `<div ngbDropdown></div>`

sendilkumarn: In that case dropdown-toggle testing functionality will be missing.

pkozlowski-opensource: It is covered by other tests already

sendilkumarn: Correct me if I am wrong. 

`expect(buttonEl).toHaveCssClass('dropdown-toggle');` 

The `dropdown-toggle` class is also set via host. Ideally there should be a test case right?

This test is not there, am I missing out anything?

pkozlowski-opensource: Right, but those are hosts of 2 different directives and we don't need to test them together!

IMO the whole change in this PR should boil down to sth like this: https://github.com/ng-bootstrap/core/commit/7dccb0c37d14c16a793f50d6e8de739952ab05fb

Could you update this PR accordingly?

sendilkumarn: Okay.

PR #368, opened 2016-07-06
refactor(index): cleanup index.ts

pkozlowski-opensource:
cc @Foxandxss

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,31 +1,35 @@
-import {NgbAccordion, NgbPanel} from './accordion/accordion';
+import {NGB_ACCORDION_DIRECTIVES} from './accordion/accordion';
+import {NGB_ALERT_DIRECTIVES, NgbAlert} from './alert/alert';
 import {NGB_CAROUSEL_DIRECTIVES} from './carousel/carousel';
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -1,31 +1,35 @@
-import {NgbAccordion, NgbPanel} from './accordion/accordion';
+import {NGB_ACCORDION_DIRECTIVES} from './accordion/accordion';
+import {NGB_ALERT_DIRECTIVES, NgbAlert} from './alert/alert';
```
Foxandxss: extra NgbAlert

Foxandxss:
Seems fine. I am confused in why you need to import some stuff like `NgbAlert` but not import the accordion parts in the previous line.

pkozlowski-opensource:
> I am confused in why you need to import some stuff like NgbAlert but not import the accordion parts in the previous line.

we need this to export `NGB_PRECOMPILE`. This is needed so people can put this array in the `precompile` attribute of a component where they want to use dynamic stuff (currently alerts, tooltips, popovers but more in the future).

Foxandxss:
Oh, missed that part. LGTM

PR #370, opened 2016-07-06
docs(demo): add more advanced tooltip demos

pkozlowski-opensource:
PTAL so I can do popovers demo (will be almost identical)


wesleycho:
LGTM

PR #374, opened 2016-07-06
chore: bundle cjs, esm and umd

Foxandxss:
This generates `cjs` in the root folder, `esm` and `umd` at `/bundles/`.

There is a problem tho, when doing the `esm` I get a lot of:

```
src/accordion/accordion.ts(10,8): error TS2307: Cannot find module '@angular/core'.
```

Weird enough, if I put all the typings I get a lot of duplicated issues, if I remove them, it looks for our custom matchers.

I had to put the custom matchers to remove some errors.

How to make TS see the @angular stuff?

/cc @robwormald


Foxandxss:
That is fixed.

Green light.

PR #376, opened 2016-07-06
chore(docs): demo and doc formatting and cleanup

icfantv:
* add support for `interface` parsing and documenting
* add `NgbTabChangeEvent` to tab documentation
* add support for formatting in description field
* updated component API documentation to use formatting where appropriate.

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -14,8 +14,9 @@ import {NgbCollapse} from '../collapse/collapse';
 let nextId = 0;
 
 /**
- * The NgbPanel directive builds on top of the NgbCollapse directive to provide a panel with collapsible body that can
- * collapsed or expanded by clicking on the panel's header.
+ * The <code>NgbPanel</code> directive builds on top of the <code>NgbCollapse</code> directive
+ * to provide a panel with collapsible body that can collapsed or expanded by clicking on the
+ * panel's header.
  */
 @Component({
   selector: 'ngb-panel',
@@ -73,8 +74,8 @@ export class NgbPanel {
 }
 
 /**
- * The NgbAccordion directive is a collection of panels.
- * NgbAccordion can assure that only panel can be opened at a time.
+ * The <code>NgbAccordion</code> directive is a collection of panels.
+ * <code>NgbAccordion</code> can assure that only panel can be opened at a time.
  */
 @Component({
   selector: 'ngb-accordion',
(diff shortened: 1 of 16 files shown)


pkozlowski-opensource:
@icfantv this is kind of big patch with many changes. Let's put it aside for alpha.0 as there are few changes in this PR that I don't think we should be doing - but I don't have enough free cycles to review it in the details.

icfantv:
Ok.  Totally understood.  Working with that script and the typescript reflection API was like doing surgery on a bowl of spaghetti.  I do think we should, at the very least, document the event object.  I'm ok with doing this on the description for the change handler for now.

If you could please list your concerns so they can be addressed, that would be good.

pkozlowski-opensource:
> If you could please list your concerns so they can be addressed, that would be good.

Will do as soon as I've got a minute. Knee-deep in final details before the release.

icfantv:
> Will do as soon as I've got a minute. Knee-deep in final details before the release.

No rush.

src/accordion/accordion.ts
```diff
@@ -73,8 +74,8 @@ export class NgbPanel {
 }
 
 /**
- * The NgbAccordion directive is a collection of panels.
- * NgbAccordion can assure that only panel can be opened at a time.
+ * The <code>NgbAccordion</code> directive is a collection of panels.
```
pkozlowski-opensource: I would _love_ that we add a markup pass on those docs instead of adding HTML tags.

pkozlowski-opensource:
@icfantv so, finally had time to get back to this PR. As mentioned previously it is a mixture of several distinct changes (I will _always_ prefer small, focused PRs as those are easier to review and merge):
* extraction of API docs (interface)
* formatting in  API docs (`<code>` tag)
* changes in spelling (GB -> US)

Let me tackle those one by one

=== extraction of interface API docs

To start this change has no tests so no-no for merging :-)

But formalities aside I think that we need to do deeper changes and extract info about both interfaces and classes (soon we are going to have a modal service). As such it requires more work and deserves a separate PR. Updating #222 to reflect this.

=== formatting in  API docs

IMO formatting using HTML tags is pain in a neck. We could easily add a step where we pass extracted doc through markup formatter. While doing so we should think about handling links.

All in all this is a bigger change and deserves a separate PR.

=== changes in spelling

I don't mind either way, just let's decide & stick to one form.

pkozlowski-opensource:
IMO this one PR is in fact a collection of several bigger changes that should be tackled in separate PRs. As such I'm going to close this one (it doesn't merge cleanly anymore). Let's keep discussing needed changes here if needed.

PR #383, opened 2016-07-07
chore(build): add gulp task to prepare npm package

pkozlowski-opensource:
cc @Foxandxss

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,7 @@ var sourcemaps = require('gulp-sourcemaps');
 var ddescribeIit = require('gulp-ddescribe-iit');
 var shell = require('gulp-shell');
 var ghPages = require('gulp-gh-pages');
+var gulpFile = require('gulp-file');
 var del = require('del');
 var merge = require('merge2');
 var clangFormat = require('clang-format');
@@ -12,7 +13,6 @@ var gulpFormat = require('gulp-clang-format');
 var runSequence = require('run-sequence');
 var tslint = require('gulp-tslint');
 var webpack = require('webpack');
-var webpackDemoConfig = require('./webpack.demo.js');
 
 var PATHS = {
   src: 'src/**/*.ts',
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -174,7 +196,7 @@ gulp.task('demo-push', function() { return gulp.src(PATHS.demoDist).pipe(ghPages
 gulp.task('clean', ['clean:build', 'clean:tests', 'clean:demo', 'clean:demo-cache']);
 
 gulp.task('build', function(done) {
-  runSequence('lint', 'enforce-format', 'ddescribe-iit', 'test', 'clean:build', 'cjs', 'esm', 'umd', done);
+  runSequence('lint', 'enforce-format', 'ddescribe-iit', 'test', 'clean:build', 'cjs', 'esm', 'umd', 'npm', done);
```
Foxandxss: Let's create a release task please, and if it can double check that you want to publish, better. This task is used as a command to see that everything is working and there is no error.

gulpfile.js
```diff
@@ -74,6 +74,28 @@ gulp.task('umd', function(cb) {
       webpackCallBack('webpack', cb));
 });
 
+gulp.task('npm', function() {
+  var pkgJson = require('./package.json');
+  var targetPkgJson = {};
+  var fieldsToCopy = ['version', 'description', 'keywords', 'author', 'repository', 'license', 'bugs', 'homepage'];
+
+  targetPkgJson['name'] = '@ng-bootstrap/ng-bootstrap';
```
Foxandxss: Are we sure this work? I mean, I have no idea in how to tell npm that we want a scoped package like that.

Foxandxss:
Sounds good to me.

gulpfile.js
pkozlowski-opensource: It should, see: https://docs.npmjs.com/getting-started/scoped-packages#initializing-a-scoped-package

We just need to remember to publish it with `npm publish --access=public`

PR #388, opened 2016-07-07
feat(positioning): implement rounding

RobJacobs:
Implements rounding returned values.  Also made some tweaks to bring the position and offset methods more inline with jQuery's latest approach:

offset: https://github.com/jquery/jquery/blob/master/src/offset.js#L78
position: https://github.com/jquery/jquery/blob/master/src/offset.js#L121


pkozlowski-opensource:
Will review after we land alpha.0. Don't want to take risks with the last-minute change.

PR #394, opened 2016-07-08
feat(scrollspy): added scrollspy and its demo

### LINKED ISSUE #5: Scrollspy
Recreate BS3 Scrollspy with ng2 component.

sendilkumarn:
fixes #5

### DIFF
--- a/src/scrollspy/scrollspy.spec.ts
+++ b/src/scrollspy/scrollspy.spec.ts
@@ -1,0 +1,102 @@
+import {inject, async, addProviders} from '@angular/core/testing';
+
+import {TestComponentBuilder} from '@angular/compiler/testing';
+
+import {Component} from '@angular/core';
+import {By} from '@angular/platform-browser';
+
+import {NgbScrollSpy, NgbScrollTarget} from './scrollspy';
+
+function expectToBeActive(componentElement, idx) {
+  let anchors = componentElement.querySelectorAll('a');
+  Array.from(anchors).forEach((anchor, aIdx) => {
+    if (aIdx === idx) {
+      expect(anchor).toHaveCssClass('active');
+    } else {
+      expect(anchor).not.toHaveCssClass('active');
+    }
+  });
+}
+
+describe('ngb-scrollspy', () => {
+  let html: string;
+
+  beforeEach(() => {
+    html = `
+      <div>
+        <navbar>
+          <nav>
+            <ul>
+              <li><a [ngbScrollSpy]="selectors[0]">Anchor 1</a></li>
+              <li><a [ngbScrollSpy]="selectors[1]">Anchor 2</a></li>
+              <li><a [ngbScrollSpy]="selectors[2]">Anchor 3</a></li>
+            </ul>
+          </nav>
+        </navbar>
+        <section ngbScrollTarget style="height: 100px;overflow: scroll;">
+          <div id="foo" style="height: 500px; width: 100px"></div>
+          <div id="bar" style="height: 500px; width: 100px"></div>
+          <div id="baz" style="height: 500px; width: 100px"></div>
+        </section>
+      </div>
+    `;
+  });
+
+  it('should have no active class when initialized', async(inject([TestComponentBuilder], (tcb) => {
+       tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         let anchors = fixture.nativeElement.querySelectorAll('a');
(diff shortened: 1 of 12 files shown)


src/scrollspy/scrollspy.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+          <div id="baz" style="height: 500px; width: 100px"></div>
+        </section>
+      </div>
+    `;
+  });
+
+  it('should have no active class when initialized', async(inject([TestComponentBuilder], (tcb) => {
+       tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
```
sendilkumarn: Am i missing anything here? It fails here that too only in firefox. It runs without an error in chrome.

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,77 @@
...
+    spies.forEach(
+        spy => this.renderer.setElementClass(spy.elementRef.nativeElement, 'active', spy.selector === this.selector));
+  }
+}
+
+@Directive({selector: '[ngbScrollTarget]'})
+export class NgbScrollTarget implements OnInit {
+  private _scrollHeight: number = 0;
+  private _offsetList: any[] = [];
+  constructor(private elementRef: ElementRef, private renderer: Renderer, private _viewContainerRef: ViewContainerRef) {
```
pkozlowski-opensource: Why do we have `_viewContainerRef`? It doesn't seem to be used?

sendilkumarn:
@pkozlowski-opensource Check this one. changed the logic slightly.

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      if (isActiveTarget) {
+        spy.setActive(this._offsetElementList[i]);
+      }
+    }
+  }
+
+  ngOnInit(): void {
+    spies.forEach(spyElements => {
+      let offsetElement = document.getElementById(spyElements.selector.toString());
```
pkozlowski-opensource: This is a no-no, we can't go down such a low-level DOM manipulation path...

src/scrollspy/scrollspy.ts
```diff
@@ -0,0 +1,110 @@
...
+      this._offsetElementList.push(offsetElement);
+    });
+    this.renderer.listen(this.elementRef.nativeElement, 'scroll', () => this.checkSpies());
```
pkozlowski-opensource: As with the previous @wesleycho PR I've got serious concerns with the fact that we are going to do massive calculations (both size calculation causing layout / reflow and dirty checking). We need to see how to optimize this...

src/scrollspy/scrollspy.ts
sendilkumarn: what can we use? 

I think queryselector is also no-no

pkozlowski-opensource:
@sendilkumarn thnx for the PR (this one and all the other ones you are sending!). Unfortunately I've got growing concerns with this directive and the way we approach it. It is not the your implementation is "wrong" - I just think that we can offer better API (without global IDs) and better performance.

I feel really bad about writing the above as at the moment I don't have any concrete counter-proposal and my priorities are on typeahead / modal. In practice it means that I complain about sth without having any better solution... It is bad, I know... 

I think that the biggest mistake that we are making here is that we didn't start with a design doc... 

Here is my proposal: let's put this PR on a back-burner till after we are done with typeahead, modal and datepicker. Unless you want to start a design doc and thing about alternative impls.

Again, sorry for coming with this feedback only, I should have asked for a design doc from the very beginning...

src/scrollspy/scrollspy.ts
pkozlowski-opensource: My current thinking is that we should have a parent directive where child directives would be injected / queried.

src/scrollspy/scrollspy.ts
wesleycho: Ideally we need to throttle this event with a 50ms limit - I wish Angular made this an observable-based API, but this usage will need to be wrapped in an observable.

pkozlowski-opensource: Yeh... We could also do most of the work outside of ng zone to avoid unnecessary dirty checking....

sendilkumarn:
@pkozlowski-opensource no problem with that 
Thanks

sendilkumarn:
Closing this. If needed we can reopen later. Thanks

Issue #205, opened 2016-06-13
Accordion: keep only active panels in the DOM


pkozlowski-opensource:
As of today we switch panel's visibility through CSS classes which means that non-active (hidden) content is still kept in the DOM / change-detected etc.

As a result of discussions with the material team we should change the strategy to keep around only active panels.

Implementing what is described here will likely require to the API. Also it would be best to consider this change alongside #204

Issue #204, opened 2016-06-13
Accordion: support HTML / directives in titles


pkozlowski-opensource:
As of today HTML titles can't have HTML / ng2 directives. The actual change to be done in the `NgbPanel` class. 

Before making changes in the code we need to see what is the impact on the API of both panels and accordions.

wesleycho:
Any reason we don't want to slot content via <ng-content slot="..."></ng-content>?

PR #403, opened 2016-07-08
refactor(accordion): align implementation with tabset

### LINKED ISSUE #228: Accordion: maintaining active panel index
In the spirit of [a solution discussed for tabs](https://github.com/ng-bootstrap/core/issues/215#issuecomment-227084987) we should decide on the best course of action for accordion. 

Normally we should use the same mechanism as for tabs but the problem is that we might have multiple active panels (in accordion that supports multiple open panels at once).

maxokorokov:
This aligns the `accordion` implementation with the `tabset` implementation:

- Allows for HTML in titles
- Introduces `activeIds: string[]` input
- adds (change) event
- physically removes content panels from DOM when not active

The markup now looks the same way as for tabset:
```html
<ngb-accordion [closeOthers]="closeOthers" [activeIds]="ids" (change)="onChange($event)">
  <ngb-panel [open]="true">
    <template ngbPanelTitle><b>Title</b></template>
    <template ngbPanelContent>Content</template>
  </ngb-panel>
</ngb-accordion>
```

Closes #228, #205 #204 

P.S. I'm not happy with the API, as there is a redundancy between `[open]` and `[activeIds]` inputs. Probably it would be better to get rid of the `[open]`


maxokorokov:
This needs further discussion as:

- there will be no more `.panel-...` classes and panels and BS4, so NgbPanel doesn't make sense for me as a standalone component
- NgbCollapse is not used anymore (for now)

Will open an issue with examples about this soon

src/accordion/accordion.ts
```diff
@@ -59,53 +62,130 @@ export class NgbPanel {
...
 })
-export class NgbAccordion implements AfterContentChecked {
+export class NgbAccordion implements OnChanges,
+    AfterContentInit {
   @ContentChildren(NgbPanel) _panels: QueryList<NgbPanel>;
 
   /**
+   * An array of panel of identifiers that should be opened
+   */
+  @Input() activeIds: string[] = [];
```
pkozlowski-opensource: Could we have it as `string | string[]` so for `closeOthers` case people don't have to specify an array?

maxokorokov:
Ok, I'm going to remove the `[open]` input, because the `toggle(id)` and `[activeIds]` is enough and it will be less code

src/accordion/accordion.ts
maxokorokov: amdended, so now you can give either `'id'` or `'id1, id2'` or `['id1', 'id2']`

maxokorokov:
Added
- `card-` based markup as in #404 
- removed the `open` input completely

Now that I'm quite pleased with the result, please review ;)

pkozlowski-opensource:
LGTM. Great stuff @maxokorokov ! 

@Foxandxss you wanted to have a look at it, feel free to review / merge.

maxokorokov:
Amended fix for `disabled` handling issues and demo documentation for `NgbPanelTitle` and `NgbPanelContent`

Foxandxss:
Why are you deleting the panel tests? Perhaps I am not following this correctly.

I am not so sure about all of this, accordion moved from a simple component to a monster component.

Do what you want @pkozlowski-opensource, but be sure to add the BREAKING CHANGE.

maxokorokov:
@Foxandxss I've only aligned it with `tabset` (if you could take a look at it's implementation) by adding HTML in the title, change events and moving some logic from NgbPanel to NgbAccordion. Plus changed markup from panels → cards.

I like that `tabset` and `accordion` have exactly the same API now, as they're almost the same thing.

As for tests, I agree that more tests are required for both tabset and accordion. Will double check for `NgbPanel` tests though

P.S. By the way `NgbPanel`is exactly the same as `NgbTab` and the same is true for `NgbTabTitle` vs `NgbPanelTitle` and `NgbTabContent` vs `NgbPanelContent`. Not sure if they should be merged in one or not... probably not

Foxandxss:
I am not saying that you are doing wrong with this PR, I just feel like this implementations are too complex.

maxokorokov:
Sure, agree with increased complexity; it's mostly due to `open` replacement with `activeIds: string | string[]` and the fact that you need to store a map of states vs only one active tab id for the `tabset`

pkozlowski-opensource:
>  just feel like this implementations are too complex.

@Foxandxss it is "more complex" since it handles cases that the previous implementation did not (most importantly - HTML in title / header). But more importantly it might be just a matter of perception. Honestly I don't find the current code complex at all and I would be hard pressed to come up with a simpler implementation.

@Foxandxss could you at least have a look at the tests and check if you agree with the cases that we are trying to cover and the API we are exposing to the users? This is the most important part. The implementation part is easy - we can always change it.

My suggestion: have another look at the tests / API. And feel free to send an alternative PR if you can think of a simpler implementation. I'm all for the simplest possible thing, provided that it has all the required functionality.

Foxandxss:
> @Foxandxss could you at least have a look at the tests and check if you agree with the cases that we are trying to cover and the API we are exposing to the users?

I did, I wouldn't write a comment if I didn't.

As I said, feel free to merge but remember to put the breaking change on the commit for the changelog generation.

pkozlowski-opensource:
> I did, I wouldn't write a comment if I didn't.

It wasn't clear from your comment what you consider as "monster" / "complex". Were you referring to the API? Or implementation? 

Are you fine with the API and it is only the implementation that bothers you? Or are you seeing the API as needing another approach?

Foxandxss:
A bit of both. ActiveIds seems more complex than simply [open] and we maintain an stack of ids, an stack of states and a stack of panels. I guess that is all needed, but I find it a bit complex just for having a collapsible menu.

But by all means, go ahead.

pkozlowski-opensource:
> A bit of both. ActiveIds seems more complex than simply [open] and we maintain an stack of ids, an stack of states and a stack of panels. I guess that is all needed, but I find it a bit complex just for having a collapsible menu.

How about this: give it a shot and propose an alternative (PR, design doc, more detailed review of this PR). If I don't hear from you by Mon I will merge this one.

How does it sound?

Foxandxss:
I can't come with a solution for Monday.

Merge this (with the breaking change message with examples of what we had before, what we have now).

src/accordion/accordion.spec.ts
```diff
@@ -4,52 +4,46 @@ import {TestComponentBuilder} from '@angular/compiler/testing';
...
 
+  const result = [];
+
   for (let i = 0; i < panels.length; i++) {
-    if (openPanelsDef[i]) {
-      expect(panels[i]).toHaveCssClass('panel-open');
-    } else {
-      expect(panels[i]).not.toHaveCssClass('panel-open');
-    }
+    result.push(panels[i].classList.contains('active'));
```
wesleycho: Couldn't you just do
```js
const result = panels.map(panel => panels[i].classList.contains('active'));
```

wesleycho:
At first I was wary, but looking at this more carefully, I like this implementation - HTML titles is a problem with the prior implementation, but this buys us dynamic title templates.

src/accordion/accordion.spec.ts
maxokorokov: Sure, it was just a quick fix of the previous code

maxokorokov:
Lastly amended test fix and checked `NgbPanel` tests that I removed: they make no sense as there is no more markup/click logic generated at the `NgbPanel` level. Restored a couple of tests on the disabled panel cases from there though...

PR #405, opened 2016-07-08
refactor(tabset): removes not active tabs content from DOM

maxokorokov:
Tab visibility is handled via `ngIf` instead of css `active` class manipulation


pkozlowski-opensource:
LGTM

PR #410, opened 2016-07-08
chore(demo): make the demo code for alerts more straightforward

jnizet:
The code used to close an alert passes the ID of the alert to close, which is then
searched for in the array of alerts.

This commit simplifies the code by just passing the alert itself and using `indexOf` to find the index of the alert to close.


pkozlowski-opensource:
LGTM, thnx!

PR #411, opened 2016-07-08
feat(rating): set cursor on stars of rating component

jnizet:
The cursor used on the stars of the rating component is a text selection cursor, which doesn't make it clear that they're clickable. The cursor stays the same if the rating component is readonly.
This commit sets a 'pointer' cursor on the stars, and a 'not-allowed' cursor when it's read-only.

### DIFF
--- a/src/rating/rating.spec.ts
+++ b/src/rating/rating.spec.ts
@@ -83,6 +83,27 @@ describe('ngb-rating', () => {
        });
      })));
 
+  it('should set pointer cursor on stars when not readonly', async(inject([TestComponentBuilder], (tcb) => {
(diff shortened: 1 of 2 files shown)


Foxandxss:
I saw this too, I haven't thought about the solution but I don't dislike that one.

jnizet:
Sorry for the failing build. I will fix the formatting and amend the commit.

pkozlowski-opensource:
I'm good with this approach. We might be re-visiting how stars are rendered in the future so things might change but IMO this is a good starting point.

+1 from me

pkozlowski-opensource:
And LGTM

PR #413, opened 2016-07-08
demo: set pointer cursor on links

jnizet:
Many components (tabs, pager, etc.) use links without any href attributes. Browsers (Chrome at least) don't set a pointer cursor by default on such links, which make them look weird and non-clikable.
This commit just adds a CSS rule in the demo to style links with a pointer cursor.

### DIFF
--- a/demo/src/style/app.scss
+++ b/demo/src/style/app.scss
@@ -181,3 +181,6 @@ section h3 {
   margin-bottom: 10px;
 }
 
+a {
+  cursor: pointer;
+}


Foxandxss:
We will probably look for a different solution, other than that.

jnizet:
OK, no problem. This was meant as a quick fix for the demo anyway. But I agree it looks like a dirty hack.

pkozlowski-opensource:
Thnx for helping out @jnizet ! Now, while this is a quick fix I agree with @Foxandxss that we need to solve it on a component level. This is actually being discussed in #142.

Thnx for the suggestion but we are going to pass on this PR. I will comment more in #142

PR #416, opened 2016-07-09
chore(tooltip): refactor tooltip and popover to share common code

jnizet:
I haven't touched the unit tests at all. That at least proves that the refactoring doesn't break existing tests, but work still needs to be done in order to cut the repetitions. I have no clear idea how to modify them to reduce repetitions without making them hard to understand, though.

Notes regarding the implementation:
 - I had to disabled tslint on one line of code because it contradicts with clang-format, making it impossible to build without one of them complaining
 - My initial idea was to declare the common `@Input` fields in the abstract base class (and use an abstract class instead of an interface for ITooltipWindow), but that doesn't work. Decorators seem to be ignored if they are in the base class. Hence the need to declare them in the subclasses and provide `getXxxInput()` methods to make them available for the base class methods.
 - if you find a better name for the common base class and interface, please suggest
 - I wasn't sure where to put the common classes, so I created a separate file in tooltip, but if you think that should go in util or somewhere else, please say.

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -3,20 +3,15 @@ import {
   Directive,
   Input,
   ChangeDetectionStrategy,
-  OnInit,
-  AfterViewChecked,
   Injector,
   Renderer,
-  ComponentRef,
   ElementRef,
   TemplateRef,
   ViewContainerRef,
-  ComponentFactoryResolver,
-  ComponentFactory
+  ComponentFactoryResolver
 } from '@angular/core';
 
-import {parseTriggers, Trigger} from '../util/triggers';
-import {Positioning} from '../util/positioning';
+import {ITooltipWindow, NgbAbstractTooltip} from '../tooltip/abstract-tooltip';
 
 @Component({
   selector: 'ngb-popover-window',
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
@jnizet thnx for this PR. Generally speaking it is a "good thing". But I was kind of putting off this refactoring since I was going back and forth between inheritance, composition and pushing things upstream to the ng2.

Your PR is good and definitively helps and as such I'm willing to merge it. But at the same time I would like to explore other avenues so let me sleep on it for a couple of days.

BTW, feel free to comment on / discuss issues you are planing to work on so we can exchange current thinking. 

Most importantly: thnx for the interest and all the contributions so far!

jnizet:
Thanks, and no problem. A PR is always food for thought, and I don't mind if it's being rejected. 

I also asked myself if inheritance was the best solution, or if composition was a better idea. I started with inheritance because it seemed correct, and not just as a way to inherit code: both components *are* tooltip components. If someone has a better design idea, having an example of inheritance-based code here will at least allow comparing the pros and cons of the two approaches.

Have a good sleep :-)

pkozlowski-opensource:
Hi @jnizet ! 

So, I did some more thinking on this one and I'm going to pass on it for now, sorry. The main reason is that I want to explore all the composition options before we go into inheritance. Also there are parts that I would like to (potentially) push into core.

In the future feel free to discuss in issues before coding to avoid any unmerged PRs.  And thnx for all contributions - keep them coming!

PR #417, opened 2016-07-09
demo: change the demo theme to the default one

### LINKED ISSUE #414: Code in black background in demo is hard to read
I have, like almost 10% of the male population, a (light) color vision deficiency.
This makes it quite hard for me to read the code in the demos, mainly because their background is black (the purple or pink or whatever color used for html elements, in particular, is hard for me to read).

I know JavaScript hipsters like their dark background :-) But in my experience, I and other people I know who have color vision deficiencies have much less problem reading code when it's more traditionally written in black or dark colors, on a white background.

Would you consider changing the color theme of the code, or providing a simple way to switch to a more traditional color theme?

jnizet:
This changes the demo theme from the dark one to the default one, and sets the background to the same color as the bootstrap card header, to make it look more integrated with bootstrap.

Also moved the vendor CSS to the vendor bundle. Not really important for our demo but I like to keep the things clean.

Fixes #414

### DIFF
--- a/demo/src/app/app.component.ts
+++ b/demo/src/app/app.component.ts
@@ -3,8 +3,6 @@ import {ROUTER_DIRECTIVES} from '@angular/router';
 
 import {SideNavComponent} from './shared';
 
-import 'prismjs/themes/prism-okaidia.css';
-import 'bootstrap/dist/css/bootstrap.css';
 import '../style/app.scss';
 
 @Component({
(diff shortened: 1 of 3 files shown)


Foxandxss:
I am not going to say that I love it, but is better from what we have.

pkozlowski-opensource:
I think I like it :-)

PR #423, opened 2016-07-11
chore: fix file names in the API extraction script

### LINKED ISSUE #419: Api-doc generating wrong fileName
Some of the API generated with our tool gives the full path instead of relative path:

```
{"NgbPanel":{"fileName":"src/accordion/accordion.ts"
```

That is good but:

```
"NgbPopover":{"fileName":"/Users/jesus/Programming/Angular2/bootstrap/core/src/popover/popover.ts"
```

This one is not.

If you go to the demo, you can see how tabset points to a path that doesn't exist.

I have debugged it a bit and seems that comes from:

```
this.program = ts.createProgram(fileNames, {});
// and then
var sourceFile = this.program.getSourceFile(fileName);
```

And that `ts` is the `typescript` library.

pkozlowski-opensource:
Closes #419

### DIFF
--- a/misc/api-doc.js
+++ b/misc/api-doc.js
@@ -24,18 +24,17 @@ class APIDocVisitor {
 
     return sourceFile.statements.reduce((directivesSoFar, statement) => {
       if (statement.kind === ts.SyntaxKind.ClassDeclaration) {
-        return directivesSoFar.concat(this.visitClassDeclaration(statement));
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@Foxandxss I was trying to see why TS switches from relative to absolute paths but it seems completely arbitrary... The only thing I can think of at this point is to _not_ get this info from TS. PTAL.

PR #425, opened 2016-07-11
chore: start 1.0.0-alpha.1 iteration

pkozlowski-opensource:


wesleycho:
LGTM - I don't think this generally needs a PR, just consider it part of the release process.

PR #429, opened 2016-07-12
feat(typeahead): add match highlighting component

pkozlowski-opensource:
This is one of the elements needed for typeahead.

### DIFF
--- a/src/typeahead/highlight.ts
+++ b/src/typeahead/highlight.ts
@@ -1,0 +1,39 @@
+import {Component, Input, OnChanges, ChangeDetectionStrategy} from '@angular/core';
+import {regExpEscape, toDefinedString} from '../util/util';
+
+@Component({
+  selector: 'ngb-highlight',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<template ngFor [ngForOf]="_parts" let-part let-isOdd="odd">` +
+      `<span *ngIf="isOdd" class="{{highlightClass}}">{{part}}</span><template [ngIf]="!isOdd">{{part}}</template>` +
+      `</template>`,  // template needs to be formatted in a certain way so we don't add empty text nodes
+  styles: [`
+    .ngb-highlight {
+      font-weight: bold;
+    }
+  `]
+})
+export class NgbHighlight implements OnChanges {
+  private _parts: string[];
+
+  @Input() highlightClass = 'ngb-highlight';
+  @Input() result: string;
+  @Input() term: string;
+
+  ngOnChanges() {
+    const resultStr = toDefinedString(this.result);
+    const resultLC = resultStr.toLowerCase();
+    const termLC = toDefinedString(this.term).toLowerCase();
+    let currentIdx = 0;
+
+    if (termLC.length > 0) {
+      this._parts = resultLC.split(new RegExp(`(${regExpEscape(termLC)})`)).map((part) => {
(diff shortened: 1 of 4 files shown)


src/typeahead/highlight.ts
```diff
@@ -0,0 +1,35 @@
...
+      `<span *ngIf="isOdd" class="{{highlightClass}}">{{part}}</span><template [ngIf]="!isOdd">{{part}}</template>` +
+      `</template>`,  // template needs to be formatted in a certain way so we don't add empty text nodes
+  styles: [`
+    .ngb-highlight {
+      font-weight: bold;
+    }
+  `]
+})
+export class NgbHighlight implements OnChanges {
+  private parts: string[];
```
maxokorokov: Should be `_parts`

src/typeahead/highlight.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  @Input() highlightClass = 'ngb-highlight';
+  @Input() result: string;
```
maxokorokov: `result` seems strange to me here as a name if you look at it as a standalone component:

```html
<ngb-highlight result="something to highlight" term="term">
```

maybe rename to something like `input`?

maxokorokov:
Looks good for me, I played with it a bit already.
One small additional point: it will generate a bunch of empty spans between each letter if `term=""`. And no documentation yet :)

pkozlowski-opensource:
> it will generate a bunch of empty spans between each letter if term="". 

Thnx, will fix & add test for it.

> And no documentation yet :)

Do we want to have docs for it? I mean, do we consider this to be part of the public API? Or rather an internal part of the typeahead's implementation?

src/typeahead/highlight.ts
```diff
@@ -0,0 +1,43 @@
...
+import {regExpEscape} from '../util/util';
+
+function toDefinedString(val: any) {
+  return (val !== undefined && val !== null) ? `${val}` : '';
+}
+
+@Component({
+  selector: 'ngb-highlight',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  template: `<template ngFor [ngForOf]="_parts" let-part let-isOdd="odd">` +
```
Foxandxss: This "certain way" you meantion is having to contact them instead of putting them in a multiline format with the backsticks?

src/typeahead/highlight.ts
```diff
@@ -0,0 +1,43 @@
+import {Component, Input, OnChanges, ChangeDetectionStrategy} from '@angular/core';
+import {regExpEscape} from '../util/util';
+
+function toDefinedString(val: any) {
```
Foxandxss: can't we move this to util?

Foxandxss:
Sounds good to me, but on the tests, you mix `result` with `[result]` (and the other attrs as well) no matter if they receive something from the "controller" or if they are a static value.

src/typeahead/highlight.ts
pkozlowski-opensource: Yes, we want to avoid blank text nodes

src/typeahead/highlight.ts
pkozlowski-opensource: Certainly!

pkozlowski-opensource:
Thnx @Foxandxss for the review. I will take care of the comments and merge this one.

PR #431, opened 2016-07-12
feat(accordion): added Card Bootstrap Color Options

### LINKED ISSUE #420: Accordian/Card Bootstrap Color Options
I think it would be great if there was an option on the accordion to have a certain type. For example:

```html
<ngb-accordion [closeOthers]="true" type="success">
</ngb-accordion>
```
-- OR --
```html
<ngb-accordion [closeOthers]="true">
        <ngb-panel type="warning" title="Collapsible Group Item #1" [open]="true"></ngb-panel>
        <ngb-panel type="success" title="Collapsible Group Item #2"></ngb-panel>
        <ngb-panel type="info" title="Collapsible Group Item #3"></ngb-panel>
        <ngb-panel type="primary" title="Collapsible Group Item #4"></ngb-panel>
        <ngb-panel type="default" title="Collapsible Group Item #5"></ngb-panel>
</ngb-accordion>
```

This way we can easily utilize the different colors that bootstrap offers (with cards and panels)

Here is a plunker, started from another discussion (Issue #404) and adapted to this request (except it only shows the ngb-accordion example):
http://plnkr.co/edit/2IDH6VDXQsjDhk8l0lHj?p=preview

sendilkumarn:

fixes #420

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -31,8 +31,9 @@ describe('ngb-accordion', () => {
 
   beforeEach(() => {
     html = `
-      <ngb-accordion #acc="ngbAccordion" [closeOthers]="closeOthers" [activeIds]="activeIds" (change)="changeCallback($event)">
-        <ngb-panel *ngFor="let panel of panels" [id]="panel.id" [disabled]="panel.disabled">
+      <ngb-accordion #acc="ngbAccordion" [closeOthers]="closeOthers" [activeIds]="activeIds" 
+        (change)="changeCallback($event)" [type]="classType">
+        <ngb-panel *ngFor="let panel of panels" [id]="panel.id" [disabled]="panel.disabled" [type]="panel.type">
           <template ngbPanelTitle>{{panel.title}}</template>       
           <template ngbPanelContent>{{panel.content}}</template>        
         </ngb-panel>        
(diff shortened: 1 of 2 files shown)


src/accordion/accordion.spec.ts
```diff
@@ -271,15 +272,34 @@ describe('ngb-accordion', () => {
          expectOpenPanels(fixture.nativeElement, [false, false, false]);
        });
      })));
+
+  it('should override the type in accordion with type in panel', async(inject([TestComponentBuilder], (tcb) => {
```
pkozlowski-opensource: This is _very_ light on tests.... We need some more:
* what happens if no type is specified? (No binding to `[type]` on accordion / panel)?
* what happens if type is specified only on accordion level and no panel has type?

I know that those might sound like trivial cases but we _need_ to have them covered.

BTW: you don't have to re-use the same html in each and every test, feel free to have test-specific markup.

sendilkumarn: > BTW: you don't have to re-use the same html in each and every test, feel free to have test-specific markup.

you caught me here. I will update now.

src/accordion/accordion.ts
```diff
@@ -102,6 +110,13 @@ export class NgbAccordion implements AfterContentChecked {
   @Input('closeOthers') closeOtherPanels: boolean = false;
 
   /**
+   *  Accordion-Panel type (CSS class). Bootstrap 4 recognizes the following types: "success", "info", "warning" and
+   * "danger".
+   */
+  @Input() type = 'default';
```
pkozlowski-opensource: If I'm not mistaken `card-default` doesn't exist in BS4 (at least not in alpha.2). As such we shouldn't be adding it. Instead we should add any type-related class when type is not specified.

src/accordion/accordion.ts
```diff
@@ -130,6 +145,9 @@ export class NgbAccordion implements AfterContentChecked {
       this._closeOthers(this.activeIds[0]);
       this._updateActiveIds();
     }
+
+    // update classes for panels
+    this._setClassType();
```
pkozlowski-opensource: Why do we do this in `ngAfterContentChecked`? This doesn't sound right. We should simply calculate proper class in a template.

src/accordion/accordion.ts
```diff
@@ -175,6 +193,16 @@ export class NgbAccordion implements AfterContentChecked {
       this._panelRefs.set(panel.id, panel);
     });
   }
+
+  private _setClassType() {
```
pkozlowski-opensource: IMO we could do this simpler.

sendilkumarn: ` [class]="'card-header card-' + (panel.type ? panel.type: type)"`

i thought of this.

src/accordion/accordion.ts
```diff
@@ -88,7 +95,8 @@ export interface NgbPanelChangeEvent {
   </div>
 `
 })
-export class NgbAccordion implements AfterContentChecked {
+export class NgbAccordion implements AfterContentChecked,
+    OnInit {
```
pkozlowski-opensource: Why do we have `OnInit` imported and used in `implements` here? I don't see it being used (I would expect TS to complain in this case...)

src/accordion/accordion.spec.ts
```diff
@@ -271,15 +272,78 @@ describe('ngb-accordion', () => {
...
+      </ngb-accordion>
+      <button *ngFor="let panel of panels" (click)="acc.toggle(panel.id)">Toggle the panel {{ panel.id }}</button>
+    `;
+       tcb.overrideTemplate(TestComponent, testHtml).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         let el = fixture.nativeElement.querySelectorAll('.card-header');
+         expect(el[0]).toHaveCssClass('card-primary');
+         expect(el[1]).toHaveCssClass('card-primary');
+         expect(el[2]).toHaveCssClass('card-primary');
```
pkozlowski-opensource: This is not correct! We shouldn't get any class other than `card-header` (and eventually `active`) by default. Run the demo and see how it looks....

src/accordion/accordion.ts
```diff
@@ -102,6 +108,13 @@ export class NgbAccordion implements AfterContentChecked {
   @Input('closeOthers') closeOtherPanels: boolean = false;
 
   /**
+   *  Accordion-Panel type (CSS class). Bootstrap 4 recognizes the following types: "success", "info", "warning" and
+   * "danger".
+   */
+  @Input() type = 'primary';
```
pkozlowski-opensource: This is not correct. By default we shouldn't have any type.

src/accordion/accordion.ts
```diff
@@ -76,7 +82,7 @@ export interface NgbPanelChangeEvent {
   template: `
   <div class="card">
     <template ngFor let-panel [ngForOf]="_panels">
-      <div class="card-header" [class.active]="_isOpen(panel.id)">
+      <div [class]="'card-header card-' + (panel.type ? panel.type: type)" [class.active]="_isOpen(panel.id)">
```
pkozlowski-opensource: You need to check if type is set and only then add `card-type` class. We are doing sth similar in progress bar: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/progressbar/progressbar.ts#L11

PR #439, opened 2016-07-14
refactor(pagination): don't use types where those can be inferred

pkozlowski-opensource:


sendilkumarn:
was doing this for tabset 👍

PR #442, opened 2016-07-14
Demo site improvements

tsm91:


Foxandxss:
I like the changes but not so sure about that "make that footer stick to the bottom of the make"

tsm91:
> I like the changes but not so sure about that "make that footer stick to the bottom of the make"

Yeah, i'm not sure if its needed. But on the [home page](https://ng-bootstrap.github.io/#/home) the content is not enough to push the footer to the bottom and it looks nasty.

The implementation is based on one of bootstrap's [official example](http://v4-alpha.getbootstrap.com/examples/sticky-footer-navbar/).

I clicked through the demo page and haven't found any visual differences / bugs.

RobJacobs:
Flexbox should do the trick: https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/

tsm91:
@RobJacobs Can you check the last commit? I hope i got it right, i'm not familiar with flexbox.

I ended up putting an `ngbd-app` css attribute selector in `app.scss` so autoprefixer runs through that piece of code. Maybe this isnt the best solution.

pkozlowski-opensource:
@tsm91 I've picked f073f33 and e05a2f9, squashed them into one and landed as 8995004d9c213b1b2d9248dea837c4ba09bb8d0a. . Thnx!

"Sticky" footer is more controversial as the last change you've proposed didn't work well on individual component demos. Let's tackle this in a separate PR (it is always a good idea to open small, focused PRs - those are much easier to review and merge).

tsm91:
alright, cool

sorry for the sticky tricky footer mess :)

PR #444, opened 2016-07-17
WIP : feat(timepicker):   Base elements

### LINKED ISSUE #24: Timepicker
Implement timepicker from UI Bootstrap

Should note, this likely will need to be coordinated with the datepicker with the implementation of a default parsing service that can be optionally included that uses the Date constructor.

This should be flexible enough to accept other parsing mechanisms, i.e. moment.js, as long as a common interface is implemented.

sendilkumarn:
Initial Commit 

let me know your comments Thanks 👍 

fixes #24


src/timepicker/timepicker.ts
```diff
@@ -0,0 +1,149 @@
...
+export class NgbTime {
+  constructor(public hour: number, public minute: number, public second: number) {}
+}
+
+@Component({
+  selector: 'ngb-timepicker',
+  exportAs: 'ngbTimepicker',
+  providers: [NGB_TIMEPICKER_VALUE_ACCESSOR],
+  template: `
+    <input [ngModel]="model?.hour" (ngModelChange)="updateHour($event)" [maxLength]=2> <span> : </span>
```
icfantv: Should these have BS form field styling?

src/timepicker/timepicker.ts
```diff
@@ -0,0 +1,149 @@
...
+      this.model.hour = 0;
+    }
+    this.onTouched();
+    this.onChange(this.model);
+  }
+
+  decrementHour() {
+    this.model.hour--;
+    if (this.model.hour < 0) {
+      this.model.hour = 23;
```
icfantv: What did we decide on supporting AM/PM?

src/timepicker/timepicker.ts
sendilkumarn: I will add them

icfantv: Input type = number?

src/timepicker/timepicker.ts
sendilkumarn: This is base skeleton. I will add them

src/timepicker/timepicker.ts
sendilkumarn: yet to decide on that I guess?

src/timepicker/timepicker.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.model.second--;
+    if (this.model.second < 0) {
+      this.model.second = 59;
+      this.decrementMinute();
+    }
+    this.propagateModelChange();
+  }
+
+  updateHour(newVal) {
+    let hour = toInteger(newVal);
```
icfantv: If we do input type of number is this utility even necessary?  I think there's also the trick of simply pretending the value with a plus sign...

sendilkumarn: >  input type of number

that has its own issues. IMO its better to have a normal input type and convert as we need.

> trick of simply pretending the value with a plus sign

we do have a check here, so we can validate with that.

pkozlowski-opensource:
@sendilkumarn thnx for starting this off. I'm going to look at the code, add more tests and see how I can progress. Will probably open a PR with code from this PR and my changes.

sendilkumarn:
Okay then @pkozlowski-opensource let me know if there is anything

pkozlowski-opensource:
Picking up commits from this PR as part of #450 - let's continue the discussion in there.

PR #449, opened 2016-07-19
Fix memory leak from un-removed event listeners

tsm91:
[Renderer](https://angular.io/docs/ts/latest/api/core/index/Renderer-class.html).listen / .listenGlobal returns a function that you need to call in order to remove the attached event listener.

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -5,6 +5,7 @@ import {
   ChangeDetectionStrategy,
   OnInit,
   AfterViewChecked,
+  OnDestroy,
   Injector,
   Renderer,
   ComponentRef,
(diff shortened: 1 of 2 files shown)


icfantv:
@tsm91 please squash these commits.  In the future, if you have to make additional changes to a PR, you can just do a `git commit --amend` to reuse the existing changelist and then do a `git push origin <branch_name> --force` to push the new changes.  This will save you from having to mess with squashing in the future.  Very handy...

@pkozlowski-opensource or @Foxandxss will look at this and decide if this is the correct behavior.  Thanks for contributing.

pkozlowski-opensource:
@tsm91 thnx, this looks like a correct. change. Did you actually see the memory leak on your side or it is just a preventive measure?

I will confirm with Misko tomorrow that the cleanup is needed.

tsm91:
@icfantv Alright, next time ill do that.

> Did you actually see the memory leak on your side or it is just a preventive measure?

@pkozlowski-opensource  I was checking the source for educational purposes and ran into this, so the latter.

Have you guys ever heard of [drool](https://github.com/samccone/drool)? I never write tests but you do, so maybe this can be useful. Its an automated memory leak detection and analysis tool.

wesleycho:
This looks correct - we need to be careful with stuff that returns subscribed observables in ng2 and make sure to gc them.

icfantv:
All I know is that async pipes are cleaned up automatically.

pkozlowski-opensource:
I'm going to merge this one as I don't see any harm here and there are potential benefits.

PR #450, opened 2016-07-20
timepicker

pkozlowski-opensource:
This is the first version of a timepicker. What remains to be done in this PR is:
* [x] - look & feel
* [x] - test improvements

After this initial version gets merged we can open issues for:
* [ ] - corner cases (invalid model etc.)
* [ ] - more features (disabled, time-range, ...)
* [ ] - timepicker as drop-down
* [ ] - duration mode

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -1,0 +1,43 @@
+import {inject, async, addProviders} from '@angular/core/testing';
+
+import {TestComponentBuilder} from '@angular/compiler/testing';
+
+import {Component} from '@angular/core';
+
+import {NgbTimepicker} from './timepicker';
+
+function getInput(nativeEl: HTMLElement) {
+  return nativeEl.querySelectorAll('input');
+}
+
+describe('ngb-timepicker', () => {
+  it('should render hour and minute inputs by default', async(inject([TestComponentBuilder], (tcb) => {
+       const html = `<ngb-timepicker> </ngb-timepicker>`;
+
+       tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+         fixture.detectChanges();
+
+         const button = getInput(fixture.nativeElement);
+         expect(button.length).toBe(2);
+       });
+     })));
+});
+
+describe('ngb-timepicker', () => {
+  it('should have seconds input when it is enabled', async(inject([TestComponentBuilder], (tcb) => {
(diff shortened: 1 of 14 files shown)


pkozlowski-opensource:
I believe that this one is ready for the review now, PTAL.

icfantv:
I'll pull down and look at styling when I get to work.

src/timepicker/timepicker.ts
```diff
@@ -0,0 +1,187 @@
...
+ */
+@Component({
+  selector: 'ngb-timepicker',
+  styles: [`
+    .chevron::before {
+      border-style: solid;
+      border-width: 0.29em 0.29em 0 0;
+      content: '';
+      display: inline-block;
+      height: 0.70em;
```
icfantv: `height` should be `0.69em`

src/timepicker/timepicker.ts
```diff
@@ -0,0 +1,187 @@
...
+      transform: rotate(-45deg);
+      vertical-align: middle;
+      width: 0.71em;
+    }
+    
+    .chevron.bottom:before {
+      top: -.3em;
+      transform: rotate(135deg);
+    }
+  `],
```
icfantv: need to add CSS:
```
    .btn-link {
      outline: 0;
    }
```
otherwise we get the button halo affect and in IMO looks weird because it's not really a button.

icfantv:
Just the two CSS changes and LGTM.

pkozlowski-opensource:
@icfantv fixed CSS per your suggestion, thnx. @Foxandxss you wanted to have another look at it, right?

src/timepicker/timepicker.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+
+function getButtons(nativeEl: HTMLElement) {
+  return nativeEl.querySelectorAll('button.btn-link');
+}
+
+function getMeridianButton(nativeEl: HTMLElement) {
+  return nativeEl.querySelector('button.btn-primary-outline');
+}
+
+function expectToDisplayTime(el: HTMLElement, time: string) {
```
Foxandxss: This is better to be done as a matcher.

Foxandxss:
Sounds good to me. Left a minor comment.

src/timepicker/timepicker.spec.ts
pkozlowski-opensource: Yes, we should move to Jasmine matchers in place of helper functions for most of our tests. I will open a separate issue for this.

Issue #338, opened 2016-07-02
Tooltip / popover: factor out common code


pkozlowski-opensource:
This should be done for both directives code and tests

pkozlowski-opensource:
We will be able to do better here as soon as https://github.com/angular/angular/pull/13032 is merged / released and we are ready to bump up minimal Angular version required (probably to 2.3). Also we would have to update our API extraction script to properly handle class inheritance.

thbt:
I have been working on this issue for some time but I am not sure I am headed the right way.

I went for a generic solution leveraging the object oriented side of typescript : Popup components extend from this class `abstract class NgbPopupAnchor<T extends NgbPopupWindow>`, this class deal with dynamically creating the popup, positioning it right, and destroying it.

While this work great for Tooltip and Popover, I am currently trying to see if it can be easily used for components such as the typeahead, the dropdown or the datepicker, and see how the class can be made even more generic. I would like to know if I'm heading in the right direction.

pkozlowski-opensource:
Now that Angular supports inheritance for classes with `@Input` / `@Output` tackling this one got easier. Still, the API extraction scripts would have to be updated - that is probably the hardest part for this PR.

PR #451, opened 2016-07-21
refactor(tooltip): factor out common triggers listening code

pkozlowski-opensource:
Part of #338


RobJacobs:
LGTM, I really like the effort you are making factoring out common code.

PR #452, opened 2016-07-21
feat(typeahead): typeahead directive

maxokorokov:
Initial version of typeahead with observables. 
Works only with strings for now, not yet with arbitrary objects.

Things yet to do:
- [x] figure out how to prevent `keyup` default behaviour
- [x] figure out how to update ngModel and input value separately
- [x] make it work with arbitrary objects as models (`string` <=> `object` conversion)

P.S. don't like the `skip` flag and the `.filter()`, but don't know how to do it otherwise


src/typeahead/typeahead-window.ts
```diff
@@ -4,8 +4,9 @@ import {NgbHighlight} from './highlight';
 @Component({
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
+  host: {'class': 'dropdown-menu', 'style': 'display: block'},
   template: `
-    <div class="dropdown-menu" aria-labelledby="dropdownMenu1">
+    <div aria-labelledby="dropdownMenu1">
```
pkozlowski-opensource: Not sure I understand this change. If we are moving things to the `host` (which is a good move IMO) let's move everything. What is a reason for this change?

BTW: here `dropdownMenu1` is hard-coded we should pass it as input.

I can do those fixes in a separate PR but I would like to know what is the reasoning behind moving this to `host`?

src/typeahead/typeahead-window.spec.ts
```diff
@@ -10,7 +10,7 @@ function normalizeText(txt: string): string {
   return txt.trim().replace(/\s+/g, ' ');
 }
 
-function expectResults(nativeEl: HTMLElement, resultsDef: string[]): void {
+export function expectResults(nativeEl: HTMLElement, resultsDef: string[]): void {
```
pkozlowski-opensource: Could we extract it to a separate file so we don't have inter-test imports?

src/index.ts
```diff
@@ -31,19 +34,23 @@ export {NgbAlert, NgbDismissibleAlert, NGB_ALERT_DIRECTIVES} from './alert/alert
 export {NgbCarousel, NgbSlide, NGB_CAROUSEL_DIRECTIVES} from './carousel/carousel';
 export {NgbCollapse, NGB_COLLAPSE_DIRECTIVES} from './collapse/collapse';
 export {NgbDropdown, NgbDropdownToggle, NGB_DROPDOWN_DIRECTIVES} from './dropdown/dropdown';
+export {NgbHighlight} from './typeahead/highlight';
```
pkozlowski-opensource: Do we want to export it as a public API? My understanding was that we might open it up in the future but not initially.

src/index.ts
```diff
@@ -31,19 +34,23 @@ export {NgbAlert, NgbDismissibleAlert, NGB_ALERT_DIRECTIVES} from './alert/alert
...
 export {NgbPager, NGB_PAGER_DIRECTIVES} from './pager/pager';
 export {NgbPagination, NGB_PAGINATION_DIRECTIVES} from './pagination/pagination';
 export {NgbProgressbar, NGB_PROGRESSBAR_DIRECTIVES} from './progressbar/progressbar';
 export {NgbRating, NGB_RATING_DIRECTIVES} from './rating/rating';
 export {NgbTabset, NgbTab, NgbTabContent, NgbTabTitle, NgbTabChangeEvent, NGB_TABSET_DIRECTIVES} from './tabset/tabset';
 export {NgbTooltip, NgbTooltipWindow, NGB_TOOLTIP_DIRECTIVES} from './tooltip/tooltip';
+export {NgbTypeahead, NGB_TYPEAHEAD_DIRECTIVES} from './typeahead/typeahead';
+export {NgbTypeaheadWindow} from './typeahead/typeahead-window';
```
pkozlowski-opensource: I don't think we want to export this one. It is a private API, isn't it?

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,211 @@
+import {inject, async} from '@angular/core/testing';
+import {TestComponentBuilder} from '@angular/compiler/testing';
+import {Component, DebugElement} from '@angular/core';
+import {NGB_TYPEAHEAD_DIRECTIVES, NgbTypeahead} from './typeahead';
+import {Observable} from 'rxjs/Rx';
+import 'rxjs/add/operator/map';
+import {NgbTypeaheadWindow} from './typeahead-window';
+import {By} from '@angular/platform-browser';
+import {expectResults} from './typeahead-window.spec';
```
pkozlowski-opensource: Yep, extract this helper function to a separate file.

src/typeahead/typeahead-window.ts
maxokorokov: Yes, better to move everything I guess. It was because of the positioning. Positioning is applied to the host, and it's because of WIP :)

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+});
+
+@Component(
+    {selector: 'test-cmp', directives: [NGB_TYPEAHEAD_DIRECTIVES], precompile: [NgbTypeaheadWindow], template: ''})
+class TestComponent {
+  model = '';
+
+  find = (text$: Observable<string>) => {
+    const values = ['one', 'one more', 'two', 'three'];
```
pkozlowski-opensource: Maybe move this is as an instance of a component? This is what would be happening in "real life", no?

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+    return text$.map(text => values.filter(v => v.startsWith(text)));
+  };
+
+  findNothing = (text$: Observable<string>) => { return text$.map(text => []); };
+
+  findObjects = (text$: Observable<string>) => {
+
+    const values = [{id: 1, value: 'one'}, {id: 10, value: 'one more'}, {id: 2, value: 'two'}, {id: 3, value: 'three'}];
```
pkozlowski-opensource: Maybe move this is as an instance of a component? This is what would be happening in "real life", no?

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+
+           getDebugInput(fixture.debugElement).triggerEventHandler('keyup.arrowUp', {});
+           fixture.detectChanges();
+           expectWindowResults(compiled, ['+one', 'one more']);
+         });
+       })));
+
+  });
+
+  xdescribe('conversion', () => {
```
pkozlowski-opensource: Remove till you've got it done so we can see what `lint` has to say.

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,120 @@
...
+    '(keyup.esc)': '_closePopup()'
+  }
+})
+export class NgbTypeahead implements OnInit {
+  private _popupService: PopupService<NgbTypeaheadWindow>;
+  private _positioning = new Positioning();
+  private _skipValueChange = false;
+  private _windowRef: ComponentRef<NgbTypeaheadWindow>;
+
+  @Input() ngbTypeahead: (text: Observable<string>) => Observable<string[]>;
```
pkozlowski-opensource: `Observable<string[]>` => `Observable<any[]>` as people might to decide push objects down the pipe.

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,120 @@
...
+          }
+          return true;
+        })
+        .let (this.ngbTypeahead)
+        .subscribe((results) => {
+
+          // no results
+          if (!results || results.length === 0) {
+            this._closePopup();
+            return;
```
pkozlowski-opensource: Could we have if / else instead of the early return? Make it more readable IMO.

src/typeahead/typeahead.spec.ts
maxokorokov: `gulp lint` is fine, no worries. Will push a couple of commits to fix the build, your comments and `keyup` after you've finished

src/util/popup.ts
```diff
@@ -18,7 +18,7 @@ export class PopupService<T> {
     this._windowFactory = componentFactoryResolver.resolveComponentFactory<T>(type);
   }
 
-  open(content: string | TemplateRef<any>): ComponentRef<T> {
+  open(content: string | TemplateRef<any> = ''): ComponentRef<T> {
```
pkozlowski-opensource: I would rather check if it is null / undefined in the `_getContentNodes` (and return `[]` in this case) to avoid unnecessary text nodes creation.

src/typeahead/typeahead.ts
pkozlowski-opensource: It will also need an input for a `formatter` (so we can pass it down to the typeahead window).

maxokorokov: Sure, and a `parser` or something as well for the `object` → `string` conversion for the `input` text

maxokorokov:
Thanks, @pkozlowski-opensource for initial review. Will go on working on it

src/index.ts
```diff
@@ -16,6 +16,8 @@ import {NgbProgressbar, NGB_PROGRESSBAR_DIRECTIVES} from './progressbar/progress
 import {NgbRating, NGB_RATING_DIRECTIVES} from './rating/rating';
 import {NgbTabset, NgbTab, NgbTabContent, NgbTabTitle, NgbTabChangeEvent, NGB_TABSET_DIRECTIVES} from './tabset/tabset';
 import {NgbTooltip, NgbTooltipWindow, NGB_TOOLTIP_DIRECTIVES} from './tooltip/tooltip';
+import {NgbTypeahead, NGB_TYPEAHEAD_DIRECTIVES} from './typeahead/typeahead';
+import {NgbTypeaheadWindow} from './typeahead/typeahead-window';
```
pkozlowski-opensource: We don't need this import

pkozlowski-opensource:
@maxokorokov sure. This is very good progress already, I guess the main remaining part is to better understand integration with the whole form ecosystem.

maxokorokov:
Rebased and added implementation for custom objects as values and proper ngModel integraton ([last commit](https://github.com/ng-bootstrap/ng-bootstrap/pull/452/commits/e05f9f91a53af87ddf3c497f0b84ad6317a84287))

maxokorokov:
Started to split it into bite-sized pieces with #460, #461 and #462 so it would be easier to review. Once those are merged, will update this pull request with the final version to review

pkozlowski-opensource:
@maxokorokov thnx for opening smaller PRs, was about to ask you for this :-) Reviewing them now!

pkozlowski-opensource:
@maxokorokov merged all the typeahead related smaller PRs. You can rebase yours now.

maxokorokov:
This looks like a final version for now.
Will open a separate pull request with demos / documentation on Monday

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,132 @@
...
+          this._windowRef.instance.resultTemplate = this.resultTemplate;
+        }
+      }
+    });
+  }
+
+  private _onKeyEnter() {
+    if (this._windowRef) {
+      const result = this._windowRef.instance.getActive();
+      this._selectResult(result);
```
pkozlowski-opensource: Does it make sense to call `event.preventDefault();` in here (so for example a form doesn't get submitted?)

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+@Directive({
+  selector: 'input[ngbTypeahead][ngModel]',
+  host: {
+    '[class.open]': '_windowRef !== null',
+    '(document:click)': '_closePopup()',
+    '(keydown.arrowDown)': '_onKeyDown($event)',
+    '(keydown.arrowUp)': '_onKeyUp($event)',
+    '(keydown.enter)': '_onKeyEnter()',
+    '(keydown.esc)': '_closePopup()'
```
pkozlowski-opensource: I think that we should have a dedicated method here (`_onKeyEsc` ?) so we could stop propagation of the event. This is needed for cases where typeahead would be placed inside a modal and not stopping propagation would close a modal as well. At least this is my theory :-)

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+       })));
+
+    it('should be opened when there are results', async(inject([TestComponentBuilder], (tcb) => {
+         const html = `<input type="text" [(ngModel)]="model" [ngbTypeahead]="find"/>`;
+
+         tcb.overrideTemplate(TestComponent, html).createAsync(TestComponent).then((fixture) => {
+           fixture.detectChanges();
+           const compiled = fixture.nativeElement;
+
+           fixture.componentInstance.model = 'one';
```
pkozlowski-opensource: Hmm, I'm puzzled by this tests... If I'm reading it correctly a typeahead window would open if a _model_ value is changed... But it shouldn't be the case! We should only be kicking-in the whole machinery in response to user actions.

Not sure if this is a bug in the implementation or just a trick in a test, but I would like to understand it better.

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+           expect(getNativeInput(compiled).value).toBe('o');
+
+           getWindowLinks(fixture.debugElement)[0].triggerEventHandler('click', {});
+           fixture.detectChanges();
+           expect(getWindow(compiled)).toBeNull();
+           expect(getNativeInput(compiled).value).toBe('one');
+         });
+       })));
+
+    it('should select the result on ENTER, close window and fill the input',
```
pkozlowski-opensource: If I recall properly we should have exactly the same behaviour for TAB press.

src/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,246 @@
...
+
+           getDebugInput(fixture.debugElement).triggerEventHandler('keydown.enter', {});
+           fixture.detectChanges();
+           expect(getWindow(compiled)).toBeNull();
+           expect(getNativeInput(compiled).value).toBe('one');
+           expect(fixture.componentInstance.model).toBe('one');
+         });
+       })));
+
+    it('should select previous/next results with up/down arrow keys', async(inject([TestComponentBuilder], (tcb) => {
```
pkozlowski-opensource: `should make active`? We are not selecting things here, are we?

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,132 @@
...
+ * NgbTypeahead directive provides a simple way of creating powerful typeaheads from any text input
+ */
+@Directive({
+  selector: 'input[ngbTypeahead][ngModel]',
+  host: {
+    '[class.open]': '_windowRef !== null',
+    '(document:click)': '_closePopup()',
+    '(keydown.arrowDown)': '_onKeyDown($event)',
+    '(keydown.arrowUp)': '_onKeyUp($event)',
+    '(keydown.enter)': '_onKeyEnter()',
```
pkozlowski-opensource: We need to add the same behaviours for tabs (minus preventing default....)

pkozlowski-opensource:
@maxokorokov I've left few more comments but it looks awesome already! Can't wait to land it - we are going to have superb typeahead directive! Splendid job here!

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private _openPopup() {
+    if (!this._windowRef) {
+      this._windowRef = this._popupService.open();
+      this._windowRef.instance.selectEvent.subscribe((result: any) => this._selectResult(result));
+
+      // position
+      const targetPosition = this._positioning.positionElements(
+          this._elementRef.nativeElement, this._windowRef.location.nativeElement, 'bottom', false);
```
pkozlowski-opensource: Shouldn't we use `bottom-left` instead of `bottom` so the popup is left-aligned instead of being centered?

src/typeahead/typeahead.spec.ts
maxokorokov: Yeah, as discussed, I removed the the `ngModel` injection completely, and updated all tests to be triggered by the `input` event. Will push in a minute

maxokorokov:
@pkozlowski-opensource thanks for the comment on the tests - it made me rethink implementation.

So now, as discussed:

- there is no more `ngModel` injection - it should work with future `formControl` too!
- no more `ngModel.control.valueChanges` usage, a dedicated emitter for value changes that is triggered only on user input
- tests now are triggered by user input too
- additional tests for TAB key and `preventDefault()` triggering
- had to merge `NgbValueAccessor` into `NgbTypeahead` because of all this though...

Now it looks much cleaner to me!
Please, take another look

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,168 @@
...
+    '[class.open]': '_windowRef !== null',
+    '(document:click)': '_closePopup()',
+    '(input)': 'onChange($event.target.value)',
+    '(keydown)': '_handleKeyDown($event)'
+  },
+  providers: [NGB_TYPEAHEAD_VALUE_ACCESSOR]
+})
+export class NgbTypeahead implements OnInit,
+    ControlValueAccessor {
+  private _onChangeNoEmit;
```
pkozlowski-opensource: What is the type of `_onChangeNoEmit` ?

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,168 @@
...
+  private _popupService: PopupService<NgbTypeaheadWindow>;
+  private _positioning = new Positioning();
+  private _valueChanges = new EventEmitter<string>();
```
pkozlowski-opensource: Shouldn't it be just an `Observable`  from RxJS?

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,168 @@
...
+  provide: NG_VALUE_ACCESSOR,
+  useExisting: forwardRef(() => NgbTypeahead),
+  multi: true
+};
+
+/**
+ * NgbTypeahead directive provides a simple way of creating powerful typeaheads from any text input
+ */
+@Directive({
+  selector: 'input[ngbTypeahead][ngModel]',
```
pkozlowski-opensource: Do we still need / want `[ngModel]` in the selector?

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,168 @@
...
+    this._popupService.close();
+    this._windowRef = null;
+  }
+
+  private _handleKeyDown(event: KeyboardEvent) {
+    if (!this._windowRef) {
+      return;
+    }
+
+    if (['ArrowDown', 'ArrowUp', 'Enter', 'Escape'].indexOf(event.key) > -1) {
```
pkozlowski-opensource: Are you sure that this is going to work fine cross-browser? This is what we are doing in the angular-ui/bootstrap: https://github.com/angular-ui/bootstrap/blob/c61d16a9330c18105d074ffd23f14db06689809b/src/typeahead/typeahead.js#L399-L431

src/typeahead/typeahead.ts
```diff
@@ -0,0 +1,168 @@
...
+        break;
+      case 'ArrowUp':
+        this._windowRef.instance.prev();
+        break;
+      case 'Enter':
+        const result = this._windowRef.instance.getActive();
+        this._selectResult(result);
+        break;
+      case 'Escape':
+      case 'Tab':
```
pkozlowski-opensource: `Tab` should work like `Enter`, not like `ESC`

pkozlowski-opensource:
@maxokorokov left few more comments, sorry... I know that it was long iteration but there are just a couple of items we need to take care of before landing this one (and this is soooo close!)

src/typeahead/typeahead.ts
maxokorokov: You're right, won't work in safari; using `event.which` for now

maxokorokov:
@pkozlowski-opensource no problem, thanks for reviewing. Addressed your comments and amended:

- `Tab` works like `Enter`
- removed `[ngModel]` and updated tests
- using `event.which` not `event.key`

PR #456, opened 2016-07-22
feat(typeahead): add support for custom templates for search results

pkozlowski-opensource:
cc @maxokorokov

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -1,27 +1,35 @@
-import {Component, Input, Output, EventEmitter} from '@angular/core';
+import {Component, Input, Output, EventEmitter, TemplateRef} from '@angular/core';
 import {NgbHighlight} from './highlight';
+
+function toStringFormatter(result) {
+  return `${result}`;
+}
+
+class ResultTplCtx {
+  constructor(public result, public term: string, public formatter: (result) => string) {}
+}
 
 @Component({
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
   template: `
+    <template #rt let-result="result" let-term="term" let-formatter="formatter">
+      <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>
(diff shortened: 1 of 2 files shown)


src/typeahead/typeahead-window.ts
```diff
@@ -47,4 +64,6 @@ export class NgbTypeaheadWindow {
   prev() { this.activeIdx = (this.activeIdx === 0 ? this.results.length - 1 : this.activeIdx - 1); }
 
   select(item) { this.selectEvent.emit(item); }
+
+  private prepareTplCtx(result: any) { return new ResultTplCtx(result, this.term, this.formatter); }
```
maxokorokov: underscore -> `_prepareTplCtx`
and also -> `_activeIdx` by the way

src/typeahead/typeahead-window.ts
```diff
@@ -37,8 +45,17 @@ export class NgbTypeaheadWindow {
    * A function used to format a given result before display. This function should return a formated string without any
    * HTML markup.
    */
-  @Input() formatter = (result) => { return `${result}`; };
+  @Input() formatter = toStringFormatter;
```
maxokorokov: just curious, why replace this?
maybe put it to utils, as we're using it elswhere?

maxokorokov:
Looks alright to me!

src/typeahead/typeahead-window.ts
pkozlowski-opensource: > just curious, why replace this?

Oh, this is due to TS / linter. TS interprets `formatter = (result) => { return `${result}`; };` as a "method" member and our linting rules are forcing methods after regular fields. Put the 2 together and you have to have `formatter` as the last field (even after outputs!).

> maybe put it to utils, as we're using it elswhere?

Makes sense. Will move it in a separate PR.

pkozlowski-opensource: @maxokorokov here: https://github.com/ng-bootstrap/ng-bootstrap/pull/464

Issue #458, opened 2016-07-22
timepicker: support disabled and readonly states


pkozlowski-opensource:
We need to find right semantics (shell we disable _all_ controls? what is the difference between disabled and readonly?).

Tests are essentials. 

Refs:
* https://github.com/angular-ui/bootstrap/issues/1113
* https://github.com/angular-ui/bootstrap/issues/2219

sendilkumarn:
we will have both. 👍 

> shall we disable all controls?

We can wrap this inside `fieldset`  only for `disable` and for `readonly` we can go for individual elements in that picker

> disable / readonly

the difference will be in the value of ng-model will be

PR #466, opened 2016-07-23
feat(timepicker): adding disabled

sendilkumarn:
#458

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -14,6 +14,10 @@ function getButtons(nativeEl: HTMLElement) {
   return nativeEl.querySelectorAll('button.btn-link');
 }
 
+function getFieldsetElement(element: HTMLElement): HTMLFieldSetElement {
+  return <HTMLFieldSetElement>element.querySelector('fieldset');
+}
+
 function getMeridianButton(nativeEl: HTMLElement) {
   return nativeEl.querySelector('button.btn-primary-outline');
 }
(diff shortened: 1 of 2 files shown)


src/timepicker/timepicker.ts
```diff
@@ -40,68 +40,70 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
...
+            </td>
+          </template>
+          <template [ngIf]="meridian">
+            <td>&nbsp;</td>
+            <td>&nbsp;</td>
+          </template>
+        </tr>
+        <tr>
+          <td>
+            <input type="text" class="form-control" maxlength="2" size="2" [readonly]="readonly"
```
pkozlowski-opensource: I don't see `readonly` field defined anywhere in the component. Did you mean to use `disabled`?

src/timepicker/timepicker.spec.ts
```diff
@@ -360,10 +360,40 @@ describe('ngb-timepicker', () => {
        })));
 
   });
+
+  describe('disable', () => {
+
+    it('should not change the value, when it is disabled', async(inject([TestComponentBuilder], (tcb) => {
```
pkozlowski-opensource: We should also check that a user is not allowed to change the value of any of the input fields directly (by entering a new value).

src/timepicker/timepicker.ts
```diff
@@ -158,22 +170,28 @@ export class NgbTimepicker implements ControlValueAccessor {
   }
 
   updateHour(newVal: string | number) {
-    this.model.updateHour(toInteger(newVal) || 0);
-    this.propagateModelChange();
+    if (!this.disabled && !this.readonly) {
```
pkozlowski-opensource: Do we need this check? If an input is disabled / readonly a user shouldn't be able to interact with a control anyway. Or is this check here for unit-tests only? If this is just for unit tests we should change them to check if a field properly gets `readonly` / `disabled` properties.

The same comment applies to all the remaining `update`: `updateHour`, `updateMinute`, `updateSecond`

src/timepicker/timepicker.ts
```diff
@@ -158,22 +170,28 @@ export class NgbTimepicker implements ControlValueAccessor {
...
+      this.model.updateSecond(toInteger(newVal) || 0);
+      this.propagateModelChange();
+    }
   }
 
   toggleMeridian() {
-    if (this.meridian) {
+    if (this.meridian && !this.disabled && !this.readonly) {
```
pkozlowski-opensource: I don't think that this check is good. A meridian button should not be active when the whole control is disabled but marking inputs as `readonly` shouldn't prevent people from clicking on a meridian button.

BTW, the same comment applies here as for `update` methods - do we really need any check here? If a button gets disabled users can't click on it anyway...

src/timepicker/timepicker.ts
```diff
@@ -133,6 +135,16 @@ export class NgbTimepicker implements ControlValueAccessor {
...
 
+  /**
+   * To disable timepicker
+   */
+  @Input() disabled = false;
+
+  /**
+   * To make timepicker readonly
+   */
+  @Input() readonly = false;
```
pkozlowski-opensource: So we need to discuss semantics of `readonly`. The thing is that is you apply `readonly` to the entire control (inputs and buttons) than it is not different from `disabled` - as such there would be no point in having both. 

Looking at the implementation of http://angular-ui.github.io/bootstrap/#/timepicker I think that the intention is to mark inputs as readonly while still allowing people to click on buttons (chevrons, meridian). As such we should probably rename this property to `readonlyInputs` to better convey what we mean here.

src/timepicker/timepicker.ts
```diff
@@ -40,68 +40,70 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
...
-        <td class="text-xs-center">
-          <button class="btn-link" (click)="changeHour(hourStep)"><span class="chevron"></span></button>
-        </td>
-        <td>&nbsp;</td>
-        <td class="text-xs-center">
-          <button class="btn-link" (click)="changeMinute(minuteStep)"><span class="chevron"></span></button>
-        </td>
-        <template [ngIf]="seconds">
-          <td>&nbsp;</td>
+    <fieldset [disabled]="disabled">
```
pkozlowski-opensource: While this is enough to disable all the controls we need to make sure that up / down chevrons are changing color when a button gets disabled. We need to have some visual clue to the users that a button is not clickable any more.

pkozlowski-opensource:
@sendilkumarn left some more comments.

src/timepicker/timepicker.spec.ts
```diff
@@ -360,10 +360,164 @@ describe('ngb-timepicker', () => {
        })));
 
   });
+
+  describe('disable', () => {
```
pkozlowski-opensource: disable => disabled

src/timepicker/timepicker.ts
sendilkumarn: `readonlyInputs` 👍

sendilkumarn:
@pkozlowski-opensource Thanks updated as per comments

PR #472, opened 2016-07-25
fix(timepicker): distinguish undefined and zero values

### LINKED ISSUE #457: timepicker: data display on empty model value
There is a lengthy discussion about the desired behavior here: https://github.com/angular-ui/bootstrap/issues/1114

Need to go over it and check if we do "the right thing".

pkozlowski-opensource:
Fixes #457

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -21,12 +21,15 @@ function getMeridianButton(nativeEl: HTMLElement) {
 function expectToDisplayTime(el: HTMLElement, time: string) {
   const inputs = getInputs(el);
   const timeParts = time.split(':');
+  let timeInInputs = [];
 
   expect(inputs.length).toBe(timeParts.length);
 
   for (let i = 0; i < inputs.length; i++) {
-    expect((<HTMLInputElement>inputs[i]).value).toBe(timeParts[i]);
+    timeInInputs.push((<HTMLInputElement>inputs[i]).value);
   }
+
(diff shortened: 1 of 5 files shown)


src/timepicker/timepicker.spec.ts
```diff
@@ -21,12 +21,15 @@ function getMeridianButton(nativeEl: HTMLElement) {
 function expectToDisplayTime(el: HTMLElement, time: string) {
   const inputs = getInputs(el);
   const timeParts = time.split(':');
+  let timeInInputs = [];
 
   expect(inputs.length).toBe(timeParts.length);
 
   for (let i = 0; i < inputs.length; i++) {
-    expect((<HTMLInputElement>inputs[i]).value).toBe(timeParts[i]);
+    timeInInputs.push((<HTMLInputElement>inputs[i]).value);
```
wesleycho: A bit minor, but it's nicer to do
```ts
let timeInInputs = inputs.map((input: HTMLInputElement) => input.value);
```

wesleycho:
Other than that minor suggestion, this LGTM - I would be fine with it merged even without changing with suggestion if preferred, as it's more of a code style thing.

src/timepicker/timepicker.spec.ts
pkozlowski-opensource: Yeh, wanted to do it but the list we get from DOM query doesn't have `map` method so we would have to convert it to an array and it is not as nice any more.

Going to keep it as-is for now but if you can see how to simplify it please send a PR

PR #475, opened 2016-07-26
demo(typeahead): typeahead demos and documentation

maxokorokov:
Initial demos for typeahead including

- simple static search
- result formatting
- wikipedia search
- custom templates for results

### DIFF
--- a/demo/src/app/app.routes.ts
+++ b/demo/src/app/app.routes.ts
@@ -16,7 +16,8 @@ import {
   NgbdRating,
   NgbdTabs,
   NgbdTimepicker,
-  NgbdTooltip
+  NgbdTooltip,
+  NgbdTypeahead
 } from './components';
 
 const routes: RouterConfig = [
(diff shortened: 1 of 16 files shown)


pkozlowski-opensource:
This LGTM generally speaking but before merging we should fix a couple of issues:
* simple example should still use `ngModel` I guess
* we are not properly handling cases of an empty field. Try this with a simple example: type `a` (observe that results are displayed), delete `a` => the same results are still displayed which is wrong. We should either close popup or display the whole array. More generally I think that we are lacking an example showing how to kick-off result display when a minimal no of characters (2?) is entered.

maxokorokov:
> simple example should still use ngModel I guess

Just wanted to show the example as simple as possible, but was thinking about it too. Will put back I guess

> delete a => the same results are still displayed

Actually it's not the case if you look closely at the last two values. And it's simply because of the 10 item limit. Now this is a question → should we treat it as some form of a `select` and display everything when there are no values? For me it's just the implementation detail up to the user...

There is an issue though with wikipedia search with an empty value - it still displays `searching...` even there obviously will be no results.

> an example showing how to kick-off result display when a minimal no of characters (2?)

Will add this one

maxokorokov:
@pkozlowski-opensource amended, PTAL

- added `ngModel` to the first example
- added empty `term` handling for all examples
- added min 2 characters for the first example

Feel free to fix formatting if you don't find it readable enough

pkozlowski-opensource:
Merged, thnx!

Issue #476, opened 2016-07-26
API docs: text overlaps for long property or type names 


pkozlowski-opensource:
As an example check http://localhost:9090/#/components/popover for which `ngbPopover` has a longish type specification (`string | TemplateRef<any>`) and creates very unpleasant look & feel (text is wrapped in an ugly way). The problem is even more visible as soon as we land typeahead.

The only idea I've got so far is to have 2 rows for each property:
* name, type and default value in the first row
* description in the second row.

But all the options are on the table...

maxokorokov:
I was playing a bit too with it. What about this layout?

![screen shot 2016-07-26 at 16 16 09](https://cloud.githubusercontent.com/assets/8074436/17141431/51cb70d6-534c-11e6-8356-f1de8f76eea6.png)

pkozlowski-opensource:
> What about this layout?

I think I like better than my proposal :+1: 

Both are similar but yours look better to me. But I would probably keep table header.

Could you send an alternative PR so I can play with it?

maxokorokov:
@pkozlowski-opensource just replace your table with the one I added in the PR. I have only played with inputs section

pkozlowski-opensource:
Solved via 30902bc

PR #477, opened 2016-07-26
demo: prevent api doc columns from clobbering each other

pkozlowski-opensource:
This is an attempt at solving #476

I'm not a big fan of this change but this is the best I can think of right now. Maybe it will trigger a better idea...

### DIFF
--- a/demo/src/app/components/shared/api-docs/api-docs.component.html
+++ b/demo/src/app/components/shared/api-docs/api-docs.component.html
@@ -6,12 +6,12 @@
   <table class="table table-sm table-hover">
     <tbody>
       <tr>
-        <td class="col-md-2">Selector</td>
-        <td class="col-md-10"><code>{{apiDocs.selector}}</code></td>
+        <td class="col-md-3">Selector</td>
+        <td class="col-md-9"><code>{{apiDocs.selector}}</code></td>
       </tr>
       <tr *ngIf="apiDocs.exportAs">
-        <td class="col-md-2">Exported as</td>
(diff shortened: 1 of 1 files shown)


maxokorokov:
Argh, thought was commenting here... See the screenshot in #476 

```html
<table class="table table-sm table-hover">
  <tbody>
    <tr *ngFor="let input of apiDocs.inputs">
      <td class="col-md-3"><code>{{input.name}}</code></td>
      <td class="col-md-9">
        <div><i>Type: </i><code>{{ input.type }}</code></div>
        <div><i>Default value: </i>{{input.defaultValue || '-'}}</div>
        <div style="margin: 10px 0">{{ input.description }}</div>
      </td>
    </tr>
  </tbody>
</table>
```

and would throw away the header completely...

pkozlowski-opensource:
> and would throw away the header completely...

We need to keep some consistency between inputs and outputs. If we throw it away we should do it for both or none.

maxokorokov:
would have also probably added `<code>` for the default value, unless it's `-`

pkozlowski-opensource:
I've pushed an update to this PR based on @maxokorokov ideas. I've also added a check around default value so we display it only when there is one. This limits amount of unused blank space.

I start to like it!

api_doc_changes

pkozlowski-opensource:
@icfantv I've pushed this one as I think it is best we've got right now. You had some ideas about using muted text which I didn't quite get so please send a follow up PR(s) if you see how to improve things. Thnx!

PR #478, opened 2016-07-26
fix(typeahead): re-position on after view check

maxokorokov:
Stumbled upon a small positioning issue while playing with validity: popup might move when page contents change.

Implementation same as for `tooltip` → reposition during `onAfterViewChecked`


pkozlowski-opensource:
This is a good change. Will merge it as soon as TravisCI is green.

maxokorokov:
@pkozlowski-opensource I don't think it was triggered for some reason, will push again

PR #484, opened 2016-07-27
demo: generate plunkers

pkozlowski-opensource:
This is still WIP but wanted to share if anyone wants to have an early look. Currently you can see it in action in the alert/basic demo only.


gulpfile.js
```diff
@@ -189,16 +189,30 @@ gulp.task('generate-docs', function() {
   return gulpFile('api-docs.ts', docs, {src: true}).pipe(gulp.dest(PATHS.demoApiDocs));
 });
 
+gulp.task('generate-plunks', function() {
+  var getPlunker = require('./misc/plunk-gen');
+  var demoGenUtils = require('./misc/demo-gen-utils');
+
+  var componentName = 'alert';
+
+  var plunks = demoGenUtils.getDemoNames('alert').reduce(function(soFar, demoName) {
```
wesleycho: This would make more sense as
```js
var plunks = demoGenUtils.getDemoNames('alert').map(demoName => {
  return {
    name: ``${componentName}/demos/${demoName}/plnkr.html`,
    source: getPlunker(componentName, demoName)
  };
});
```

wesleycho:
LGTM

demo/src/app/components/datepicker/demos/customday/datepicker-customday.ts
```diff
@@ -25,7 +25,7 @@ import {NgbDateStruct} from '@ng-bootstrap/ng-bootstrap';
     }
   `]
 })
-export class NgbdDatepickerCustomDay {
+export class NgbdDatepickerCustomday {
```
maxokorokov: I would probably prefer to rename files to `datepicker-custom-day.ts`
(but not important really)

Issue #469, opened 2016-07-24
Investigate compatibility with reactive forms


pkozlowski-opensource:
Today several directives (buttons, timepicker, typeahead) are assuming presence of the `ngModel` directive. But for reactive forms people are not going to use `ngModel` so we need to see how we could support both template-driven and model-driven forms.

crooksey:
If this issue is now closed, is there any documentation to support reactive forms?

jasonhjohnson:
@crooksey Were you able to get this working correctly with reactive forms?

crooksey:
Yes I had/have this working, 

```
this.oForm = this.fb.group({
    oHForm: this.fb.group({ // create nested formgroup to pass to child
        selectTypeahead: ['', Validators.required],
        ohDatePicker: [],
        }),	
})
```
Shows how to setup a form group for typeahead and datePickers

Template, datepicker:
```
<input class="form-control" placeholder="dd-mm-yyyy"
     id="oh_date_picker"
     formControlName="ohDatePicker"
     name="dp"  
     ngbDatepicker #d="ngbDatepicker">
```

Template typeahead:
```
<input type="text" class="form-control" 
    formControlName=" selectTypeahead"
    [ngbTypeahead]="search" 
    
    (selectItem)="taMatch($event)"
    [resultTemplate]="rt" [inputFormatter]="formatter" autofocus tabindex="0" #msgFocous/>
```

Hope this helps someone.

rajeshmahendran41:
How to use time picker with reactive forms ?

PR #486, opened 2016-07-27
test: add minimal tests for form validation

maxokorokov:
Minimal validation tests for `required` validatior (both template-driven and model-driven forms) for:

- buttons
- timepicker
- typeahead

as a part of investigation for #469


maxokorokov:
Actually, maybe it makes sense to isolate these tests into a separate file like `forms.spec.ts`...

PR #489, opened 2016-07-28
fix: rename outline buttons class names for alpha.3 migration

pkozlowski-opensource:
See https://github.com/twbs/bootstrap/pull/20391/files


icfantv:
LGTM

PR #491, opened 2016-07-28
fix(pagination): display proper cusrsor on each page

pkozlowski-opensource:
Part of #142

### DIFF
--- a/src/pagination/pagination.ts
+++ b/src/pagination/pagination.ts
@@ -7,41 +7,46 @@ import {getValueInRange, toInteger, toBoolean} from '../util/util';
 @Component({
   selector: 'ngb-pagination',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  styles: [`
+    button.page-link {
+      outline: 0;
+    }
+  `],
   template: `
     <nav>
       <ul [class]="'pagination' + (size ? ' pagination-' + size : '')">
         <li *ngIf="boundaryLinks" class="page-item" [class.disabled]="!hasPrevious()">
-          <a aria-label="First" class="page-link" (click)="selectPage(1)">
+          <button aria-label="First" class="page-link" (click)="selectPage(1)">
             <span aria-hidden="true">&laquo;&laquo;</span>
             <span class="sr-only">First</span>
-          </a>                
+          </button>                
         </li>
       
-        <li *ngIf="directionLinks"class="page-item" [class.disabled]="!hasPrevious()">
-          <a aria-label="Previous" class="page-link" (click)="selectPage(page-1)">
+        <li *ngIf="directionLinks" class="page-item" [class.disabled]="!hasPrevious()">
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.ts
```diff
@@ -7,41 +7,46 @@ import {getValueInRange, toInteger, toBoolean} from '../util/util';
 @Component({
   selector: 'ngb-pagination',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  styles: [`
+    button.page-link {
```
pkozlowski-opensource: I'm not sure this is the best selector and I'm a bit worried that this one is going to "leak" outside of this widget. Any other ideas?

src/pagination/pagination.ts
```diff
@@ -7,41 +7,46 @@ import {getValueInRange, toInteger, toBoolean} from '../util/util';
...
+      outline: 0;
```
Foxandxss: This kills a11y completely.

pkozlowski-opensource: Using buttons? Or removing outline?

What is your alternative proposal? (hint: if you don't remove outline it looks really ugly - try clicking on prev / next buttons).

Foxandxss: I hear you, but without outline, an user tabbing through the page won't ever notice those buttons.

pkozlowski-opensource: I think that we all know that there are no perfect solutions (that is - or at least we can't think of them atm... ) - longish discussion in #142 is the evidence. So I think that we are at the point where we need to choose _something_ - even if not perfect we need to choose the least bad approach. And we need to do something for alpha.3 since links without `href` are styled black.

So, this is the decision time. I can see the following options:
* this PR
* variant of this where we remove outline only from the focused buttons (to be tested)
* what I've proposed in https://github.com/ng-bootstrap/ng-bootstrap/issues/142#issuecomment-231587330 - empty `href` with prevent default

I can't think of anything else atm. So please, let's have a discussion about what _can_ be done. Either pick one of the above or propose an alternative, otherwise it will be hard progress on this one.

pkozlowski-opensource:
Alternative PR: #492

pkozlowski-opensource:
Update:
* removing `outline` is not an option
* we can't use buttons as those are not styled properly in BS4 => going to open an issue for them
* closing this PR and going with the shot-term solution of using empty `href`s

Thnx for pointing out the issue with this PR @Foxandxss !

mendeza:
I was on vacation thus could not comment in a timely fashion. You are correct that removing outline is not something ng-bootstrap should do. Leave it up to consuming projects to improve the outline style or to hide it. Leaving it visible is correct A11y.

I do think that eventually moving to buttons (or even being able to choose either) will be better. Buttons will be more mobile friendly. 

The nice thing about using `href` as a magic attribute is that it is not some arbitrary thing - it is descriptive and canonical.

PR #493, opened 2016-07-28
fix: use proper styling on accordions and tabs

pkozlowski-opensource:
I've tried to use `<button>` elements styled with `btn-link` and it works _mostly_ fine. The only real downside is that `<button>`s leave an unpleasant "halo effect" when a given button is clicked. 

My attempt at removing this effect might have a11y ramifications - see discussion in https://github.com/ng-bootstrap/ng-bootstrap/pull/491#discussion-diff-72609319

As the result of all the discussions we had so far I'm leaning to the approach from this PR as:
* it improves the situation (proper styling and cursor)
* doesn't introduce _new_ a11y issues

The bitter truth is that I'm not a11y expert so I can't fully assess all the consequences of choices we are making here. Alternative proposals are more than welcomed.

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -82,7 +82,7 @@ export interface NgbPanelChangeEvent {
   <div class="card">
     <template ngFor let-panel [ngForOf]="_panels">
       <div [class]="'card-header ' + (panel.type ? 'card-'+panel.type: type ? 'card-'+type : '')" [class.active]="_isOpen(panel.id)">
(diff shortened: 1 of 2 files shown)


icfantv:
LGTM

mendeza:
When I did a dev-tools hack demo to use buttons as tab headers the `outline` was evident. I saw this as a plus. Developers should not be afraid to mold this behavior in their `app.css`. Not everything can or should be delivered on a silver platter.

PR #500, opened 2016-07-28
demo(typeahead): add form-control class to inputs

maxokorokov:
Minor fix → adding `form-control` class to demos with typeahead to inputs

### DIFF
--- a/demo/src/app/components/typeahead/demos/basic/typeahead-basic.html
+++ b/demo/src/app/components/typeahead/demos/basic/typeahead-basic.html
@@ -5,6 +5,6 @@ A typeahead example that gets values from a static <code>string[]</code>
   <li>limits to 10 results</li>
 </ul>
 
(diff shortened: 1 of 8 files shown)


pkozlowski-opensource:
I'm not sure... I think I like the version before this change better... Let's see what others think...

maxokorokov:
It looks horrifying in IE :) 
It shouldn't fill the whole width probably though

pkozlowski-opensource:
> It shouldn't fill the whole width probably though

Yep, that's my point. I was briefly checking if we can make it smaller without too much pain. Size?

maxokorokov:
```javascript
@Component({
  ...
  styles: [`.form-control { width: 300px; }`]
})
```

screen shot 2016-07-28 at 19 23 52

pkozlowski-opensource:
Yep, if you could update the PR so an input gets smaller size this would be great.

maxokorokov:
done

PR #508, opened 2016-07-29
chore: remove api-docs.ts from linting

### LINKED ISSUE #504: chore: lint fails on the generated API doc
Need to find a way of dealing with this. One of:
* make it pass lint validation
* exclude from linting

Foxandxss:
Since this is a generated file, we just exclude it. We shouldn't worry to lint something that is generated with a tool that is not under our control.

Fixes #504


pkozlowski-opensource:
LGTM if TravisCI is green

PR #516, opened 2016-07-29
fix(timepicker): propagate only valid model values

### LINKED ISSUE #513: timepicker: make the component invalid if wrong input
Using the demo with custom validation, if I enter something invalid (like `aa`, for example) in one of the fields, or if I clear one of the fields, I would expect a validation error other than "It's a bit too early", and the model to become null, since `aa:<blank><blank>` is not a valid input and should thus not, IMO, be saved as `00:00` in the model.

pkozlowski-opensource:
Fixes #513


Foxandxss:
LGTM

PR #517, opened 2016-07-29
fix(timepicker): format input values on blur

### LINKED ISSUE #511: timepicker: impossible to enter a two-digit number
Using the demo, if I try to enter, for example, `10` in the hour input field, the timepicker immediately formats the entered `1` as `01`, and then doesn't let me enter the `0`.

pkozlowski-opensource:
Fixes #511

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -27,6 +27,10 @@ function getMeridianButton(nativeEl: HTMLElement) {
   return nativeEl.querySelector('button.btn-outline-primary');
 }
 
+function createChangeEvent(value: string) {
+  return {target: {value: value}};
+}
+
 function expectToDisplayTime(el: HTMLElement, time: string) {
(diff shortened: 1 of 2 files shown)


wesleycho:
So this doesn't really format on blur as far as I can tell, but on the change output. Is that sufficient? As far as I can tell, this theoretically shouldn't change anything.

Also, should this use the input output if we go the output route? Or do we need to use `(blur)`?

pkozlowski-opensource:
> So this doesn't really format on blur as far as I can tell, but on the change output

That's the intention. I can change the commit message.

> Is that sufficient? As far as I can tell, this theoretically shouldn't change anything.

Actually it changes everything :-) So far we were using `ngModel` which updates on `input` (each key stroke). This PR changes to the reaction to the `change` event (user "done" editing - in case of input fields it is equivalent to `blur`).

> Also, should this use the input output if we go the output route?

Not sure what you mean here...

PR #519, opened 2016-07-29
feat(timepicker): allow hiding the spinners

### LINKED ISSUE #510: timepicker: support removing the chevrons
Especially when the keyboard and mousewheel support is there, I think the chevrons are redundant and should be removed, or at least removable:

 - they take a whole lot of vertical space, much more than traditional form inputs
 - they're less efficient than the keyboard to increase/decrease the values

So it would be nice if there was a way to hide them. Even nicer would be to have a service allowing to configure defaults, and be able to hide them globally.

jnizet:
fix #510


pkozlowski-opensource:
Merged, thnx!

PR #524, opened 2016-07-30
demo(timepicker): improve validation demo

jnizet:
- use a form-group with .has-danger or .has-success to have a red or green border around the inputs
- use .form-control-feedback instead of an alert for error/success message, since that's what it's designed for


pkozlowski-opensource:
Thnx!

PR #530, opened 2016-08-02
reafctor: use forms 0.2.0

maxokorokov:
This is an experiment with migration to new forms 0.2.0. Everything works except for one thing:

Main issue is with tests now, as NgModel update became asynchronous → the recommended solution for now is to do the following until there is a solution in `core/testing`...

```javascript
fixture.detectChanges();
fixture.whenStable().then(() => {
    fixture.detectChanges();
    // expect here...
```

For now this makes tests absolutely ugly, especially for the `timepicker`

### DIFF
--- a/demo/src/app/components/timepicker/demos/validation/timepicker-validation.html
+++ b/demo/src/app/components/timepicker/demos/validation/timepicker-validation.html
@@ -1,13 +1,12 @@
 <p>Illustrates custom validation, you have to select time between 12:00 and 13:59</p>
 
-<div class="form-group" [class.has-success]="c.valid" [class.has-danger]="!c.valid">
-  <ngb-timepicker [(ngModel)]="time" #c="ngForm" [ngFormControl]="lunchControl" required></ngb-timepicker>
+<div class="form-group" [class.has-success]="ctrl.valid" [class.has-danger]="!ctrl.valid">
+  <ngb-timepicker [(ngModel)]="time" [formControl]="ctrl" required></ngb-timepicker>
   <div class="form-control-feedback">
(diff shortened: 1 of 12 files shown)


pkozlowski-opensource:
@maxokorokov this is the best we can do for the tests for now so I think we should merge it in to better prepare for rc5. Do you see anything else that would have to go into this PR before it can land?

maxokorokov:
@pkozlowski-opensource that is sad. But ok

Just two things:

- should I add `@angular/forms` to `vendor.ts`?
- don't you want to merge #516 / #517 first?

pkozlowski-opensource:
> should I add @angular/forms to vendor.ts

This won't hurt so yes. I will open a separate issue for re-considering current files split

> don't you want to merge #516 / #517 first?

So someone will have to deal with merge conflicts :-) If you are willing to suffer through conflicts resolution after #516 / #517 lands this would help me a lot!

maxokorokov:
@pkozlowski-opensource I think it will be simpler this way, so merge those first and then I'll realign this one

pkozlowski-opensource:
>  I think it will be simpler this way, so merge those first and then I'll realign this one

@maxokorokov thnx - I've just landed both PRs. It might make your life a _bit_ easier since one of PRs eliminated `ngModel` from internals of the timepicker. So maybe it will reduce number of async calls needed...

>

maxokorokov:
@pkozlowski-opensource that's what I thought too

maxokorokov:
@pkozlowski-opensource done, just one last thing. Do you think I should leave `//TODO`s in the tests?

PR #531, opened 2016-08-02
fix(demo): clarify typeahead demo variable name

Brocco:
The demo specified `request` while the parameter was actually the `response`


pkozlowski-opensource:
Merged, thnx!

PR #534, opened 2016-08-03
chore: extract methods that a part of the public API

pkozlowski-opensource:
PR adds extraction of methods that can be called on exported directive instances. I've added a sample doc to accordion to see how it displays. Will add more docs if this approach is validated.

@icfantv could you please have a look at it from the l&f points of view? What worries me that now a user needs to scroll really a lot to see samples. Maybe we should re-organize the page in the way that demos are displayed first? Or display one flashy demo first (with a link to more demos), then API and then more demos? 

Part of #222.

### DIFF
--- a/demo/src/app/components/shared/api-docs/api-docs.component.html
+++ b/demo/src/app/components/shared/api-docs/api-docs.component.html
@@ -53,5 +53,22 @@
       </table>
     </section>
   </template>
+
+  <template [ngIf]="apiDocs.exportAs && apiDocs.methods.length">
+    <section>
+      <h3 id="methods">Methods</h3>
+      <table class="table table-sm table-hover">
+        <tbody>
+        <tr *ngFor="let method of apiDocs.methods">
+          <td class="col-md-3"><code>{{method.name}}</code></td>
+          <td class="col-md-9">
+            <div><i>Signature: </i><code>{{ _methodSignature(method) }}</code></div>
+            <div style="margin: 10px 0">{{ method.description }}</div>
+          </td>
+        </tr>
+        </tbody>
+      </table>
+    </section>
+  </template>
 </div>
 <hr/>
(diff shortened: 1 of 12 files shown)


demo/src/app/components/shared/api-docs/api-docs.component.html
```diff
@@ -53,5 +53,22 @@ <h3 id="outputs">Outputs</h3>
...
+
+  <template [ngIf]="apiDocs.exportAs && apiDocs.methods.length">
+    <section>
+      <h3 id="methods">Methods</h3>
+      <table class="table table-sm table-hover">
+        <tbody>
+        <tr *ngFor="let method of apiDocs.methods">
+          <td class="col-md-3"><code>{{method.name}}</code></td>
+          <td class="col-md-9">
+            <div><i>Signature: </i><code>{{ _methodSignature(method) }}</code></div>
```
wesleycho: Does ng2 support [tagged template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals)? I wonder if this would provide a more elegant implementation. I've no problem with it either way though.

pkozlowski-opensource: Nope, sadly not.

demo/src/app/components/shared/api-docs/api-docs.component.ts
```diff
@@ -13,4 +13,9 @@ export class NgbdApiDocs {
   @Input() set directive(directiveName) {
     this.apiDocs = docs[directiveName];
   };
+
+  private _methodSignature(method) {
+    const args = method.args.map((arg) => { return `${arg.name}: ${arg.type}`; }).join(', ');
```
pkozlowski-opensource: Will change, thnx.

pkozlowski-opensource:
This is ready for review now.

wesleycho:
LGTM

Issue #518, opened 2016-07-29
Introduce global config for all widgets


pkozlowski-opensource:
Good start described here: https://github.com/ng-bootstrap/ng-bootstrap/issues/510#issuecomment-236221937

We just need to refine it to have a consistent naming convention / way of coding it.

jnizet:
We will also have to decide how these configuration services/classes are documented, and how the documentation of the components themselves must be written in order to identify which component inputs can be configured from the global configuration.

pkozlowski-opensource:
So, finally had some more time to think about this item. Before diving into the proposal (which, in broad strokes, is identical to @jnizet one) here are the desired properties of the final solution:
* users should be able to configure (globally or in a given module / component tree) default values of inputs (those not provided when using a widget)
* we should still be able to properly generate API docs
* we shouldn't duplicate docs between config object and inputs 
* applying config in directives shouldn't result in code explosion (shouldn't be royal pain in the @#*&#)

Given the above and assuming that we want to have a global config for a directive named `NgbFoo` we should:
* have a class named `NgbFooConfig` that list all defaults for inputs (their names, types and default values)
* `NgbFoo` should have a constructor that optionally injects `NgbFooConfig` and propagates config values to input feilds

Example using `NgbProgressbar`: 

```typescript
export class NgbProgressbarConfig {
	animated = false;
	max = 100;
	striped = false
}
```

```typescript
export class NgbProgressbar {
    constructor(@Optional() config: NgbProgressbarConfig = new NgbProgressbarConfig()) {
        this.animated = config.animated; 
        ....  
    }
}
```

### Open questions
* I'm not sure where we should keep API docs... On inputs only? On config only? On both?
* ...

### Infrastructure changes
* change API generation script to extract info from classes
* change API generation script to merge info / docs from config classes and directives themselves
* decide how to present info about config in the API doc

pkozlowski-opensource:
@jnizet I'm going to poke around infrastructure changes in the coming days but if you would like to send a PR introducing config for one of the widgets it would help us iterate on what needs to be done and answer all the remaining questions (timepicker is not the best candidate probably as there is quite a bit of changes in it atm)

jnizet:
OK. I'll do that as quickly as possible (probably this evening, UTC+2). Just a note: in my attempt of doing that on the time picker, I first tried the syntax you suggest above:

```
    const DEFAULT_CONFIG = new NgbProgressbarConfig();
[...]

PR #536, opened 2016-08-03
feat(progressbar): add config service to provide default values

jnizet:
refs #518

The first commit keeps the initialization of the inputs in the component as is. 
Also note that I changed the documentation of the animated input in order to indicate what it *will* do once the bootstrap bug is fixed, according to bootstrap's documentation.

The second commit changes the infrastructure code to
 - generate documentation for public properties of services in addition to methods
 - add a component to display the api docs of services in the demo
 - get the default value and description of inputs from the matching config service property if not specified in the directive itself
 - indicate which properties are initialized with properties of the config service
 - use all this in the documentation of NgbProgressbar

### DIFF
--- a/demo/src/app/components/progressbar/demos/config/progressbar-config.ts
+++ b/demo/src/app/components/progressbar/demos/config/progressbar-config.ts
@@ -1,0 +1,17 @@
+import {Component} from '@angular/core';
+import {NgbProgressbarConfig} from '@ng-bootstrap/ng-bootstrap';
+
+@Component({
+  selector: 'ngbd-progressbar-config',
+  templateUrl: './progressbar-config.html',
+  providers: [NgbProgressbarConfig] // add the NgbProgressbarConfig to the component providers
+})
+export class NgbdProgressbarConfig {
+  constructor(config: NgbProgressbarConfig) {
+    // customize default values of progress bars used by this component tree
+    config.max = 1000;
+    config.striped = true;
+    config.animated = true;
+    config.type = 'success';
+  }
+}
--- a/src/progressbar/progressbar-config.ts
+++ b/src/progressbar/progressbar-config.ts
@@ -1,0 +1,30 @@
+import {Injectable} from '@angular/core';
+
+/**
+ * Configuration service for the NgbProgressbar component.
+ * You can inject this service, typically in your root component, and customize the values of its properties in
+ * order to provide default values for all the progress bars used in the application.
+ */
+@Injectable()
+export class NgbProgressbarConfig {
+  /**
+   * Maximal value to be displayed in the progressbar.
+   */
+  max = 100;
+
+  /**
+   * A flag indicating if the stripes of the progress bar should be animated. Takes effect only for browsers
+   * supporting CSS3 animations, and if striped is true.
+   */
+  animated: boolean = false;
+
+  /**
+   * A flag indicating if a progress bar should be displayed as striped.
+   */
+  striped: boolean = false;
+
+  /**
+   * Type of progress bar, can be one of "success", "info", "warning" or "danger".
+   */
+  type: string;
+}
--- a/src/progressbar/progressbar-config.ts
+++ b/src/progressbar/progressbar-config.ts
@@ -7,24 +7,8 @@ import {Injectable} from '@angular/core';
  */
 @Injectable()
 export class NgbProgressbarConfig {
-  /**
-   * Maximal value to be displayed in the progressbar.
-   */
   max = 100;
-
-  /**
-   * A flag indicating if the stripes of the progress bar should be animated. Takes effect only for browsers
-   * supporting CSS3 animations, and if striped is true.
-   */
-  animated: boolean = false;
-
-  /**
-   * A flag indicating if a progress bar should be displayed as striped.
-   */
-  striped: boolean = false;
-
-  /**
-   * Type of progress bar, can be one of "success", "info", "warning" or "danger".
-   */
+  animated = false;
+  striped = false;
   type: string;
 }
--- a/demo/src/app/components/progressbar/demos/config/progressbar-config.html
+++ b/demo/src/app/components/progressbar/demos/config/progressbar-config.html
@@ -1,0 +1,5 @@
+<p>This progress bar uses the customized default values.</p>
+<ngb-progressbar value="250"></ngb-progressbar>
+
+<p>This progress bar uses the customized default values, but changes the type using an input.</p>
+<ngb-progressbar value="500" type="info"></ngb-progressbar>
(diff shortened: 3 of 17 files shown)


pkozlowski-opensource:
@jnizet thnx for this PR. It looks good overall but I will hold its merging till we've got angular rc.5 sorted out. The reason is that RC5 brings changes to the way we configure / consume providers.

jnizet:
@pkozlowski-opensource I rebased this branch on master.

pkozlowski-opensource:
@jnizet thnx so much for this one, it looks really good! I'm too tired to review this properly today but will have a fresh look & land it tomorrow morning. Thnx again!

demo/src/app/components/progressbar/demos/config/progressbar-config.ts
```diff
@@ -0,0 +1,17 @@
+import {Component} from '@angular/core';
+import {NgbProgressbarConfig} from '@ng-bootstrap/ng-bootstrap';
+
+@Component({
+  selector: 'ngbd-progressbar-config',
+  templateUrl: './progressbar-config.html',
+  providers: [NgbProgressbarConfig] // add the NgbProgressbarConfig to the component providers
```
maxokorokov: I don't think we need this one now

src/progressbar/progressbar-config.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * A flag indicating if a progress bar should be displayed as striped.
+   */
+  striped: boolean = false;
+
+  /**
+   * Type of progress bar, can be one of "success", "info", "warning" or "danger".
+   */
+  type: string;
```
maxokorokov: could be `type: string = 'success' | 'info' | ...`

maxokorokov:
Looks good! I was just thinking after the looking at the demo site, that it's a shame that we have a complete duplication of information for `NgbProgressbar` and `NgbProgressbarConfig` documentation.

Do we need and explicit descriptions and types in the `NgbProgressbarConfig` at all? Everything is kinda listed in the `NgbProgressbar` section already...

pkozlowski-opensource:
> Looks good! I was just thinking after the looking at the demo site, that it's a shame that we have a complete duplication of information for NgbProgressbar and NgbProgressbarConfig documentation.

> Do we need and explicit descriptions and types in the NgbProgressbarConfig at all? Everything is kinda listed in the NgbProgressbar section already...

After looking at the demo site I tend to agree with @maxokorokov - there is too much duplication on the demo site... @jnizet maybe we should re-think the initial strategy and:
* keep description on the directive level (not sure how it will play out with default values / types...)
* only list input names for the config object?

WDYT?

pkozlowski-opensource:
Oh, if you don't mind I'm going to pick up your changes to the API generation script as I need those for modals.

demo/src/app/components/progressbar/demos/config/progressbar-config.ts
jnizet: If we don't add the provider here, the config will be the global one, and this demo will thus change the global config used to initialize all the other demos.

src/progressbar/progressbar-config.ts
jnizet: I left it as it was before, in the component, but I think this is a good idea.

jnizet:
No problem, of course, to pick the api generation script bits.

Regarding the redundancy of information, I think that, as long as it's not repeated in the code, it's fine. In the API doc, IMHO, it doesn't hurt: if I look at the documentation of a service, I'd prefer to have all the documentation there rather than having to realize that I must look into the matching property of the matching directive to find the information. But if you prefer to only list the input names in the config, of course, it's easily doable: I just need to remove all the other stuff from the template. I'm afraid it will look like we were just too lazy to document things, though. We will also have to distinguish between the properties of Config services that match with a directive property, and the other, non-Config services, because those ones should be documented properly.

I agree, however, that the demo pages are becoming a bit crowded as it is. I think it has already been mentioned in another issue. Once the directives become more complex, the demos are not immediately visible, because you have to scroll to find them. 

Maybe each demo page should have two tabs, each with their own URL, and put the demo on one tab (that would be the default one), and the api in the other. I also thought about an accordion, but I think it's important to be able to link directly to the documentation. BTW, it would be nice to generate real unique HTML IDs for each class and each property/method, to be able to link directly to a specific piece of the documentation. Not sure how that would work with the hashbang routing used in the demo, though.

pkozlowski-opensource:
Thnx @jnizet - I've opened #628 for the API extraction bits.

Let me chew over your latest comment.

pkozlowski-opensource:
@jnizet OK, after spending a weekend without looking at any code and having a fresh look here are my thoughts:
* I hear what you are saying about docs duplication not hurting but IMO the current proposal overloads demo page and benefits doesn't justify it. My proposal: let's keep docs on the directive level and for config objects just list: class name, its description and property names.
* I'm not excited about type and default value moved to the config classes - this makes working on directives much harder as now I need to jump between 2 classes to understand what is going on... I'm afraid that I don't have good counter-proposal ATM....
* I'm not sure if having providers on our module level will cover a scenario where users  would like to have a config class specified as a provider on their module level. Did you check that providers specified on my module override providers on an imported module (probably yes, but just to be sure).

Sorry for all those back-and-forth but there are still few wrinkles I would like to see ironed out.... Just let me know if I'm asking for too much so I can take it over... Otherwise we really do appreciate your work on this stuff!

jnizet:
OK. No problem to make the evolutions.

Just to be crystal clear on what to do:

- leave the generated documentation of directives as it is in the current status of this PR: description of each input, default value coming from the NgbProgressbarConfig, and indication that it's initialized from the matching property of NgbProgressbarConfig
- in the source code, move back the documentation of the property to the directive, rather than having it in the NgbProgressbarConfig class
- In the generated documentation of the NgbConfig service, only list the properties, without type, and without documentation.

Is that right?

If that's right, my plan is to use a documentation component dedicated to Config services, because "regular" services will need to have their fields and methods documented. Is that OK?

Regarding the provider, I will add a test which covers this scenario.

jnizet:
I went ahead: implemnted the required changes (hopefully), and rebased on master. 

I also removed the `isDirective` flag from the NgbdApiDocs component, since we now use a separate component for services, classes and intervals that are not directives.

jnizet:
Note: I have a commit ready on top of this that adds types (interfaces) describing the api doc model, and uses them in the now less simple api doc components. Tell me if you would like this commit to be part of this PR.

pkozlowski-opensource:
@jnizet I've merged it with a couple of minor changes (conflict resolution). I think that now we can use this pattern to add config classes to other directives!

jnizet:
Great! I'll start doing it component by component.

PR #538, opened 2016-08-03
demo(typeahead): add error recovery to http demo

maxokorokov:
Add failed requests handling to wikipedia demo 

- in case it doesn't work 
- to explicitly demonstrate error recovery

### DIFF
--- a/demo/src/app/components/typeahead/demos/http/typeahead-http.html
+++ b/demo/src/app/components/typeahead/demos/http/typeahead-http.html
@@ -1,12 +1,20 @@
 A typeahead example that gets values from the <code>WikipediaService</code>
 <ul>
-  <li>remote data retrieval</li>
   <li><code>debounceTime</code> operator</li>
-  <li><code>do</code> operator</li>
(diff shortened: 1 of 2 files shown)


maxokorokov:
As an alternative (because the example starts to get huge), might add only the `catch()` to show where to handle errors and throw away all the visual:

```javascript
search = (text$: Observable<string>) =>
  text$
    .debounceTime(300)
    .distinctUntilChanged()
    .do(term => { this._searching = term.length > 0; })
    .switchMap(term => term === '' ? Observable.of([]) :
      this._service.search(term)
        .catch(() => {                         // <-----
          // handle any errors here            // <-----
          return Observable.of([]);            // <-----
        })
    )
    .do(() => { this._searching = false; });
```

pkozlowski-opensource:
Shouldn't `catch()` go to the service instead?

maxokorokov:
@pkozlowski-opensource it might of course, but how would you handle it later to update the UI in this case?

pkozlowski-opensource:
Let's put it on a back-burner for now as there are things we still don't fully understand / need to explore more.

PR #549, opened 2016-08-05
chore: remove forms, http and rxjs from UMD bundle

### LINKED ISSUE #548: UMD bundles contain Angular and RxJS
By quickly checking size of our bundles I've realized that we are bundling Angular and RxJS as part of what gets produced in `dist/bundles`. This is wrong - we should have Angular 2 and RxJS as externals.

Foxandxss:
I used the same technique (more or less) than Angular so it should be good.

If we start adding more operators, I will simplify the setup a bit with a `rxjsExternal` or similar.

Fixes #548

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -74,7 +74,19 @@ gulp.task('umd', function(cb) {
         entry: './dist/index.js',
         output: {filename: 'dist/bundles/ng-bootstrap.js', library: 'ngb', libraryTarget: 'umd'},
         devtool: 'source-map',
-        externals: {'@angular/core': ngExternal('core'), '@angular/common': ngExternal('common')}
+        externals: {
+          '@angular/core': ngExternal('core'),
+          '@angular/common': ngExternal('common'),
+          '@angular/forms': ngExternal('forms'),
+          'rxjs/Rx': {root: 'Rx', commonjs: 'rxjs/Rx', commonjs2: 'rxjs/Rx', amd: 'rxjs/Rx'},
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -74,7 +74,20 @@ gulp.task('umd', function(cb) {
         entry: './dist/index.js',
         output: {filename: 'dist/bundles/ng-bootstrap.js', library: 'ngb', libraryTarget: 'umd'},
         devtool: 'source-map',
-        externals: {'@angular/core': ngExternal('core'), '@angular/common': ngExternal('common')}
+        externals: {
+          '@angular/core': ngExternal('core'),
+          '@angular/common': ngExternal('common'),
+          '@angular/forms': ngExternal('forms'),
+          '@angular/http': ngExternal('http'),
```
pkozlowski-opensource: We are not using `http` in the library itself (only in the demo) so it is not strictly needed. But it doesn't hurt either so I'm fine with keeping it.

pkozlowski-opensource:
Did you check that no other parts of Angular are pulled in? I saw `compiler` showing up, but maybe it was transitive dependency of forms (would be wired, still)

pkozlowski-opensource:
LGTM otherwise.

gulpfile.js
Foxandxss: Oh, thought we did. Let me fix.

Foxandxss:
Left the terminal open with the fix but never pushed it...

Issue #550, opened 2016-08-06
boolean inputs: handle them all the same way


jnizet:
Maybe I'm missing something. If so, pardon me.

Reading the source code of NgbProgressbar, I noticed that the boolean inputs are of type `boolean | string`, and that a method is used to transform this value to a boolean (where the empty string is returned as true) to use it in the template.

My guess is that it's designed this way to allow using 

    <ngb-progressbar striped></ngb-progressbar>

instead of the more verbose 

    <ngb-progressbar [striped]="true"></ngb-progressbar>. 

But the boolean inputs in the other components don't use the same strategy, making the library inconsistent.

If there is no reason for progress bar to be different from all the other ones, then it should be modified. Otherwise, all the other ones should be modified. Anyway, a comment on the `toBoolean` method would be helpful to understand what's going on.

If my guess on the rationale of this design is correct, my vote would be to use boolean everywhere. It makes the usage a bit more verbose, but also clearer. It makes the component code simpler, lighter and more obvious. And there is less chance of introducing another incoherence in future components.

pkozlowski-opensource:
> Maybe I'm missing something. 

You are not!

> My guess is that it's designed this way to allow using ... instead of the more verbose...

Correct.

> But the boolean inputs in the other components don't use the same strategy, making the library inconsistent.

Yes, we've got the same issue for numbers, BTW.

As for question regarding what we should do to be consistent - I don't know. My hope was that we can get some sort of solution from the core, see: #8968. BTW, material folks got the same problem and hence this issue.

> my vote would be to use boolean everywhere. It makes the usage a bit more verbose, but also clearer.

I'm fine with removing all the type conversion "magic" till we've got a good pattern for this in the Angular2 itself.

jnizet:
Regarding the numbers, there are several issues:

1. some components (like rating) accept a `number`, others (like pagination) accept a `number | string`, and use `toInteger()` to coerce the value to an integer
2. `toInteger()` actually does two things: transform strings to numbers, and transform decimal numbers to integers. The intent is not always very clear. For example, the intent in pagination seems to have both transformations applied, whereas in NgbTime, the intent seems to just transform the potentially decimal or null numbers into integers. 
3. In the components, like rating or progressbar, which have inputs of type number, the spec and/or the demo *sometimes* uses `value="50"` instead of `[value]="50"` to provide the input. The actual value of the input in the component, which is supposed to be a number, is thus a actually a string. 

[...]

PR #552, opened 2016-08-07
chore: don't accept string as boolean input

jnizet:
This is a first commit for #550, dealing only with boolean inputs, and not changing anything to number inputs.


src/pagination/pagination.ts
```diff
@@ -48,46 +48,31 @@ import {getValueInRange, toInteger, toBoolean} from '../util/util';
...
-  set ellipses(value: boolean) {
-    this._ellipses = toBoolean(value);
-  }
-
-  get ellipses(): boolean { return this._ellipses; }
+  @Input() ellipses = true;
+  private _maxSize = 0;
+  private _page = 0;
+  private _pageCount = 0;
+  private _pageSize = 10;
```
pkozlowski-opensource: Could we move those private fields UP, to the top of the file where they were initially?

pkozlowski-opensource: `_pageSize` should be a regular `@Input`.

jnizet: I agree, but the idea of this PR was to only focus on boolean inputs, and to keep numbers as is for the moment, mainly because I wasn't sure what to do with all these number inputs.

I will definitely do another PR to deal with numbers, but I wasn't sure about what to do with all the number inputs here. 

We agreed that `number | string` should become `number`, but the usage of `toInteger()` here has another effect: it transforms decimal numbers to integers. And I am not sure that this is desired or not, or if we should just accept the user inputs as is.

Once I have an answer for that question (sorry if that wasn't clear in the issue), I will make another PR for numbers, if you don't mind.

jnizet:
Moved non-input fields to the top as before, and rebased on master.

wesleycho:
This LGTM.

pkozlowski-opensource:
@jnizet just landed this one, thnx! You can kill the number part now!

PR #554, opened 2016-08-07
fix(typeahead): specify any type for formatters argument

jnizet:
The generated .d.ts file doesn't specify `any` as the type of the formatter argument (`value`) when it is defined as

    inputFormatter: (value) => string;

Declaring it with an explicit `any` type fixes the issue:

    inputFormatter: (value: any) => string;

That is quite strange, because for the attributes of `NgbTypeaheadWindow` which don't have a explicit type declared, like for example

    @Input() results;

The corresponding attribute in the generated .d.ts file does have an explicit `any` type declared:

    results: any;

So this looks like a bug or limitation of the typescript compiler.

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -62,7 +62,7 @@ export class NgbTypeahead implements OnInit,
   /**
    * A function to convert a given value into string to display in the input field
    */
-  @Input() inputFormatter: (value) => string;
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
> So this looks like a bug or limitation of the typescript compiler

Agreed, this looks fishy. Let's get this PR in as the change doesn't hurt.

PR #555, opened 2016-08-08
refactor(typeahead): TS type adjustements

pkozlowski-opensource:
This PR adjusts TS types. It also gets rid of the `ariaLabelledBy` as it is not used (window is a private API and we are not using this input; our users got no way of using it at all).

@maxokorokov PTAL

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -4,7 +4,7 @@ import {NgbHighlight} from './highlight';
 import {toString} from '../util/util';
 
 export type ResultTplCtx = {
-  result: string,
+  result: any,
   term: string,
   formatter: (_: string) => string
 };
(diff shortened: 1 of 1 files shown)


maxokorokov:
I don't get what types should be specified and what shouldn't. 
Ok for immediate initialisation and inferred types, but why leave nothing in `@Input() results;` for instance?

Otherwise looks good to me.

pkozlowski-opensource:
> Ok for immediate initialisation and inferred types, but why leave nothing in @Input() results; for instance?

We should have something for `results` as well! Naive approach could be `any[]` but AFAIK `ngFor` is smarter and can go over any iterable and I don't know how to express this in TS (and even `ngFor` [doesn't know](https://github.com/angular/angular/blob/74b57dfa7d1aafd79ee385b8274537b1aa83a5c2/modules/%40angular/common/src/directives/ng_for.ts#L81)).

So, yeh, for `results` we should have type, I just don't know how to express it. It will default to `any` and we don't need to add this explicitly.

wesleycho:
Just a note, the commit has a typo :) (should be `adjustments`).

The only question I have is why remove the aria stuff? We should probably have a mechanism for uniquely identifying it for accessibility purposes.

Otherwise, this LGTM.

pkozlowski-opensource:
> The only question I have is why remove the aria stuff? We should probably have a mechanism for uniquely identifying it for accessibility purposes.

What we've got today is wrong and not used so no point in keeping it around. We should have extensive accessibility review / fixes as soon as we are done with base functionality.

> Just a note, the commit has a typo :) (should be adjustments).

Will fix, thnx!

PR #559, opened 2016-08-09
fix(buttons): don't interfere with 'active' class

### LINKED ISSUE #557: Buttons: NgbRadioLabel directive is too "agressive"
The `NgbRadioLabel` as implemented today prevents checkbox demo from working correctly. It is implementation is very simple:

```typescript
@Directive({selector: 'label.btn', host: {'[class.active]': 'checked'}})
export class NgbRadioLabel {
  @Input() checked: boolean;
}
```

But if we move to NgModules this directive gets activated on all labels with the `btn` class (quite common scenario). And it can mess up scenarios like this:

 ```html
  <label class="btn btn-primary" [class.active]="model.left">
    <input type="checkbox" [(ngModel)]="model.left"> Left (pre-checked)
  </label>
```

In this scenario ` [class.active]="model.left"` will be overridden by the NgbRadioLabel directive and will never take effect. We can't remove `active` label when the directive is not used / manipulated from outside.

pkozlowski-opensource:
Fixes #557

### DIFF
--- a/src/buttons/radio.spec.ts
+++ b/src/buttons/radio.spec.ts
@@ -27,6 +27,22 @@ function getInput(nativeEl: HTMLElement, idx: number): HTMLInputElement {
   return <HTMLInputElement>nativeEl.querySelectorAll('input')[idx];
 }
 
+describe('NgbActiveLabel', () => {
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({imports: [NgbButtonsModule, FormsModule, ReactiveFormsModule]});
+  }));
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@maxokorokov would you mind reviewing this one?

wesleycho:
LGTM

Foxandxss:
LGTM as well.

maxokorokov:
@pkozlowski-opensource I'm late, but LGTM too

Issue #563, opened 2016-08-10
Tests: migrate to TestBed


pkozlowski-opensource:
`TestComponentBuilder` is deprecated and will be removed in rc.6. We need to migrate to  `TestBed`. Check https://github.com/angular/angular/pull/10621 for more details.

pkozlowski-opensource:
I've just opened a PR that can serve as an example for migration: #566

If anyone from the community would like to help out here please send a single PR per component. Thnx!

sendilkumarn:
I will take components from 
Accordion / Alerts / Buttons / Carousel / Collapse and Dropdown. 

@wesleycho let me know if you already started in any of these.

wesleycho:
I'm recovering from coming back home from Europe (9 hour time difference), so don't wait on me :) .

pkozlowski-opensource:
Components to take care of:

- [x] Accordion
- [ ] Alert 
- [ ] Buttons
- [ ] Carousel 
- [ ] Collapse
- [ ] Dropdown 
- [ ] Modal
- [ ] Pagination 
- [ ] Popover 
- [x] Progressbar
- [ ] Rating 
- [ ] Tab 
- [ ] Timepicker 
- [ ] Tooltip 
- [ ] Typeahead

jnizet:
I started working on typeahead, but it's a big one, and it's time to go to bed now :-)
Bad news: I was hoping to be able to get rid of the `whenStable()` calls, but tests still don't pass without them.

sendilkumarn:
Lets me finish off the remaining three.

pkozlowski-opensource:
@sendilkumarn I'm on tabs atm. Let me push a PR for this as there is an even simpler pattern :-)

sendilkumarn:
Okay then let me wait for that.

pkozlowski-opensource:
@sendilkumarn see #579

sendilkumarn:
got it 👍

sendilkumarn:
Why can't we move `createTestComponent` into `util.ts` file and then use it.

I will also remove `async` from other tests. leave only `fakeAsync`. This will `standardize` the list

wesleycho:
It cannot be moved to util.ts, that is a script of utils bundled into the library itself. We would need to create a separate testUtil.ts or something like that. We should abstract that out, although I think @pkozlowski-opensource mentioned something about having a better pattern in mind.

pkozlowski-opensource:
[...]

PR #566, opened 2016-08-10
test(progressbar): migrate to TestBed

pkozlowski-opensource:
Part of #563


wesleycho:
I like it - I take it this needs to be done for all components?

pkozlowski-opensource:
@wesleycho wait a sec, there is actually a simpler way. Pushing an update.

pkozlowski-opensource:
@wesleycho so it seems that we don't need to call compile components if we use template (vs. templateUrl). As such the whole tests becomes sync and much simpler as the result.

Please take another look.

And yes, it needs to be done for all components - all help would be more than appreciated!

PR #567, opened 2016-08-10
test(accordion): migrate to TestBed

sendilkumarn:
Part of #563

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -1,10 +1,10 @@
-import {inject, async, addProviders} from '@angular/core/testing';
+import {async, TestBed} from '@angular/core/testing';
 
 import {TestComponentBuilder} from '@angular/core/testing';
 
 import {Component} from '@angular/core';
 
-import {NGB_ACCORDION_DIRECTIVES} from './accordion';
+import {NgbAccordionModule} from './index';
 
 function getPanels(element: HTMLElement): HTMLDivElement[] {
   return <HTMLDivElement[]>Array.from(element.querySelectorAll('div .card-header'));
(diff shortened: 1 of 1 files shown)


src/accordion/accordion.spec.ts
```diff
@@ -1,9 +1,10 @@
-import {inject, async, addProviders} from '@angular/core/testing';
+import {async, TestBed} from '@angular/core/testing';
 
 import {TestComponentBuilder} from '@angular/core/testing';
 
 import {Component} from '@angular/core';
 
+import {NgbAccordionModule} from './index';
 import {NGB_ACCORDION_DIRECTIVES} from './accordion';
```
pkozlowski-opensource: `NGB_ACCORDION_DIRECTIVES` is not used any more and could be removed.

src/accordion/accordion.spec.ts
```diff
@@ -27,304 +28,346 @@ function expectOpenPanels(nativeEl: HTMLElement, openPanelsDef: boolean[]) {
...
+        <template ngbPanelContent>{{panel.content}}</template>
+      </ngb-panel>
+    </ngb-accordion>
+    <button *ngFor="let panel of panels" (click)="acc.toggle(panel.id)">Toggle the panel {{ panel.id }}</button>
+  `;
+
+  beforeEach(() => { TestBed.configureTestingModule({declarations: [TestComponent], imports: [NgbAccordionModule]}); });
+
+  it('should have no open panels', async(() => {
+       TestBed.overrideComponent(TestComponent, {set: {template: html}});
```
pkozlowski-opensource: In this particular test suite a template is shared by many (all?) test cases so I think that we could have the `TestBed.overrideComponent(TestComponent, {set: {template: html}});` in the `beforeEach` block only and not repeat it in all the `it` blocks.

src/accordion/accordion.spec.ts
```diff
@@ -27,304 +28,346 @@ function expectOpenPanels(nativeEl: HTMLElement, openPanelsDef: boolean[]) {
...
+
+       tc.activeIds = '';
+       fixture.detectChanges();
+
+       expectOpenPanels(el, [false, false, false]);
+
+       tc.activeIds = 'wrong id,one';
+       fixture.detectChanges();
+
+       expectOpenPanels(el, [true, false, false]);
```
pkozlowski-opensource: Why did you change the new lines in this test? Could you please revert?

pkozlowski-opensource: Same goes for other tests.

pkozlowski-opensource:
LGTM, generally speaking. Left some comments that I would like to see addressed before this one lands, though.

Thnx for helping out!

PR #568, opened 2016-08-10
test(rating): migrate to TestBed

jnizet:
refs #563

I've taken the liberty to adapt the template by extracting a utility function to create the test component fixture. That improves DRYness and make the tests simpler, IMHO.


pkozlowski-opensource:
> I've taken the liberty to adapt the template by extracting a utility function to create the test component fixture. That improves DRYness and make the tests simpler, IMHO.

Good move! I think that we could even extract into a separate util (this would require introduction of the `Type` argument) and shared between tests.

pkozlowski-opensource:
Thnx for helping out @jnizet - this is very, very much appreciated!

PR #573, opened 2016-08-11
test(radio): migrate to testbed

sendilkumarn:
part of #563


src/buttons/radio.spec.ts
```diff
@@ -309,24 +304,25 @@ describe('ngbRadioGroup', () => {
...
+     }));
 });
 
 @Component({selector: 'test-cmp', template: ''})
 class TestComponent {
   form = new FormGroup({control: new FormControl('', Validators.required)});
 
   model;
-  values = ['one', 'two', 'three'];
+  values: any = ['one', 'two', 'three'];
```
wesleycho: Any particular reason for this change?

sendilkumarn: Yeah. 
In one of the test cases values was used as an object

PR #576, opened 2016-08-11
test(typeahead): migrate to TestBed

jnizet:
refs #563

### DIFF
--- a/src/typeahead/highlight.spec.ts
+++ b/src/typeahead/highlight.spec.ts
@@ -1,8 +1,9 @@
-import {inject, async, TestComponentBuilder, ComponentFixture} from '@angular/core/testing';
+import {async, TestBed, ComponentFixture} from '@angular/core/testing';
 
 import {Component} from '@angular/core';
 
 import {NgbHighlight} from './highlight';
+import {NgbTypeaheadModule} from './index';
 
 
 /**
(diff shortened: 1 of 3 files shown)


src/typeahead/highlight.spec.ts
```diff
@@ -28,153 +29,109 @@ function highlightHtml(fixture) {
...
-       const html = '<ngb-highlight result="123" [term]="2" highlightClass="my"></ngb-highlight>';
-       tcb.overrideTemplate(TestComponent, html)
-           .createAsync(TestComponent)
-           .then((fixture: ComponentFixture<TestComponent>) => {
-             fixture.detectChanges();
-             expect(highlightHtml(fixture)).toBe('1<span class="my">2</span>3');
-           });
-     })));
+  beforeEach(() => {
+    TestBed.overrideModule(NgbTypeaheadModule, {set: {exports: [NgbHighlight]}});
```
pkozlowski-opensource: Is this needed?

src/typeahead/typeahead.spec.ts
```diff
@@ -441,7 +384,7 @@ class TestComponent {
   private _objects =
       [{id: 1, value: 'one'}, {id: 10, value: 'one more'}, {id: 2, value: 'two'}, {id: 3, value: 'three'}];
 
-  model = '';
+  model: any = '';
```
pkozlowski-opensource: Why this change? Here the type will be inferred as `string`. Please revert.

jnizet: One of the tests has

    comp.model = {};

That does not compile anymore if the model is a string.

src/typeahead/typeahead.spec.ts
```diff
@@ -51,360 +51,291 @@ function expectWindowResults(element, expectedResults: string[]) {
...
+         fixture.whenStable()
+             .then(() => {
+               expectInputValue(el, 'text');
 
-             comp.model = null;
-             fixture.detectChanges();
-             fixture.whenStable().then(() => {
+               comp.model = null;
+               fixture.detectChanges();
+               return fixture.whenStable();
```
pkozlowski-opensource: This is good stuff :+1:

src/typeahead/highlight.spec.ts
jnizet: Yes, otherwise all the tests fail. My understanding is that the test module is just like any regular module: it imports NgbTypeaheadModule, but NgbHighlight is not exported by NgbTypeaheadModule, so it's an internal component of NgbTypeaheadModule, that may not be used outside of NgbTypeaheadModule, including in the test module. 

Overriding the exports of NgbTypeaheadModule allows exporting the internal NgbHighlight component (i.e. do as if it was exported), and thus using it in the test module.

src/typeahead/typeahead.spec.ts
jnizet: I had a hard time understanding promise chaining with angular 1. Now is the time to capitalize. I couldn't resist flattening the pyramid of callbacks :-)

pkozlowski-opensource:
Merged, thnx!

PR #581, opened 2016-08-11
test(timepicker): migrate to testbed

sendilkumarn:
part of #563

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -1,6 +1,6 @@
 import {Component} from '@angular/core';
 
-import {inject, async, TestBed, TestComponentBuilder} from '@angular/core/testing';
+import {TestBed} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
 
 import {Validators, FormControl, FormGroup, FormsModule, ReactiveFormsModule} from '@angular/forms';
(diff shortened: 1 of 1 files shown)


src/timepicker/timepicker.spec.ts
```diff
@@ -45,722 +45,699 @@ function expectToDisplayTime(el: HTMLElement, time: string) {
...
-           });
-         });
-       })));
+    it('should update inputs value on model change', () => {
+      const html = `<ngb-timepicker [ngModel]="model"></ngb-timepicker>`;
+
+      const fixture = createTestComponent(html);
+      fixture.componentInstance.model = {hour: 13, minute: 30};
+      fixture.detectChanges();
+      fixture.whenStable().then(() => {
```
jnizet: This pyramid of callbacks could be flattened with promise chaining. I did it in the typeahead test. That makes things easier to read IMHO, and reduces the indentation level.

```
    fixture.whenStable().then(() => {
        ...
        return fixture.whenStable();
    }).then(() => {
        ...
        return fixture.whenStable();
    }).then(() => {
        ...
    });
```

pkozlowski-opensource: Yep. But let's do it in a separate PR. It is not directly related to `TestBed`

Issue #541, opened 2016-08-04
Adjust visibility of fields / methods accessed by templates


pkozlowski-opensource:
As of today we are accessing fields / methods in templates where those fields / methods are marked as `private`. This won't work with offline compilation which means that we need to make sure that all fields / methods accessed from templates are not private

pkozlowski-opensource:
Here are some more details on what need to be done for each and every component:
* [ ] - make sure that all fields accessed from templates are public and not prefixed with `_`
* [ ] - make sure that all methods called from templates are public and not prefixed with `_`
* [ ] - if a directive is not exported all public methods called from a template need to be annotated with `@internal` in JSDoc
* [ ] - if a directive is exported all public methods not meant to be part of the public API need to be made private or annotated with `@internal` in JSDoc if used from templates.

At the end of the day the only private fields / methods need to be the ones that are _not_ accessed in a template.

Foxandxss:
I can start by removing the underscores everywhere.

pkozlowski-opensource:
> I can start by removing the underscores everywhere.

IMO it is better to do it on the widget-per-widget basis. Easier to review. I'm just sending an example PR for a carousel.

pkozlowski-opensource:
#565 is an example of this for a carousel. Would be great if people could help out with other widgets.

jnizet:
I'm on typeahead

jnizet:
tooltip looks fine to me. I'm on timepicker.

pkozlowski-opensource:
Components to take care of:

- [ ] Accordion
- [ ] Alert 
- [ ] Buttons
- [ ] Carousel 
- [ ] Collapse
- [x] Dropdown 
- [ ] Modal
- [ ] Pagination 
- [ ] Popover 
- [ ] Progressbar
- [ ] Rating 
- [ ] Tab 
- [ ] Timepicker 
- [ ] Tooltip 
- [x] Typeahead

pkozlowski-opensource:
I believe that this one is fixed now. Thnx for all the awesome help @jnizet and @chenyuzhcy !

PR #582, opened 2016-08-11
refactor(typeahead): use public members in a template

jnizet:
part of #541

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -1,5 +1,4 @@
 import {Component, Input, Output, EventEmitter, TemplateRef} from '@angular/core';
-import {NgbHighlight} from './highlight';
 
 import {toString} from '../util/util';
 
@@ -18,7 +17,7 @@ export type ResultTplCtx = {
       <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>
     </template>
     <template ngFor [ngForOf]="results" let-result let-idx="index">
-      <button class="dropdown-item" [class.active]="idx === _activeIdx" 
+      <button class="dropdown-item" [class.active]="idx === activeIdx" 
         (mouseenter)="markActive(idx)" 
         (click)="select(result)">
           <template [ngTemplateOutlet]="resultTemplate || rt" 
@@ -28,7 +27,7 @@ export type ResultTplCtx = {
   `
 })
 export class NgbTypeaheadWindow {
-  private _activeIdx = 0;
+  activeIdx = 0;
 
   /**
    * Typeahead match results to be displayed
@@ -41,7 +40,7 @@ export class NgbTypeaheadWindow {
   @Input() term: string;
 
   /**
-   * A function used to format a given result before display. This function should return a formated string without any
+   * A function used to format a given result before display. This function should return a formatted string without any
    * HTML markup.
    */
   @Input() formatter = toString;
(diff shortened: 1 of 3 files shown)


src/typeahead/typeahead.ts
```diff
@@ -57,7 +57,8 @@ export class NgbTypeahead implements OnInit,
   private _popupService: PopupService<NgbTypeaheadWindow>;
   private _positioning = new Positioning();
   private _valueChanges = new Subject<string>();
-  private _windowRef: ComponentRef<NgbTypeaheadWindow>;
+
+  windowRef: ComponentRef<NgbTypeaheadWindow>;
```
pkozlowski-opensource: Does it need to be public? It is not accessed from a template, is it?

pkozlowski-opensource: Ah, it is used in host... Actually I'm not sure what the rule for the `host` should be :-)

jnizet: Hah! I just blindly followed your carousel example, to be honest. You made keyPrev() and keyNext() public, and they're also only used in host :-)

src/typeahead/typeahead-window.ts
```diff
@@ -56,13 +56,13 @@ export class NgbTypeaheadWindow {
    */
   @Output('select') selectEvent = new EventEmitter();
 
-  getActive() { return this.results[this._activeIdx]; }
+  getActive() { return this.results[this.activeIdx]; }
 
-  markActive(_activeIdx: number) { this._activeIdx = _activeIdx; }
+  markActive(activeIdx: number) { this.activeIdx = activeIdx; }
```
jnizet: Wait a moment. This directive is not exported, so this method should be `@internal`, because it's called from the template, right? And same for `select()`, right?

jnizet: Actually, I think I understood "exported" wrong. Do you mean "exported from the NgModule", "exported using an exportAs attribute in the component", or both? In fact I concentrated on the two first rules, and I have a hard time understanding the rationale and the end goal behind two last ones:

- if a directive is not exported all public methods called from a template need to be annotated with @internal in JSDoc
- if a directive is exported all public methods not meant to be part of the public API need to be made private or annotated with @internal in JSDoc if used from templates.

pkozlowski-opensource: If a given component is not exported from our module we don't care about `@internal` too much since we will never have API docs for those components. But still good to have it.

If a given component doesn't have `exportAs` we don't care much eather as for those components don't have methods description as part of their API. So in this sense every method called from a template should be marked as `@internal`. We don't care too much today, but will be good to have things clean if we decide to export a given component one day.

Hope it clarifies.

jnizet: Yes, somewhat, thanks. 

My main doubt is with `next()`, `prev()` and `getActive()` in typeahead-window. 

As you said, this component is not exported from the module, so it doesn't matter much if these methods are internal or not.

These methods, which are not called from the template, exist and are public because typeahead needs to call them. Not because they're meant to be part of the public external API of the component (at least not now). So I would tend to make the internal, too, even though they're not called from the template. 

And in fact, the same goes with all the methods implementing Angular interfaces: they're not called from the template either, and are not supposed to be called by anything other than the framework, so these should be internal, too, in my mind.

Anyway, I made `markActive()` and `select()` internal, now, so these changes should follow the rules.

Tell me if anything else must be changed.

src/typeahead/typeahead.ts
pkozlowski-opensource: Maybe we should introduce a public method, sth like `isOpen()`, `hasPopup()` or something? Marking `WindowRef` makes me nervous...

jnizet: Sure, no problem. Will do.

Foxandxss:
LGTM, but are we only removing the underscore on what goes to the template?

jnizet:
@Foxandxss The issue was not about removing underscores from private fields and methods. It was about making sure all fields and methods used from the template were public. Making them public implies removing the underscore (which is a marker for private). But the fields that stay private keep their underscores.

I'm not a big fan of underscore for private, but that is the convention here, and I don't think there was a desire to change it.

Foxandxss:
It will change, bu that is a different topic.

PR #583, opened 2016-08-11
test(all): standardizing spec files

sendilkumarn:
currently `createTestComponent` gets two args.

They will be helpful when we move this function to `testUtils`

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -1,6 +1,4 @@
-import {async, TestBed} from '@angular/core/testing';
-
-import {TestComponentBuilder} from '@angular/core/testing';
+import {TestBed, ComponentFixture} from '@angular/core/testing';
 
 import {Component} from '@angular/core';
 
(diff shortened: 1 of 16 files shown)


sendilkumarn:
#563

src/util/util.ts
```diff
@@ -1,3 +1,5 @@
+import {ComponentFixture, TestBed} from '@angular/core/testing';
```
pkozlowski-opensource: Wait, this is clearly wrong.

src/accordion/accordion.spec.ts
```diff
@@ -26,6 +24,13 @@ function expectOpenPanels(nativeEl: HTMLElement, openPanelsDef: boolean[]) {
   expect(result).toEqual(openPanelsDef);
 }
 
+function createTestComponent(html: string, TestComponent: any): ComponentFixture<any> {
```
pkozlowski-opensource: We can't just return `ComponentFixture<any>` as we would loose types.

src/util/util.ts
sendilkumarn: oops I will remove it. Initially I had it in that util file. 

My bad.

src/accordion/accordion.spec.ts
pkozlowski-opensource: The proper signature should probably be sth like `createTestComponent<T>(html: string, TestComponent: T): ComponentFixture<T>` but this would require that we invoke it like `createTestComponent<TestComponent>(html, TestComponent);` which is mouthful....

pkozlowski-opensource:
@sendilkumarn I don't think extracting helper function will buy us much as of today... I think that proper solution is sth like this: https://github.com/angular/angular/issues/10241

Let's put the helper extraction on a back burner for now. Could you change this PR to just remove `async()`?

src/accordion/accordion.spec.ts
sendilkumarn: This would not fit in `TestBed.overrideComponent` as it requires `concreteType<any>`

sendilkumarn:
Okay then.

sendilkumarn:
> proper solution is sth like this: angular/angular#10241

@pkozlowski-opensource  that makes sense and more logical. we will wait for that

PR #586, opened 2016-08-11
feat(tabset): add two-way binding with activeId

Quinn-L:

This also fixes the demo 'Select an active tab by id' where the bound id would not be in-sync if the user clicked on the tab.

### DIFF
--- a/demo/src/app/components/tabset/demos/selectbyid/tabset-selectbyid.html
+++ b/demo/src/app/components/tabset/demos/selectbyid/tabset-selectbyid.html
@@ -1,4 +1,4 @@
-<ngb-tabset [activeId]="selectedId">
+<ngb-tabset [(activeId)]="selectedId">
   <ngb-tab title="Simple" id="foo">
     <template ngbTabContent>
       <p>Raw denim you probably haven't heard of them jean shorts Austin. Nesciunt tofu stumptown aliqua, retro synth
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
Hmm, but we've got already the 'change' `@Output`: https://ng-bootstrap.github.io/#/components/tabs

How is this proposal different from what exists? Is it only about the `[(activeId)]` syntax? For sure we shouldn't support both of those events...

Quinn-L:
The two-way binding solves the activeId in-sync, otherwise you're expecting users to manually update this themselves in every place they use this. In the demo page, click the 'Fancy title' tab with the mouse, then click 'Toggle selected tab' (misspelled toogle). Nothing seems to happens but that's because when you click 'Fancy title' activeId doesn't get updated.

To me, the change output is only really for the Prevent tab change functionality, ie. stop the change before it actually happens.

jnizet:
I *think* @Quinn-L has a point on this one. In fact, I would maybe even go one step further, and make the select method internal. Here are my thoughts:

- the activeId of the tabset and the activeId in the caller model should be easy to keep in sync
- the select() method is useful internally, but the user doesn't really need it: he/she can simply change the value of `activeId` in its model
- just setting `activeId` won't prevent him/her from making a disabled tab active, but since he/she is doing it programmatically, that might be his/her choice, and he/she can prevent it by him/herself anyway. By the way, I don't think anything prevents the user from initially selecting a disabled tab. 
- the current change event is useful to prevent navigation to a tab (although I wonder why not just disable it), but it would be useful to have an event emitted when the change has effectively been done (I remember having to do that to reflow charts with angular 1). Currently, a subscriber to the change event has no guarantee that another subscriber won't prevent the event right after.

Quinn-L:
Feature on backburner unless more users justify this use case.

PR #587, opened 2016-08-12
refactor(dropdown): use public members in template

sendilkumarn:
part of #541

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -63,8 +63,8 @@ describe('ngb-dropdown', () => {
 
   it('should allow toggling dropdown from outside', () => {
     const html = `
-      <button (click)="drop.open(); $event.stopPropagation()">Open</button>
-      <button (click)="drop.close(); $event.stopPropagation()">Close</button>
+      <button (click)="drop.openDropdown(); $event.stopPropagation()">Open</button>
+      <button (click)="drop.closeDropdown(); $event.stopPropagation()">Close</button>
       <button (click)="drop.toggle(); $event.stopPropagation()">Toggle</button>
       <div ngbDropdown #drop="ngbDropdown"></div>`;
 
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -8,9 +8,9 @@ import {Directive, Input, Output, HostListener, EventEmitter} from '@angular/cor
   exportAs: 'ngbDropdown',
   host: {
     'class': 'dropdown',
-    '[class.open]': '_open',
-    '(keyup.esc)': '_closeFromOutside()',
-    '(document:click)': '_closeFromOutside()'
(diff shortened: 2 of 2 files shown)


src/dropdown/dropdown.spec.ts
```diff
@@ -63,8 +63,8 @@ describe('ngb-dropdown', () => {
 
   it('should allow toggling dropdown from outside', () => {
     const html = `
-      <button (click)="drop.open(); $event.stopPropagation()">Open</button>
-      <button (click)="drop.close(); $event.stopPropagation()">Close</button>
+      <button (click)="drop.openDropdown(); $event.stopPropagation()">Open</button>
+      <button (click)="drop.closeDropdown(); $event.stopPropagation()">Close</button>
```
pkozlowski-opensource: Those renames would break the existing public API and make it more wordy. See my alternative proposal in the other comment.

src/dropdown/dropdown.ts
```diff
@@ -8,9 +8,9 @@ import {Directive, Input, Output, HostListener, EventEmitter} from '@angular/cor
   exportAs: 'ngbDropdown',
   host: {
     'class': 'dropdown',
-    '[class.open]': '_open',
```
pkozlowski-opensource: Could we use `isOpen()` here? This way we can keep `_open` as private.

src/dropdown/dropdown.ts
```diff
@@ -22,7 +22,7 @@ export class NgbDropdown {
   /**
    *  Defines whether or not the dropdown-menu is open initially.
    */
-  @Input('open') private _open = false;
```
pkozlowski-opensource: Revert, see above.

src/dropdown/dropdown.ts
```diff
@@ -34,21 +34,21 @@ export class NgbDropdown {
...
    */
-  isOpen() { return this._open; }
+  isOpen() { return this.open; }
 
   /**
    * Opens the dropdown menu of a given navbar or tabbed navigation.
    */
-  open() {
-    this._open = true;
+  openDropdown() {
```
pkozlowski-opensource: Revert the rename, see above.

src/dropdown/dropdown.ts
```diff
@@ -34,21 +34,21 @@ export class NgbDropdown {
...
+    this.open = true;
     this.openChange.emit(true);
   }
 
   /**
    * Closes the dropdown menu of a given navbar or tabbed navigation.
    */
-  close() {
-    this._open = false;
+  closeDropdown() {
```
pkozlowski-opensource: Revert the rename, see above.

sendilkumarn:
This can be closed now. I think it was handled.

pkozlowski-opensource:
> This can be closed now. I think it was handled.

Woops, actually no. I've sent #653 to finish it off

PR #588, opened 2016-08-12
docs(tabset): make the doc of `select()` clearer

jnizet:
The name of the argument, `tabIdx`, lead the reader to think that what must be passed is the index of the tab, instead of its unique id. Renaming to `tabId` and indicating that the id of the tab is what is expected by the method makes it clearer.

The same confusion is also fixed in the tests.


pkozlowski-opensource:
Thnx. This was a left-over from the times where we've used real indexes. Good catch!

PR #597, opened 2016-08-15
refactor(buttons): use public members in a template

chenyuzhcy:
part of #541


pkozlowski-opensource:
Merged. Thnx for helping out!

PR #600, opened 2016-08-16
chore: switch from es6-shim to core-js

mlaval:
Core-js happens to be more performant in IE browsers.

See also https://github.com/angular/angular/issues/8112


Foxandxss:
I don't mind, we swapped that in angular.io as well.

wesleycho:
Seems like everywhere is moving to use core.js - I approve

PR #604, opened 2016-08-17
feat: support IE9

mlaval:
This PR adds support for IE9,. which doesn't require that many changes.
Some more details about what was needed to make the demo app to work and the full test suite to pass:

**Source:**
 - Time picker: in IE9, a disabled `<fieldset>` doesn't disable children.

**Demo:**
 - Added classList polyfill to the demo website.

**Testing:**
 - Added Angular 2's testing polyfills to Karma's setup.
 - Increased Karma's timers for slow IE9.

**Spec:**
 - Pagination: in IE9, with the polyfill, classList need to be retrieved each time.
 - Progress bar: no `value` attribute in IE9 (not sure why attribute was used).
 - Radio group: some race condition was happening, making this test to fail. Note: it was occuring also in Chrome when focusing this particular test suite, but not when running the full campaign.
 - Typeahead: IE9 requires a different way to fire events.


maxokorokov:
Actually instead of adding classlist polyfill I would throw it away altogether and stick to more usual `expect(...).toHaveCssClass('...')`. I think it was one of the first tests I wrote and forgot to fix that...

pkozlowski-opensource:
@maxokorokov I think that we need a polyfill for the `core` of Angular itself (ex. `[class.foo]` is using `classList` behind the scenes). But yes, we could / should change this test for sure.

karma.conf.js
```diff
@@ -38,6 +39,10 @@ module.exports = function(config) {
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: ['Chrome'],
-    singleRun: false
+    singleRun: false,
+    captureTimeout: 60000,
+    browserDisconnectTimeout : 60000,
+    browserDisconnectTolerance : 3,
+    browserNoActivityTimeout : 60000
```
wesleycho: Super minor, but probably should remove the spaces before `:` here

src/progressbar/progressbar.spec.ts
```diff
@@ -87,12 +87,12 @@ describe('ng-progressbar', () => {
       // this might fail in IE11 if attribute binding order is not respected for the <progress> element:
       // <progress [value]="" [max]=""> will fail with value = 1
       // <progress [max]="" [value]=""> will work with value = 10
```
wesleycho: Is this comment still relevant if we are using the `value` property on the element?

pkozlowski-opensource: Yes, I believe that this is still relevant. But we should add SL tests with IE* to be sure :-)

karma.conf.js
mlaval: Fixed

Issue #3, opened 2015-09-15
Modal


BobbieBarker:
Recreate BS4 modal with ng2 component.

wesleycho:
Some notes:

- API
  - Investigate observable-based modal to allow cancellation
  - Suggestion made to be more declarative with a component vs. imperative by passing objects into a service method
    - @wesleycho: Personally, I feel like this would result in an API that is too rigid, and this can be made possible on the user's side with a custom modal component consuming this service. Any points/counterpoints welcome on this

pkozlowski-opensource:
@wesleycho are you working / planning to work actively on this one? I've started to look into what is possible so can either take it over or bounce off ideas.

Foxandxss:
No one is really having anything assigned. Except the stuff we already have as PR, so feel free on bounce ideas. At the end you are the one who knows more :P

pkozlowski-opensource:
> At the end you are the one who knows more :P

@Foxandxss this is _very_ dangerous assumption, don't make this mistake :-)

Foxandxss:
No, we are counting on that, don't let us down.

wesleycho:
I've done looking into this before, but there is an unfortunate thing here - the DynamicComponentLoader kind of sucks to use. It feels like it's underpowered.

We should perhaps talk about what we need the modal to do.

The modal needs to ultimately slot the modal dialog into a container of a user's choice. Angular Material just decided to go the single insertion way (which may be best initially) by forcing the user to use a container element in the template, and then used a service to insert it into the container while storing a ref to the backdrop and modal in the service after inserting via DCL. On destruction, it then gets cleaned up.

I'm not the biggest fan of putting a container right in the template, but it seems to be necessary due to how the DCL works. The DCL seems necessary for this sort of problem, unless we want to go the declarative route and force the user to declare their modal right in the template.

I personally prefer imperative declaration of components such as modals through the usage of a service, but the framework seems to make declarative usages easier. However, if we go the declarative route, we then also run into the same problem as alerts with self destruction & reseting of state.

pkozlowski-opensource:
@wesleycho I hear what you are saying. I'm playing with something right now that might work for the "service" usage without too much pain for the end users. Will shore tomorrow or day after tomorrow. 

But yeh, DCL needs more love, Misko was showing me some ideas for future evolution. Wish try to push this thing forward.

pkozlowski-opensource:
[...]

PR #609, opened 2016-08-18
feat(modal): add initial version of the modal service

pkozlowski-opensource:
Very rough implementation of the `NgbModal` service. Opening this PR early on so we can discuss the details. Check discussion in https://github.com/ng-bootstrap/ng-bootstrap/issues/3#issuecomment-240734294 for more details / remaining work. 

Even if there is still more work that needs to be done I believe that this is the first good shoot witch should have stable API (I mean, remaining work shouldn't have non-backward compatible API changes). As such I think that we can go ahead and land it so users can play with it. 

Please review.


src/index.ts
```diff
@@ -16,13 +17,14 @@ import {NgbTooltipModule} from './tooltip';
 import {NgbTypeaheadModule} from './typeahead';
 
 export {NgbPanelChangeEvent} from './accordion';
+export {NgbModal} from './modal';
```
jnizet: Shouldn't NgbModalRef and NgbModalOptions be exported too? They're the return type and the argument type of `NgbModal.open()`, which is the public API, so client code should be able to refer to these types, too, shouldn't they?

pkozlowski-opensource: They should, I thought I've exported them. This is still WIP so thx for catching this one!

src/modal/modal_container.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  private _applyWindowOptions(windowInstance: NgbModalWindow, options: Object) {
+    ['backdrop', 'keyboard', 'size'].forEach((optionName: string) => {
+      if (isDefined(options[optionName])) {
+        windowInstance[optionName] = options[optionName];
+      }
+    });
+  }
+
+  private _getContentNodes(content: string | TemplateRef<any>, context: ModalContentContext) {
```
wesleycho: What type of object is this returning?

pkozlowski-opensource: An array of DOM nodes. Will add return type.

src/modal/modal_ref.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (this._windowCmptRef) {
+      this.reject(reason);
+      this.removeModalElements();
+    }
+  }
+
+  private removeModalElements() {
+    this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._windowCmptRef.hostView));
+    if (this._backdropCmptRef) {
+      this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._backdropCmptRef.hostView));
```
wesleycho: Are we creating separate backdrops for each modal when working with stacked modals? This differs from the UI Bootstrap behavior, although we have had a feature request in UI Bootstrap to create separate backdrops. I'd be fine with that, just want to verify intention here.

wesleycho:
This LGTM - it's possible I missed something, so one other person reviewing probably would be a good idea.

src/modal/modal_ref.ts
pkozlowski-opensource: This is only temporally here till we add support for stacked modals (if we decide to do so!). But yes, I agree that we will have to move this logic if we decide to support stacked modals.

demo/src/app/components/modal/demos/basic/modal-basic.ts
```diff
@@ -0,0 +1,21 @@
...
+export class NgbdModalBasic {
+  closeResult: string;
+
+  constructor(private modalService: NgbModal) {}
+
+  open(content) {
+    this.modalService.open(content).result.then((result) => {
+      this.closeResult = `Closed with: ${result}`;
+    }, (reason) => {
+      this.closeResult = `Dismissed with: ${reason}`;
```
maxokorokov: escape and backdrop click return `1` and `0` as dismiss reasons which are `NgbDismissReasons` values. But it looks strange in the demo → should switch over enum values and display text?

src/modal/modal.module.ts
```diff
@@ -1,8 +1,19 @@
...
 
-@NgModule({declarations: [NgbModalBackdrop, NgbModalWindow], entryComponents: [NgbModalBackdrop, NgbModalWindow]})
+export {NgbModal, NgbModalOptions} from './modal';
+export {NgbModalRef} from './modal_ref';
+
+@NgModule({
+  declarations: [NgbModalContainer, NgbModalBackdrop, NgbModalWindow],
+  entryComponents: [NgbModalBackdrop, NgbModalWindow],
+  exports: [NgbModalContainer],
+  providers: [NgbModal, NgbModalStack]
```
maxokorokov: Do we need `NgbModalStack` here now?

src/modal/modal.ts
```diff
@@ -0,0 +1,26 @@
...
+  size?: 'sm' | 'lg';
+}
+
+/**
+ * A service to open modal windows. Creating modals is straightforward: create a template and pass it as an argument to
+ * the "open" method!
+ */
+@Injectable()
+export class NgbModal {
+  constructor(private modalStack: NgbModalStack) {}
```
maxokorokov: _modalStack

src/modal/modal.module.ts
```diff
@@ -1,8 +1,19 @@
 import {NgModule} from '@angular/core';
 
+import {NgbModalContainer} from './modal_container';
```
maxokorokov: elswere in the project all the files are named with `-`, not `_`, ex. `ngb-time.ts`, `ngb-typeahead-window.ts`

src/modal/modal.ts
```diff
@@ -0,0 +1,26 @@
+import {Injectable, TemplateRef} from '@angular/core';
+
+import {NgbModalStack} from './modal_stack';
+import {NgbModalRef} from './modal_ref';
+
+export interface NgbModalOptions {
+  backdrop?: boolean | 'static';
+  keyboard?: boolean;
+  size?: 'sm' | 'lg';
```
maxokorokov: The above is not documented neither here, nor in the `NgbModalWindow` → no documentation on the demo site

src/modal/modal_container.ts
```diff
@@ -0,0 +1,74 @@
...
+    const windowCmptRef = this._viewContainerRef.createComponent(this._windowFactory, 0, this._injector, nodes);
+    let backdropCmptRef;
+    let ngbModalRef;
+
+    if (options.backdrop !== false) {
+      backdropCmptRef = this._viewContainerRef.createComponent(this._backdropFactory, 0, this._injector);
+    }
+    ngbModalRef = new NgbModalRef(this._viewContainerRef, windowCmptRef, backdropCmptRef);
+
+    modalContentContext.close = (reasult: any) => { ngbModalRef.close(reasult); };
```
maxokorokov: typo: result

src/modal/modal_ref.ts
```diff
@@ -0,0 +1,45 @@
+import {ComponentRef, ViewContainerRef} from '@angular/core';
+
+import {NgbModalBackdrop} from './modal_backdrop';
+import {NgbModalWindow} from './modal_window';
+
+export class NgbModalRef {
+  private resolve: (result?: any) => void;
+  private reject: (reason?: any) => void;
```
maxokorokov: _reject, _resolve

src/modal/modal_ref.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  dismiss(reason?: any) {
+    if (this._windowCmptRef) {
+      this.reject(reason);
+      this.removeModalElements();
+    }
+  }
+
+  private removeModalElements() {
```
maxokorokov: _removeModalElements

src/modal/modal_ref.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class NgbModalRef {
+  private resolve: (result?: any) => void;
+  private reject: (reason?: any) => void;
+  result: Promise<any>;
+
+  constructor(
+      private _viewContainerRef: ViewContainerRef, private _windowCmptRef: ComponentRef<NgbModalWindow>,
+      private _backdropCmptRef?: ComponentRef<NgbModalBackdrop>) {
+    _windowCmptRef.instance.dismissEvent.subscribe((reason: any) => { this.dismiss(reason); });
```
maxokorokov: not sure if need to unsubscribe or not...

src/modal/modal_stack.ts
```diff
@@ -0,0 +1,19 @@
+import {Injectable, TemplateRef} from '@angular/core';
+
+import {NgbModalRef} from './modal_ref';
+
+@Injectable()
+export class NgbModalStack {
+  private modalContainer: any;
```
maxokorokov: `NgbModalContainer`?

maxokorokov:
LGTM, some comments in the code plus:

- had a feeling that `any` vs concrete type usage looks a bit messy in general, feel free to ignore :)
- dynamic component creation looks like a lot of boilerplate and similar to `PopupService`, but what can you do with it right now?...

src/modal/modal.module.ts
pkozlowski-opensource: Yes, as I need to share the same instance between `NgbModal` and `NgbModalContainer`

demo/src/app/components/modal/demos/basic/modal-basic.ts
pkozlowski-opensource: changed

src/modal/modal.module.ts
pkozlowski-opensource: renamed

src/modal/modal.ts
pkozlowski-opensource: Yes, I know. We need to update the API extraction script for this.

src/modal/modal_stack.ts
pkozlowski-opensource: I didn't do it initially since it is creating circular deps between imports. But apparently this is no causing any harm so we've got proper types now.

src/modal/modal-container.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    ngbModalStack.registerContainer(this);
+  }
+
+  open(content: string | TemplateRef<any>, options): NgbModalRef {
+    const modalContentContext = new ModalContentContext();
+    const nodes = this._getContentNodes(content, modalContentContext);
+    const windowCmptRef = this._viewContainerRef.createComponent(this._windowFactory, 0, this._injector, nodes);
+    let backdropCmptRef;
+    let ngbModalRef;
```
Foxandxss: Can't we type this var and the previous one?

Foxandxss:
Nothing against, minus a minimal nitpick.

Issue #613, opened 2016-08-19
Typeahead: 4 unit tests failing in Safari 9


mlaval:
When running the full campaign, 4 unit tests are failing in Safari 9:
```
Safari 9.0.0 (Mac OS X 10.11.0) ngb-typeahead window should be opened when there are results FAILED
	Expected 4 to equal 2.
	expectResults
	expectWindowResults
	Expected '' to be 'one'.
	stack@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1577:26
	buildExpectationResult@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1547:19
	expectationResultFactory@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:638:40
	addExpectationResult@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:330:58
	addExpectationResult@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:588:41
	node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1501:32
	attemptSync@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1886:28
	run@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1874:20
	execute@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1859:13
	queueRunnerFactory@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:697:42
	execute@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:359:28
	fn@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:2479:44
	attemptAsync@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1916:28
	run@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1871:21
	node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1898:19
	node_modules/jasmine-core/lib/jasmine-core/jasmine.js:1842:11
	complete@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:371:19
Safari 9.0.0 (Mac OS X 10.11.0) ngb-typeahead window should select the result on click, close window and fill the input FAILED
	Expected 4 to equal 2.
	expectResults
	expectWindowResults
Safari 9.0.0 (Mac OS X 10.11.0) ngb-typeahead window should select the result on ENTER, close window and fill the input FAILED
[...]

PR #614, opened 2016-08-19
chore(CI): set up Saucelabs

mlaval:
This is the initial setup.

Note that Safari 9 is disabled due to https://github.com/ng-bootstrap/ng-bootstrap/issues/613


pkozlowski-opensource:
Awesome sauce @mlaval - Thnx!

PR #616, opened 2016-08-20
refactor(test): make component creation common for tests

maxokorokov:
Moving common component creation code into a generic function:

- new `util/tests.ts` for common test code (unless anybody suggests a better place)
- generic `createGenericTestComponent(html, componentType)` function
- input order harmonisation for tests


Foxandxss:
Sounds good.

PR #617, opened 2016-08-22
chore: extract API docs from services

pkozlowski-opensource:
Needed for the Modal service.


jnizet:
Shouldn't it also document the return type, if any (and modal.open() has one)?

pkozlowski-opensource:
> Shouldn't it also document the return type, if any (and modal.open() has one)?

Yeh, we probably should. Added return type extraction as well. We will have to figure out how to display it but it is a separate issue...

Foxandxss:
Sounds good to me.

Issue #20, opened 2015-09-22
Datepicker


wesleycho:
A clean & simpler datepicker

maxokorokov:
Hey guys,

I've spent a couple of days playing with datepicker and it starts to look promising, at least to me :)

Here are some things we could start with:
- A small [design doc](https://docs.google.com/document/d/1bqa2jmn0o0R85pnol2cJ_DbO48uO61bQTaAHl8LTRLE/edit?usp=sharing)
- And a [plunker](http://plnkr.co/edit/FnHbFOssFw2NeZDDsi74?p=preview) to play with (please see the doc first, it's just a proof of concept and far from working good)

Notes:
- There is no proper selection and ngModel integration yet, but I suppose it to be the same as timepicker
- Only days view is in the example

Please let me know what you think!

PR #618, opened 2016-08-22
feat(datepicker): initial version of datepicker component

maxokorokov:
First datepicker version (WIP)

- see the [design doc](https://docs.google.com/document/d/1bqa2jmn0o0R85pnol2cJ_DbO48uO61bQTaAHl8LTRLE/edit?usp=sharing)
- play with [plunker](http://plnkr.co/edit/FnHbFOssFw2NeZDDsi74?p=preview)

Is inline datepicker, not in a popup and only the month view (this will easily extended later)

Current checklist:
- [x] finish tests
- [x] decide on and cleanup CSS/HTML for day and navigation templates
- [x] extract i18n from the calendar into separate class (+ maybe use native APIs?)
- [x] finish ngModel integration (now works only one way)
- [x] add documentation
- [x] at least one demo to start with
- [x] hide exposed NgbDate from the user
- [x] replace default navigation with month/year select boxes

Please leave feedback.

WIP for #20


src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,85 @@
+import {NgbDate} from './ngb-date';
+
+export interface NgbCalendar {
+  getDaysPerWeek(): number;
+  getWeeksPerMonth(): number;
+  getFirstDayOfWeek(): number;
```
wesleycho: Would it make more sense to make these three functions getters?

src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,85 @@
...
+  getWeekDayName(date: NgbDate): string;
+  getMonthName(date: NgbDate): string;
+
+  getNextDate(date: NgbDate): NgbDate;
+  getPrevDate(date: NgbDate): NgbDate;
+
+  getNextMonth(date: NgbDate): NgbDate;
+  getPrevMonth(date: NgbDate): NgbDate;
+
+  getToday(): NgbDate;
```
wesleycho: Ditto about a getter

src/datepicker/ngb-calendar.ts
maxokorokov: Hmm, I don't think you can have getters in interfaces in Typescript

src/datepicker/datepicker-month-view.ts
```diff
@@ -0,0 +1,54 @@
...
+    .weekday {
+      text-align: center;
+      font-weight: bold;
+      width: 30px;
+      height: 30px;
+    }
+    .weeknumber {
+      text-align: center;
+      font-size: 12px;
+      color: #444;
```
pkozlowski-opensource: This would be no-go as it would kill custom themes. Sizes expressed in pixels will also kill zoom in / out no? 

More generally we should try to limit custom styling here as users won't have any way of overriding / customizing it. As discussed one way would be to use buttons. 

This part needs more investigation / work.

src/datepicker/datepicker-month-view.ts
```diff
@@ -0,0 +1,54 @@
+import {Component, Input, TemplateRef, Output, EventEmitter} from '@angular/core';
+import {MonthViewModel, DayViewModel, DayTemplateContext} from './datepicker-view-model';
+
+@Component({
+  selector: 'tbody[ngb-datepicker-month-view]',
```
pkozlowski-opensource: Change to `ngbDatepickerMonthView`. Do we care that it is on `tbody`? Maybe we should relax this constraint so we leave open doors to supporting different markup?

src/datepicker/datepicker-navigation.ts
```diff
@@ -0,0 +1,32 @@
+import {Component, Input, Output, EventEmitter} from '@angular/core';
+import {NavigationEvent} from './datepicker-view-model';
+
+@Component({
+  selector: 'tbody[ngb-datepicker-navigation]',
```
pkozlowski-opensource: Change to `ngbDatepickerNavigation`. Do we care that it is on `tbody`? Maybe we should relax this constraint so we leave open doors to supporting different markup?

src/datepicker/datepicker-navigation.ts
```diff
@@ -0,0 +1,32 @@
...
+  styles: [`
+    td {
+      text-align: center;
```
pkozlowski-opensource: AFAIK there is a Bootstrap utlity CSS class for this, might be a good idea to use it instead?

src/datepicker/datepicker.ts
```diff
@@ -0,0 +1,153 @@
...
+    .day.selected {
+      background-color: blue;
+      color: white;
+    }
+    .day.selected:hover {
+      background-color: blue;
+      color: white;
+    }
+    .day.greyed {
+      color: #d3d3d3;
```
pkozlowski-opensource: Yeh, we definitively can't have those styles hard-coded here...

src/datepicker/ngb-calendar.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+      const calendar = new GregorianCalendar();
+
+      const jsToday = new Date();
+      const today = new NgbDate(jsToday.getFullYear(), jsToday.getMonth(), jsToday.getDate());
+
+      expect(calendar.getToday().equals(today)).toBeTruthy();
+    });
+  });
+
+});
```
pkozlowski-opensource: You will have to rebase your PR and rename this file to take e348323af37af42e8bec029855775c0c52ae77de into account.

src/datepicker/ngb-calendar.ts
pkozlowski-opensource: But you can have fields :-)

src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export class GregorianCalendar implements NgbCalendar {
```
pkozlowski-opensource: As discussed let's split into 2 services: pure calendar logic and localization-focused part.

maxokorokov:
@pkozlowski-opensource thanks for the initial review. I know the HTML markup and CSS are a mess now - it migrated here from my proof-of-concept prototype. I'll leave only the bare minimum for the layout and use bootstrap styles wherever possible.

maxokorokov:
- rebased to have `datepicker.module.ts`
- cleaned up CSS significantly (removed custom colors and pixel sizes)
- extracted  default day template into `datepicker-day-view.ts` and made it use bootstrap classes
- cleaned up model and argument names

src/datepicker/ngb-date.ts
```diff
@@ -0,0 +1,25 @@
+export class NgbDate {
+  static from(date: {year: number, month: number, date?: number}) {
```
sendilkumarn: can we make `month` also optional ?

maxokorokov:
If anybody was following current pull request / issue #20, looks like the initial version is ready. Last changes are:

- split i18n and calendar
- hidden internal `NgbDate` from the user
- introduced `datepicker-navigation-selection` - a simpler navigation for the first version using two select boxes for month and year.  'Zoom' style navigation (as in angular-ui) could be added later on, as it's too much already.
- introduced `min/maxDates` and `navigateTo(date)` in the API
- updated the plunker
- more tests and docs

So it looks like this for now (see demo for minimal API example):
screen shot 2016-08-31 at 18 58 42. 

Reviews are welcome. It's big, but the minimal working component is kinda big.
Ping @pkozlowski-opensource, @wesleycho, @Foxandxss and whoever is interested

src/datepicker/ngb-date.ts
maxokorokov: Sure, but `NgbDate` is and not user-exposed, so it wasn't necessary for now. See the public API in the `NgbDatepicker`

wesleycho: This PR should stick to minimal features as possible, that would be a feature request for the future.

demo/src/app/components/datepicker/demos/basic/datepicker-basic.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Component({
+  selector: 'ngbd-datepicker-basic',
+  template: require('./datepicker-basic.html')
+})
+export class NgbdDatepickerBasic {
+
+  model;
+
+  minDate = {year: 2000, month: 0, date: 7};
```
pkozlowski-opensource: Shouldn't it be `day` instead of `date`? For me `date` indicated the whole object. Could we please rename it?

src/datepicker/ngb-date.ts
```diff
@@ -0,0 +1,25 @@
...
+                    this.year < other.year;
+  }
+
+  after(other: NgbDate) {
+    return !other ? false : this.year === other.year ?
+                    this.month === other.month ? this.date === other.date ? false : this.date > other.date :
+                                                 this.month > other.month :
+                    this.year > other.year;
+  }
+}
```
pkozlowski-opensource: Those before / after methods are kind of long-winded. Maybe it would be easier to rewrite them with if / else instead of using tenary?

Other thought: how about implementing a `compare`   method with standard contract (-1, 0, 1 as return values). Not saying we need to do this in the PR, just sth to think about.

src/datepicker/ngb-date.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+describe('ngb-date', () => {
+
+  it('equals', () => {
+    const date = new NgbDate(2016, 8, 18);
+
+    expect(date.equals(new NgbDate(2016, 8, 18))).toBeTruthy();
+
+    expect(date.equals(null)).toBeFalsy();
+    expect(date.equals(undefined)).toBeFalsy();
```
pkozlowski-opensource: Could you split it into multiple `it` blocks, ex: one for undefined / null, other for "real" dates etc.

src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,61 @@
+import {NgbDate} from './ngb-date';
+import {Injectable} from '@angular/core';
+
+/**
+ * Default implementation of a Gregorian calendar
+ */
+@Injectable()
+export class NgbCalendar {
```
pkozlowski-opensource: We should have an abstract class `NgbCalendar` and the concentrate class `NgbGregorianCalendar` that extends it. We should then only use `NgbCalendar` in DI / other parts of the code.

src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,61 @@
...
+  private toDate(date: NgbDate) { return new Date(date.year, date.month, date.date); }
+
+  private fromDate(jsDate: Date) { return new NgbDate(jsDate.getFullYear(), jsDate.getMonth(), jsDate.getDate()); }
```
pkozlowski-opensource: I think that those 2 methods belong to NgbDate as static members.

src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,61 @@
...
+  getMonths() { return [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]; }
+
+  getWeeksPerMonth() { return 6; }
+
+  getWeekday(date: NgbDate) {
+    let jsDate = this.toDate(date);
+    return jsDate.getDay();
+  }
+
+  getNextDate(date: NgbDate): NgbDate {
```
pkozlowski-opensource: `getNextDate` => `getNextDay`, same for prev

src/datepicker/ngb-calendar.ts
```diff
@@ -0,0 +1,61 @@
...
+    return this.fromDate(jsDate);
+  }
+
+  getPrevMonth(date: NgbDate): NgbDate {
+    let jsDate = this.toDate(date);
+    jsDate = new Date(jsDate.getFullYear(), jsDate.getMonth() - 1, 1);
+    return this.fromDate(jsDate);
+  }
+
+  getWeekNumber(week: NgbDate[], firstDayOfWeek: number) {
```
pkozlowski-opensource: shouldn't we have a default for `firstDayOfWeek` here?

demo/src/app/components/datepicker/demos/basic/datepicker-basic.ts
maxokorokov: Yeah, you're right, will rename, it will be better. Got confused by the standard js `Date` where there is a `day` field but `getDate()` method that returns it and `getDay()` that returns the day of the week...

src/datepicker/ngb-date.ts
maxokorokov: It's shorter and I have tests :) Auto-formatting is kinda crappy, otherwise it's readable. Will note the `compare` and see how `if`s look

src/datepicker/datepicker-day-view.ts
```diff
@@ -0,0 +1,20 @@
...
+@Component({
+  selector: '[ngb-datepicker-day-view]',
+  styles: [`
+    :host {      
+      text-align: center;
+      padding: 0.185rem 0.25rem;      
+      border-radius: 0.25rem;
+    }
+  `],
+  host: {'[class.bg-primary]': 'selected', '[class.text-muted]': 'day.date.month !== month.number || day.disabled'},
```
pkozlowski-opensource: How about extracting `day.date.month !== month.number || day.disabled` to a method (`isDisabled()`?). Probably would be easier to read / follow / understand.

src/datepicker/datepicker-i18n.ts
```diff
@@ -0,0 +1,11 @@
+import {Injectable} from '@angular/core';
+
+const WEEKDAYS = ['Su', 'Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa'];
+const MONTHS = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
+
+@Injectable()
+export class NgbDatepickerI18n {
```
pkozlowski-opensource: As with the calendar we need to split into an abstract class and a concrete implmentation(`NgbDatepickerI18nENLocale` or sth)

src/datepicker/ngb-calendar.ts
maxokorokov: Ok, will change that and move to getters I guess

src/datepicker/datepicker-day-view.ts
```diff
@@ -0,0 +1,20 @@
+import {Component, Input} from '@angular/core';
+import {MonthViewModel, DayViewModel} from './datepicker-view-model';
+
+@Component({
+  selector: '[ngb-datepicker-day-view]',
```
pkozlowski-opensource: Why is this one using attribute as a selector? If we go down this path it should be camelCase instead of snake-case.

pkozlowski-opensource:
@maxokorokov did another pass on the code and left few comments. Nothing major but there are few things we need to care of. Please have a look and ping me when you've updated the PR.

I'm switching to RC6 now so we can rebase this PR on top of latest ng2 version

src/datepicker/ngb-calendar.ts
maxokorokov: why? in the end it's `ngbDatepicker` that picks a default one anyway

maxokorokov:
@pkozlowski-opensource thanks for the review, did the following in respect to your comments:

- split `i8n` and `calendar` into abstract classes
- renamed `date` -->`day` in the data model and `getNextDate` --> `getNextDay`
- renamed directives not to use dashes
- cleaned up templates
- split some tests into to be more precise
- aligned with RC6

maxokorokov:
Also added the documentation for the `NgbDatepickerDayTemplateContext` - data model for the day template

src/datepicker/datepicker-navigation.ts
```diff
@@ -0,0 +1,56 @@
...
+@Component({
+  selector: '[ngbDatepickerNavigation]',
+  styles: [`
+    td {
+      text-align: center;
+      padding-bottom: 0.25rem;
+    }
+  `],
+  template: `
+    <tr>
```
wesleycho: This restricts this component to being inside a `tbody` element - is this what we want?

src/datepicker/datepicker-service.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+         year: 2000,
+         weeks: [{number: 1, days: [{date: new NgbDate(2000, 1, 1), disabled: true}]}],
+         weekdays: [1]
+       });
+     }));
+
+  it('should mark dates out of min/max bounds as disabled',
+     inject([NgbDatepickerService, NgbCalendar], (service, calendar) => {
+       let monthViewModel = service.generateMonthViewModel(
+           calendar.getToday(), new NgbDate(2000, 0, 1), new NgbDate(2000, 1, 10), 1, () => { return true; });
```
wesleycho: fwiw the callback could just be written as `() => true`

src/datepicker/datepicker-service.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+         year: 2000,
+         weeks: [{number: 1, days: [{date: new NgbDate(2000, 1, 1), disabled: true}]}],
+         weekdays: [1]
+       });
+     }));
+
+  it('markDisabled callback should not override date bounds',
+     inject([NgbDatepickerService, NgbCalendar], (service, calendar) => {
+       let monthViewModel = service.generateMonthViewModel(
+           calendar.getToday(), new NgbDate(2000, 0, 1), new NgbDate(2000, 0, 10), 1, () => { return false; });
```
wesleycho: Similarly here with false

wesleycho:
I have a question about `NgbCalendar` - do all calendar systems have concepts of months, weeks, and days?

demo/src/app/components/datepicker/demos/basic/datepicker-basic.ts
```diff
@@ -0,0 +1,19 @@
+import {Component} from '@angular/core';
+
+const now = new Date();
+
+@Component({
+  selector: 'ngbd-datepicker-basic',
+  template: require('./datepicker-basic.html')
```
wesleycho: This should just be `templateUrl: './datepicker-basic.html'`

src/datepicker/datepicker-month-view.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    spyOn(fixture.componentInstance, 'onClick');
+
+    const dates = getDates(fixture.nativeElement);
+    dates[0].click();
+
+    expect(fixture.componentInstance.onClick).not.toHaveBeenCalledWith();
+  });
+
+  // CURSOR TESTS FAIL IN IE 9
```
wesleycho: Is it possible to have these tests run in the other browsers but IE9, i.e. browser sniffing? Not the greatest solution, but better to have the test information than not IMO.

icfantv:
TBH, it's been 6 weeks since I've done ng2 development and am not familiar w/ the changes past RC3.  All this looks fine to me, but that should be taken w/ a grain of salt.

wesleycho:
After talking with @pkozlowski-opensource, this LGTM as is minus the demo template if we want to get in. My comments were more minor/affirming the decisions being made.

src/datepicker/datepicker-navigation.ts
maxokorokov: I know it's restrictive as of now. Not only the `ngbDatepickerNavigation`, but also `ngbDatepickerMonthView` as well. It's because the simplest layout for the calendar I've came up with for now is a table - so all column alignment would be synchronised easily between navigation and dates, weekdays, etc. I would probably migrate to a better one if I knew how to do it easily with purer CSS.

Having said that, the only thing that's exposed now to the user is the template inside of a `<td>` element so the internals should be replaceable in the longer run. Now each section is just a table's `<tbody>`

demo/src/app/components/datepicker/demos/basic/datepicker-basic.ts
maxokorokov: Oh, it's the script that generates the demo skeleton apparently. Good catch, thanks. Will update the script in a separate PR.

maxokorokov:
Thanks for taking a look, @wesleycho and @icfantv. 

About calendars: I'm no calendar expert here, but I think it would be quite challenging to support calendars without day/month/year notion as well in one component. Or it will just be a parallel implementation inside.

Plus current implementation will need tuning for sure (ex. `getWeeksPerMonth()` in the calendar should probably be `getWeeksPerMonth(month, year)` as some calendars have different number of days in different months, etc)

src/datepicker/datepicker-day-view.spec.ts
```diff
@@ -0,0 +1,77 @@
+import {TestBed} from '@angular/core/testing';
+
+import {Component} from '@angular/core';
+import {NgbDatepickerModule} from './datepicker.module';
+import {NgbDatepickerDayView} from './datepicker-day-view';
+
+function getElement(element: HTMLElement): HTMLElement {
```
pkozlowski-opensource: Could you rename it to sth more descriptive? What kind of an element we are trying to get?

src/datepicker/datepicker-month-view.spec.ts
pkozlowski-opensource: We are going to add a test utility for browser sniffing and skip those tests in IE9

pkozlowski-opensource:
LGTM apart from a small nit (don't really need to be fixed) and a missing browser sniffing. Great stuff @maxokorokov - datepickers are not easy but I think that we've got really nice API and architecture in here! Let's land this baby so people can play with it.

sendilkumarn:
@maxokorokov great one really 👏 👏 👏 👍

jnizet:
That looks great. 👍 

I come a bit late to the game, but after just looking at the demo and playing a little bit with it, here are a few remarks:

- some classes aren't exported, but should be, like DayTemplateContext, NgbDatepickerI18n. 
- NgbDatepickerI18n should be documented in the demo page
- Just as for timepicker, I wonder if we shouldn't export an interface `NgbDateStruct` that would give a name to the year-month-day structure used as the model of the component. 
- If you specify a minDate but no maxDate, you get an exception. Not a big deal, but unexpected. The exception message is very clear though 👍 
- Saying that months are 0-based in the documentation wouldn't hurt. That's a never-ending source of bugs
- It feels strange to have mandatory limits on the navigation. I understand that the use of a select box makes it mandatory, and I don't have any idea of how to change that, but still. The very first use-case that comes to mind is the selection of a birth date, and having only 10 years before now as the limit forces the user to set a min date... and a max date (see above). Not a big deal.

Again, very good job!

maxokorokov:
Thanks for the review @jnizet 

- `DayTemplateContext` is an interface purely for documentation, user doesn't need it
- concerning min/max dates: default values are  -10 and +10 yrs just to provide some default values and yes, you have to provide some limits for select boxes... 
- There is a different kind of 'zoom' navigation that might be implemented as in [here](https://eternicode.github.io/bootstrap-datepicker/?markup=input&format=&weekStart=&startDate=&endDate=&startView=0&minViewMode=0&maxViewMode=4&todayBtn=false&clearBtn=false&language=en&orientation=auto&multidate=&multidateSeparator=&keyboardNavigation=on&forceParse=on#sandbox) for instance that will make navigation infinite. Can be done later
- will allow setting only one of `minDate` or `maxDate` and leaving the second one as default value, you're right about this one

icfantv:
IMO, this is good.  @maxokorokov have you made [do you want to make] any changes per @jnizet's recommendations, specifically, the last bullet above?  If not, I'm inclined to merge this.  We can tweak the padding after it lands.

maxokorokov:
Amended ie9 cursor test ignore and `min/maxDate` are completely optional now as in the comment above

PR #624, opened 2016-08-24
chore: rename index.ts for systemjs

Foxandxss:
Having `index.js` per component forces the end user to have an entry at `systemjs config`. So if we have 20 components, we force the users to have 20 entries.

I chose `foo.module.ts` as the name.


src/index.ts
```diff
@@ -1,22 +1,19 @@
...
-import {NgbPopoverModule} from './popover';
-import {NgbProgressbarModule} from './progressbar';
-import {NgbRatingModule} from './rating';
-import {NgbTabsetModule} from './tabset';
-import {NgbTimepickerModule} from './timepicker';
-import {NgbTooltipModule} from './tooltip';
-import {NgbTypeaheadModule} from './typeahead';
-
-export {NgbPanelChangeEvent} from './accordion';
-export {NgbTabChangeEvent} from './tabset';
```
pkozlowski-opensource: Those exports are still needed. Did you remove them by accident?

Foxandxss: No, just added them with the other imports, line 3 and 13.

pkozlowski-opensource: Huh? Those are _exports_....

Foxandxss: Ah, completely blind.

pkozlowski-opensource:
I've changed the remaining parts and merged.

PR #625, opened 2016-08-25
fix(positioning): offset element position in a scrollable element

### LINKED ISSUE #621: Tooltip position issue
ng-bootstrap: 1.0.0-alpha.2
Bootstrap: 4.0.0-alpha.3

I am using the modal CSS from bootstrap, and have noticed the position of the tooltip does not get calculated properly when the modal content has scroll bars.

Excuse the lack of plunkr, however this can be demonstrated, even without the complete modal setup, on the tooltip demo page by wrapping the first button in the `Quick and easy tooltips` section with the following elements like this:
```
<div class="modal-content" style="height:200px; overflow-y:auto">
  <div style="height:180px">
  </div>
  <button type="button" class="btn btn-secondary" placement="top" ngbTooltip="Tooltip on top">
    Tooltip on top
  </button>
</div>
```
The gap between the target element and the tooltip gets larger as depending on how far down you scroll down.

Quinn-L:

Fixes #621

### DIFF
--- a/src/util/positioning.spec.ts
+++ b/src/util/positioning.spec.ts
@@ -7,26 +7,40 @@ describe('Positioning', () => {
   const bodyHeight = document.body.style.height;
   const bodyWidth = document.body.style.width;
 
-  function createElement(height: number, width: number, marginTop: number, marginLeft: number): HTMLElement {
+  function getStyleValue(value: number): string {
+    if (value) {
+      return `${value}px`;
+    } else {
+      return '0';
+    }
+  }
+
+  function createElement(
+      height: number, width: number, marginTop: number, marginLeft: number, marginBottom: number,
+      marginRight: number): HTMLElement {
     let element = document.createElement('div');
     element.style.display = 'inline-block';
-    element.style.height = height + 'px';
-    element.style.width = width + 'px';
-    element.style.marginTop = marginTop + 'px';
-    element.style.marginLeft = marginLeft + 'px';
+    element.style.height = getStyleValue(height);
+    element.style.width = getStyleValue(width);
+    element.style.marginTop = getStyleValue(marginTop);
+    element.style.marginLeft = getStyleValue(marginLeft);
+    element.style.marginBottom = getStyleValue(marginBottom);
+    element.style.marginRight = getStyleValue(marginRight);
 
     return element;
   }
 
-  let element = createElement(200, 300, 100, 150);
-  document.body.appendChild(element);
-  let targetElement = createElement(50, 100, 10, 20);
-  document.body.appendChild(targetElement);
+  let containerElement = createElement(500, 500, 0, 0, 0, 0);
+  document.body.appendChild(containerElement);
+  let element = createElement(200, 300, 100, 150, 100, 150);
+  containerElement.appendChild(element);
+  let targetElement = createElement(50, 100, 10, 20, 10, 20);
+  containerElement.appendChild(targetElement);
 
   document.documentElement.style.margin = '0';
   document.body.style.margin = '0';
-  document.body.style.height = '2000px';
-  document.body.style.width = '2000px';
+  document.body.style.height = '1000px';
+  document.body.style.width = '1000px';
 
   it('should calculate the element offset', () => {
     let position = positioning.offset(element);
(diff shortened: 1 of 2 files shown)


RobJacobs:
A couple of things to consider, the uib position and offset methods were intended to follow the jQuery position and offset methods.  The jQuery position method includes margins, so an element with margin-left: 100px would have a position left of 0.  When positioning elements, we need to ignore margins so an element with margin-left: 100px needs to have position left of 100.

The problem you are reporting actually needs to be fixed in the position method, not the offset method.  Doing a compare of the previous [version](https://github.com/angular-ui/bootstrap/blob/07c31d0731f7cb068a1932b8e01d2312b796b4ec/src/position/position.js#L52) you will notice scroll offset is accounted for in the position method.  This version is missing that piece.  The following [plunk](https://plnkr.co/edit/6mNt5BVH5GGWJjq9QmWk?p=preview) demonstrates a proposed fix with a test to compare the jQuery and uib position/offset results.  Code changes are:

Add includeMargin parameter to make the position method behave like jQuery.
Move parentOffset adjustments to only work with non-body parent elements.
Subtract scrollTop/Left values for non-body parent elements.

    position(element: HTMLElement, round = true, includeMargin = true): ClientRect {
      let elPosition: ClientRect;
      let parentOffset: ClientRect = {width: 0, height: 0, top: 0, bottom: 0, left: 0, right: 0};

      if (this.getStyle(element, 'position') === 'fixed') {
          elPosition = element.getBoundingClientRect();
      } else {
          const offsetParentEl = this.offsetParent(element);
          elPosition = this.offset(element, false);
  
          if (offsetParentEl !== document.documentElement) {
            parentOffset = this.offset(offsetParentEl, false);
            parentOffset.top += offsetParentEl.clientTop - offsetParentEl.scrollTop;
            parentOffset.left += offsetParentEl.clientLeft - offsetParentEl.scrollLeft;
          }
      }

      elPosition.top -= parentOffset.top;
      elPosition.bottom -= parentOffset.top;
      elPosition.left -= parentOffset.left;
      elPosition.right -= parentOffset.left;

      if (includeMargin) {
          elPosition.top -= parseFloat(this.getStyle(element, 'margin-top'));
          elPosition.bottom += parseFloat(this.getStyle(element, 'margin-bottom'));
          elPosition.left -= parseFloat(this.getStyle(element, 'margin-left'));
          elPosition.right += parseFloat(this.getStyle(element, 'margin-right'));
      }

      if (round) {
          elPosition.height = Math.round(elPosition.height);
          elPosition.width = Math.round(elPosition.width);
          elPosition.top = Math.round(elPosition.top);
          elPosition.bottom = Math.round(elPosition.bottom);
          elPosition.left = Math.round(elPosition.left);
          elPosition.right = Math.round(elPosition.right);
      }
      return elPosition;
    }

In positionElements method, call position method with margins excluded:

    const hostElPosition = appendToBody ? this.offset(hostElement, false) : this.position(hostElement, false, false);

Quinn-L:
Thanks @RobJacobs, I've included your suggestions.
FYI, as you're well versed with the code, I don't mind cancelling this pull if you can get a push out faster or have better code (not that this is really urgent, though).

RobJacobs:
@Quinn-L I'm fine with using this PR to address the problem.  I'm still seeing unexpected results so I'm afraid my suggestions still need some work.  The uib position top is not matching the jQuery position top when the container element has scrolled.  I'm going to do some more testing and post when I have this solved.

RobJacobs:
The jQuery position method does not adjust for scroll.  This updated [plunk](http://plnkr.co/edit/6mNt5BVH5GGWJjq9QmWk?p=preview) demonstrates the uib position method producing the same results as the jQuery position method and the uib positionElements method working correctly when appendToBody is true or false.  The scroll adjustments have been moved to the positionElements method.  I'd like to get @pkozlowski-opensource opinion on how to approach this.  Should the scroll adjustment reside in the uib position method and diverge from jQuery (or have another parameter to include scrollOffset)?  Or leave the scroll adjustment in the positionElements method?

pkozlowski-opensource:
@RobJacobs it would take me a bit of time to dive into this code and be able to provide reasonable input. One thought here: wouldn't this case be properly covered by the `appendToBody` option that we didn't implement yet (but it wouldn't be difficult).

As far as I understand then we would switch to the absolute positioning and calculations would be different. 

If needed I will go over the current code / try to suggest sth but for now throwing-in semi-random ideas :-)

RobJacobs:
@pkozlowski-opensource The appendToBody is implemented and working.  The downfall of using appendToBody is that you would need to call the positionElements to reposition the target element on the scroll event of the container div.  When the target element is added to the same container element as the host element, the positionElements method only needs to be called once (or when the host element moves within the container element).  Scrolling on the container element has no impact on the relative position of the target element to host element.  I apologize for making this sound more complex than it really is, to simplify my question, should the uib position method produce the same results as the jQuery position method?

Quinn-L:
Newest pull includes your changes, however, having played around a bit more with the tooltip/popups, unfortunately I think the `appendToBody` should be the correct approach (including recalculating the positions on scroll events). The problem with the current tooltips/popups is that they get clipped when placed in a nested div that has scrollbars.

Playing around with the official bootstrap demo (haven't looked into their source code, though), they append to the body and recalculate on scrolling.

Quinn-L:
Closing this pull because this fix doesn't fix the underlying issue with tooltips/popups (ie. it being clipped).
The official bootstrap uses tether to resolve their positioning issues. I'm not sure what the official stance on third party libraries are for this project, but it's a bit out of scope to reinvent this myself.

PR #627, opened 2016-08-25
test: mark async tests as async

pkozlowski-opensource:
We need to mark asynchronous tests with `async(...)` otherwise things fail miserably in RC6.


Foxandxss:
Sounds good.

PR #635, opened 2016-08-30
chore(api): properly extract type info

### LINKED ISSUE #633: doc: incorrect type of property
In the generated documentation of `NgbModalRef`, the `result` property is of type `any`, although its actual type defined in the source is `Promise<any>`.

pkozlowski-opensource:
Fixes #633


jnizet:
Sorry to be too late for the release. LGTM anyway. 

Adding a second script to package.json to run the apidoc tests in "non-TDD" mode would be nice. Not being a NodeJS guy, I had to investigate to find out how to run these tests.

PR #640, opened 2016-08-30
feat(typeahead): add support for the showHint option

### LINKED ISSUE #639: Typeahaed: add support for the show-hint option
When this option is activated navigating via available option (keyboard) should show a about-to-be-selected item as highlighted in the input box.

pkozlowski-opensource:
Closes #639

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -1,5 +1,5 @@
 import {TestBed, ComponentFixture, async} from '@angular/core/testing';
-import {createGenericTestComponent} from '../test/common';
+import {createGenericTestComponent, isBrowser} from '../test/common';
 import {expectResults, getWindowLinks} from '../test/typeahead/common';
 
 import {Component, DebugElement} from '@angular/core';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
OH man, IE...

maxokorokov:
Might that just be `fixture.whenStable()` issue in the dismissal tests?

src/typeahead/typeahead.ts
```diff
@@ -61,6 +62,7 @@ export class NgbTypeahead implements OnInit,
   private _positioning = new Positioning();
   private _subscription: Subscription;
   private _valueChanges = new Subject<string>();
+  private _userInput: string;
```
maxokorokov: ... S T **U** V W ... :)

maxokorokov:
@pkozlowski-opensource I fixed the build issue, it should work now

pkozlowski-opensource:
Thnx @maxokorokov - rebased and it is green now :-) Needs review.

wesleycho:
The comment is still valid - @maxokorokov was pointing out the mistake in alphabetical order of variables

src/typeahead/typeahead.ts
```diff
@@ -118,44 +125,45 @@ export class NgbTypeahead implements OnInit,
...
-        if (this.resultFormatter) {
-          this._windowRef.instance.formatter = this.resultFormatter;
-        }
-        if (this.resultTemplate) {
-          this._windowRef.instance.resultTemplate = this.resultTemplate;
-        }
-      }
-    });
+    this._subscription =
+        this._valueChanges.do(value => this._userInput = value).let (this.ngbTypeahead).subscribe((results) => {
```
wesleycho: This looks strange - we should probably not be putting a space after `let` there. Ideally, I think we should probably shorten this line by dropping to the next, but we might be dealing with clang-format here...

src/typeahead/typeahead.ts
```diff
@@ -194,11 +204,39 @@ export class NgbTypeahead implements OnInit,
...
+        this._renderer.invokeElementMethod(
+            this._elementRef.nativeElement, 'setSelectionRange', [this._userInput.length, formattedVal.length]);
+      } else {
+        this.writeValue(this._windowRef.instance.getActive());
+      }
+    }
+  }
+
+  private _formatItemForInput(item: any): string {
+    return item && this.inputFormatter ? this.inputFormatter(item) : toString(item);
```
wesleycho: I know we're expecting `this.inputFormatter` to be a function here, but this could be problematic if a user is using ES6 - are we fine with there being an error in that situation?

src/typeahead/typeahead.spec.ts
```diff
@@ -395,6 +395,102 @@ describe('ngb-typeahead', () => {
     });
   });
 
+  if (!isBrowser(['ie']) && !isBrowser(['edge'])) {
```
maxokorokov: should work as `!isBrowser(['ie', 'edge'])`

src/typeahead/typeahead.ts
pkozlowski-opensource: This is clang... We should probably try to update to the latest version, although it would be probably massive...

PR #654, opened 2016-09-01
chore: add types to the api-docs components

jnizet:
The doc components are getting more complex. Adding types makes the code easier to understand and maintain.


demo/src/app/components/shared/api-docs/api-docs-class.component.ts
```diff
@@ -12,13 +13,10 @@ import docs from '../../../../api-docs';
...
+  apiDocs: ClassDesc;
+  @Input() set type(typeName: string) {
     this.apiDocs = docs[typeName];
   };
 
-  methodSignature(method) {
-    const args = method.args.map(arg => `${arg.name}: ${arg.type}`).join(', ');
-    return `${method.name}(${args}): ${method.returnType}`;
-  }
+  methodSignature(method: MethodDesc) { return signature(method); }
```
sendilkumarn: It should have a return type here right ?

PR #655, opened 2016-09-01
feat(accordion): add config service to provide default values

jnizet:
refs #518

### DIFF
--- a/demo/src/app/components/shared/api-docs/api-docs.component.ts
+++ b/demo/src/app/components/shared/api-docs/api-docs.component.ts
@@ -38,16 +38,12 @@ export class NgbdApiDocs {
   };
 
   /**
-   * Returns the default value of the given directive input. If falsy, returns the default value of the matching
-   * config service property
+   * Returns the default value of the given directive input by first looking for it in the matching config service
+   * property. If there is no matching config property, it reads it from the input.
    */
   defaultInputValue(input: InputDesc): string {
-    if (input.defaultValue !== undefined) {
-      return input.defaultValue;
-    }
-
     const configProperty = this._configProperties[input.name];
-    return (configProperty && configProperty.defaultValue);
+    return configProperty ? configProperty.defaultValue : input.defaultValue;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


src/progressbar/progressbar.ts
```diff
@@ -23,18 +23,18 @@ export class NgbProgressbar {
   /**
    * Maximal value to be displayed in the progressbar.
    */
-  @Input() max = 100;
+  @Input() max: number;
```
pkozlowski-opensource: If we do this then we are loosing default values info from the docs, no?

jnizet: No. The api-docs component now gets the default value here, and if undefined, looks for it in the matching  property of the associated config service. So the default value should in fact not be defined here:
 - it's useless code, since it's immediately overwritten by the constructor, which copies the default value from the property
 - it's misleading code, since it makes the reader think that the default value is 100, whereas it's in fact copied from the config property in the constructor
 - it's also misleading because the generated documentation reads the default value from here although the actual default value comes from the config property

The last point makes me realize that the api-docs component should do the opposite of what it does now: try to find the default value in the matching config property, and only if not found, read it from here.

I'll amend the PR.

pkozlowski-opensource:
@jnizet could you please rebase? I think it has a conflict with your other PR I've just merged :-) Thnx!

jnizet:
@pkozlowski-opensource rebase done

demo/src/app/components/shared/api-docs/api-docs.component.ts
```diff
@@ -38,16 +38,12 @@ export class NgbdApiDocs {
   };
 
   /**
-   * Returns the default value of the given directive input. If falsy, returns the default value of the matching
-   * config service property
+   * Returns the default value of the given directive input by first looking for it in the matching config service
+   * property. If no matching config property, reads it from the input.
```
icfantv: "If there is no..., it reads it from..."

icfantv:
@jnizet just the one grammar change and LGTM.

PR #661, opened 2016-09-02
feat(alert): add config service to provide default values

jnizet:
refs #518

I suffered with the tests of NgbDismissibleAlert because creating an instance of this component using new or with the TestBed is harder than with other components: it has many dependencies thatare used in the constructor. Instead of trying to mock them all or to change the code of the component just to ease that test, I thus used a slightly different test strategy than for the other components.

### DIFF
--- a/src/alert/alert.spec.ts
+++ b/src/alert/alert.spec.ts
@@ -1,10 +1,11 @@
-import {fakeAsync, tick, TestBed, ComponentFixture} from '@angular/core/testing';
+import {fakeAsync, tick, TestBed, ComponentFixture, inject} from '@angular/core/testing';
 import {createGenericTestComponent} from '../util/tests';
 
 import {Component} from '@angular/core';
 
 import {NgbAlertModule} from './alert.module';
 import {NgbAlert, NgbDismissibleAlert} from './alert';
+import {NgbAlertConfig, NgbDismissibleAlertConfig} from './alert-config';
 
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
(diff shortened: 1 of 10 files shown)


src/alert/alert.spec.ts
```diff
@@ -21,79 +22,179 @@ describe('ngb-alert', () => {
...
+    beforeEach(() => { TestBed.configureTestingModule({declarations: [TestComponent], imports: [NgbAlertModule]}); });
 
-       tick(800);
-       fixture.detectChanges();
-       expect(alertEl.getAttribute('role')).toEqual('alert');
-       expect(getCloseButton(alertEl)).toBeTruthy();
+    it('should initialize inputs with provided config', () => {
+      let config: NgbDismissibleAlertConfig;
+      const fixture = createTestComponent(`<template ngbAlert>Hello, {{name}}!</template>`);
+      inject([NgbDismissibleAlertConfig], (c: NgbDismissibleAlertConfig) => {
```
wesleycho: This seems weird to me - does it need to be injected here? Can you explain a little more in detail about what situation this is simulating?

wesleycho:
This LGTM, just want a little more clarification on the test scenario.

src/alert/alert.spec.ts
jnizet: The idea is to simulate what the config service documentation suggests to do to globally change the default values for alerts (and all the other components):

 - NgbAlertModule declares an NgbAlertConfig provider 
 - an application main module (simulated by the test module here) imports NgbAlertModule (or NgbModule). A global NgbAlertConfig service, containing mutable default values, is thus available to the whole application
 - the root application component gets a reference to this global NgbAlertConfig config thanks to dependency injection. That is simulated by the inject() in the test
 - the root application component sets custom default values in the injected NgbAlertConfig service. That is simulated by the body of the function passed to inject().
 - A template of the application ends up using an NgbAlert component. This is simulated by the createTestComponent() call in the test
 - The created alert should use the default values set before in the NgbAlertConfig service. This is what the test checks.

Does that clarify? I agree the test is more convoluted than the other similar tests, but checking for the right timeout can't be done by just inspecting the dom, and, to be fair, I haven't found a way to just getting an instance of the NgbDismissibleAlert in the test.

Using

    TestBed.createComponent(NgbDismissibleAlert)

leads to the following error message, which looks completely wrong to me:
    
    Cannot create the component NgbDismissibleAlert as it was not imported into the testing module!

icfantv:
@wesleycho are you satisfied with the explanation?
@jnizet please rebase and fix the conflicts.  thanks.

wesleycho:
Thanks for the very detailed explanation - I'm ok with this then.

jnizet:
Rebase done. I also added the missing api doc for the two config service un the demo and fixed the label of the config demo. :-/

PR #663, opened 2016-09-02
test: ability to exclude certain tests for certain browsers

maxokorokov:
Ability to exclude certain tests from execution in certain browsers needed by #618 and #640 

```javascript
it_disableFor(['ie9', 'ie10'], 'test description' () => {
  // ...
});

xit_disableFor(...)

fit_disableFor(...)
```
Simplest version, might be enhanced to support async, timeouts, later, etc

### DIFF
--- a/src/util/tests.spec.ts
+++ b/src/util/tests.spec.ts
@@ -1,0 +1,52 @@
+import {getBrowser, isBrowser} from './tests';
+
+const sampleAgents = {
+  ie9: 'Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 7.1; Trident/5.0)',
+  ie10: 'Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; WOW64; Trident/6.0)',
+  ie11: 'Mozilla/5.0 (Windows NT 6.1; WOW64; Trident/7.0; AS; rv:11.0) like Gecko',
+  edge:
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
> Simplest version, might be enhanced to support async, timeouts, later, etc

@maxokorokov maybe we are making more complex than necessary? How about a simple function `isBrowser(Browser | Browser[])` that we can stick into if /else around certain tests? This way we don't need to mess around with Jasmine.

Oh, and could we have a catch-all Browser for `ie`?

maxokorokov:
@pkozlowski-opensource on the one hand I like that it generates `xit` now for unsupported browsers, on the other I don't want to mess with Jasmine too.

Catch-all for `ie` is without `edge` meaning `ie9-11`, right?

pkozlowski-opensource:
>  on the one hand I like that it generates xit now for unsupported browsers, on the other I don't want to mess with Jasmine too.

OK, whatever :-) Both are fine for me. Pick what you feel is the best.

> Catch-all for ie is without edge meaning ie9-11, right?

Yep this would work.

Oh, and the Travis is failing atm, not sure if this is a flake or a legit issue.

maxokorokov:
Decided not to support Jasmine - its simpler and usage this stuff should be an exception :) 
Added proper tests and ie-catchall as well. 
Hopefully Travis was a flake, because Edge was failing without any error messages, will see.

pkozlowski-opensource:
Thnx, merged. Now you can use it in the datepicker PR :-)

PR #668, opened 2016-09-02
chore: fix build

jnizet:
Without this fix, the build fails with `error TS4078: Parameter 'browsers' of exported function has or is using private name 'Browser'`.

### DIFF
--- a/src/util/tests.ts
+++ b/src/util/tests.ts
@@ -7,7 +7,7 @@ export function createGenericTestComponent<T>(html: string, type: {new (...args:
   return fixture as ComponentFixture<T>;
 }
 
-type Browser = 'ie9' | 'ie10' | 'ie11' | 'ie' | 'edge' | 'chrome' | 'safari' | 'firefox';
+export type Browser = 'ie9' | 'ie10' | 'ie11' | 'ie' | 'edge' | 'chrome' | 'safari' | 'firefox';
 
 export function getBrowser(ua = window.navigator.userAgent) {
   let browser = 'unknown';


jnizet:
Shouldn't we execute `gulp build` instead of just `gulp` on Travis? It would catch such failures.

icfantv:
I don't know enough about gulp to make any sort of educated statement here.  @Foxandxss or @pkozlowski-opensource, can you comment?  I didn't want to merge this without checking...

icfantv:
@jnizet rebase please.  thanks.

jnizet:
@icfantv rebase done.

pkozlowski-opensource:
@jnizet this is a good change so I'm going to merge it. But this is not the right fix for the build problem since the root cause of the issue is that we are taking this test helper file into account when bundling and we should not. Similarly we are transpiling test files when building demo and we should not. I will open a separate issue for those cases.

tl;dr; this is a good code change but we need proper fix for the issue you've identified.

pkozlowski-opensource:
As for running `gulp build` on TravisCI - why not (it will catch more pbs but will be slower as well), but before we should: 
* understand why tests build didn't catch this PB
* properly exclude test files from cjs packaging

jnizet:
Regarding the first point: the error is triggered by the `declaration: true` compiler option which is set on the cjsProject (in the gulpfile.js file) and in the esmProject (in the tsconfig-es2015.json file), but not on the testProject. When generating the declarations, typescript realizes that an exported function uses a non-exported type in its arguments.

So an alternative to running `gulp build` in Travis would be to set `declaration: true` on the testProject. It would catch this specific error. But still, most of the time in the travis build is spent running the tests on saucelabs, and adding the build tasks would only add a few seconds (around 7 seconds on my machine).

Regarding the second point, I've submitted #672.

PR #672, opened 2016-09-03
chore(build): avoid bundling test utilities

### LINKED ISSUE #670: Build: exclude test artifacts from bundling
Currently the https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/util/tests.ts file is not excluded from bundling and it should.

My take on this would be to create a separate folder for test-specific stuff so we don't keep bumping into this - file by file exclusion as we do in https://github.com/ng-bootstrap/ng-bootstrap/blob/925a3cfc9a19b5125fe627b732b3a96c0289f16a/gulpfile.js#L45-L58  is just too brittle.

jnizet:
- use our own version of typescript with gulp-typescript rather than the bundled 1.8 version
- move all test-related files (except spec.ts files) to a test directory that is excluded from the cjs and esm gulp tasks

fix #670


icfantv:
@jnizet LGTM.  Needs rebasing.  Are the final three lines that changed in the gulpfile what tell gulp to use the typescript version defined in our package.json rather than its internal typescript version?

jnizet:
@icfantv yes, that's it. See https://www.npmjs.com/package/gulp-typescript#typescript-version.

jnizet:
@icfantv rebased (but will probably have to rebase again once the other PR is merged)

icfantv:
@jnizet yep, please rebase.  sorry you have to keep doing this.

jnizet:
@icfantv no problem really :-). Rebase done

PR #676, opened 2016-09-04
feat(alert): self-closing alerts are non-dismissible by default

### LINKED ISSUE #660: [alert] rename NgbDismissibleAlert, make it non-dismissible
NgbDismissibleAlert has a strange name: it implies that regular alerts, i.e. NgbAlerts, are not dismissible, although they are. And bizarrely, NgbDismissibleAlert is always dismissible with a close button, even though the whole point of the directive is to have an alert that self-closes itself after a while, rendering the close button redundant.

We're still in alpha version, so renaming a component is not a big deal.

What I propose:

 - rename NgbDismissibleAlert to NgbSelfClosingAlert
 - add a boolean dismissible input to NgbSelfClosingAlert, aligning it with NgbAlert
 - set its default value to false

The last point is more debatable than the others. I typically use such alerts for sort-lived success messages where the close button is redundant. Maybe others use it for long-lived messages where having the close button by default is better. But anyway, since the default values will soon be customizable globally, that's not a big deal.

jnizet:
- refactor NgbDismissibleAlert to NgbSelfClosingAlert
- make it possible to hide the close button on self-closing alerts
- make the self-closing alert tests a bit less convoluted

fix #660

### DIFF
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -16,7 +16,7 @@ import {
 } from '@angular/core';
 
 import {PopupService} from '../util/popup';
-import {NgbAlertConfig, NgbDismissibleAlertConfig} from './alert-config';
+import {NgbAlertConfig, NgbSelfClosingAlertConfig} from './alert-config';
 
 /**
  * Alerts can be used to provide feedback messages.
@@ -36,7 +36,7 @@ import {NgbAlertConfig, NgbDismissibleAlertConfig} from './alert-config';
 export class NgbAlert {
   /**
    * A flag indicating if a given alert can be dismissed (closed) by a user. If this flag is set, a close button (in a
-   * form of a cross) will be displayed.
+   * form of an ×) will be displayed.
    */
   @Input() dismissible: boolean;
   /**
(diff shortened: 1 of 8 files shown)


src/alert/alert.ts
```diff
@@ -60,11 +60,16 @@ export class NgbAlert {
...
  */
 @Directive({selector: 'template[ngbAlert]'})
-export class NgbDismissibleAlert implements OnInit, OnDestroy {
+export class NgbSelfClosingAlert implements OnInit, OnDestroy {
   private _popupService: PopupService<NgbAlert>;
   private _timeout;
 
   /**
+   * A flag indicating if the alert can be dismissed (closed) by a user. If this flag is set, a close button (in a
+   * form of a cross) will be displayed.
```
icfantv: `&times;`? "times"?  "x"?  just a stupid nit here on my part, but i think of "cross" as a plus.

icfantv: ok, i'm confused.  how are these lines showing up as "new" when they are already present in [the original file](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/alert/alert.ts#L38-L41)?

icfantv:
@jnizet just the one question and LGTM.

src/alert/alert.ts
jnizet: The lines appear in the source of NgbAlert (the "regular" alert component). This is a new input for NgbSelfClosingAlert, previously named NgbDismissibleAlert, which didn't have such an input: the close button was always visible in such alerts, despite the fact that they can close themselves automatically. 

So I added the equivalent input here and copy-pasted the api doc from NgbAlert, to be able to show or hide the close buttons on self-closing alerts too. I also chose to make the button invisible by default, since this component is self-closing and thus doesn't need a close button most of the time.

I'll amend the PR with "an ×" instead of "a cross".

PR #677, opened 2016-09-04
feat(datepicker): add config service to provide default values

jnizet:
refs #518

### DIFF
--- a/src/datepicker/datepicker.spec.ts
+++ b/src/datepicker/datepicker.spec.ts
@@ -1,12 +1,15 @@
-import {TestBed, ComponentFixture, async} from '@angular/core/testing';
+import {TestBed, ComponentFixture, async, inject} from '@angular/core/testing';
 import {createGenericTestComponent} from '../test/common';
 import {getMonthSelect, getYearSelect, getNavigationLinks} from '../test/datepicker/common';
 
-import {Component} from '@angular/core';
+import {Component, TemplateRef} from '@angular/core';
 import {FormsModule, ReactiveFormsModule, FormGroup, FormControl, Validators} from '@angular/forms';
 
 import {NgbDatepickerModule} from './datepicker.module';
 import {NgbDate} from './ngb-date';
+import {NgbDatepickerConfig} from './datepicker-config';
+import {NgbDatepicker} from './datepicker';
+import {DayTemplateContext} from './datepicker-day-template-context';
 
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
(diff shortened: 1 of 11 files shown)


icfantv:
@jnizet needs rebasing

src/datepicker/datepicker.spec.ts
```diff
@@ -23,13 +26,43 @@ function getDatepicker(element: HTMLElement): HTMLElement {
...
+  expect(datepicker.showWeekNumbers).toBe(config.showWeekNumbers);
+  expect(datepicker.startDate).toBe(config.startDate);
+}
+
+function customizeConfig(config: NgbDatepickerConfig) {
+  config.dayTemplate = {} as TemplateRef<DayTemplateContext>;
+  config.firstDayOfWeek = 2;
+  config.markDisabled = (date) => false;
+  config.minDate = {year: 2000, month: 0, day: 1};
+  config.maxDate = {year: 2030, month: 1, day: 31};
```
icfantv: I don't think 31FEB is a valid day.  :-)

icfantv: Do we want to hard code this as a value or can we calculate it as added time from the `minDate`?  The reason I ask is because in other calendar systems, 01JAN is not the first of the year.

This _may_ mean that we need to have a base config object and have different calendars have different extensions of the config - which I'm ok with.  @pkozlowski-opensource @jnizet, thoughts?

icfantv:
@jnizet I don't see any default values for the config.  Was this intentional?

src/datepicker/datepicker.spec.ts
jnizet: If I read the code correctly:

By default, if left undefined, the min date is computed by using the 1st of January of ten years before the start date (or now if there is no start date).

By default, if left undefined, the max date is computed by using the 31 of December of ten years after the start date (or now if there is no start date).

@maxokorokov can you confirm?

But this is done by the Calendar abstraction, which by default is Gregorian, but can be changed to another implementation. So I think we're fine here. 

Note that the above function is just used in unit tests to check that custom config values are honored by the component. I didn't change anything to the min/max date selection logic.

jnizet:
The config service uses the same default values as the original datepicker component. Except for the first day of week and for the boolean flags, the default value is `undefined`, because nothing is mark disabled by default, and the start, min and max dates are computed automatically from the current date if not explicitly specified.

src/datepicker/datepicker.spec.ts
jnizet: If I read the code correctly:

By default, if left undefined, the min date is computed by using the 1st of January of ten years before the start date (or now if there is no start date).

By default, if left undefined, the max date is computed by using the 31 of December of ten years after the start date (or now if there is no start date).

@maxokorokov can you confirm?

But this is done by the Calendar abstraction, which by default is Gregorian, but can be changed to another implementation. So I think we're fine here. 

Note that the above function is just used in unit tests to check that custom config values are honored by the component. I didn't change anything to the min/max date selection logic.

maxokorokov:
Hey @jnizet, sorry for a late reply. On vacations here :) 
But yes, I confirm your comment about the `maxDate`. Its the last day of the year in 10 years from today or start date.

Issue #591, opened 2016-08-12
Analytics + router broken


pkozlowski-opensource:
It looks like I've broken Google analytics when we navigate between rotues. Probably happened as part of the RC5 migration.

icfantv:
I'd be willing to bet that it's angulartics

pkozlowski-opensource:
> I'd be willing to bet that it's angulartics

Yes and no... The problem is that it is not updated to modules / RC5 and we are hacking around it anyway... So I'm less and less excited about using it at all...

Anyway, the root cause is that it seems like in RC5 providers specified on a module level are not instantiated until those are injected _somewhere_. In this sense providers are lazy-instantiated... I don't recall exactly how it worked in RC4 but here we are with the change... 

I'm going to send a PR that injects `Angulartics2GoogleAnalytics` into the `AppComponent` so it is instantiated. But I believe that in the longer run we need to update to a new version of angulartics or drop its usage altogether.

icfantv:
@pkozlowski-opensource, ok, i'll investigate that on my end.

pkozlowski-opensource:
@icfantv I've sent a PR with a work around - #599

icfantv:
They have not cut a new release yet and are still building against RC2.

PR #679, opened 2016-09-04
feat(demo): enable Google Analytics

### LINKED ISSUE #649: Demo site: introduce fine-grained analytics
Due to #646 we had to remove angulartics2 and with all the history / experience we had with this library I'm reluctant to use it in the future. I think we should hand-roll sth simple that is specific to google analytics (angulartics2 is trying to be multi-provider and this just adds bloat to our demo page).

jnizet:
fix #649

### DIFF
--- a/demo/src/app/app.component.ts
+++ b/demo/src/app/app.component.ts
@@ -1,4 +1,5 @@
-import {Component} from '@angular/core';
+import {Component, OnInit} from '@angular/core';
+import {Analytics} from './shared/analytics/analytics';
 
 import '../style/app.scss';
 
@@ -6,5 +7,11 @@ import '../style/app.scss';
   selector: 'ngbd-app',
   templateUrl: './app.component.html'
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
+  constructor(private _analytics: Analytics) {
+  }
+
+  ngOnInit(): void {
+    this._analytics.trackPageViews();
+  }
 }
--- a/demo/src/app/components/shared/api-docs/api-docs-class.component.ts
+++ b/demo/src/app/components/shared/api-docs/api-docs-class.component.ts
@@ -1,6 +1,7 @@
 import {Component, ChangeDetectionStrategy, Input} from '@angular/core';
 import docs from '../../../../api-docs';
 import {ClassDesc, MethodDesc, signature} from './api-docs.model';
+import {Analytics} from '../../../shared/analytics/analytics';
 
 /**
  * Displays the API docs of a class, which is not a directive.
@@ -14,9 +15,16 @@ import {ClassDesc, MethodDesc, signature} from './api-docs.model';
 })
 export class NgbdApiDocsClass {
   apiDocs: ClassDesc;
+
+  constructor(private _analytics: Analytics) {}
+
   @Input() set type(typeName: string) {
     this.apiDocs = docs[typeName];
   };
 
   methodSignature(method: MethodDesc): string { return signature(method); }
+
+  trackSourceClick() {
+    this._analytics.trackEvent('Source File View', this.apiDocs.className);
+  }
 }
--- a/demo/src/app/components/shared/api-docs/api-docs-config.component.ts
+++ b/demo/src/app/components/shared/api-docs/api-docs-config.component.ts
@@ -1,6 +1,7 @@
 import {Component, ChangeDetectionStrategy, Input} from '@angular/core';
 import docs from '../../../../api-docs';
 import {ClassDesc} from './api-docs.model';
+import {Analytics} from '../../../shared/analytics/analytics';
 
 const CONFIG_SUFFIX_LENGTH = 'Config'.length;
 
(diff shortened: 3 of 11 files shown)


demo/src/app/components/shared/api-docs/api-docs-class.component.ts
```diff
@@ -14,9 +15,16 @@ import {ClassDesc, MethodDesc, signature} from './api-docs.model';
...
+  constructor(private _analytics: Analytics) {}
+
   @Input() set type(typeName: string) {
     this.apiDocs = docs[typeName];
   };
 
   methodSignature(method: MethodDesc): string { return signature(method); }
+
+  trackSourceClick() {
+    this._analytics.trackEvent('Source file view', this.apiDocs.className);
```
icfantv: Initial caps, please, as these are event categories.

demo/src/app/components/shared/api-docs/api-docs-config.component.ts
```diff
@@ -20,8 +21,14 @@ export class NgbdApiDocsConfig {
...
 
+  constructor(private _analytics: Analytics) {}
+
   @Input() set type(typeName: string) {
     this.apiDocs = docs[typeName];
     this.directiveName = typeName.slice(0, -CONFIG_SUFFIX_LENGTH);
   };
+
+  trackSourceClick() {
+    this._analytics.trackEvent('Source file view', this.apiDocs.className);
```
icfantv: ditto

icfantv:
@jnizet this is really clever.  i like it.  please fix the event names to be initial caps as they are event categories and we'll get this in.  appreciate all your work on the lib.

sendilkumarn:
👏 this is great 👍 @jnizet

jnizet:
I haven't been able to test it end-to-end, obviously. 

I'm a bit concerned about [this line](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/demo/src/public/index.html#L15), which I haven't changed, but which is probably redundant, because the Analytics service will send the page view even after the initial page load. But since it was there and since I use the same strategy as angulartics, I didn't want to remove it.

icfantv:
@jnizet were you seeing one or two calls to GA in the network console?  As long as we only have one on page load/route change, we're ok.

I can pull the PR and see as well.

jnizet:
There are two calls, but the first one, from the inline script in index.html, doesn't post the path as the second one does. So the first one could be used to track actual page loads. Or it could make things more confusing. I'll let you guys decide.

demo/src/app/app.component.ts
```diff
@@ -1,10 +1,17 @@
...
   selector: 'ngbd-app',
   templateUrl: './app.component.html'
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
+  constructor(private _analytics: Analytics) {
+  }
+
+  ngOnInit(): void {
+    this._analytics.trackPageViews();
```
pkozlowski-opensource: Why do we need it? I would expect that simply including the `Analytics` as a provider should do the trick.

demo/src/app/components/shared/api-docs/api-docs-class.component.ts
```diff
@@ -14,9 +15,16 @@ import {ClassDesc, MethodDesc, signature} from './api-docs.model';
...
+  constructor(private _analytics: Analytics) {}
+
   @Input() set type(typeName: string) {
     this.apiDocs = docs[typeName];
   };
 
   methodSignature(method: MethodDesc): string { return signature(method); }
+
+  trackSourceClick() {
+    this._analytics.trackEvent('Source File View', this.apiDocs.className);
```
pkozlowski-opensource: Shouldn't we have more info (like which component / class is concerned)?

pkozlowski-opensource: Scratch this, it is already there. Sorry for the noise....

demo/src/app/shared/analytics/analytics.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(private _location: Location, private _router: Router) {
+    this._enabled = window.location.href.indexOf('ng-bootstrap.github.io') >= 0;
+  }
+
+  /**
+   * Intended to be called only once. Subscribes to router events and sends a page view
+   * after each ended navigation event.
+   */
+  trackPageViews() {
```
pkozlowski-opensource: Can't we do this in a constructor? What is the advantage of having this is a separate method?

pkozlowski-opensource:
@jnizet Looks good, thnx! This PR confirms that we can get away with a simple in-house impl instead of bringing deps on 3rd parties. 

I've got one comment essentially - what is the advantage of having a separate method to enable page views tracking? The downside I can see it that someone forgets to call it one day :-) Personally I would move this logic to the constructor.

Oh, and I think that we should remove the call we've got in index.html

demo/src/app/app.component.ts
jnizet: I thought that #591 still applied here: services are lazily instantiated and must be injected somewhere. Doesn't it?

demo/src/app/shared/analytics/analytics.ts
jnizet: First, I just thought injecting the service was necessary to instantiate it, and injecting it to do nothing with it looked weird. So I thought leaving the responsibility to the main component to decide to start tracking page views looked better to me.

Second, although this is untested, not putting this logic in the constructor makes it more easily testable. Basically the same reason why the angular doc encourages the use of ngOnInit rather than the constructor to start interacting with services injected in the constructor: https://angular.io/docs/ts/latest/tutorial/toh-pt4.html#!#the-ngoninit-lifecycle-hook

pkozlowski-opensource: OK, let's go with your proposal for now.

PR #681, opened 2016-09-05
chore(modal): fixes typos

sendilkumarn:
Few really simple typos

### DIFF
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -8,13 +8,13 @@ import {NgbModalRef} from './modal-ref';
  */
 export interface NgbModalOptions {
   /**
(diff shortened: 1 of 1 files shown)


src/modal/modal.ts
```diff
@@ -25,7 +25,7 @@ export interface NgbModalOptions {
 }
 
 /**
- * A service to open modal windows. Creating modals is straightforward: create a template and pass it as an argument to
+ * A service to open modal windows. Creating modal is straightforward: create a template and pass it as an argument to
```
jnizet: English is not my mother tongue, but shouldn't it be "Creating **a** modal"?

sendilkumarn: that makes sense.

PR #685, opened 2016-09-05
chore(pagination): update to show page count

### LINKED ISSUE #683: Update pagination demo to show how to access page count
Per #682, we should expose the page count on the pagination component.

icfantv:
* update basic pagination demo to show page count.

Fixes #683

### DIFF
--- a/demo/src/app/components/pagination/demos/basic/pagination-basic.html
+++ b/demo/src/app/components/pagination/demos/basic/pagination-basic.html
@@ -1,5 +1,5 @@
 <div>Default pagination</div>
-<ngb-pagination [collectionSize]="70" [(page)]="page"></ngb-pagination>
+<ngb-pagination #pagination [collectionSize]="70" [(page)]="page"></ngb-pagination>
 
 <div>directionLinks = false</div>
 <ngb-pagination [collectionSize]="70" [(page)]="page" [directionLinks]="false"></ngb-pagination>
@@ -10,3 +10,4 @@
 <hr>
 
 <pre>Current page: {{page}}</pre>
+<pre>Total pages: {{pagination.pageCount}}</pre>


icfantv:
Not entirely sure how I feel about this because on one hand, it's purely an ng2 feature not a library one.  But on the other hand, it's inevitable we will get questions on how to do this if we don't have an example.

jnizet:
What is missing, actually, is the documentation for the public properties and methods:

 - the public methods are documented only if `apiDocs.exportAs` is truthy. This looks wrong since, apparently, we can assign a variable even if there is no `exportAs` defined.
 - the properties that are public, but are neither inputs nor outputs are never documented.

@icfantv do you want to make these changes as part of this PR, or should I open an issue (and a PR) about that? 

I think we went through all the properties and methods that shouldn't be exposed before to make them private or internal, but this should be checked if we start exposing them in the api doc.

Regarding the demo, I think it doesn't hurt to have it. Instead of the two `<pre>` blocks at the end, I would show something more realistic such as a message looking like:

*Displaying page {{ page }} of {{ pagination.pageCount }}*

wesleycho:
The proper commit message for this is `docs(pagination): ...`

wesleycho:
After reading this more carefully & thinking about it, it may be worth considering exporting the pagination component. What are your thoughts @pkozlowski-opensource?

pkozlowski-opensource:
tl;dr; I don't we should support / document this use-case. There are more corner cases that we need to take with Misko / others in the core team.

So reading through this discussion I think that we've got several things mixed together here. Let me tackle those one by one.

Firstly we should ask ourselves if it makes sense to export pagination component. My assumption was that _not_ as I didn't see use cases for this. The one that started this discussion (`pageCount`) isn't valid IMO since users of this directive already know total no of items and items per page so the only "logic" we would expose would be `Math.ceil(this.collectionSize / this.pageSize);` - not worth larger API surface at this point, IMO.

Now, the fact that users can get a hand on a component instance without it being exported comes as a surprise for me and I will take it with Misko today to check if this is a desired behavior or not. 

Lastly the documentation concerns - it is true that we should also document public members, at least in theory. But as soon as we do it we are stepping into very, very dangerous territory. If users are permitted to interact with public fields (without going through a method) we've got more or less no way of assuring that the internal state of a component is consistent (please observe that we do synchronize internal state inside the `ngOnChanges()` lifecycle hook and as far as I can see this hook wouldn't be be called if we interact with component instance directly,

jnizet:
I just noticed that getting a hand on a component even though it doesn't has an `exportAs` is the very first thing that is explained in our github wiki : https://github.com/ng-bootstrap/ng-bootstrap/wiki/Code-conventions#instance-variables :-/

icfantv:
Closing this per slack discussion.

PR #695, opened 2016-09-07
fix(dropdown): don't emit duplicate change events

### LINKED ISSUE #694: dropdown: open/close events should only be fired if necessary
Currently, close events are emitted everytime the dropdown is closed, even if it's in fact already closed. In particular, since the dropdown adds a click listener to the document to close itself when a click happens outside the dropdown, a close event is emitted by the dropdown every time we click anywhere on the document. 

The same goes for open events: an event is emitted when the dropdown is opened, even if it's already open.

This doesn't look right to me.

pkozlowski-opensource:
Fixes #694


jnizet:
LGTM

PR #701, opened 2016-09-08
refactor(timepicker): use disabled state from a control

pkozlowski-opensource:
New forms have an additional callback (`setDisabledState`) on the `ControlValueAccessor` - let's use this hook to properly support disabled controls (both template-driven and reactive).

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -728,8 +728,16 @@ describe('ngb-timepicker', () => {
       fixture.detectChanges();
       fieldset = getFieldsetElement(fixture.nativeElement);
       expect(fieldset.hasAttribute('disabled')).toBeFalsy;
-
     });
+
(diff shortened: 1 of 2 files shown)


src/timepicker/timepicker.spec.ts
```diff
@@ -728,8 +728,16 @@ describe('ngb-timepicker', () => {
       fixture.detectChanges();
       fieldset = getFieldsetElement(fixture.nativeElement);
       expect(fieldset.hasAttribute('disabled')).toBeFalsy;
-
     });
+
+    it('should have disabled class, when it is disabled using reactive forms', async(() => {
```
jnizet: You mean disabled *attribute*, right?

jnizet:
Other than that, LGTM.

src/timepicker/timepicker.spec.ts
pkozlowski-opensource: Yeh, good catch, will fix. Thnx

PR #703, opened 2016-09-08
fix(typeahead): properly support disabled inputs

pkozlowski-opensource:
Since `ngModel` was taking over `disabled` binding before this change it was impossible to disable an input containing the typeahead control.


jnizet:
LGTM

PR #708, opened 2016-09-09
refactor: use convenience positioning function

pkozlowski-opensource:
We had the same code block duplicated in 3 directives and with popups for date / timepicker this would be duplicated in even more places. 

This is is simple / mechanic code extract + move.

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -16,7 +16,7 @@ import {
 } from '@angular/core';
 
 import {listenToTriggers} from '../util/triggers';
-import {Positioning} from '../util/positioning';
+import {positionElements} from '../util/positioning';
 import {PopupService} from '../util/popup';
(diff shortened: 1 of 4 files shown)


wesleycho:
Would it be better to make this an injectable, or do you think keeping it more in pure JS is better for this service?

If it is preferred to go with this approach, this LGTM.

pkozlowski-opensource:
> Would it be better to make this an injectable, or do you think keeping it more in pure JS is better for this service?

ATM injectable would be more hassle with no apparent benefit. We might make it injectable in the future if there are new dependencies / state in there but for now it would just force us to inject / assign things without buying us anything.

wesleycho:
I'm fine with that argument - I was wondering if the weight of that is enough to go with this approach.

This is good to merge.

PR #711, opened 2016-09-09
feat(timepicker): propagate second as 0 if seconds not displayed

jnizet:
also make sure the model doesn't stay null when invalid due to incorrect second and seconds are hidden

### DIFF
--- a/demo/src/app/components/timepicker/demos/seconds/timepicker-seconds.ts
+++ b/demo/src/app/components/timepicker/demos/seconds/timepicker-seconds.ts
@@ -9,6 +9,6 @@ export class NgbdTimepickerSeconds {
   seconds = true;
 
   toggleSeconds() {
-      this.seconds = !this.seconds;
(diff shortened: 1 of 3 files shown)


demo/src/app/components/timepicker/demos/seconds/timepicker-seconds.ts
```diff
@@ -9,6 +9,6 @@ export class NgbdTimepickerSeconds {
   seconds = true;
 
   toggleSeconds() {
-      this.seconds = !this.seconds;
+    this.seconds = !this.seconds;
```
pkozlowski-opensource: Is this clang-format, manual edit or sth else? Just wondering how this edit happened.

jnizet: I think it's manual edit. I must have been bothered by the unusual indentation level.

Issue #706, opened 2016-09-09
datepicker: visual improvements


pkozlowski-opensource:
Datepicker's look & feel could use some love. Few ideas:
- [ ] get more space between days and different sections
- [x] we should be changing day's background in datepickers to light gray or sth when a user mouse-hovers over a given day.
- [ ] make navigation part visually separated from the calendar part

c1rus:
What about adding some extra classes on current date and weekends?

skendrot:
Looking for this as well. The current spacing is too tight and makes it unable to view the year in the dropdown

PR #713, opened 2016-09-09
feat(datepicker): add background color when hovering day

jnizet:
part of #706

### DIFF
--- a/src/datepicker/datepicker-day-view.ts
+++ b/src/datepicker/datepicker-day-view.ts
@@ -9,7 +9,7 @@ import {Component, Input} from '@angular/core';
       border-radius: 0.25rem;
     }
   `],
-  host: {'[class.bg-primary]': 'selected', '[class.text-muted]': 'isMuted()'},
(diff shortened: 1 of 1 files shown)


src/datepicker/datepicker-month-view.ts
```diff
@@ -17,6 +17,9 @@ import {DayTemplateContext} from './datepicker-day-template-context';
       height: 100%;
       cursor: pointer;
     }
+    .day:not(.disabled):hover {
+      background-color: rgba(0, 0, 0, 0.075);
```
pkozlowski-opensource: We can't do this as it would kill theming... We need to find a BS class instead to apply on mouseover...

jnizet: Oh, right. Let me try something else.

pkozlowski-opensource:
OK, this helps. IMO we will need to do bigger re-design to HTML / CSS structure to improve look & feel as well as have proper accessibility, but this PR improves the situation. Thnx.

PR #715, opened 2016-09-10
feat(datepicker): datepickers in popups

pkozlowski-opensource:
This is ready for the review now. There are still a number of corner cases that we should take care about but IMO we should get it to the user hands ASAP.

I will open small, focused issues for all the remaining items as soon as this one lands.


src/datepicker/datepicker-input.ts
```diff
@@ -0,0 +1,204 @@
...
+  /**
+   * @internal
+   */
+  manualDateChange(value: string) {
+    this._model = this._parserFormatter.parse(value);
+    this._onChange({year: this._model.year, month: this._model.month, day: this._model.day});
+    this._writeModelValue(this._model);
+  }
+
+  isOpen() { return !!this._cRef; }
```
wesleycho: This would probably be better as a getter

src/datepicker/datepicker-input.spec.ts
```diff
@@ -0,0 +1,220 @@
+import {TestBed, ComponentFixture, fakeAsync, tick} from '@angular/core/testing';
+import {By} from '@angular/platform-browser';
+import {createGenericTestComponent} from '../test/common';
+
+import {Component} from '@angular/core';
+import {FormsModule} from '@angular/forms';
+
+import {NgbDatepickerModule} from './datepicker.module';
+import {NgbInputDatepicker} from './datepicker-input';
```
jnizet: Shouldn't the directive be named NgbInputDatepicker, or the file be named input-datepicker?

wesleycho: I'm fine with this - the regular datepicker is `NgbDatepicker`, we namespace all of these for clarity that these are third party components & it creates a nice user cognition that they should be prefixing the ngb directives/components.

src/datepicker/datepicker-input.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * First day of the week, 0=Sun, 1=Mon, etc.
+   */
+  @Input() firstDayOfWeek: number;
+
+  /**
+   * Callback to mark a given date as disabled
+   */
+  @Input() markDisabled: (date: {year: number, month: number, day: number}) => boolean;
```
jnizet: Couldn't we create an interface for this structure. What would be the disadvantage of doing so? It would make our code easier to read (IMO), and the client code safer too, because it could use that interface too, and thus avoid typos like `seconds` instead of `second`?

wesleycho: Isn't it basically a NgbDate?

src/datepicker/datepicker-input.ts
```diff
@@ -0,0 +1,204 @@
...
+  /**
+   * Whether to display week numbers
+   */
+  @Input() showWeekNumbers: boolean;
+
+  /**
+   * Date to open calendar with. If nothing provided, calendar will open with current month.
+   * Use 'navigateTo(date)' as an alternative
+   */
+  @Input() startDate: {year: number, month: number};
```
jnizet: I think this should be renamed to `startMonth` (and in datepicker, too).

src/datepicker/datepicker-input.ts
pkozlowski-opensource: Guys, this is not part of this PR, it is moved over from the NgbDatePicker. If you want to tackle this let's do it in a separate issue - this PR is a big one already. My quick take on it:
* yes, let's create an interface
* no, it is not the same as `NgbDate` - `NgbDate` is meant to be internal and shouldn't be exposed to the end users

src/datepicker/datepicker-input.ts
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  close() {
+    if (this.isOpen()) {
+      this._vcRef.remove(this._vcRef.indexOf(this._cRef.hostView));
+      this._cRef = null;
+    }
+  }
+
+  toggle() {
```
jnizet: These methods should be documented.

src/datepicker/datepicker-input.ts
pkozlowski-opensource: Once again, this is not really part of this PR. If this needs to be discussed please create a separate issue.

src/datepicker/datepicker-input.ts
pkozlowski-opensource: Oh yes! Good catch, thnx!

demo/src/app/components/datepicker/demos/popup/datepicker-popup.html
```diff
@@ -0,0 +1,14 @@
+<form class="form-inline">
+  <div class="form-group">
+    <div class="input-group">
+      <input class="form-control" placeholder="yyyy-mm-dd"
+             name="dp" [(ngModel)]="model" ngbDatepicker #d="ngbDatepicker">
+      <div class="input-group-addon" (click)="d.toggle()" >
+        <img src="img/calendar-icon.svg" style="width: 1.2rem; height: 1.2rem; cursor: pointer;"/>
```
jnizet: Here, on Firefox and Chrome, the height of the addon is a bit too large. Setting the height to 1 rem makes it perfectly aligned with the input.

jnizet:
This looks really great. Congrats. 👍 

I'm not sure if it's one of the known bugs that you plan to fix later, so I'll report it here. From the demo, open the popup, pick a date, erase the last two digits from the input field, click elsewhere in the page, then click the addon button again: an exception is thrown (`Cannot read property 'year' of null at NgbInputDatepicker.manualDateChange`), and it's not possible to open the popup anymore without refreshing the page.

pkozlowski-opensource:
> This looks really great. Congrats.

thnx @jnizet 

> I'm not sure if it's one of the known bugs that you plan to fix later, so I'll report it here. From the demo, open the popup, pick a date, erase the last two digits from the input field, click elsewhere in the page, then click the addon button again: an exception is thrown (Cannot read property 'year' of null at NgbInputDatepicker.manualDateChange), and it's not possible to open the popup anymore without refreshing the page.

There are still many, many corner cases around invalid dates / inputs, validation and popup opening. I will fix the one you've found but there are many more tests we should add to make it really robust. In any case we are still in alpha so as long as the happy path works I want people to play with it.

But yeh, it definitively needs more polish.

demo/src/app/components/datepicker/demos/popup/datepicker-popup.html
maxokorokov: I would say just remove height altogether and leave width 1.2rem

maxokorokov:
Looks quite good for the initial version!
Sure the validation/conversion/focus/blur should be improved later.

One thing that needs to be fixed for sure for me is when you type something invalid like `aaa` in the input and try to open, it crashes the datepicker forever with `TypeError: Attempted to assign to readonly property.` To me the first version should eat all invalid inputs silently and probably avoid formatting `2016` to `2016--`

Another is that there is no documentation on the `NgbInputDatepicker` in the demo currently.

It's sad that we have all datepicker inputs in three places, but I can't see what to do with it right away...

src/datepicker/datepicker-input.ts
pkozlowski-opensource: `isOpen()` is consistent with what we've got in other directives (dropdown, typeahead, ...). Please open a separate issue if you think that we should change it across the project.

pkozlowski-opensource:
@jnizet @maxokorokov @wesleycho once again thnx for all your reviews! I went through all the comments and fixed pbs or commented on your comments :-)

I've also fixed bugs you've seen - I'm sure that more remain but yeh, that's just a starting point.

wesleycho:
I'm fine with this as it is for now.

PR #718, opened 2016-09-10
feat(datepicker): add support for disabled datepicker

### LINKED ISSUE #702: datepicker: support disabled datepickers
Same as #701 but for datepickers.

jnizet:
fix #702

### DIFF
--- a/src/datepicker/datepicker-month-view.spec.ts
+++ b/src/datepicker/datepicker-month-view.spec.ts
@@ -104,7 +104,23 @@ describe('ngbDatepickerMonthView', () => {
     const dates = getDates(fixture.nativeElement);
     dates[0].click();
 
-    expect(fixture.componentInstance.onClick).not.toHaveBeenCalledWith();
+    expect(fixture.componentInstance.onClick).not.toHaveBeenCalled();
+  });
+
+  it('should not send date selection events if disabled', () => {
+    const fixture = createTestComponent(`
+        <template #tpl let-date="date">{{ date.day }}</template>
+        <tbody ngbDatepickerMonthView [month]="month" [dayTemplate]="tpl" [disabled]="true" (select)="onClick($event)"></tbody>
+      `);
+
+    fixture.detectChanges();
+
+    spyOn(fixture.componentInstance, 'onClick');
+
+    const dates = getDates(fixture.nativeElement);
+    dates[0].click();
+
+    expect(fixture.componentInstance.onClick).not.toHaveBeenCalled();
   });
 
   if (!isBrowser('ie9')) {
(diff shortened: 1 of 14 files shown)


pkozlowski-opensource:
Wooohooo, thnx for this, it was blocking me for another PR :-) Will review tomorrow when fresh.

pkozlowski-opensource:
@maxokorokov would you mind reviewing this one?

demo/src/app/components/datepicker/demos/basic/datepicker-basic.ts
```diff
@@ -9,8 +9,13 @@ const now = new Date();
 export class NgbdDatepickerBasic {
 
   model;
+  disabled = false;
 
   selectToday() {
     this.model = {year: now.getFullYear(), month: now.getMonth(), day: now.getDate()};
   }
+
+  toggleDisabled() {
```
pkozlowski-opensource: We could probably in-line this in a template. Having said this I'm not sure it makes sense to demonstrate it as part of the very basic demo. If I recall properly we want to have to the basic demo as simple as possible (that is, use default values for all props and provide only ones that are must-have for the most rudimentary demo).

src/datepicker/datepicker.ts
```diff
@@ -41,6 +41,7 @@ const NGB_DATEPICKER_VALUE_ACCESSOR = {
         [dayTemplate]="dayTemplate || dt"
         [showWeekdays]="showWeekdays"
         [showWeekNumbers]="showWeekNumbers"
+        [disabled]="disabled"
```
pkozlowski-opensource: shouldn't we disable navigation as well?

jnizet: I wasn't sure, so I chose the easy route :-)

But yes, that's another thing we have to decide. It isn't strictly necessary, but it might make things clearer. If we choose to do it, I guess it would also be a good idea to switch to the month of the selected date when it becomes disabled.

jnizet:
Rebased and added a specific demo for the disabled state.
Added a second commit to also disable the navigation. Not sure yet if it's a good idea or not, but since it's in a separate commit, it'll be easy to pick one or both depending on what we choose.

maxokorokov:
LGTM. And I think there is no point in leaving navigation enabled. But no strong opinions on this one, as I guess that the datepicker will rarely be used inline. And in the case of an input + datepicker in the popup its the input that should be disabled, right?

src/datepicker/datepicker-month-view.spec.ts
```diff
@@ -107,6 +107,22 @@ describe('ngbDatepickerMonthView', () => {
...
+      `);
+
+    fixture.detectChanges();
+
+    spyOn(fixture.componentInstance, 'onClick');
+
+    const dates = getDates(fixture.nativeElement);
+    dates[0].click();
+
+    expect(fixture.componentInstance.onClick).not.toHaveBeenCalledWith();
```
maxokorokov: better be `not.toHaveBeenCalled()` but I guess it's my mistake from the test case above (line 107) :)

jnizet:
Replaced `toHaveBeenCalledWith()` by `toHaveBeenCalled()`.

PR #721, opened 2016-09-11
WIP(accordion): allow having parts of the header that do not toggle. Do not merge.

### LINKED ISSUE #717: Accordion Feature Request - Title Content that does not cause expand/collapse

For example...

``` html
<ngb-accordion>
  <ngb-panel>
    <template ngbPanelTitle>
      <span>Non toggle text</span>
    </template>
    <template ngbPanelContent>
        Some Settings
    </template>
  </ngb-panel>
</ngb-accordion>
```

``` html
  <ngb-panel id="1">
    <template ngbPanelHeader>
      <span>Non toggle text</span>
    </template>
    <template ngbPanelContent>
        Some Settings
    </template>
  </ngb-panel>
</ngb-accordion>
```

jnizet:
fix #717

This is an early PR to fix #717. Since the decision on what the best design should be has not been taken yet, consider this as a WIP, or as an experiment to implement a solution.

If this solution is adopted, we should also decide if it's worth keeping NgbPanelTitle, since everything it does can be done with NgbPanelHeader, without too much hassle.

### DIFF
--- a/demo/src/app/components/timepicker/demos/seconds/timepicker-seconds.ts
+++ b/demo/src/app/components/timepicker/demos/seconds/timepicker-seconds.ts
@@ -9,6 +9,6 @@ export class NgbdTimepickerSeconds {
   seconds = true;
 
   toggleSeconds() {
-      this.seconds = !this.seconds;
+    this.seconds = !this.seconds;
   }
 }
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -861,6 +861,95 @@ describe('ngb-timepicker', () => {
       expectSameValues(timepicker, config);
     });
   });
+
+  describe('Seconds handling', () => {
+    it('should propagate seconds to 0 in model if seconds not shown and no second in initial model', async(() => {
+         const html = `<ngb-timepicker [(ngModel)]="model" [seconds]="false"></ngb-timepicker>`;
+
+         const fixture = createTestComponent(html);
+         fixture.componentInstance.model = {hour: 10, minute: 30};
+         fixture.detectChanges();
+         fixture.whenStable()
+             .then(() => {
+               fixture.detectChanges();
+               return fixture.whenStable();
+             })
+             .then(() => {
+               const inputs = fixture.debugElement.queryAll(By.css('input'));
+
+               inputs[1].triggerEventHandler('change', createChangeEvent('40'));
+               fixture.detectChanges();
+               expectToDisplayTime(fixture.nativeElement, '10:40');
+               expect(fixture.componentInstance.model).toEqual({hour: 10, minute: 40, second: 0});
+             });
+       }));
+
+    it('should propagate second as 0 in model if seconds not shown and null initial model', async(() => {
+         const html = `<ngb-timepicker [(ngModel)]="model" [seconds]="false"></ngb-timepicker>`;
+
+         const fixture = createTestComponent(html);
+         fixture.detectChanges();
+         fixture.whenStable()
+             .then(() => {
+               fixture.detectChanges();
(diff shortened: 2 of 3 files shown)


jhuntoo:
On the topic of whether to keep NgbPanelTitle or not, I think it makes some sense to remove it.

- The quick out-of-box experience is covered by
```html
<ngb-panel title="Simple">
```
-  Like @jnizet said, it's easy enough to achieve the same with NgbPanelHeader for those folks who are customising the header anyway.

- Number of supported methods for defining the header goes from 3 to 2 👍.

src/accordion/accordion.ts
```diff
@@ -94,8 +125,10 @@ export interface NgbPanelChangeEvent {
   <div class="card">
     <template ngFor let-panel [ngForOf]="panels">
       <div [class]="'card-header ' + (panel.type ? 'card-'+panel.type: type ? 'card-'+type : '')" [class.active]="isOpen(panel.id)">
-        <a tabindex="0" href (click)="!!toggle(panel.id)" [class.text-muted]="panel.disabled">
-          {{panel.title}}<template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></template>          
+        <template *ngIf="panel.headerTpl" [ngTemplateOutlet]="panel.headerTpl?.templateRef" [ngOutletContext]="getHeaderContext(panel)">
+        </template>
+        <a *ngIf="!panel.headerTpl" tabindex="0" href (click)="!!toggle(panel.id)" [class.text-muted]="panel.disabled">
```
pkozlowski-opensource: With this change people will have to manage `[class.text-muted]="panel.disabled"` on their side, right? But currently this is not part of the context.

We should explore if we can move `[class.text-muted]="panel.disabled"` higher up or expose it as part of the context.

src/accordion/accordion.ts
```diff
@@ -32,6 +32,36 @@ export class NgbPanelContent {
...
+ */
+@Directive({selector: 'template[ngbPanelHeader]'})
+export class NgbPanelHeader {
+  constructor(public templateRef: TemplateRef<any>) {}
+}
+
+/**
+ * The context available to the NgbPanelHeader directive
+ */
+export class NgbPanelHeaderContext implements NgbPanelHeaderContext {
```
pkozlowski-opensource: Class and interface with the same name? We should distinguish those to avoid confusion. `NgbPanelHeaderContextImpl` for the implementation class?

jnizet: That's a mistake. There should only be a class. I just forgot to remove the interface. Not even sure why that compiles BTW. I originally started with just a class, then noticed that regular classes without any decorator were not documentetd, so introduced an interface. Then I thought that I should rather improve the documentation process and removed the interface, but forgot this implements clause.

src/accordion/accordion.ts
jnizet: Indeed. I went back and forth between exposing the NgbPanel itself in the context, and not doing it. It's only useful for this `disabled` flag, but the disabled flag itself is normally coming from the model, which is accessible from the template. So I finally decided to not expose the panel in the context.

pkozlowski-opensource:
This looks to me generally speaking apart from the comment I've left about the text-mutted class. At the same time I must say that I don't feel like I can evaluate all the impact at this point as we are still missing:
* full accessibility support
* keyboard navigation support
* animations support

Given the above my worry is that a user will have to handle more and more on its side. Once again, I can't see any specific problems, just don't know how the code is going to look like if we add support for all the mentioned points.

@jnizet how about you open a separate PR for the API doc script changes and we think a bit more about this one?

jnizet:
OK, not problem. I'll do that a bit later.

PR #725, opened 2016-09-12
chore: document interface methods and commented classes

jnizet:
Documenting methods in interfaces is not needed yet but will probably be in the future.
Documenting classes which don't have any decorator is useful for context classes, that could be used later, in accordion for example


pkozlowski-opensource:
Thnx!

PR #726, opened 2016-09-13
chore: use ViewChild to make some tests simpler

jnizet:
Just discovered that ViewChild allowed getting a reference to the actual component under test, which thus allows to check its state without going through DOM manipulation to check that it has the expected state.

### DIFF
--- a/src/alert/alert.spec.ts
+++ b/src/alert/alert.spec.ts
@@ -1,7 +1,7 @@
 import {fakeAsync, tick, TestBed, ComponentFixture, inject} from '@angular/core/testing';
 import {createGenericTestComponent} from '../test/common';
 
-import {Component} from '@angular/core';
+import {Component, ViewChild} from '@angular/core';
 
 import {NgbAlertModule} from './alert.module';
 import {NgbAlert, NgbSelfClosingAlert} from './alert';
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
@jnizet I don't think we should do this. Your change indeed makes tests simpler but at the expense of testing less. From what I understand now we will just assert state of a component but not the fact that this state was reflected in the DOM. If, for example I forget to add DOM or host binding a test would pass but a component would be broken, right?

In short: we are not going through the DOM becouse there is no easier way - we do so to assert that the proper DOM gets rendered.

Unless I've missed sth in my hasty review :-)

jnizet:
I agree in general. But for the tests I modified, the effect on the DOM of the inputs has already been tested by other tests. The actual point of those test is just to check that the config configured on the module level affects the inputs of the component correctly. The DOM effects of these inputs are tested by other tests already.

It could be applied elsewhere. For example, the datepicker component uses a sub-component for the navigation. The navigation component itself is unit-tested, and checks, for example, that the `disabled` input disables the buttons and the select boxes of the navigation in the DOM. When testing the parent datepicker component, all we need to check is that disabling the datepicker disables the navigation component (and thus, transitively, the DOM of the navigation component). Testing the DOM of the navigation component in the datepicker component doesn't hurt, especially if it's easy to do so, but it is redundant. Just checking that disabling the datepicker correctly sets the disabled input of the navigation to true should be enough.

pkozlowski-opensource:
Oh, I see - you are doing it only for the config part. LGTM then!

PR #733, opened 2016-09-14
feat(typeahead): allow veto of an item selection

### LINKED ISSUE #699: Typeahead: allow selection veto
In some more advanced use-cases a typeahead might display more options than are supposed to be selected by a user. An example of this can be seen on https://www.kayak.com/flights where in addition to flight proposal there is the "Sign up to access recent searches anywhere". It looks like a "normal" option (is displayed in a dropdown, can be iterated over with keyboard etc.) except that it can't be selected and shouldn't result in model binding.

We could easily cover this use-case by giving people ability to veto selection in the `selectItem` callback (ex. by interpreting `false` as a return value or changing event we pass to the `selectItem` function).

pkozlowski-opensource:
Closes #699

BREAKING CHANGE: payload of the typeahead's `selectItem` event was changed - now it is
an object implementing the `NgbTypeaheadSelectItemEvent`. You can easily migrate your existing
code by changing:

`<input [ngbTypeahead]="find" (selectItem)="onSelect($event)"/>`

to

`<input [ngbTypeahead]="find" (selectItem)="onSelect($event.item)"/>`

### DIFF
--- a/demo/src/app/components/typeahead/typeahead.component.ts
+++ b/demo/src/app/components/typeahead/typeahead.component.ts
@@ -6,6 +6,7 @@ import {DEMO_SNIPPETS} from './demos';
   template: `
     <ngbd-content-wrapper component="Typeahead">
       <ngbd-api-docs directive="NgbTypeahead"></ngbd-api-docs>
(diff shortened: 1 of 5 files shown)


wesleycho:
LGTM

PR #745, opened 2016-09-16
fix(tooltip): hide, or avoid opening the tooltip if its value is falsy

### LINKED ISSUE #737: Feature request: Tooltip enable check
LIke ui-bootstrap, having an `enable` function would help in the conditionally showing the tooltip.

My personal use case is checking if there is anything to display, otherwise I end up seeing just a tooltip arrow.

jnizet:
fix #737

### DIFF
--- a/src/tooltip/tooltip.spec.ts
+++ b/src/tooltip/tooltip.spec.ts
@@ -76,6 +76,30 @@ describe('ngb-tooltip', () => {
       expect(getWindow(fixture)).toBeNull();
     });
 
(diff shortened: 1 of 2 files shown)


anrepppp:
I am still seeing the popover arrow when there is no content.

anrepppp:
<template *ngIf="product.pdqLocationList.length > 0" #pdqContent>
</template>
<div class="col-md-4"
                            [ngbPopover]="pdqContent">
                            {{product.qtyDC}}
                          </div>
I tried the above approach.

PR #747, opened 2016-09-16
fix(tabset): rename change event to avoid conflicts with native ones

### LINKED ISSUE #741: Tabset (change) event gets fired in non-related situations.
### Bug description:

Basically what happens is that if an <input type="text"> is inside a tab, any method associated with the tabset's (change) event will get fired if the input's value changes (and the input loses focus).

So the event gets triggered if either a selected tab changes or if an input field's value changes inside the tab changes.

My personal case is the following. On a (change) event I run code that changes the route to include the open tab in the parameters. One of the tabs has a form. By changing a form input value in that form, the route change function runs and another tab opens.

To make it work I'm using this hack:
```
onTabSelect(data: NgbTabChangeEvent) {
if (data.hasOwnProperty('nextId')) { 
            this._router.navigate(...);
        }   
}
```

Angular: 2.0.0

ng-bootstrap: 1.0.0.alpha-4

Bootstrap: 4.0.0.alpha-4

EDIT: I now realize I probably should have replied here instead: #594

pkozlowski-opensource:
Fixes #741

### DIFF
--- a/demo/src/app/components/tabset/demos/preventchange/tabset-preventchange.html
+++ b/demo/src/app/components/tabset/demos/preventchange/tabset-preventchange.html
@@ -1,4 +1,4 @@
-<ngb-tabset  (change)="beforeChange($event)">
+<ngb-tabset  (tabChange)="beforeChange($event)">
   <ngb-tab title="Simple" id="foo">
     <template ngbTabContent>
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
> but it's a breaking change, so the commit comment should mention it, shouldn't it?

Definitively should! Thnx. Doing way too many things at the some time knowing perfectly that multi-tasking doesn't work...

Issue #754, opened 2016-09-17
NgbDate format and initial model?


asadsahi:
Two questions as I cannot find details anywhere:

1) how can I enforce NgbDate to follow dateformat like 'dd/MM/yyyy'.
2) I set initial model to empty string or '2016-12-10', both ended up in this initial date:

![image](https://cloud.githubusercontent.com/assets/3470892/18607423/c8e06582-7cc3-11e6-8de1-f1fe507d8c74.png)
 
How I can set the initial date based on a certain set format, e.g in my case its en-gb culture?

Thanks

pkozlowski-opensource:
There are 2 separate, although related parts of your question, let me tackle it one by one.

1) How to control date parsing (from user input) and formatting (in the input field).

As of today this is abstracted away behind the `NgbDateParserFormatter` contract, see here: https://github.com/ng-bootstrap/ng-bootstrap/blob/64eca2d4df317ce211e32f8e5357dc6df0d02b2c/src/datepicker/ngb-date-parser-formatter.ts#L4-L7

By default the datepicker comes with the basic implementation of this interface that just accepts dates in the ISO format. If you want to handle a different format (or multiple formats) you can provide your own implementation of the  `NgbDateParserFormatter` and register it as a provider in your module.

We might have a separate implementation in the future (say `NgbMomentJSParserFormatter` or a completely custom one) depending on the interest from the community. In any case you've got a flex point today. 

2) How do use strings as a model 

There is not great story for this as there is a missing robust support for model transformation in the angular itself. See discussion in the timepicker-related issue for more details: #545

pkozlowski-opensource:
Related: http://stackoverflow.com/q/39546645/1418796

asadsahi:
Thanks @pkozlowski-opensource great explanation as always. Will try to implement custom formatter as you mentioned.

jnizet:
@pkozlowski-opensource if that is the recommended way (for now at least) to do that, shouldn't we export and document NgbDateParserFormatter?

pkozlowski-opensource:
>  shouldn't we export and document NgbDateParserFormatter?

We definitively should!!!

jnizet:
OK. I'll do it, unless you want to.

pkozlowski-opensource:
> OK. I'll do it, unless you want to.

Go ahead @jnizet - PR would be great!

nrobinaubertin:
I've created a small example for those who are looking for an actual implementation.
[...]

PR #755, opened 2016-09-17
feat(datepicker): allow parsing and formatting dates in a custom way

jnizet:
refs #754

Note that I had to change the method definitions of NgbDateParserFormatter, because they used NgbDate, which is supposed to stay internal, although NgbDateParserFormatter is supposed to be exported. I thus used NgbDateStruct instead, and added the necessary transformation from NgbDateStruct to NgbDate in the directive. Transforming NgbDate to NgbDateStruct when formatting isn't necessary, because NgbDate implements NgbDateStruct (loosely).

Also, I tried my best to explain what implementations should do, but didn't add a demo, because I think it's a little soon to promote the usage of that class. Easier custom formatting will probably come sooner or later, and the contract of the interface might change once we polish the directive, for example to make the model null and the control invalid when the entered date is partial (2016-01) or invalid (2016-01-40).


pkozlowski-opensource:
> Note that I had to change the method definitions of NgbDateParserFormatter, because they used NgbDate, which is supposed to stay internal, although NgbDateParserFormatter is supposed to be exported.

This is a good move.

PR #758, opened 2016-09-17
feat(alert): remove self-closing alert component

jnizet:
BREAKING CHANGE: the `NgbSelfClosingAlert` component has been removed.
Check the self-closing alert demo to know how to achieve the same thing with `NgbAlert`.


Hongbo-Miao:
The old `dismissOnTimeout` is actually quite convenient way...

Hongbo-Miao:
@jnizet can I ask why you remove the `@Input` of `dismissOnTimeout`? Thanks

Since it is actually quite convenient way especially when having an array of Alerts and using with ngrx/store.

PR #763, opened 2016-09-18
demo(tabset): illustrate proper method of switching active tab

### LINKED ISSUE #594: Tabset API discussion
Following my pull request #586 has brought in some interesting design discussion that I think should be discussed as a whole.

* `activeId` should support 'banana-in-the-box' two-way binding. It is simply intuitive and no boiler plating required for keeping the user's bound field in-sync when the tab changes via mouse click.
* the `select` method should probably be internal use only, since it's main usage is for user's clicking the tab via mouse and the code tests whether it can change or not.
* `change` output event, at a first glance, appears redundant for preventing users from selecting a tab, since the tab itself has a `disable` field. However, I think if we modify `change` to handle promise/observable callback prevention this will support a more advance case (eg. a modal dialog asking whether the user wants to change tab or not).
* 'banana-in-the-box' can be split out to listen to the post-change event separately, however adding a `post-change` event that provides more in-depth details about the tab change that occurred (eg. from/to id) might also be desired. I personally don't have a use case with this, I can only think of is a user wants to track their user's behavior.

### LINKED ISSUE #762: activeId is not updated
I'm not sure if this is a bug or "_by design_", the `activeId` is not updated automatically when client click any tab!
This will force developer to add `change` event to keep track of the `activeId` with something like:

```
beforeChangeTab(e) {
    this.selectedTab = e.nextId;
  }
```

You can see this in your demo('Select an active tab by id'), for example try:

1. Open [Tabs demo](https://ng-bootstrap.github.io/#/components/tabs).
2. Scroll to 'Select an active tab by id' demo.
3. Click on 'Fancy title' tab.
4. Click 'Toggle selected tab'. **This should select 'Simple' tab, but in reality nothing happens.**

pkozlowski-opensource:
Closes #762
Closes #594

### DIFF
--- a/demo/src/app/components/tabset/demos/selectbyid/tabset-selectbyid.html
+++ b/demo/src/app/components/tabset/demos/selectbyid/tabset-selectbyid.html
@@ -1,4 +1,4 @@
-<ngb-tabset [activeId]="selectedId">
+<ngb-tabset #t="ngbTabset">
   <ngb-tab title="Simple" id="foo">
     <template ngbTabContent>
       <p>Raw denim you probably haven't heard of them jean shorts Austin. Nesciunt tofu stumptown aliqua, retro synth
@@ -21,5 +21,5 @@
 </ngb-tabset>
 
 <p>
-  <button class="btn btn-outline-primary" (click)="toggleSelected()">Toogle selected tab</button>
+  <button class="btn btn-outline-primary" (click)="t.select('bar')">Selected tab with "bar" id</button>
 </p>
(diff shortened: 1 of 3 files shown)


jnizet:
LGTM. But I wonder if it wouldn't be a good idea to demo how to use ViewChild to get a reference to the tabset from the demo component. 

I agree it's standard Angular stuff, but I think it's not well-known, and that in most use-cases where the user wants to programmatically select a tab, this will be done from a method of the component, where passing a reference to the tabset is not ideal or possible.

pkozlowski-opensource:
> But I wonder if it wouldn't be a good idea to demo how to use ViewChild to get a reference to the tabset from the demo component.

We can definitively do this later on if we see that there is sth that people are struggling with.

PR #764, opened 2016-09-18
demo(datepicker): add demo for custom day view

### LINKED ISSUE #761: datepicker: customize styling cells based on cell date ("i.e. customClass datepicker-options")
### Bug description:

In the "Ui Bootstrap" datepicker, it was possible to customize the look and feel of a single cell, based on a callback method. From the `[documentation](https://angular-ui.github.io/bootstrap/#/datepicker)`:

> datepicker-options $ - An object to configure the datepicker in one place. 
    customClass ({date: date, mode: mode}) - An optional expression to add classes based on passing an object with date and current mode properties.

I'd like this functionality, so as to use the datepicker component as a sort of 'minimal' calendar, where certain kind of events are displayed in a certain color. Clicking that date would then show more details of that event.

jnizet:
fix #761

### DIFF
--- a/demo/src/app/components/datepicker/datepicker.component.ts
+++ b/demo/src/app/components/datepicker/datepicker.component.ts
@@ -26,6 +26,11 @@ import {DEMO_SNIPPETS} from './demos';
                         [tsSnippet]="snippets.i18n.code">
         <ngbd-datepicker-i18n></ngbd-datepicker-i18n>
       </ngbd-example-box>
+      <ngbd-example-box demoTitle="Custom day view" 
+                        [htmlSnippet]="snippets.customday.markup" 
+                        [tsSnippet]="snippets.customday.code">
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
@jnizet thnx for the PR - we definitively need this demo. But I would have 2 remarks:
* how about we use the popup version of the datepicker as this is what most people are going to use in practice?
* IMO we should do the visual difference much more pronounced. Italic part is hardly visible and border is very light

Maybe we should simplify it and just change the background color for, say Sundays or weekends? I would also probably avoid creating a directive for this demo - how about we've got sth simple that focuses on mechanics of supplying a custom day template?

jnizet:
OK, will do that.

jnizet:
@pkozlowski-opensource I made it simpler and more obviously modified. Here's how it looks:

capture d ecran 2016-09-18 a 17 22 09

PR #766, opened 2016-09-18
chore(build): remove internal on ControlValueAccessor methods

jnizet:
The `@internal` comment annotation causes the methods not to be included in the generated .d.ts file, which in turn cause error messages when building the demo:

```
error TS2420: Class 'NgbDatepicker' incorrectly implements interface 'ControlValueAccessor'.
  Property 'writeValue' is missing in type 'NgbDatepicker'.
```

I thus removed the `@internal`, and added these methods in the list of angular lifecycle methods, since we don't want them to be documented.


wesleycho:
I'm fine with this - thoughts @pkozlowski-opensource?

Issue #748, opened 2016-09-16
Prevent auto select of first item in typehead


doorman02:
Is it possible to prevent auto select of the first item in the typehead component? I need this because I want the user to be able to press enter if he wants to search normally and only if an item in the list is selected by the user e.g. using the mouse or arrow keys he should navigate to the item.

pkozlowski-opensource:
Yes, we should add an option for this. It would be rather easy to implement so PRs welcomed (would be wise to wait till #746 lands otherwise you are looking for funny rebases).

kjartanvalur:
Here is a pull request for this issue
https://github.com/ng-bootstrap/ng-bootstrap/pull/767

PR #767, opened 2016-09-19
(Typeahead) New configuration - Auto select item

kjartanvalur:
See issue #748

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -66,7 +66,7 @@ export class NgbTypeaheadWindow {
    */
   @Output('select') selectEvent = new EventEmitter();
 
-  getActive() { return this.results[this.activeIdx]; }
+  getActive() { return this.activeIdx >= 0 ? this.results[this.activeIdx] : null; }
 
   /**
    * @internal
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -37,7 +37,8 @@ export interface ResultTemplateContext {
 })
 export class NgbTypeaheadWindow {
   
-  activeIdx: number;
+  activeIdx: number = 0;
+  autoSelectItem: boolean = true;
 
   /**
    * Typeahead match results to be displayed
@@ -74,7 +75,7 @@ export class NgbTypeaheadWindow {
 
   next() { this.activeIdx = (this.activeIdx + 1) % this.results.length; }
 
-  prev() { this.activeIdx = (this.activeIdx === -1 ? this.results.length - 1 : this.activeIdx - 1); }
+  prev() { this.activeIdx = ( (this.autoSelectItem === true && this.activeIdx === 0) || (this.autoSelectItem === false && this.activeIdx === -1) ? this.results.length - 1 : this.activeIdx - 1); }
 
   /**
    * @internal
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -75,7 +75,9 @@ export class NgbTypeaheadWindow {
 
   next() { this.activeIdx = (this.activeIdx + 1) % this.results.length; }
 
-  prev() { this.activeIdx = ( (this.autoSelectItem === true && this.activeIdx === 0) || (this.autoSelectItem === false && this.activeIdx === -1) ? this.results.length - 1 : this.activeIdx - 1); }
+   prev() { this.activeIdx = ( (this.autoSelectItem === true && this.activeIdx === 0) ||
+   (this.autoSelectItem === false && this.activeIdx === -1)
+   ? this.results.length - 1 : this.activeIdx - 1); }
 
   /**
    * @internal
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -1,6 +1,7 @@
 import {Component, Input, Output, EventEmitter, TemplateRef} from '@angular/core';
 
 import {toString} from '../util/util';
+import {NgbTypeaheadConfig} from './typeahead-config';
 
 /**
  * Context for the typeahead result template in case you want to override the default one
@@ -36,7 +37,6 @@ export interface ResultTemplateContext {
   `
 })
 export class NgbTypeaheadWindow {
-  
   activeIdx: number = 0;
   autoSelectItem: boolean = true;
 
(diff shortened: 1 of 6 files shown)


pkozlowski-opensource:
@kjartanvalur thnx for the PR. In order to consider it for merging please add tests (you can use the existing ones as an example). Check our contrib guidelines for more info: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md

kjartanvalur:
I did add a test to typeahead-config.spec.ts :)

kjartanvalur:
Updated tests, and fixed conflicts with other tests

src/typeahead/typeahead-window.ts
```diff
@@ -73,7 +75,13 @@ export class NgbTypeaheadWindow {
 
   next() { this.activeIdx = (this.activeIdx + 1) % this.results.length; }
 
-  prev() { this.activeIdx = (this.activeIdx === 0 ? this.results.length - 1 : this.activeIdx - 1); }
+  prev() {
+    this.activeIdx =
+        ((this.autoSelectItem === true && this.activeIdx === 0) ||
+                 (this.autoSelectItem === false && this.activeIdx === -1) ?
+             this.results.length - 1 :
+             this.activeIdx - 1);
```
wesleycho: There is way too much logic compacted in these lines - find a way to simplify readabiliity

wesleycho:
This needs tests for all the conditional branching that it introduces in the parse function so the behavior is codified in tests.

src/typeahead/typeahead.ts
```diff
@@ -108,6 +108,11 @@ export class NgbTypeahead implements OnInit,
   @Input() showHint: boolean;
 
   /**
+  * Auto select the first item in the result list
+  */
+  @Input() autoSelectItem: boolean = null;
```
maxokorokov: I don't think you need initialization here, especially to `null`. I think it defaults to `true` in the config

src/typeahead/typeahead-window.ts
```diff
@@ -36,7 +37,8 @@ export interface ResultTemplateContext {
   `
 })
 export class NgbTypeaheadWindow {
-  activeIdx = 0;
+  activeIdx: number = 0;
```
pkozlowski-opensource: Revert this change. We don't need type here is it is inferred.

src/typeahead/typeahead-window.ts
```diff
@@ -36,7 +37,8 @@ export interface ResultTemplateContext {
...
+  autoSelectItem: boolean = true;
```
pkozlowski-opensource: We don't need type here is it is inferred. Remove `boolean`

src/typeahead/typeahead.spec.ts
```diff
@@ -473,7 +473,7 @@ describe('ngb-typeahead', () => {
            const fixture = createTestComponent(`<input type="text" [(ngModel)]="model" 
                 [ngbTypeahead]="findAnywhere" 
                 [inputFormatter]="uppercaseFormatter" 
-                [showHint]="true"/>`);
+                [showHint]="true" [autoSelectItem]="true" />`);
```
pkozlowski-opensource: We should _not_ modify existing tests. `autoSelectItem` should be default. Revert.

src/typeahead/typeahead.spec.ts
```diff
@@ -496,7 +496,7 @@ describe('ngb-typeahead', () => {
 
       it('should restore hint when results window is dismissed', async(() => {
            const fixture = createTestComponent(
-               `<input type="text" [(ngModel)]="model" [ngbTypeahead]="findAnywhere" [showHint]="true"/>`);
+               `<input type="text" [(ngModel)]="model" [ngbTypeahead]="findAnywhere" [showHint]="true" [autoSelectItem]="true" />`);
```
pkozlowski-opensource: We should _not_ modify existing tests. `autoSelectItem` should be default. Revert.

pkozlowski-opensource:
As @wesleycho pointed out this needs much more tests, indeed. We should have at least:
* no item is selected with the new option switched on
* navigating with arrow down properly gets us to the next item
* navigating with arrow up / down properly wraps

src/typeahead/typeahead-window.ts
kjartanvalur: done

src/typeahead/typeahead.ts
kjartanvalur: done

src/typeahead/typeahead-window.ts
kjartanvalur: done

src/typeahead/typeahead-window.ts
kjartanvalur: done

src/typeahead/typeahead.spec.ts
kjartanvalur: done

src/typeahead/typeahead.spec.ts
kjartanvalur: done

src/typeahead/typeahead.spec.ts
```diff
@@ -473,7 +473,7 @@ describe('ngb-typeahead', () => {
            const fixture = createTestComponent(`<input type="text" [(ngModel)]="model" 
                 [ngbTypeahead]="findAnywhere" 
                 [inputFormatter]="uppercaseFormatter" 
-                [showHint]="true"/>`);
+                [showHint]="true" />`);
```
pkozlowski-opensource: Why this change?

src/typeahead/typeahead.spec.ts
```diff
@@ -525,21 +525,26 @@ describe('ngb-typeahead', () => {
...
         expect(typeahead.showHint).toBe(true);
+        expect(typeahead.autoSelect).toBe(false);
       });
     });
 
     describe('Custom config as provider', () => {
       beforeEach(() => {
         const config = new NgbTypeaheadConfig();
         config.showHint = true;
+        config.autoSelect = false;
```
pkozlowski-opensource: I think that the more interesting test would exercise non-default path

src/typeahead/typeahead.ts
```diff
@@ -147,6 +153,10 @@ export class NgbTypeahead implements OnInit,
             this._closePopup();
           } else {
             this._openPopup();
+            if (this.autoSelect !== null && this.autoSelect !== undefined) {
+              this._windowRef.instance.activeIdx = this.autoSelect ? 0 : -1;
+              this._windowRef.instance.autoSelect = this.autoSelect;
```
pkozlowski-opensource: Based on my previous comment I don't think we need `this._windowRef.instance.autoSelect = this.autoSelect`

pkozlowski-opensource:
@kjartanvalur I did another pass on this PR and unfortunately it is not in the state where we could consider it for merging :-( 

Besides comments in the code the PR doesn't merge cleanly (needs rebasing) and we are still tests for several use-cases I've brought up [previously](https://github.com/ng-bootstrap/ng-bootstrap/pull/767#issuecomment-249209093)

Given all the above and the fact that this PR haven't seen any activity in almost 2 weeks I'm going to close it for now. Feel free to open a new PR when you've got time to work on this. Thnx!

PR #777, opened 2016-09-21
fix(tooltip): properly handle components using OnPush strategy

pkozlowski-opensource:
Fixes #772

PR #781, opened 2016-09-21
fix(popover): properly handle components using OnPush strategy

pkozlowski-opensource:
@wesleycho this is essentially the same as #777 - would be great if you could LGTM this one.


wesleycho:
LGTM

PR #785, opened 2016-09-22
fix(dropdown): support autoClose when several dropdowns exist

### LINKED ISSUE #687: Opens several drop-down lists
Is this a bug o new feature?
image
How to disable it?

pkozlowski-opensource:
The essence of the trick is that now we can distinguish clicks coming from the dropdown toggle element and not interpret those as external clicks.

Fixes #687

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -1,4 +1,4 @@
-import {Directive, Input, Output, HostListener, EventEmitter} from '@angular/core';
+import {Directive, Input, Output, EventEmitter, ElementRef} from '@angular/core';
 import {NgbDropdownConfig} from './dropdown-config';
 
 /**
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.ts
```diff
@@ -82,28 +84,47 @@ export class NgbDropdown {
...
       this.close();
     }
   }
+
+  /**
+   * @internal
+   */
+  set toggleElement(toggleElement: any) { this._toggleElement = toggleElement; }
+
+  private _isEventFromToggle($event) { return $event.target === this._toggleElement; }
```
wesleycho: You probably want this to be `$event.target.contains(this._toggleElement)`

pkozlowski-opensource: Nope. I want to precisely capture clicks on a toggle button. Changing this to `this._toggleElement.contains($event.target)` would also capture menu items clicks which is not what we want.

wesleycho:
This LGTM

PR #789, opened 2016-09-23
feat(typeahead): add support for the "editable" option

### LINKED ISSUE #788: Detect typeahead validity with formGroup
It is impossible to detect when a typeahead is valid (meaning that an item as been selected or match one of the list - and not just a random input). The input becomes valid whenever something is inserted. I'm using a model driven approach to forms.

Angular: 2.0.0

ng-bootstrap: 1.0.0-alpha.5

Bootstrap: 4-alpha 4

pkozlowski-opensource:
Closes #788


wesleycho:
LGTM

PR #791, opened 2016-09-23
refactor: prepare code for AoT

mlaval:
The Angular compiler is more strict when compiling (ngc way) the code.
This is also true when compiling (tsc way) the code it produces (the `ngfactory.ts` files).


pkozlowski-opensource:
LGTM, thnx @mlaval ! Will merge it in as soon as TravisCI is green.

PR #794, opened 2016-09-23
chore: upgrade webpack to 2.1.0-beta

mlaval:
Some quick comments:
- with webpack 2.1.0-beta.25, the build bugs (super slow/hangs), so using 2.1.0-beta.22 for now
- couldn't make `alias` work with existing loaders, so all the demo and src code are copied in a `tmp` folder before processing (this is needed for AoT anyway)
- there was an issue to load reflect-metadata, but it is useless as already available in core-js
- vendors and polyfills could be merged

PR #795, opened 2016-09-23
Aot demo2

mlaval:
Follows #794, only the second commit is to be reviewed.

With AoT, loading the demo site goes down from 2.5s to 0.9s (in Chrome, local server).

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@angular/platform-browser-dynamic": "2.0.0",
     "@angular/platform-server": "2.0.0",
     "@angular/router": "3.0.0",
+    "@ngtools/webpack": "^1.1.4",
     "@types/core-js": "^0.9.34",
     "@types/jasmine": "^2.5.37",
     "@types/node": "^6.0.46",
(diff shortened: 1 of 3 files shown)


mlaval:
Updated to use `@ngtools/webpack`, much better!

src/modal/modal.spec.ts
```diff
@@ -7,7 +7,7 @@ import {NgbModalModule, NgbModal, NgbActiveModal, NgbModalRef} from './modal.mod
 const NOOP = () => {};
 
 @Injectable()
-class SpyService {
+export class SpyService {
```
pkozlowski-opensource: Curious as why this change was necessary?

pkozlowski-opensource:
@mlaval I've merged the WebPack 2.0 but now this one needs rebasing (just drop one commit) :-/ 
Also I've left one minor comment. Can't wait to land this one :-)

src/modal/modal.spec.ts
mlaval: At some point, the Angular compiler was yelling at me for this.
But it was a bad configuration where tests were compiled, not needed anymore.

mlaval:
PR rebased and updated

PR #799, opened 2016-09-27
fix(modal): add `forRoot` to modules

### LINKED ISSUE #784: Modals and lazy-loaded NgbModule
Ref: http://stackoverflow.com/questions/39629827/opening-a-modal-from-a-lazy-loaded-module

We need to investigate / fix (if there is anything to fix).

k-schneider:
I converted all modules to follow the same pattern as pretty much all of them expose a configuration service of some kind.

BREAKING CHANGE: The root app module must now import using the forRoot() static method.

Closes #784

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ import {NgbModule} from '@ng-bootstrap/ng-bootstrap';
 
 @NgModule({
   declarations: [AppComponent, ...],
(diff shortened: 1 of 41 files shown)


pkozlowski-opensource:
@Foxandxss this looks fine for me, generally speaking, but I would like you to have a deeper look at this one.

PR #810, opened 2016-09-28
fix(dropdown): change to contains

### LINKED ISSUE #803: ngbDropdownToggle double triggers on child element
### Bug description:
When you have the ngbDropdownToggle on a button element, and a span inside that button element, clicking on the span (inside the button) does not toggle the dropdown (or, more specifically, I think it double-toggles it back to its original state).

This bug is new as of alpha 6--it may be related to the change from HostListener to (click) for handling the event. The bug was observed using chrome version 53.

http://plnkr.co/edit/TyaAkbqHeLq7HidFzpNf?p=preview

Angular: 2.0.0? whatever the sample plunker defaults to--it does not specify

ng-bootstrap: 1.0.0-alpha.6

Bootstrap: 4.0.0-alpha.4

wesleycho:
- Change to check if event target is contained in toggle to catch all child element clicks

This should fix #803.

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -190,6 +190,30 @@ describe('ngb-dropdown-toggle', () => {
     expect(buttonEl.getAttribute('aria-expanded')).toBe('false');
   });
 
+  it('should toggle dropdown on click of child of toggle', () => {
+    const html = `
+      <div ngbDropdown>
+          <button ngbDropdownToggle>
+            <span class="toggle">Toggle dropdown</span>
+          </button>
+      </div>`;
+
+    const fixture = createTestComponent(html);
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.spec.ts
```diff
@@ -190,6 +190,30 @@ describe('ngb-dropdown-toggle', () => {
     expect(buttonEl.getAttribute('aria-expanded')).toBe('false');
   });
 
+  it('should toggle dropdown on click of child of toggle', () => {
```
maxokorokov: phrasing confused me a little when I first read it, but can't come up with a better one :)

pkozlowski-opensource:
LGTM. Make sure that you add reference to the issue being closed in the commit message while merging.

Issue #802, opened 2016-09-27
Carousel should have _getSlideIdxById as public function


wesleycho:
I was going through programmatic selection of slides in the carousel with a co-worker, and it turns out that for programmatically setting the current slide, the `_getSlideIdxById` function would be very useful for getting the information needed to pass into the `carousel.select` function. Currently we ended up having to rewrite that functionality on the user side, and I see this being a common problem.

Conversely, it can be useful to get the information on what is the currently selected slide from the carousel - if not by making public `_getSlideById`, at the least having an appropriate output that emits the new value every time it changes.

icfantv:
I have no issue with this.

wesleycho:
Adding comment here - the problem apparently is that the `select` function in the carousel has its argument misnamed. It should be `slideId`, not `slideIdx`.

PR #812, opened 2016-09-28
feat(carousel): make getSlideIdxById public

wesleycho:
- Make `getSlideIdxbyId` a public method for easier programmatic selection of slides with the carousel

This should address #802

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -155,6 +155,11 @@ export class NgbCarousel implements AfterContentChecked,
   }
 
   /**
+   * Gets the slide index from the id for use in programmatically selecting a slide.
+   */
+  getSlideIdxById(slideId: string): number { return this.slides.toArray().indexOf(this._getSlideById(slideId)); }
+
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@wesleycho I would like to dig a bit deeper into this one when back from AC

pkozlowski-opensource:
@wesleycho I've just checked the PR and I think that there is some confusion here. The thing is that the `select` method gets slide identifier as an argument and _not_ the index. I think this is the good API.

But I think I know from where the confusion steams from! We've got good documentation:
> Navigate to a slide with a specified identifier.

but the method signature is misleading: `select(slideIdx: string)`. We should change it to `select(slideId: string)`.

Provided that the `select` method takes a string equals to slide id (and not index) do you think that you've got all the APIs you need? Or do we still miss things?

wesleycho:
Oh, if that's the case then yeah, we don't need this public.

wesleycho:
Closing this as unneeded

Issue #801, opened 2016-09-27
Rating: allow stars customisation and fine-grained ratings


maxokorokov:
The rating component currently has two issues:
- it uses hardcoded &#9733; and &#9734; symbols
- it doesn't allow half- and fine-grained ratings like `3.75 / 5`

I think that default solution should stay the same as now → hardcoded html symbols

My suggestion to fix problems above is threefold:

- [x] Allow an optional template to display a star and `3.75`rating the way user wants
- [x] Allow decimal numbers as model values via `[(rate)]` binding
- [x] Allow `<template>` to be a child of `<ngb-rating>`
- [ ] Allow optional mouse selection of half-stars via the `[halfStars]` flag (as in #669)

So a user would be able to select either `1, 2, 3 (default)` or `1, 1.5, 2, 2.5, 3 (halfStars=true)` with the mouse and display any rating he wants like `4.06`. It will also allow using external libs like glyphicons or font awesome for star display.

The proposed API is:

```html
<template #t let-fill="fill">
  <!-- fill is a number between 0-100 to represent how the star is filled -->
</template>

<ngb-rating [(rate)]="currentRate" [starTemplate]="t" [halfStars]="true"></ngb-rating>
```

Let me know what you think. I might start with adding custom templates.

pkozlowski-opensource:
I believe that with #833 landed we are done here.

Hongbo-Miao:
@pkozlowski-opensource I think "allow optional mouse selection of half-stars" it still not implemented.

joaopgrassi:
Yes.. the half-stars only works in a "read-only" way. You cannot hover the mouse and select half of a start. Would be really nice feature to have.

mpaolor:
> Yes.. the half-star only works in a "read-only" way. You cannot hover the mouse and select half of a star. Would be really nice feature to have.. is this going to come any time?

yeah it would be awesome! is it gonna come?

PR #815, opened 2016-09-28
feat(rating): allow custom star templates

maxokorokov:
Part of #801 

Adds a template to allow custom rating display
Default rendering still uses &#9733; and &#9734; symbols

```html
<template #t let-fill="fill">
  <!-- fill is a number between 0-100 to represent how the star is filled -->
</template>

<ngb-rating [(rate)]="currentRate" [starTemplate]="t"></ngb-rating>
```

screen shot 2016-09-29 at 00 05 32

### DIFF
--- a/demo/src/app/components/rating/demos/index.ts
+++ b/demo/src/app/components/rating/demos/index.ts
@@ -1,8 +1,9 @@
 import {RatingBasicComponent} from './basic/basic.component';
 import {RatingConfigComponent} from './config/config.component';
+import {RatingTemplateComponent} from './template/template.component';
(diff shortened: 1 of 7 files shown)


wesleycho:
LGTM

maxokorokov:
Rebased

pkozlowski-opensource:
This LGTM as well. What we could / should do in a separate PR is to allow star templates as child elements of the `ngb-rating`:

```html
<ngb-rating ...>
    <template>...</template>
</ngb-rating>
```

This is the "trick" that ngFor is using so it would be cool to support this as well. But we can totally do this in a separate PR.

PR #817, opened 2016-09-29
feat(rating): allow decimal numbers as rating values

maxokorokov:
Part of #801 

Allows for fine-grained rating values like `3.72`. 
Properly handles decimal numbers and template `fill` value when updated via `[rate]` binding. 
Plus a small readonly demo.

screen shot 2016-09-29 at 15 39 54

### DIFF
--- a/demo/src/app/components/rating/demos/decimal/decimal.component.html
+++ b/demo/src/app/components/rating/demos/decimal/decimal.component.html
@@ -1,0 +1,12 @@
+<template #t let-fill="fill">
+  <span *ngIf="fill === 100" class="star full">&hearts;</span>
+  <span *ngIf="fill === 0" class="star">&hearts;</span>
+  <span *ngIf="fill < 100 && fill > 0" class="star">
+    <span class="half" [style.width.%]="fill">&hearts;</span>&hearts;
+  </span>
+</template>
+<ngb-rating [(rate)]="currentRate" [starTemplate]="t" [readonly]="true" max="5"></ngb-rating>
+<hr>
+<pre>Rate: <b>{{currentRate}}</b></pre>
+<button class="btn btn-sm btn-outline-primary" (click)="currentRate = 1.35">1.35</button>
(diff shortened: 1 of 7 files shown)


src/rating/rating.ts
```diff
@@ -85,11 +96,30 @@ export class NgbRating implements OnInit {
     this.hover.emit(value);
   }
 
-  ngOnInit(): void {
-    this._oldRate = this.rate;
-    this.range = this._buildTemplateObjects();
+  /**
+   * internal
+   */
+  getFillValue(index: number): number {
```
pkozlowski-opensource: Remove the JS doc comment. The way it is written right now it would show up in the API doc.

pkozlowski-opensource:
LGTM apart from the comment I've left.

demo/src/app/components/rating/demos/decimal/decimal.component.html
```diff
@@ -0,0 +1,12 @@
+<template #t let-fill="fill">
+  <span *ngIf="fill === 100" class="star full">&hearts;</span>
+  <span *ngIf="fill === 0" class="star">&hearts;</span>
+  <span *ngIf="fill < 100 && fill > 0" class="star">
+    <span class="half" style.width="{{fill + '%'}}">&hearts;</span>&hearts;
```
wesleycho: `style.width` works like this so that one wouldn't need to use a binding?

pkozlowski-opensource: One could also write: `[style.width.%]="fill"` if I'm not mistaken.

maxokorokov: Kinda works both ways `[style.width]=""` and `style.width='{{}}'`, but will try the one @pkozlowski-opensource is suggesting!

maxokorokov: `[style.width.%]` works fine, thanks

PR #821, opened 2016-09-30
docs(carousel): clarify arguments of the select() method

### LINKED ISSUE #802: Carousel should have _getSlideIdxById as public function

pkozlowski-opensource:
Closes #802

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -101,10 +101,10 @@ export class NgbCarousel implements AfterContentChecked,
   ngOnDestroy() { clearInterval(this._slideChangeInterval); }
 
   /**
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
cc @wesleycho

wesleycho:
LGTM

PR #824, opened 2016-09-30
docs(README): remove http://issuestats.com/ badges

pkozlowski-opensource:
http://issuestats.com/ seems to be broken most of the time
and makes our landing page look half-baked.


wesleycho:
LGTM

PR #826, opened 2016-10-01
fix(modal): properly destroy content views

### LINKED ISSUE #806: Modal doesn't call ngOnDestroy when closing
When using modal with an external component like this:

```
<template #addEntity="" let-c="close">
  <add-entity [close]="c"></add-entity>
</template>
```

`ngOnInit` is being called but `ngOnDestroy` is never called (backdrop close or close function...).

pkozlowski-opensource:
Fixes #806


wesleycho:
LGTM

PR #828, opened 2016-10-03
fix(buttons): use disabled state from a control

### LINKED ISSUE #805: NgbRadioGroup doesn't support disabled state
The `ControlValueAccessor` interface has the optional `setDisabledState` state method which we don't implement correctly in the `NgbRadioGroup`. We need to add impl for this method and make sure that proper flags / styles are set on corresponding elements. 

As a ref, we might bump into https://github.com/twbs/bootstrap/issues/20799

maxokorokov:
Adds the `.disabled` class on the label and `disabled` attribute on the input of radios in case form control becomes disabled.

Fixes #805

### DIFF
--- a/src/buttons/radio.spec.ts
+++ b/src/buttons/radio.spec.ts
@@ -32,6 +32,10 @@ function getInput(nativeEl: HTMLElement, idx: number): HTMLInputElement {
   return <HTMLInputElement>nativeEl.querySelectorAll('input')[idx];
 }
 
+function getLabel(nativeEl: HTMLElement, idx: number): HTMLElement {
+  return <HTMLElement>nativeEl.querySelectorAll('label')[idx];
+}
+
 describe('NgbActiveLabel', () => {
   beforeEach(() => {
     TestBed.configureTestingModule(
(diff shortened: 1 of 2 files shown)


src/buttons/radio.spec.ts
```diff
@@ -312,11 +316,35 @@ describe('ngbRadioGroup', () => {
     expect(getGroupElement(fixture.nativeElement)).toHaveCssClass('ng-valid');
     expect(getGroupElement(fixture.nativeElement)).not.toHaveCssClass('ng-invalid');
   });
+
+  it('should disable label and input when it is disabled using reactive forms', () => {
```
pkozlowski-opensource: Could you also add a test for template-driven forms? I _know_ that the current implementation should work, but better to be safe than sorry...

pkozlowski-opensource:
LGTM but I would be even more happy if we could have a test for template-driven forms :-)

src/buttons/radio.spec.ts
maxokorokov: Amended the test for template-driven forms

PR #842, opened 2016-10-04
feat(popover): add isOpen() and shown/hidden events

### LINKED ISSUE #841: Popover: Detect When Popover is Open
See #837 for original issue and some early commentary

Would be great to have the capability to detect when the popover is open. An example use case would be to apply a class to the element that triggers the popover.

### Expose `@Outputs` for `close` and `open` events

**Pros**
 - Components can react to popover state changes more easily (additional customization opportunities)

**Cons**
 - Increases the API footprint

**Example Implementation**

```
<button [ngbPopover]="template" [class.open]="opened" (open)="opened = true" (close)="closed = true"></button>
```

===

### Expose an `isOpen()` method on the popover instance

**Pros**
 - Simplest way to attack the proposed use case, and more declarative in nature

**Cons**
 - Limited customization, and/or more difficult to implement other behavior aside from the proposed use case

**Example Implementation**

```
<button [ngbPopover]="template" #popover="ngbPopover" [class.open]="ngbPopover.isOpen()"></button>
```

===

### Other Considerations

Certainly for the initially proposed use case, the latter is easiest to understand and implement, and makes the most sense. But as soon as you think outside of that use case, the first option is certainly more flexible. For example, if the component wants to react by performing some action when the popover is opened (analytics, maybe, or changing some other part of the page?). The latter aims for a declarative approach, which makes it harder to detect _when it changes_, but easier to detect _what it changed to_. 

Note that a workaround for receiving events when the popover is opened/closed can be accomplished through the use of a child component:

```
class TestComponent implements OnInit, OnDestroy {

  @Output('init')
  init = new EventEmitter();

  @Output('destroy')
  destroy = new EventEmitter();

  ngOnInit() {
    this.init.emit();
  }

  ngOnDestroy() {
    this.destroy.emit();
  }

}
```

```
<template #template>
  <test-component (init)="opened()" (destroy)="closed()"></test-component>
</template>

<button [ngbPopover]="template" #popover="ngbPopover" [class.open]="ngbPopover.isOpen()"></button>
```

colinmorelli:
Adds an `isOpen()` method to the popover, as well as shown and hidden events. Putting this up to get feedback on it, as I work on the same for tooltip and add documentation.

Closes #841

### DIFF
--- a/demo/src/app/components/popover/demos/index.ts
+++ b/demo/src/app/components/popover/demos/index.ts
@@ -1,9 +1,10 @@
 import {NgbdPopoverBasic} from './basic/popover-basic';
 import {NgbdPopoverTplcontent} from './tplcontent/popover-tplcontent';
 import {NgbdPopoverTriggers} from './triggers/popover-triggers';
+import {NgbdPopoverVisibility} from './visibility/popover-visibility';
 import {NgbdPopoverConfig} from './config/popover-config';
 
-export const DEMO_DIRECTIVES = [NgbdPopoverBasic, NgbdPopoverTplcontent, NgbdPopoverTriggers, NgbdPopoverConfig];
+export const DEMO_DIRECTIVES = [NgbdPopoverBasic, NgbdPopoverTplcontent, NgbdPopoverTriggers, NgbdPopoverVisibility, NgbdPopoverConfig];
 
 export const DEMO_SNIPPETS = {
   basic: {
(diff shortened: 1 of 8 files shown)


pkozlowski-opensource:
Merged, thnx! If you happen to send us PRs in the feature (which would be awesome!) try to keep them small and focused (in this particular example a separate PR for a tooltip and another one for a popover). Smaller PRs are easier to review and gets merged faster. Thnx again.

Issue #818, opened 2016-09-29
[feature request] Pass custom class to modal 


jspizziri:
Perhaps I missed something when looking through the code; however, it would be nice if it were possible to pass custom classes on the modal with [NgbModalOptions](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal.ts#L9), that would then be applied to the modal template.

Thanks!

pkozlowski-opensource:
I'm very, very hesitant to go down this path. The thing is that modal window and backdrop creates many elements so if we want to allow users to customise most things the number of options would explode.

I understand the need but at this point your best option is just to fork the directive and maintain your own fork that would suit your particular needs. We want to support some form of customisation in the future but probably sth more sophisticated than just adding CSS classes.

Going to keep this one open as I understand the need but don't hold your breath and expect movement on this issue any time soon. What we need is design for something more flexible.

jspizziri:
@pkozlowski-opensource, I appreciate the desire for an abstract solutuon. Perhaps we can begin a discussion on the design by articulating all the requirements of customization. I would rather begin a discussion and potentially try and contribute then form my own little tribe by forking, as I'm entirely disinterested in maintaining my own fork 😊

pkozlowski-opensource:
@jspizziri sure, feel free to start a design doc and link it here. You can find an example of a design doc for other modal part [here](https://docs.google.com/document/d/1puwnSknaW3bLg_Ww2ewuILneCom1wbMTArvfFq0YFV8/edit?usp=sharing)

As for requirements / flexibility we will have really hard time predicting what / where users might want to add. So ideally we would allow them to provide a custom window component and the custom one should replace the default one: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal-window.ts

As you can see the current modal window has quite a bit of functionality so this would have to be moved to some other place so providing custom components is easy. Ideally users should just provide the template bit (https://github.com/ng-bootstrap/ng-bootstrap/blob/54ae058767c67b0f66c7d908c031d0af56ea7614/src/modal/modal-window.ts#L26-L28) but I'm not sure how to do this, technically speaking

Of course there are other ways of solving the customization problem so let your creativity go wild :-)

[...]

PR #845, opened 2016-10-05
feat(modal): add support for custom windowClass on modals

colinmorelli:
Adds the ability to add a custom class to a modal when it's opened.

Note that this does not (yet) add support for adding a custom backdrop class. That may need more discussion to talk about what should happen when, for example, multiple modals are open. Whose backdrop gets applied?

Fixes (at least part of) #818

### DIFF
--- a/demo/src/app/components/modal/demos/customclass/modal-customclass.ts
+++ b/demo/src/app/components/modal/demos/customclass/modal-customclass.ts
@@ -1,0 +1,33 @@
+import {Component, ViewEncapsulation} from '@angular/core';
+
+import {NgbModal, ModalDismissReasons} from '@ng-bootstrap/ng-bootstrap';
+
+@Component({
+  selector: 'ngbd-modal-customclass',
+  templateUrl: './modal-customclass.html',
+  encapsulation: ViewEncapsulation.None,
+  styleUrls: [ './modal-customclass.css' ]
+})
+export class NgbdModalCustomClass {
+  closeResult: string;
+
+  constructor(private modalService: NgbModal) {}
+
+  open(content) {
+    this.modalService.open(content, { windowClass: 'dark-modal' }).result.then((result) => {
+      this.closeResult = `Closed with: ${result}`;
+    }, (reason) => {
+      this.closeResult = `Dismissed ${this.getDismissReason(reason)}`;
+    });
+  }
+
+  private getDismissReason(reason: any): string {
+    if (reason === ModalDismissReasons.ESC) {
+      return 'by pressing ESC';
+    } else if (reason === ModalDismissReasons.BACKDROP_CLICK) {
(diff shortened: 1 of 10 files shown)


demo/src/app/components/modal/demos/customclass/modal-customclass.ts
```diff
@@ -0,0 +1,33 @@
+import {Component, ViewEncapsulation} from '@angular/core';
+
+import {NgbModal, ModalDismissReasons} from '@ng-bootstrap/ng-bootstrap';
+
+@Component({
+  selector: 'ngbd-modal-customclass',
+  templateUrl: './modal-customclass.html',
+  encapsulation: ViewEncapsulation.None,
```
colinmorelli: I don't love disabling view encapsulation on this, but I wanted to keep the associated styles close to the component. Modals are tough in that (by design) they render content outside of the component's view hierarchy.

demo/src/app/components/modal/demos/customclass/modal-customclass.ts
```diff
@@ -0,0 +1,33 @@
...
+  styleUrls: [ './modal-customclass.css' ]
+})
+export class NgbdModalCustomClass {
+  closeResult: string;
+
+  constructor(private modalService: NgbModal) {}
+
+  open(content) {
+    this.modalService.open(content, { windowClass: 'dark-modal' }).result.then((result) => {
+      this.closeResult = `Closed with: ${result}`;
```
pkozlowski-opensource: I don't think we need to repeat all the demo here. Just `this.modalService.open(...)` should be enough, IMO

colinmorelli:
Apologies if I'm missing something obvious but I'm not seeing comments anywhere

pkozlowski-opensource:
> Apologies if I'm missing something obvious but I'm not seeing comments anywhere

@colinmorelli no, it is me failing to push the right buttons :-/

src/modal/modal-window.ts
```diff
@@ -15,7 +15,7 @@ import {ModalDismissReasons} from './modal-dismiss-reasons';
 @Component({
   selector: 'ngb-modal-window',
   host: {
-    'class': 'modal fade in',
+    '[class]': '"modal fade in" + (windowClass ? " " + windowClass : "")',
```
colinmorelli: This is part of the host selectors applied to the angular template. AFAIK there's no way to use ES6 string concatenation here. It follows the same pattern that's used on line 26: `<div [class]="'modal-dialog' + (size ? ' modal-' + size : '')" role="document">`

pkozlowski-opensource: You are right, sorry for the noise.

realappie:
Really a shame this commit only supports a string for the window class property. It would have been more helpful if it was named windowClass and accepted an array of strings !

PR #846, opened 2016-10-06
feat(modal): allow components as content

### LINKED ISSUE #680: Modal as encapsulated component

### Bug description:
It's nice to have the Modal as Component now.

But what I'm missing or maybe don't know is how to make it as external Component.

So I got a Component with HTML and a Controller with the functions for that modal inside it.
And from another Component I then open that Modal.

What I did now was the HTML template inside the Component HTML where I wanna open the Modal.
Like the example is doing, but what I really want is an excapsulated Modal. 
What if you need to open the same Modal at 5 places?
What if you wanna lazy load the modal?

-

**Angular:**
RC5

**ng-bootstrap:**
Alpha 3

**Bootstrap:**
Alpha 3

### LINKED ISSUE #727: Modal: string argument to the open method is confusing
See http://stackoverflow.com/q/39464345/1418796

Really, the string argument is only good for unit-testing but not for any real-life applications. We should  remove it altogether or at least clarify in the docs.

On top of this we need to support components as arguments to the `open` method, see #680

pkozlowski-opensource:
Closes #680
Closes #727

### DIFF
--- a/src/modal/modal-ref.ts
+++ b/src/modal/modal-ref.ts
@@ -3,6 +3,24 @@ import {Injectable, ComponentRef, ViewRef, ViewContainerRef} from '@angular/core
 import {NgbModalBackdrop} from './modal-backdrop';
 import {NgbModalWindow} from './modal-window';
 
+
+/**
+ * A reference to an active (currently opened) modal. Instances of this class
+ * can be injected into components passed as modal content.
+ */
+@Injectable()
+export class NgbActiveModal {
+  /**
+   * Can be used to close a modal, passing an optional result.
+   */
+  close(result?: any): void {}
+
+  /**
+   * Can be used to dismiss a modal, passing an optional reason.
+   */
+  dismiss(reason?: any): void {}
+}
+
 /**
  * A reference to a newly opened modal.
  */
--- a/demo/src/app/components/modal/demos/component/modal-component.html
+++ b/demo/src/app/components/modal/demos/component/modal-component.html
@@ -1,0 +1,4 @@
+<p>You can pass an existing component as content of the modal window. In this case remember to add content component
+as an <code>entryComponents</code> section of your <code>NgModule</code>.</p>
+
+<button class="btn btn-lg btn-outline-primary" (click)="open()">Launch demo modal</button>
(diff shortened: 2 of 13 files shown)


colinmorelli:
How would you propose handling inputs to the components? As far as I can tell it doesn't look like this allows for them.

Another question: while I think this is a lot cleaner, is there any functional difference between passing a component to a modal, versus using the below code?

```
<template #modal>
  <my-custom-component [input]="value()"></my-custom-component>
</template>
```

I've done that for popovers, and it has worked well.

pkozlowski-opensource:
> How would you propose handling inputs to the components? As far as I can tell it doesn't look like this allows for them.

This PR leaves out Input / Output interactions. We can add it in the (near) future if needed.

> Another question: while I think this is a lot cleaner, is there any functional difference between passing a component to a modal, versus using the below code?

The real difference is for cases where you don't have access to any template (ex. a general confirmation dialog, reusable dialogs, a sign-in service etc.). 

> I've done that for popovers, and it has worked well.

That's great, continue to use `TemplateRef`s if this works for you! We are not removing this functionality, just adding another flex point so users can use components for cases where there is no easy access to a template.

pkozlowski-opensource:
@jnizet would be awesome if you could test-drive / review this one!

src/modal/modal-ref.ts
```diff
@@ -3,11 +3,29 @@ import {Injectable, ComponentRef, ViewRef, ViewContainerRef} from '@angular/core
...
+   */
+  dismiss(reason?: any): void {}
+}
+
 /**
  * A reference to a newly opened modal.
  */
 @Injectable()
-export class NgbModalRef {
+export class NgbModalRef extends NgbActiveModal {
```
pkozlowski-opensource: Agree. I was going back and forth on this one. Somehow `NgbModalRef` and `NgbActiveModal` are very similar yet different... Will split them up.

deepu105:
Its a good start :+1: and thank you for that. Though it addresses part of the ask, one of the very common need would be to pass inputs (more so than outputs I guess) hence would love to see support for input/output to the component. Hope it will be implemented soon

src/modal/modal-ref.ts
pkozlowski-opensource: Done.

pkozlowski-opensource:
> Implementing my "reusable confirm service" usecase is still not possible due to the lack of support for injected values/inputs, but we're getting much closer, now.

Yep, I will open a separate issue to track this.

Issue #847, opened 2016-10-06
Datepicker and months with only 5 rows needed (instead of 6 in most cases)


pkozlowski-opensource:
Check this plunk: http://plnkr.co/edit/Z9tecsJTv7tnintVJ3gU?p=preview (or navigate to August 2016 using our demo site).

You will notice that the last row has all days belonging to September 2016 so we could safely drop / not display it. But it would mean that for some months we would have 5 rows only instead of usual 6 so it might create an unpleasant flickering effect in the UI... 

Not sure if we should change this or not but opening so we can discuss / decide.

pkozlowski-opensource:
For the record, both kayak and booking.com would drop the last row in this case.

maxokorokov:
See my comment https://github.com/ng-bootstrap/ng-bootstrap/issues/753#issuecomment-252188117
It's doable now, but we might hide it behind a flag

maxokorokov:
Added an example of how to hide these days in the pull request above. 
Now the question is: do we need to hide this functionality behind a flag in the API or leave like this?

pkozlowski-opensource:
> Now the question is: do we need to hide this functionality behind a flag in the API or leave like this?

Let's put it on a back-burner till we've got more feedback on the current API.

Issue #753, opened 2016-09-17
Range selection support in Datepicker


tonystar:
We would like to use and help to maintain Datepicker component, but what is blocking us — range selection support. Attached is a sketch what we are looking for. Do you have any plans for this feature? Or any ideas how to implement with existing component?

<img width="429" alt="screenshot 2016-09-17 11 02 23" src="https://cloud.githubusercontent.com/assets/5911853/18606936/68cbacb2-7cc6-11e6-9820-196e0abb4db4.png">

pkozlowski-opensource:
I think that you can implement a range Datepicker with the building blocks we've got in place. Some hints to implement:
* you would create a new control (implement `ControlValueAccessor`)
* you could stick 2 datepickers as part of your control
* you would have to provide a custom template to a day cell to mark it as "in range"

I would expect it to not take more than 1 day of dev with tests etc. 

We will be putting more demos on our website and probably will have some examples with range datepicker. But not planning to have a specific date-range-picker, no. I think that there are just too many UI variations for this so we would rather focus on providing basic building blocks and flex points so people can craft they own range pickers.

I'm going to keep it open so we can add more examples to our demo page.

pkozlowski-opensource:
@tonystar don't hesitate to come up with a more detailed question if you bump into any roadblocks. We would be keen on adding necessery APIs / flex points so your use-case is covered. And of course would very much appreciate contributions to the datepicker (or any other component).

tonystar:
Wow! Such comprehensive answer! I will invite our developers in this discussion and let's work together on the demo! We are adopting the ng-bootstrap from time it was created and looking forward for the new opportunities!

PEsteves8:
I had the same question and I'm glad you're focused on the basic blocks that would allow developers to implement everything they might need. 

I actually did a variation of ng2-bootstrap that allowed me to selected multiple ranges (as long as the developer doesn't use the single date option). It has a lot of customization, like number of datepickers, default month selection controls removed, etc.

Here's the demo(the code needs serious refactoring):

https://pesteves8.github.io/dayselector/

Now to be honest, I would prefer to use the ng-bootstrap library (instead of the ng2-bootstrap) so I'll see if I can follow your advice to implement the same functionality, hopefully in less time that my first attempt.
I'll let you know if I run into any issues with the library :) .

pkozlowski-opensource:
[...]

PR #849, opened 2016-10-07
feat(datepicker): pass current month value to 'markDisabled' callback

maxokorokov:
Small datepicker enhancement: now able to access current month value both in custom template and `markDisabled` callback.

That allows for instance to disable completely days that are outside of current month and collapse them (see 'custom day view' demo):

```typescript
[markDisabled]='markDisabled'

...

markDisabled = (date, current: {month: number}) {
  return date.month != current.month;
};

...

<template let-date='date' let-currentMonth='currentMonth'>
  <span [class.hide_day]='date.month !== currentMonth' ...>
</template>

...

.hide_day {
  display: none; // or visibility: hidden; to avoid line collapse
}

```

Contains an example for #847 and #753


pkozlowski-opensource:
LGTM

PR #850, opened 2016-10-07
fix(datepicker): replace not-allowed cursor with default one

maxokorokov:
Replaces the `not-allowed` cursor with a `default` one for disabled dates. 

- `not-allowed` cursor looks way too restrictive
- All major sites do this


pkozlowski-opensource:
LGTM when TravisCI gets green.

Issue #339, opened 2016-07-02
Tooltip / popover: add support for the "container" @Input


pkozlowski-opensource:
Ref.: http://v4-alpha.getbootstrap.com/components/popovers/#options

wesleycho:
How would we approach this? Would we need a component ref to insert into and usage of the component resolver for a more restrictive version? We could require the user to pass the ref into the tooltip/popover for this feature.

pkozlowski-opensource:
> Would we need a component ref to insert into and usage of the component resolver for a more restrictive version? 

Nope, since the target place could be outside of the bootstrapped ng2 app. The solution is to simply move DOM nodes to the target place. It is should be relatively  straightforward to implement.

pkozlowski-opensource:
We've been supporting the `container` option for quite a while now, not sure while this one was left open :-/

PR #853, opened 2016-10-08
WIP - feat(popover): Add container input, support 'body' selector

### LINKED ISSUE #852: appendToBody option for Popover
It looks like the positioning service is ready to accept an `appendToBody` parameter, but this option isn't included in the NgbPopoverConfig. I took a crack at editing the transpiled javascript in the npm module to see if it was possible and sort of got it working, but I'm wondering if there's a known issue standing in the way of it getting implemented.

spongessuck:
Adds an `appendToRoot` input to the popover component to allow for placement of the popover in the root element. Addresses #852.

I'm not sure about the cross-platform implications of this solution, but it works in the browser, is analogous to the ui-bootstrap `appendToBody` option, and makes popover DOM placement more flexible until #339 is implemented.


pkozlowski-opensource:
@spongessuck thnx for the PR, it looks like a first good step. Having said this I had a slightly different plan for this. My intention was to introduce `@Input` with the name `container` and type `string` and default value being undefined. For now this input would only kick in for the value equal to `body`. In the future we could extend / open up it to support any selector / element reference.

With my proposal we would be not introducing APIs that would be redundant / deprecated in the future.

Would you be willing to do the above changes in your PR?

spongessuck:
Sure, it shouldn't be hard.

spongessuck:
What about `container="root"`, to be consistent? The positioning service measures the offset based on the `html` element.

pkozlowski-opensource:
> What about container="root", to be consistent? The positioning service measures the offset based on the html element.

I would prefer '"body"' for now as it will still work when we add support for all selectors.  

> Sure, it shouldn't be hard.

Awesome!

spongessuck:
I'll do "html" then. I still think it's a bad idea to call it "body" when it's not measuring from the body.

pkozlowski-opensource:
> I'll do "html" then. I still think it's a bad idea to call it "body" when it's not measuring from the body

Could we please stick to `body`? I understand your arguments but I want to keep `body` as default as this is what is shown in BS docs. And as things stand today it doesn't really matter what is measured from but rather where the element is attached.

spongessuck:
If the body has any padding or margin I don't think it will be placed properly if it's in the body.

spongessuck:
I guess it'd have to be positioned `relative` too... I'll do some tests, but do you see my concern?

spongessuck:
Did some tests and it doesn't seem to matter where the body is- the position service is smarter than I thought :). I'll use "body" as requested.

src/popover/popover.ts
```diff
@@ -95,6 +109,7 @@ export class NgbPopover implements OnInit, OnDestroy {
       this._windowRef = this._popupService.open(this.ngbPopover);
       this._windowRef.instance.placement = this.placement;
       this._windowRef.instance.title = this.popoverTitle;
+      this._windowRef.instance.container = this.container;
```
pkozlowski-opensource: Does popover window need to be aware of the `container` option?

src/popover/popover.ts
```diff
@@ -34,6 +34,7 @@ import {NgbPopoverConfig} from './popover-config';
 export class NgbPopoverWindow {
   @Input() placement: 'top' | 'bottom' | 'left' | 'right' = 'top';
   @Input() title: string;
+  @Input() container: string;
```
pkozlowski-opensource: Why do we need this input?

src/popover/popover.spec.ts
```diff
@@ -382,6 +403,7 @@ describe('ngb-popover', () => {
       config = c;
       config.placement = 'bottom';
       config.triggers = 'hover';
+      config.container = 'body';
```
pkozlowski-opensource: Why do we need to modify this test?

src/popover/popover.ts
```diff
@@ -77,12 +83,20 @@ export class NgbPopover implements OnInit, OnDestroy {
...
 
     this._zoneSubscription = ngZone.onStable.subscribe(() => {
       if (this._windowRef) {
-        positionElements(this._elementRef.nativeElement, this._windowRef.location.nativeElement, this.placement);
+        positionElements(
+            this._elementRef.nativeElement, this._windowRef.location.nativeElement, this.placement,
+            this.container === 'body');
+
+        if (this.container === 'body') {
+          let windowEl = this._windowRef.location.nativeElement;
```
pkozlowski-opensource: Nit: maybe drop the temporary `windowEl` variable?

src/popover/popover.spec.ts
```diff
@@ -191,6 +193,25 @@ describe('ngb-popover', () => {
     });
   });
 
+  describe('placement', () => {
```
pkozlowski-opensource: Nit: 'placement' => 'container'?

pkozlowski-opensource:
Thnx for all the updates @spongessuck - it all starts to come together 👍 
I've just left some minor comments that we need to sort out before merging.

src/popover/popover.ts
spongessuck: I guess it doesn't.

src/popover/popover.ts
spongessuck: Ditto.

src/popover/popover.spec.ts
spongessuck: Doesn't this apply to popover and not popoverwindow? Shouldn't it test that it takes custom config properly?

src/popover/popover.spec.ts
spongessuck: Sure.

src/popover/popover.ts
pkozlowski-opensource: Whatever you prefer, I'm good either way.

spongessuck:
Not sure if you got notifs from my commits, but I think this is good to go.

pkozlowski-opensource:
Thnx @spongessuck - squashed and merged as d1a22c0. Would be awesome if you could send a similar (well, almost identical) PR for tooltips :-)

PR #860, opened 2016-10-09
chore: don't import from catch-all rxjs/Rx

pkozlowski-opensource:
Changing rxjs imports as in this PR reduces our bundle size
from ~190k -> ~45k (before minification / gzip). Non-scientific
testing on my machine shows that the site load time drops by
around 300 - 400ms.


icfantv:
LGTM.

PR #867, opened 2016-10-11
feat(tooltip): Add container input

spongessuck:
Added support for "body" value to append tooltip to body element.


pkozlowski-opensource:
@spongessuck the change itself lgtm but is seems like something went wrong with commits in this PR. Could you please tidy up things a bit so we've got only d9b9c185f788c4bdb12b407246e12bbfe59fc6e7 in this PR? Thnx!

spongessuck:
I'm not sure if I can do it- I might have to blow this PR away and submit another.

pkozlowski-opensource:
> I'm not sure if I can do it- I might have to blow this PR away and submit another.

Go ahead if this is easier for you.

PR #874, opened 2016-10-12
fix(radio): Checked attribute of native input element doesn't work

mschoudry:
Fix for bug Radio button check attribute not showing.

### DIFF
--- a/src/buttons/radio.ts
+++ b/src/buttons/radio.ts
@@ -90,6 +90,16 @@ export class NgbRadio implements OnDestroy {
     }
   }
 
+  @Input('checked')
+  set checked(value: any) {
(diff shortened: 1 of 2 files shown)


src/buttons/radio.ts
```diff
@@ -90,6 +90,11 @@ export class NgbRadio implements OnDestroy {
     }
   }
 
+  @Input('checked')
+  set checked(value: any) {
+    this._checked = true;
```
pkozlowski-opensource: Wouldn't this fail for ` <input type="radio" name="State" [checked]="false" />`? Could you please add this case to the unit test?

mschoudry: yes..

mschoudry: i made the changes.

PR #883, opened 2016-10-13
chore(build): moving to tsc from gulp-typescript

maxokorokov:
Karma in tdd mode doesn't play nicely with gulp-typescript and gulp-watch.
It would be simpler and much faster to use the combination of `tsc -w` and karma watch

The only open question I'm not sure about is do we have to create a separate `tsconfig-test.json` or use change the exising one...

P.S. Tested on win and mac


pkozlowski-opensource:
OMG, this is awesome, thnx @maxokorokov - merged. I wanted to do it for sooooooo long but never had courage / time :-) Much appreciated.

PR #884, opened 2016-10-13
fix(build): properly reference rxjs from the UMD bundle

pkozlowski-opensource:
alpha.7 has a subtle bug where externals were not configured
properly and as a result our bundle doesn't work correctly
in certain environements (SystemJS).


wesleycho:
This makes me sad :( . LGTM

pkozlowski-opensource:
Why it makes you sad?

wesleycho:
It looks a little hacky is all.

Foxandxss:
Sounds good to me. Externals in webpack is a bit ugly like that.

PR #889, opened 2016-10-13
demo(API): don't extract API docs from typings

pkozlowski-opensource:
Till now we were extracting all the class info / docs
from all the classes referenced in typings (Promise etc.).
This is obviously not needed and wasteful.


wesleycho:
LGTM

Issue #643, opened 2016-08-31
Modal: consider supporting stacked modals


pkozlowski-opensource:
That is - modals that can be opened from modals. The current thinking is:
- we are very reluctant to support those as it is very, very bad UX
- "real life" sometimes "requires" those anyway....

Will try to resist us long as possible before implementing stacked modals, but let's have this issue to track interest / real-life usage scenarios.

natcohen:
I think it should be supported (for your record)

Quinn-L:
I use modal for yes/no confirmation dialogs.
However, if stacked modals aren't supported this means you can't have 'normal' modals with data entry which then uses modal confirmation dialogs. I guess one could argue data entry in modals are a bad UX?

pkozlowski-opensource:
Yeh, I'm kind of resigning to the reality of many projects and bracing for supporting this... But won't get to it before we land datepicker.

pkozlowski-opensource:
Voice in the discussion: https://github.com/twbs/bootstrap/issues/20607#issuecomment-244127770

asadsahi:
I also agree with @pkozlowski-opensource  stacked modal aren't good from UX perspective. Also, we had huge number of problems in one of large SPA build in Angular 1. Where there were confirmation dialogs, terms dialogs on top of existing dialogs, which caused more complexity, especially when in a web app reload happens, it is difficult to maintain state for those. We totally scrappeed the idea of having functional templates in modal. only for confirmation dialogs etc.

tytskiy:
Is there any plans to support backdrop position in between stacked modals? 
It turns that for each inner modal its backdrop doesn't cover previous modal.

There is quick reproduction: http://plnkr.co/edit/FG03lIbHjzM9wAZMoLWL?p=preview

dneuer:
@pkozlowski-opensource If this is considered by the team to be "very, very bad UX," what is the suggested alternative when several closely associated components need to operate on the same data? Especially with data that's already loaded? @ViewChild and direct subcomponent access? Child routes (and the terrible API of having to have a singleton service to pass references to in-memory data)?

ASHFAQPATWARI:
@pkozlowski-opensource  As you already said, sometimes we need stacked modals anyway. What is the alternative approach suggested by you till the library support stacked modals ?

DocteurEz:
@pkozlowski-opensource is there any progress on this matter?

ravtakhar:
[...]

PR #901, opened 2016-10-17
Stack Modals in Opening Order

colinmorelli:
Small change to attach modal elements in the order that they were opened (rather than always prepending everything)

Actually, this change alone seems to allow most of what's necessary for #643 - at least in terms of presenting multiple modals. There's still work to do for visually making it better (hiding inactive modals, backdrop management, etc)


colinmorelli:
Alright, I retract my earlier comment - there's still a lot to be done for proper stack modal support. That said, I can't see how this isn't an improvement, as it allows for multi modals in simple cases.

pkozlowski-opensource:
Agreed, much more would have to be done to fully support stacked modals. At the same time I don't see harm in merging this one so going to land it, thnx.

PR #903, opened 2016-10-18
feat(modal): allow interactions with components passed as content

### LINKED ISSUE #861: Feature request: Support Injection of inputs to Modal
Related to https://github.com/ng-bootstrap/ng-bootstrap/pull/846
Its a good start that now Modal dialogs can use components, but its still not enough if the component cant be passed some content which can be displayed/used inside the dialog.

We are using ui-bootstrap with[ JHipster](https://jhipster.github.io) With ng1 version there was resolve which solved this and it enabled us to open Modal dialogs right from onEnter of a route which was nice. But now we dont see a way to do [this here](https://github.com/jhipster/generator-jhipster/blob/angular-2/generators/client-2/templates/src/main/webapp/app/admin/user-management/_user-management.state.ts#L81)

This is becoming a blocker for our Angular2 migration would appreciate a fix/timeline/workaround or advice

pkozlowski-opensource:
Closes #861

### DIFF
--- a/src/modal/modal-ref.ts
+++ b/src/modal/modal-ref.ts
@@ -3,6 +3,8 @@ import {Injectable, ComponentRef, ViewRef, ViewContainerRef} from '@angular/core
 import {NgbModalBackdrop} from './modal-backdrop';
 import {NgbModalWindow} from './modal-window';
 
+import {ContentRef} from '../util/popup';
+
 /**
  * A reference to an active (currently opened) modal. Instances of this class
  * can be injected into components passed as modal content.
@@ -29,13 +31,23 @@ export class NgbModalRef {
   private _reject: (reason?: any) => void;
 
   /**
+   * The instance of component used as modal's content.
+   * Undefined when a TemplateRef is used as modal's content.
+   */
+  get componentInstance(): any {
+    if (this._contentRef.componentRef) {
+      return this._contentRef.componentRef.instance;
+    }
+  }
+
+  /**
    * A promise that is resolved when a modal is closed and rejected when a modal is dismissed.
    */
   result: Promise<any>;
 
   constructor(
       private _viewContainerRef: ViewContainerRef, private _windowCmptRef: ComponentRef<NgbModalWindow>,
-      private _backdropCmptRef?: ComponentRef<NgbModalBackdrop>, private _contentViewRef?: ViewRef) {
+      private _contentRef: ContentRef, private _backdropCmptRef?: ComponentRef<NgbModalBackdrop>) {
     _windowCmptRef.instance.dismissEvent.subscribe((reason: any) => { this.dismiss(reason); });
 
     this.result = new Promise((resolve, reject) => {
@@ -70,12 +82,12 @@ export class NgbModalRef {
     if (this._backdropCmptRef) {
       this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._backdropCmptRef.hostView));
     }
-    if (this._contentViewRef) {
-      this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._contentViewRef));
+    if (this._contentRef && this._contentRef.viewRef) {
+      this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._contentRef.viewRef));
     }
 
     this._windowCmptRef = null;
     this._backdropCmptRef = null;
-    this._contentViewRef = null;
+    this._contentRef = null;
   }
 }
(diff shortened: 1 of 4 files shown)


jnizet:
I have some questions regarding this design:

- why is the name attribute in the demo decorated with `@Input()`? This seems unnecessary, and even couter-intuitive since it kind of implies that the framework will populate this attribute, whereas it must actually be set explicitly by the developer code. Moreover, the value of this input is not available in ngOnInit like for traditional inputs, changes do not trigger ngOnChanges calls, etc., so it's really not an input in the angular meaning, IMHO.
- if that's how we pass values to the modal inner component, as opposed to populating a `data` field of the injectable NgbActiveModal service, isn't there at least a way to make the open() method and the NgbModalRef class generic, so that the returned component instance is no `any`, but is an instance of the actual type of component passed to open()? That would make the code more type-safe, and would allow code completion, etc.

pkozlowski-opensource:
> why is the name attribute in the demo decorated with @Input()? This seems unnecessary, and even couter-intuitive since it kind of implies that the framework will populate this attribute, whereas it must actually be set explicitly by the developer code. Moreover, the value of this input is not available in ngOnInit like for traditional inputs, changes do not trigger ngOnChanges calls, etc., so it's really not an input in the angular meaning, IMHO.

This is the biggest issue with dynamically created components, in general. IMO we can add or remove `@Input` and it won't matter much as a member marked with `@Input` is still a member. The idea was to show how you can use a pre-existing component in a modal but then again, I'm totally cool with removing it.

> isn't there at least a way to make the open() method and the NgbModalRef class generic, so that the returned component instance is no any, but is an instance of the actual type of component passed to open()? That would make the code more type-safe, and would allow code completion, etc.

TBH it is beyond my current TS knowledge. Feel free to build on top of this PR to add type safety.

jnizet:
> The idea was to show how you can use a pre-existing component in a modal

That's my biggest fear with this design: I fear that people use components not intended to be used in a modal originally, and then complain to us because the component uses standard lifecycle callbacks on inputs and they don't work as usual. But you're in a better position than I am to choose the right solution.

> Feel free to build on top of this PR to add type safety

I'll try doing that when I have some time. But I'm not sure it's even possible.

pkozlowski-opensource:
> I fear that people use components not intended to be used in a modal originally, and then complain to us because the component uses standard lifecycle callbacks on inputs and they don't work as usual. 

That's the danger, I agree. But given that there are no great solutions I prefer to err on the flexibility side. 

> But you're in a better position than I am to choose the right solution.

Thnx for the trust, but I don't think we are in the position of choosing "right" solution. The way I feel about it is that we are just balancing different tradeoffs. I'm planning to take this issue with Misko and Material folks but at the same time don't want to block people from using modals.

Event if the solution proposed here isn't perfect it is simple, flexible and we are still in alpha so nothing is set in stone. Let's get some real-World feedback.

src/modal/modal-ref.ts
```diff
@@ -29,13 +31,23 @@ export class NgbModalRef {
   private _reject: (reason?: any) => void;
 
   /**
+   * The instance of component used as modal's content.
```
jnizet: This property doesn't appear in the generated doc in the demo.

jnizet:
Regarding generic open() and NgbModalRef, here's an implementation: https://github.com/jnizet/ng-bootstrap/commit/75aa2b664f363fdb09ecde8950a09c5795083f70

Advantages:
- the signature of open() is clearer (if you can understand it): you can pass a TemplateRef or a type 
- `NgbModalRef.componentInstance` is of the right type, so setting `naame` instead of `name`, for example, is detected by the compiler.

pkozlowski-opensource:
So, after long and painful deliberations I'm going to make an "executive decision" and merge this one as-is. Being painfully aware of all the imperfections of this PR I believe that this is a good compromise for now. Let's iterate on it.

karptonite:
You were looking for real-world feedback, so here is some.

As you say, this approach is quite flexible, but I're run into some of the issues raised above. I'm showing a form in a modal that I am building from data that has to be passed to the modal. I was passing the data in as an `Input()`, and, as predicted, was surprised when the `OnChanges` lifecycle hook wasn't triggered. I'm reevaluating my options, and may just make a `buildForm` function that I call directly on `modalRef.componentInstance` after opening the modal and setting the appropriate variables, but it feels like a pretty heavy handed solution.

But at least I'm able to use the modal for my use case, so that is a plus.

PR #916, opened 2016-10-19
chore: update bootstrap to ^4.0.0-alpha.5

### LINKED ISSUE #915: switch to bootstrap alpha-5
Bootstrap alpha-5 is out. 

I noticed the following change that should affect the template of popover and tooltip: https://github.com/twbs/bootstrap/pull/17238

jnizet:
Closes #915

BREAKING CHANGE: ng-boostrap now uses bootstrap 4.0.0-alpha.5, which modified the markup used to generate tooltip and popover arrows (see https://github.com/twbs/bootstrap/pull/17238). Using ng-bootstrap with an older version of bootstrap will thus not display tooltip and popover arrows anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Check all the directives in action at https://ng-bootstrap.github.io
 
 ## Dependencies
 * [Angular 2](https://angular.io) (tested with 2.0.0)
-* [Bootstrap 4](https://v4-alpha.getbootstrap.com) (tested with 4.0.0-alpha.4)
+* [Bootstrap 4](https://v4-alpha.getbootstrap.com) (tested with 4.0.0-alpha.5)
 
 ## Installation
 After installing the above dependencies, install `ng-bootstrap` via:
--- a/demo/src/app/app.component.html
+++ b/demo/src/app/app.component.html
@@ -10,7 +10,7 @@
           <a class="nav-link" [routerLink]="['/components']">Components</a>
         </li>
       </ul>
-      <p class="github-buttons pull-lg-right">
+      <p class="github-buttons float-lg-right">
         <a class="github-button"
           href="https://github.com/ng-bootstrap/ng-bootstrap"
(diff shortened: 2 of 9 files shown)


pkozlowski-opensource:
From my quick testing datepicker was a bit broken as well. Or was it only my env?

jnizet:
I'll check that.

jnizet:
There is indeed a small visual glitch in the datepicker navigation. I'll investigate. If anyone has an idea...
capture d ecran 2016-10-19 a 19 16 58

pkozlowski-opensource:
I was just wondering if we can keep the `<div class="tooltip-arrow"></div>` element around for a release or 2 to avoid breaking changes. If I understand the change correctly it wouldn't hurt (I mean - there should be no visual effect, right?).

jnizet:
I've added a css rule to the navigation select to keep its original height, and reverted the removal of the .tooltip-arrow and .popover-arrow divs, but added a TODO to remember removing them later.

pkozlowski-opensource:
@jnizet thnx, this looks good. I'm planning to cut another release tomorrow and after I will re-test entire site again with your fix.

pkozlowski-opensource:
Just noticed one more thing - there is sth wrong with the header: gh and twitter buttons got pulled to the left instead staying on the right. There were many changes to the navbar so we are probably hitting one of them.

jnizet:
I'll try looking at this tomorrow. Sorry I completely missed that.

jnizet:
@pkozlowski-opensource I couldn't reproduce this problem. It does look a little bit different in small screen sizes (where it doesn't look nice in both old and new versions), but I can't see the buttons going to the left as you described.

pkozlowski-opensource:
@jnizet sorry for the noise, I had a messed-up env when I saw "the bug". It is all good, merged. Thnx!

PR #935, opened 2016-10-20
fix(timepicker): properly display chevrons in older IEs

pkozlowski-opensource:

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -26,8 +26,9 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
       left: 0.05em;
       position: relative;
       top: 0.15em;
-      -webkit-transform: rotate(-45deg);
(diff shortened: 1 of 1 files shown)


src/timepicker/timepicker.ts
```diff
@@ -26,15 +26,17 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
       left: 0.05em;
       position: relative;
       top: 0.15em;
-      -webkit-transform: rotate(-45deg);
       transform: rotate(-45deg);
+      -webkit-transform: rotate(-45deg);
+      -ms-transform: rotate(-45deg);
```
wesleycho: Shouldn't the regular `transform` go after all of these vendor prefixes in order to override them?

pkozlowski-opensource: Maybe in the future, but this is not the case today....

PR #936, opened 2016-10-20
demo: switch logo png -> svg

pkozlowski-opensource:
Our logo in the png format is not optimized but
even after optimization it would be larger than svg.
On top of this SVG looks better so let's switch!


wesleycho:
LGTM

PR #944, opened 2016-10-21
feat(typeahead): allow selection of null item

### LINKED ISSUE #877: Feature (Typeahead): Keydown.Enter without focused item to trigger selectItem event
Use case:

Solution:

deeg:
Closes #877

Allows selection event to be fired when focus is not on a result from the typeahead. This allows you to fire the selection event with an empty item when focus is on the search box (focusFirst=false)


src/typeahead/typeahead.ts
```diff
@@ -187,6 +192,12 @@ export class NgbTypeahead implements ControlValueAccessor,
 
   handleKeyDown(event: KeyboardEvent) {
     if (!this._windowRef) {
+      if (event.which === Key.Enter) {
```
deeg: I did not want to do it for tab also, since I would still want tab to move focus to the next focusable item.

I'm not sure I would ever really want tab to select an item from the autocomplete. Maybe this is something we can make configurable in the future?

deeg:
Not sure why CI failed. Everything seems to pass on my machine. The clang errors in the log are on files I did not change.

deeg:
Messed up this PR, going to close and open another clean one.

PR #950, opened 2016-10-23
demo: pin versions of dependencies in plunkr

### LINKED ISSUE #949: plunkers: pin versions of ng-bootstrap, angular and its deps
As of today generated plunkers point to the latest version of modules on npm. This poses different problems:
* `unpkg.com` is caching aggressively so it might not "see" the newest release of ng-bootstrap / angular
* it _might_ happen that dependencies (angular, RxJS etc.) loaded in a plunker might be different from the ones we are using and it _might_ break things in an unexpected way.
* users might open an issue with one version of dependencies and with new releases plunkers might show different behaviour (or just break on load).

Remedy: pin versions of ng-bootstrap, angular, RxJS etc. in the generated plunks. Versions used in `config.js` should be taken from `package.json`

jnizet:
Closes #949

Notes:
- I also set the versions in scripts of the generated index.html file
- except for reflect-metadata, which is not referenced in package.json.

Please tell if you think I should add reflect-metadata to the dev dependencies.


misc/plunk-gen.js
```diff
@@ -89,4 +90,40 @@ ${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName
...
+  Object.keys(versions).forEach((lib) => {
+    result = result.replace(new RegExp(`%${lib}%`, 'g'), versions[lib]);
+  });
+  return result;
+}
+
+function getVersions() {
+  return {
+    angularCore: getVersion(false, '@angular/core'),
+    angularCommon: getVersion(false, '@angular/common'),
```
pkozlowski-opensource: IMO we could use the `angularCore` var everywhere (except of router of course) as all the other modules (common, platform-browser, compiler etc.) will follow the same release cycle and numbers.

jnizet: You're right of course. I'll rename it to angularVersion and use it everywhere we can.

misc/plunker-builder-templates/index.html
```diff
@@ -5,10 +5,10 @@
   <base href="." />
   <title>ng-bootstrap demo</title>
   <link rel="stylesheet" href="http://v4-alpha.getbootstrap.com/dist/css/bootstrap.min.css" />
-  <script src="https://unpkg.com/zone.js/dist/zone.js"></script>
-  <script src="https://unpkg.com/zone.js/dist/long-stack-trace-zone.js"></script>
+  <script src="https://unpkg.com/zone.js@%zoneJs%/dist/zone.js"></script>
```
pkozlowski-opensource: Maybe we could move content of this file into code so we could use ES6 template strings instead of inventing our own tempting syntax?

pkozlowski-opensource:
> Please tell if you think I should add reflect-metadata to the dev dependencies

Yeh, we should probably add it.

misc/plunk-gen.js
```diff
@@ -89,4 +90,40 @@ ${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName
...
+    angularForms: getVersion(false, '@angular/forms'),
+    typescript: getVersion(true, 'typescript'),
+    rxjs: getVersion(true, 'rxjs'),
+    ngBootstrap: packageJson.version,
+    zoneJs: getVersion(true, 'zone.js'),
+    systemjs: getVersion(true, 'systemjs')
+  };
+}
+
+function getVersion(dev, name) {
```
jnizet: No problem doing that.

misc/plunker-builder-templates/index.html
jnizet: No problem.

Issue #898, opened 2016-10-17
Pagination. Call pageChange to many times on Init


kolkov:
1.0.0-alpha.8 
why pageChange calls to much at view init with NAN as event?
![image](https://cloud.githubusercontent.com/assets/3740898/19439507/02324dfc-9487-11e6-9c46-f3cda5252a2c.png)

What could be the problem?

pkozlowski-opensource:
> What could be the problem?

No idea without a minimal reproduce scenario: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/README.md#you-think-youve-found-a-bug

kolkov:
Once this code here affects, it is not clear ...
```
getActorsList(): void{
      this.actorService
        .getActorsList(this.currentPage, this.itemsPerPage)
        .subscribe(actorList => {
          this.actors = actorList.Actors;
          this.totalItems = actorList.Total;
      });
[...]

PR #956, opened 2016-10-24
refactor(pagination): clarify edge cases and throw errors

maxokorokov:
* `ngOnChanges` suspicious calls clarified
* pagination throws errors to avoid confusion when providing non-numeric values for `page`, `pageSize`, `maxSize` and `collectionSize` (as is #898). This will also avoid emitting `NaN` values in `(pageChange)`
* added more tests for edge cases

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -440,6 +440,74 @@ describe('ngb-pagination', () => {
       fixture.detectChanges();
       expectPages(fixture.nativeElement, ['« Previous', '1', '2', '3', '4', '+5', '6', '7', '» Next']);
     });
+
+    describe('invalid inputs', () => {
+
+      ['page', 'collectionSize', 'maxSize', 'pageSize'].forEach((input) => {
+
+        it(`should throw for invalid "${input}" value`, () => {
+          const fixture = createTestComponent(`<ngb-pagination [collectionSize]="collectionSize" 
+            [maxSize]="maxSize" [pageSize]="pageSize" [page]="page"></ngb-pagination>`);
+          fixture.componentInstance[input] = NaN;
+          expect(() => { fixture.detectChanges(); }).toThrow();
+        });
+      });
+    });
+
+    it('should display one page for empty collection', fakeAsync(() => {
+         const html =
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.ts
```diff
@@ -224,4 +199,40 @@ export class NgbPagination implements OnChanges {
       this.pageChange.emit(this.page);
     }
   }
+
+  private _updatePages(newPage: number) {
+    // re-calculate new length of pages, should always have at least 1
```
pkozlowski-opensource: Why `1` and not `0`?

src/pagination/pagination.spec.ts
```diff
@@ -440,6 +440,74 @@ describe('ngb-pagination', () => {
...
+
+    describe('invalid inputs', () => {
+
+      ['page', 'collectionSize', 'maxSize', 'pageSize'].forEach((input) => {
+
+        it(`should throw for invalid "${input}" value`, () => {
+          const fixture = createTestComponent(`<ngb-pagination [collectionSize]="collectionSize" 
+            [maxSize]="maxSize" [pageSize]="pageSize" [page]="page"></ngb-pagination>`);
+          fixture.componentInstance[input] = NaN;
+          expect(() => { fixture.detectChanges(); }).toThrow();
```
pkozlowski-opensource: Wouldn't it be good to verify the exact message here? At least for one of inputs.

src/pagination/pagination.ts
maxokorokov: I think you always have at least one page in pagination, otherwise it would be only `<` and `>` arrows, which looks strange

pkozlowski-opensource:
@maxokorokov had another look at this PR and I would like to go back to the discussion on 2 items:
* throwing errors on invalid inputs vs. auto-correcting it
* displaying page "1" vs. not displaying anything when the collection size is 0

I'm kind of on the fence when it comes to those 2 items but more leaning towards not throwing and 0 vs. 1. But open to be convinced otherwise :-)

More generally I think I would prefer that we split this PR into 2 parts:
* actual bug fix
* change in behavior

This would make it easier to discuss / review and prioritize.

maxokorokov:
No pb, will open two separate requests

PR #958, opened 2016-10-24
feat(typeahead): allow selection of null item

### LINKED ISSUE #877: Feature (Typeahead): Keydown.Enter without focused item to trigger selectItem event

deeg:
Allows selection of empty item when focusFirst = false and focus is not on any options from the typeahead results.

Closes #877

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -46,7 +46,7 @@ const NGB_TYPEAHEAD_VALUE_ACCESSOR = {
  */
 export interface NgbTypeaheadSelectItemEvent {
   /**
-   * An item about to be selected
+   * An item about to be selected. Null if selecting without focus on a result.
    */
   item: any;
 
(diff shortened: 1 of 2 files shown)


src/typeahead/typeahead.ts
```diff
@@ -187,6 +192,12 @@ export class NgbTypeahead implements ControlValueAccessor,
 
   handleKeyDown(event: KeyboardEvent) {
     if (!this._windowRef) {
+      if (event.which === Key.Enter) {
```
deeg: I did not want to do it for tab also, since I would still want tab to move focus to the next focusable item.

I'm not sure I would ever really want tab to select an item from the autocomplete. Maybe this is something we can make configurable in the future?

deeg:
@pkozlowski-opensource,  am I missing anything for this to get started being reviewed?

Mind if I get a needs review tag so someone can take a look when they have time?

pkozlowski-opensource:
@deeg sorry for the delay. I'm batching several items for typeahead and will work on this early next week.

pkozlowski-opensource:
@deeg would you mind having a look at https://github.com/ng-bootstrap/ng-bootstrap/issues/877#issuecomment-259536376 ?

Issue #952, opened 2016-10-24
Show the rating value in readonly mode


Hongbo-Miao:
### Bug description:
Right now it shows the value your mouse hovered, even the rating component is readonly.

For readonly rating component, I think it will make sense to show the value that this rating component stands for. In this case, it should be 3.14

OR just provide the choice not showing the "tooltip".

![image](https://cloud.githubusercontent.com/assets/3375461/19634161/86e5ef28-9986-11e6-9460-3885282091f4.png)

### Link to minimally-working plunker that reproduces the issue:
http://plnkr.co/edit/V6VdESRFzbCobOnfcHvd?p=preview

### Version of Angular, ng-bootstrap, and Bootstrap:
Angular: 2.1.1
ng-bootstrap: 1.0.0-alpha.9
Bootstrap: 4.0.0-alpha.5

maxokorokov:
@pkozlowski-opensource is there any particular reason of having `title` set for each span? 

I think I would 

- either move both `title` and `aria-valuetext` to the span that contains all the star templates (next to `aria-valuenow`)
- or remove both of them completely as they're apparenlty redundant to `aria-valuenow` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuetext_attribute

pkozlowski-opensource:
@maxokorokov maybe @Foxandxss will remember. But yeh, I agree that we should do review of accessibility here.

pkozlowski-opensource:
Fixed by 9b5f99937b7ad06fac461dad8a269e591da8a451

PR #962, opened 2016-10-25
fix(rating): improves rating accessibility

maxokorokov:
Two changes for (#952):

- removing the title from each span with the star, as it looks confusing. It's still easily possible to add a title to the rating for the whole component if necessary
- adding `role=slider` and moving `[attr.aria-valuetext]` next to other attributes. Tested with VoiceOver on the Mac → it doesn't work without these two

https://www.w3.org/TR/wai-aria/roles#slider
https://www.w3.org/TR/wai-aria/states_and_properties#aria-valuetext

### DIFF
--- a/src/rating/rating.ts
+++ b/src/rating/rating.ts
@@ -30,11 +30,11 @@ export interface StarTemplateContext {
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
     <template #t let-fill="fill">{{ fill === 100 ? '&#9733;' : '&#9734;' }}</template>
-    <span tabindex="0" (mouseleave)="reset()" aria-valuemin="0" [attr.aria-valuemax]="max" [attr.aria-valuenow]="rate">
-      <template ngFor let-r [ngForOf]="range" let-index="index">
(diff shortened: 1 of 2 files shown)


src/rating/rating.ts
```diff
@@ -135,7 +137,7 @@ export class NgbRating implements OnInit,
   private _buildTemplateObjects(): number[] {
     let range = [];
     for (let i = 1; i <= this.max; i++) {
-      range.push({title: i});
+      range.push(i);
     }
     return range;
```
wesleycho: Maybe this is nitpicking, but one can squash all of this to one line
```ts
return Array.from({length: this.max}, (v, k) => v + 1);
```

wesleycho:
Other than that one nitpick, LGTM

src/rating/rating.ts
maxokorokov: Thanks, amended

wesleycho: It might be k + 1 - double check :)

maxokorokov:
No idea why travis is failing. I haven't even touched `popover.ts` here...

wesleycho:
LGTM

PR #974, opened 2016-10-27
fix(modal): make it easier to use components from lazy-loaded modules

### LINKED ISSUE #947: Cannot use components as content for modals in lazy loaded modules
### Bug description:

The feature to open components as content for modals does not work in lazy loaded modules.
The plunker below throws the error:  No component factory found for NgbdModalContent

http://plnkr.co/edit/O8bkySFwvCS1zt9emQSV?p=preview

Angular: 2.1.0

ng-bootstrap: 1.0.0-alpha.9

Bootstrap: 4.0.0-alpha.4

pkozlowski-opensource:
Soooooo, this is the first time when I'm kind of puzzled on how to write a test for this.

The issue here is that each lazy loaded module gets its own instance of `ComponentFactoryResolver` so to test this situation we would have to simulate lazy-loaded modules and I'm not sure sure how to do this under a test (not that I've researched this extensively...)

Fixes #947

### DIFF
--- a/src/modal/modal-container.ts
+++ b/src/modal/modal-container.ts
@@ -32,9 +32,9 @@ export class NgbModalContainer {
     ngbModalStack.registerContainer(this);
   }
 
-  open(content: string | TemplateRef<any>, options): NgbModalRef {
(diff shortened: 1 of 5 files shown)


wesleycho:
LGTM

Issue #862, opened 2016-10-10
Datepicker evolution


maxokorokov:
The datepicker will evolve. This is the issue to track down future enhancements, progress, discussion, etc:
- [x] Date range selection (#753)
- [x] Multiple months display (#977)
- [x] Visual improvements for default view (#706)
- [x] Ensure additional calendars support (#656)
- [x] Keyboard navigation (#1551)
- [x] Allow for custom date model (#1816)
- [x] Custom date models (#1979)
- [ ] Navigation customization / replacement
- [ ] Week numbers, weekdays and month names custom display (#1060)
- [x] Add content to datepicker template (#1197)

A little more details in the following [document](https://docs.google.com/document/d/1QFBpWNLH4YdxTE_gtYmHlBdknjpqTXJAjYuFHg32Q3k/edit?usp=sharing). Feel free to suggest and comment.

ravipunjwani:
@maxokorokov I would suggest a few more. Please consider if possible.

- [ ] Configuration of HTML datepicker fallback for mobile devices (optional configuration).
- [ ] The current datepicker works great with larger screens but in mobile if the input-box is placed to the right corner, the datepicker popover gets trimmed due to screen boundaries.
- [ ] Custom formatting of date. Current workaround is way longer and would be great if it just allows with simple date format strings like "dd-MMM-yyyy" so it would just work.

aitboudad:
can we have also support for `Date & Time Picker` ?

ErgoSoftUK:
Are there any plans to implement a "month-year" picker (http://jsfiddle.net/brunodd/k5zookLt/20/)

maxokorokov:
@ErgoSoftUK I was thinking about it **very** long-term, but definitely not in the nearest future, as there are plenty more priority use cases. Unless someone wants to help and do it of course :)

ErgoSoftUK:
@maxokorokov Fair enough, thanks for the response max.  I think I also spotted a bug in there.  In the demo for the popup (with input box), if you enter an invalid year (just enter "33333") then click the pop-up icon, chrome tends to completely freeze.

PEsteves8:
+1 to the suggestion of custom date formats on the input view. It would be nice to have full names for months (or event possibly weekdays), for instance.

rodrigofns:
@ravipunjwani What's the current workaround to custom format a date, like "dd-mm-yyyy"?

coutin:
[...]

PR #977, opened 2016-10-28
feat(datepicker): ability to display several months

maxokorokov:
Multiple month display support for the datepicker (part of #862):
Not that many changes here, but many files touched...
- `[displayMonths]` → new input to specify number of months
- `[showNavigationSelect]` → new input to toggle month/year select boxes in navigation

Looks like this with the select boxes:

screen shot 2016-10-28 at 15 12 15

And like this without them:

screen shot 2016-10-28 at 15 11 16

### DIFF
--- a/src/datepicker/datepicker-config.ts
+++ b/src/datepicker/datepicker-config.ts
@@ -10,12 +10,13 @@ import {NgbDateStruct} from './ngb-date-struct';
 @Injectable()
 export class NgbDatepickerConfig {
   dayTemplate: TemplateRef<DayTemplateContext>;
+  displayMonths = 1;
   firstDayOfWeek = 1;
   markDisabled: (date: NgbDateStruct, current: {year: number, month: number}) => boolean;
   minDate: NgbDateStruct;
   maxDate: NgbDateStruct;
+  navigation: 'select' | 'arrows' | 'none' = 'select';
   outsideDays: 'visible' | 'collapsed' | 'hidden' = 'visible';
-  showNavigation = true;
   showWeekdays = true;
   showWeekNumbers = false;
   startDate: {year: number, month: number};
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -1,4 +1,4 @@
-import {Component, Input, OnChanges, TemplateRef, forwardRef, OnInit, SimpleChanges} from '@angular/core';
+import {Component, Input, OnChanges, TemplateRef, forwardRef, OnInit, SimpleChange, SimpleChanges} from '@angular/core';
 import {NG_VALUE_ACCESSOR, ControlValueAccessor} from '@angular/forms';
 import {NgbCalendar} from './ngb-calendar';
 import {NgbDate} from './ngb-date';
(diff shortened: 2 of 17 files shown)


src/datepicker/datepicker.ts
```diff
@@ -222,8 +260,21 @@ export class NgbDatepicker implements OnChanges,
...
+      if (force || index === -1) {
+        newMonths.push(
+            this._service.generateMonthViewModel(
+                newDate, this._minDate, this._maxDate, toInteger(this.firstDayOfWeek), this.markDisabled));
+      } else {
+        newMonths.push(this.months[index]);
+      }
+    }
+
+    this.months = newMonths;
```
wesleycho: Not sure if it's preferred, but this can be rewritten as
```ts
Array.from({length: this.displayMonths}, (_, i) => {
  const newDate = this._calendar.getNext(this._date, 'm', i);
  const index = this.months.findIndex(month => month.firstDate.equals(newDate));
  return force || index === -1 ?
    this._service.generateMonthViewModel(
                newDate, this._minDate, this._maxDate, toInteger(this.firstDayOfWeek), this.markDisabled)) :
   this.months[index];
});
```

maxokorokov: Hmm, me neither. Not sure if it hurts readability or not. It looks nice for real array-like objects and as a shortcut for simple initialization as in rating.

src/datepicker/datepicker-config.ts
```diff
@@ -10,12 +10,14 @@ import {NgbDateStruct} from './ngb-date-struct';
 @Injectable()
 export class NgbDatepickerConfig {
   dayTemplate: TemplateRef<DayTemplateContext>;
+  displayMonths = 1;
```
pkozlowski-opensource: Maybe `noOfMonths` or sth?

src/datepicker/datepicker-config.ts
```diff
@@ -10,12 +10,14 @@ import {NgbDateStruct} from './ngb-date-struct';
...
   firstDayOfWeek = 1;
   markDisabled: (date: NgbDateStruct, current: {year: number, month: number}) => boolean;
   minDate: NgbDateStruct;
   maxDate: NgbDateStruct;
   outsideDays: 'visible' | 'collapsed' | 'hidden' = 'visible';
   showNavigation = true;
+  showNavigationSelect = true;
```
maxokorokov: hmm...

`showNavigation` → hides / shows navigation completely, leaving only months to be displayed
`showNavigationSelect` → toggles month and year select boxes

Actually you're right, now I think of merging them into:

```ts
navigation: 'select' | 'arrows' | 'hidden';
```

P.S. not sure about naming... ideas?

src/datepicker/datepicker-config.ts
maxokorokov: Yeah, I know... naming things. Maybe just `months: number;` ?

src/datepicker/datepicker-config.ts
benouat: May I suggest:

`navigation: true / false (default: true)` completely show hide the navigation header
`quickNavigation: true / false (default: false)` activate the quick/rapid navigation with selects

maxokorokov:
Amended the breaking change: `showNavigation` and `showNavigationSelect` now replaced with:

```ts
navigation: 'select' | 'arrows' | 'none';
```

src/datepicker/datepicker-config.ts
maxokorokov: Oh, haven't seen your comment there in time. 
I think I prefer to have one field in case we want to add other navigation types later (meaning `quickNavigation` won't work)

PR #979, opened 2016-10-28
fix(pagination): better protect against invalid inputs

maxokorokov:
Second attempt to improve situation with pagination invalid inputs (related to #898)

- Added tests for edge cases for `collectionSize`, `page`, `maxSize` and `pageSize` to see the pagination behaviour
- Cleaned up `ngOnChanges`
- Essentially the main change is:

```ts
if (!isNumber(this.pageCount)) {
  this.pageCount = 0;
}
```


pkozlowski-opensource:
@maxokorokov does it fix #898 ?

maxokorokov:
@pkozlowski-opensource yes it should fix it, as we won't emit `NaN`

Oops, forgot to add the test that checks that 'pageChange' is not emitted for incorrect values when splitting previous pull request. Amended.

PR #997, opened 2016-11-02
refactor(modal): removal usage of 'fade in' classes

pkozlowski-opensource:
Bootstrap [renamed](https://github.com/twbs/bootstrap/commit/aa11f002182877fe545933d6ac6009cd5363072c) 'in' to 'active' so moving to next alpha
would break modal. But more importantly we won't need those
classes anyway as soon as we move to Angular-powered animations

### DIFF
--- a/src/modal/modal-backdrop.ts
+++ b/src/modal/modal-backdrop.ts
@@ -1,5 +1,5 @@
 import {Component} from '@angular/core';
 
-@Component({selector: 'ngb-modal-backdrop', template: '', host: {'class': 'modal-backdrop fade in'}})
+@Component({selector: 'ngb-modal-backdrop', template: '', host: {'class': 'modal-backdrop', '[style.opacity]': '0.5'}})
 export class NgbModalBackdrop {
 }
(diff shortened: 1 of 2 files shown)


benouat:
@pkozlowski-opensource maybe a stupid question: what would be the added value to move away from CSS-based bootstrap animations ? Angular-powered one are great, but in this particular use-case I am not buying it... Any highlights ?

pkozlowski-opensource:
@benouat highlights would be:
* cross-platform (server-side rendering, web-worker etc.)
* you can't really do stuff with pure CSS only (things like cancelling, animating height: auto etc. require JS anyway and IMO it is better to use ng2 infrastructure instead of rolling out our custom JS)

benouat:
Actually my main concern here is simple:
The whole advantage of such a library (ng-bootstrap) is to get rid of all the query stuff and re-implement them using ng2.
To me, the CSS part should not be touched. And by touched I mean it should not be short-circuited either.

Bootstrap already comes with animations, written in CSS, not in javascript (jquery).
I would assume the project to stick to it.
I would not be against plugins, such as a ng-bootstrap-animate. but not baked in the default lib.

That's why I asked the question in the first place. 😄

pkozlowski-opensource:
> Bootstrap already comes with animations, written in CSS, not in javascript (jquery).

True, but CSS is enough - you still need JS that will add / remove CSS classes at the right time - and this is the part that we need to re-implement anyway.

PR #1002, opened 2016-11-04
feat(datepicker): add navigation notification with 'navigate' output

### LINKED ISSUE #986: NgbDatepicker: Provide possibility to hook into date navigation
### Bug description:

I'd like to know when the user navigates the datepicker by clicking the arrows, selecting a month or selecting a year. 

My datepicker more or less functions as an "appointments calendar": when a user has an appointment on a certain day, that day is shown in a dedicated color (I am using the "custom day view" for this).

The problem now is that I have to fetch all appointments beforehand. It would make sense if I can fetch the appointments _after_ the user has navigated.

Something like this would be great:

`<ngb-datepicker (navigated)=onNavigate($event)></ngb-datepicker>
`

Thanks for this great library!

Angular: 2.1.0

ng-bootstrap: 1.0.0-alpha.5

Bootstrap: 4.0.0-alpha.5

maxokorokov:
Adds the `navigate` output to get notified when current month changes

```ts
<ngb-datepicker (navigate)=onNavigate($event)></ngb-datepicker>
```

Fixes #986

### DIFF
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -1,4 +1,14 @@
-import {Component, Input, OnChanges, TemplateRef, forwardRef, OnInit, SimpleChange, SimpleChanges} from '@angular/core';
+import {
+  Component,
+  Input,
+  OnChanges,
+  TemplateRef,
+  forwardRef,
+  OnInit,
+  SimpleChanges,
+  EventEmitter,
+  Output
+} from '@angular/core';
 import {NG_VALUE_ACCESSOR, ControlValueAccessor} from '@angular/forms';
 import {NgbCalendar} from './ngb-calendar';
 import {NgbDate} from './ngb-date';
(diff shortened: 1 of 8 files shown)


src/datepicker/datepicker.ts
```diff
@@ -139,6 +149,12 @@ export class NgbDatepicker implements OnChanges,
    */
   @Input() startDate: {year: number, month: number};
 
+  /**
+   * An event fired when navigation happens and currently displayed month changes.
+   * The payload is the new month we're navigating to.
+   */
+  @Output() navigate = new EventEmitter<{year: number, month: number}>();
```
pkozlowski-opensource: Don't we want previous and next month so people could figure out direction of navigation? Just asking since changing it in the future would be a bc.

maxokorokov: On the one hand additional info won't hurt, on the other hand people can save previous month on their own, so not sure.

Would be smth like:
```ts
export interface NgbDatepickerNavigationEvent {
  current: {year: number, month: number};
  next: {year: number, month: number};
}
```

pkozlowski-opensource: `preventDefault()` could potentially make sense as well :-) So yeh, maybe let's do an event with `current` and `next` to keep our options open?

maxokorokov: Amended `NgbDatepickerNavigateEvent`

PR #1007, opened 2016-11-05
chore: add coverage

Foxandxss:
It generates a .json (unreadable for humans) with test or tdd, but only test does a remap to generate the html reports.

There is also a `gulp remap-coverage` that generates the html reports from that .json.

I could remove the .json generation with TDD But requires more changes on karma. Perhaps for a next pr.


gulpfile.js
```diff
@@ -159,26 +161,34 @@ gulp.task('build:tests', ['clean:tests'], (cb) => {
...
-    }
+gulp.task('test', ['build:tests'], function() {
+  startKarmaServer(false, false, () => {
+    return gulp.src(PATHS.coverageJson).pipe(remapIstanbul({reports: {'html': 'coverage/html'}}));
   });
 });
 
+gulp.task(
+        'remap-coverage',
+        function() { return gulp.src(PATHS.coverageJson).pipe(remapIstanbul({reports: {'html': 'coverage/html'}})); })
```
pkozlowski-opensource: Sth went wrong here. A missing `;`?

Foxandxss: let me see, thought it was just clang :P

PR #1008, opened 2016-11-05
chore: adjust some test helpers

Foxandxss:
on common.ts, that return browser is never reached, that hurts the coverage.

On matchers, the message is a bit overcomplex and fails the coverage as well, now it pass coverage and works as expected as well.


pkozlowski-opensource:
LGTM

PR #1009, opened 2016-11-05
feat: add a NgbMediaQuery service

### LINKED ISSUE #957: Feature Request: Media Query Service
# Background

One of the core concepts in bootstrap are the media queries (`sm`, `md`, `lg`, `xl`) where you can conditionally show/hide elements based upon the browser size.
# Proposal

Expose a service where users can get the media query as an observable

``` TypeScript
// media query types
export enum NgbMediaQueryOption { xs, sm, md, lg }

// service to provide access
export class NgbMediaQueryService {
  getMediaQuery(): Observable<NgbMediaQueryOption> { ... }
}

// sample usage
@Component({ ... })
export class SampleComponent implements OnInit{
  isXs: Observable<boolean>
  constructor(private media: ngbMediaQueryService) {}
  ngOnInit() {
    this.isXs = this.media.getMediaQuery()
      .map(m => m === NgbMediaQueryOption.xs);
  }
}
```
# Future Considerations

Create corresponding directives that could internally use this above service for example...
`ngbIfXs` - to only show in xs view
`ngbIfHideXs` - to hide only in xs view

Allow configuration values to override the default media query breakpoints

jnizet:
Even though it's not in the top priorities, I wanted to try implementing this media query service, which looked simple enough for me, and was requested in #957.

Here are some design choices I made, which are of course open to discussion:
- make the breakpoint names and values configurable, as I think this is supported by bootstrap; use the standard bootstrap breakpoints by default. 
- since the breakpoints are configurable, use a string for the names ('md', 'lg', etc.) instead of an enum. Strings are also easier to use from the templates
- do not create special directives like `ngbIfXs`. The number of those directives would be huge (sm, sm-up, sm-down, etc.), and they would not support custom breakpoints anyway. Reimplementing ngIf is also something I think we shouldn't do. The demo shows how to use ngIf in combination with the service, and being able to combine media query conditions with other conditions in a larger boolean expression is useful anyway
- a single resize listener is used and shared by all the subscribers. Not sure if it's the best thing to do or not. The disadvantage is that once a component uses the service, the listener is added to the window and never removed. I'm not sure this is a good idea. Maybe I'm pre-optimizing here.

Problem faced during the implementation: the ngc compilation failed as soon as I referenced the Window type. I have no idea if it's intentional to fail when this type is used or if it's a configuration problem.

Potential problems: Since the service obviously gets the window width, using it on the server (with universal) will most probably not work. I don't know much about universal, but I guess this problem should be handled by the application, and not by ng-bootstrap.

fix #957


jnizet:
Also: I added a MediaQuery "Component" in the demo, but it's not really a component. Not sure if we should add a "Services" section, which would only contain this service.

jnizet:
The build failed because of a clang-format problem on popover, which I didn't touch :-/

pkozlowski-opensource:
As discussed we will skip this feature as outside of the scope of this project. This could be a separate min-library living outside of this repo but under ng-bootstrap organization.

PR #1010, opened 2016-11-06
docs(accordion): distinguish from global and specific types

Foxandxss:
Right now both types explanation is the same and they don't tell you that one of them is applied globally and the other is applied per panel basis.

Feel free to suggest a different wording.


pkozlowski-opensource:
@Foxandxss could you make `lint` happy?

Foxandxss:
Done.

PR #1013, opened 2016-11-07
fix(modal): don't stop propagation of click events from content

### LINKED ISSUE #1011: Autoclose in dropdown does not work when dropdown is inside a modal

Angular: 2.1.0

ng-bootstrap: 1.0.0-alpha.11

Bootstrap: 4.0.0-alpha.5

pkozlowski-opensource:
Fixes #1011

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -7,7 +7,8 @@ import {
   Renderer,
   OnInit,
   AfterViewInit,
-  OnDestroy
+  OnDestroy,
+  ViewChild,
 } from '@angular/core';
 
 import {ModalDismissReasons} from './modal-dismiss-reasons';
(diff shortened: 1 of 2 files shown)


src/modal/modal-window.ts
```diff
@@ -41,8 +44,8 @@ export class NgbModalWindow implements OnInit,
 
   constructor(private _elRef: ElementRef, private _renderer: Renderer) {}
 
-  backdropClick(): void {
-    if (this.backdrop === true) {
+  backdropClick($event): void {
+    if (this.backdrop === true && !this._isNodeChildOfAnother(this.contentEl.nativeElement, $event.target)) {
```
wesleycho: The block
```ts
this._isNodeChildOfAnother(this.contentEl.nativeElement, $event.target)
```
is longer than how one would do it natively, i.e.
```ts
this.contentEl.nativeElement.contains($event.target);
```
Maybe this should be changed to a shorter name like `isContainedBy`, or just use the native version.

wesleycho:
LGTM, just have a minor comment.

pkozlowski-opensource:
Thnx @wesleycho - I've killed the `_isNodeChildOfAnother` abstraction and merged.

PR #1015, opened 2016-11-07
feat(rating): add keyboard support

maxokorokov:
Adds rating support for the keyboard as follows:

- arrow left, arrow down, page down → rate--
- arrow up, arrow right, page up → rate++
- home → rate = 0
- end → rate = max

Mostly based on https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

### DIFF
--- a/src/rating/rating.ts
+++ b/src/rating/rating.ts
@@ -11,6 +11,16 @@ import {
   ContentChild
 } from '@angular/core';
 import {NgbRatingConfig} from './rating-config';
+import {toString, getValueInRange} from '../util/util';
+
+enum Key {
+  End = 35,
+  Home = 36,
+  ArrowLeft = 37,
+  ArrowUp = 38,
+  ArrowRight = 39,
+  ArrowDown = 40
+}
 
 /**
  * Context for the custom star display template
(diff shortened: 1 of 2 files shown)


src/rating/rating.ts
```diff
@@ -28,6 +40,7 @@ export interface StarTemplateContext {
 @Component({
   selector: 'ngb-rating',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {'(keydown)': 'handleKeyDown($event)'},
```
maxokorokov: Not sure if should be 
- `keyup` - only increments by 1 once you release the key
- `keydown` - increments continuously by 1 while you hold the key

deeg: I feel keyup makes more sense. I feel for a rating, I would want to press the button x amount of times for the rating I'm giving.

Holding down seems like it would be harder to select the exact number and would cause more mistakes than precise selections.

maxokorokov: Yeah, I'm still uncertain about this one, but the [wai-aria slider demo](https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-2.html) and bootstrap js plugin use keydown apparently.

Plus from what I understand it's quite common to have this behaviour for navigation keys, so I'll leave it as is for now.

maxokorokov:
Amended. Removed `PgUp` and `PgDown` as they don't make much sense for the rating case

More links with examples:
- https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/slider/slider-1.html
- https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/slider/slider-2.html

maxokorokov:
no idea why it's failing again with some `popover.ts` issues

pkozlowski-opensource:
> no idea why it's failing again with some popover.ts issues

No idea either, we need to update clang-format version. But for this we need to merge all the in-flight PRs :-)

Issue #1012, opened 2016-11-07
Dropdown append to body


marcalj:
### Bug description:
Support for dropdown append-to-body.

### Use case
In a list inside an overflowed element, the dropdown is shown "inside" the main element, and you cannot seen it completely.

![image](https://cloud.githubusercontent.com/assets/63667/20058281/76cc7286-a4f0-11e6-86b4-b93e7fce5636.png)

Working with ng1 version:
![image](https://cloud.githubusercontent.com/assets/63667/20058295/88e6347a-a4f0-11e6-88a3-eba735254b91.png)

Thanks!

marcalj:
Coming from this issue: https://github.com/ng-bootstrap/ng-bootstrap/issues/584#issuecomment-258015148

@scttcper

PEsteves8:
I also need something like this. Any progress on it?

bharding777:
would really love this feature - constantly getting caught with dropdowns under other elements....

phillip-haydon:
Really needing this feature :(

codeMonkeysBe:
I need this feature as well. Can I help with the implementation of this?

clarkj:
My most common use case for dropdown-append-to-body is for building dropdown links in navbars.  Does anyone know a workaround for that?

ryanohearn:
I really need this feature. Any word on when this will be addressed?

Alukar:
+1

nponna:
+1

OzoTek:
Does anyone have a workaround for that?

kotmatpockuh:
+1

RonNewcomb:
+1

alesce:
+1

iwebd:
+1

nathan059:
+1

joedjc:
+1

OClyde:
+1

maarten-vandevelde:
+1

Asharma86:
This is a very important feature for this component. Stacking context is big concern and a problem very hard to solve when working with complex data-grids and such. I loved it how I was able to use the your popover with such ease. If you could use the same utility here it will be awesome!

@pkozlowski-opensource +1

florentsuc:
+1

mbstuart:
@pkozlowski-opensource, I have been working on adding this feature for use at my work. Would you be interested in a PR for my solution to this?

bluecaret:
+1

[...]

PR #1019, opened 2016-11-09
feat(dropdown): add support for appending to body

wesleycho:
- Add support for appending dropdown to body

Initial pass at #1012.

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -1,5 +1,18 @@
-import {Directive, Input, Output, EventEmitter, ElementRef} from '@angular/core';
+import {
+  Directive,
+  Input,
+  Output,
+  EventEmitter,
+  AfterContentInit,
+  OnDestroy,
+  ContentChild,
+  Renderer,
+  ComponentRef,
+  ElementRef,
+  NgZone
+} from '@angular/core';
 import {NgbDropdownConfig} from './dropdown-config';
+import {positionElements} from '../util/positioning';
 
 /**
  * Transforms a node into a dropdown.
@@ -15,7 +28,8 @@ import {NgbDropdownConfig} from './dropdown-config';
     '(document:click)': 'closeFromOutsideClick($event)'
   }
 })
-export class NgbDropdown {
+export class NgbDropdown implements AfterContentInit,
+    OnDestroy {
   private _toggleElement: any;
 
   /**
@@ -29,6 +43,12 @@ export class NgbDropdown {
   @Input() autoClose: boolean;
 
   /**
+   * A selector specifying the element the dropdown should be appended to.
+   * Currently only supports "body".
+   */
+  @Input() container: string;
+
+  /**
    *  Defines whether or not the dropdown-menu is open initially.
    */
   @Input('open') _open = false;
@@ -39,9 +59,16 @@ export class NgbDropdown {
    */
   @Output() openChange = new EventEmitter();
 
-  constructor(config: NgbDropdownConfig) {
+  @ContentChild(NgbDropdownMenu) private _dropdownMenu: NgbDropdownMenu;
+
+  private _zoneSubscription: any;
+
+  constructor(
+      private _elementRef: ElementRef, config: NgbDropdownConfig, private _renderer: Renderer,
+      private _ngZone: NgZone) {
     this.up = config.up;
     this.autoClose = config.autoClose;
+    this.container = config.container;
   }
 
 
@@ -55,6 +82,12 @@ export class NgbDropdown {
    */
   open(): void {
     if (!this._open) {
+      if (this.container === 'body') {
+        let container = window.document.querySelector(this.container);
+        this._renderer.attachViewAfter(container, this._dropdownMenu.elementRef.nativeElement);
+        this._renderer.setElementClass(container, 'open', true);
+      }
+
       this._open = true;
       this.openChange.emit(true);
     }
@@ -65,6 +98,12 @@ export class NgbDropdown {
    */
   close(): void {
     if (this._open) {
+      if (this.container === 'body') {
+        this._renderer.attachViewAfter(this._elementRef.nativeElement, this._dropdownMenu.elementRef.nativeElement);
+        let container = window.document.querySelector(this.container);
+        this._renderer.setElementClass(container, 'open', false);
+      }
+
       this._open = false;
       this.openChange.emit(false);
     }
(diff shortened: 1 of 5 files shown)


src/dropdown/dropdown.ts
```diff
@@ -1,5 +1,21 @@
...
+  EventEmitter,
+  AfterContentInit,
+  OnDestroy,
+  ComponentRef,
+  ElementRef,
+  NgZone
+} from '@angular/core';
 import {NgbDropdownConfig} from './dropdown-config';
+import {positionElements} from '../util/positioning';
+import {PopupService} from '../util/popup';
```
pkozlowski-opensource: Why do we need this import?

src/dropdown/dropdown.ts
```diff
@@ -1,5 +1,21 @@
...
+
+export class NgbDropdownWindow {
```
pkozlowski-opensource: Why do we need this class?

src/dropdown/dropdown.ts
```diff
@@ -15,10 +31,15 @@ import {NgbDropdownConfig} from './dropdown-config';
...
   }
 })
-export class NgbDropdown {
+export class NgbDropdown implements AfterContentInit, OnDestroy {
   private _toggleElement: any;
 
   /**
+   * Placement of a dropdown. Accepts: "top", "bottom", "left", "right"
+   */
+  @Input() placement: 'top' | 'bottom' | 'left' | 'right' = 'bottom';
```
pkozlowski-opensource: This is in conflict with the existing `up` option. IMO we should not add the `placement` option (do we really need it on left or right???) but instead position as `top-left` / `bottom-left` based on the `up` flag being set or not.

src/dropdown/dropdown.ts
```diff
@@ -93,12 +126,36 @@ export class NgbDropdown {
...
+      this._zoneSubscription = this._ngZone.onStable.subscribe(() => {
+        positionElements(
+            this._elementRef.nativeElement, this._dropdownMenu, this.up ? 'top-left' : 'bottom-left',
+            this.container === 'body');
+      });
+    }
+  }
+
+  ngOnDestroy() {
+    this._removeMenu();
```
pkozlowski-opensource: @wesleycho wouldn't it be enough to call `this.close()` and let Angular destroy nodes? This way we could drop the `_removeMenu()` method altogether, no?

src/dropdown/dropdown.ts
```diff
@@ -93,12 +126,36 @@ export class NgbDropdown {
     }
   }
 
+  ngAfterContentInit() {
+    this._dropdownMenu = this._elementRef.nativeElement.querySelector('.dropdown-menu');
```
pkozlowski-opensource: I'm not sure how I feel about querying for  `.dropdown-menu` like this... Maybe we should create a directive for this? The benefit would be that we could automatically add `dropdown-menu` class as well as all the required `aria` attributes.

src/dropdown/dropdown.ts
```diff
@@ -65,6 +93,11 @@ export class NgbDropdown {
    */
   close(): void {
     if (this._open) {
+      if (this.container === 'body') {
+        this._elementRef.nativeElement.appendChild(this._dropdownMenu);
+        window.document.querySelector(this.container).classList.remove('open');
```
pkozlowski-opensource: Could we go through a renderer here? More generally I'm a bit nervous about the amount of DOM manipulation that is going on here. I would definitively like to make sure that we are still properly rendered on the server-side with Universal.

pkozlowski-opensource:
@wesleycho did another pass on this one and left more comments. Sorry for not catching those things on the first pass....

wesleycho:
It's alright, I'll take a look sometime later today or this weekend.

About some of the DOM manipulation - I basically mirrored what was going on in the tooltip module. Perhaps we should investigate switching that too at some point.

wesleycho:
Updated PR, but looking like I'm having trouble mocking out Renderer - I could create a test util mock for Renderer for situations like this...not great, but might be necessary since Renderer is an abstract class.

WDYT?

deeg:
@wesleycho, just came across needing this feature. Where do we stand on this. Need any help wrapping it up?

src/dropdown/dropdown.ts
```diff
@@ -55,6 +82,12 @@ export class NgbDropdown {
    */
   open(): void {
     if (!this._open) {
+      if (this.container === 'body') {
+        let container = window.document.querySelector(this.container);
+        this._renderer.attachViewAfter(container, this._dropdownMenu.elementRef.nativeElement);
```
intellix: This will append the dropdown-menu AFTER body. This works for me with: `container.lastElementChild` but I don't know much about browser-support.
2nd parameter of attachViewAfter should be an array as well:
`this._renderer.attachViewAfter(container, [this._dropdownMenu.elementRef.nativeElement]);`

src/dropdown/dropdown.ts
```diff
@@ -65,6 +98,12 @@ export class NgbDropdown {
    */
   close(): void {
     if (this._open) {
+      if (this.container === 'body') {
+        this._renderer.attachViewAfter(this._elementRef.nativeElement, this._dropdownMenu.elementRef.nativeElement);
```
intellix: I guess same comment as above

src/dropdown/dropdown.ts
```diff
@@ -93,12 +132,44 @@ export class NgbDropdown {
     }
   }
 
+  ngAfterContentInit() {
+    if (this.container === 'body') {
+      this._zoneSubscription = this._ngZone.onStable.subscribe(() => {
+        positionElements(
```
intellix: I've got around 500 elements on page with dropdowns hidden inside them for when I click and this is causing quite an issue with performance as it's updating all of their positions all the time.
I fixed this locally by wrapping this in a check for open:
```
if (this._open) {
 ...
}
```

As that's only really when it needs to stick

src/dropdown/dropdown.ts
```diff
@@ -93,12 +132,44 @@ export class NgbDropdown {
...
   /**
    * @internal
    */
   set toggleElement(toggleElement: any) { this._toggleElement = toggleElement; }
 
   private _isEventFromToggle($event) { return !!this._toggleElement && this._toggleElement.contains($event.target); }
+
+  private _removeMenu() {
+    if (this._dropdownMenu) {
+      this._dropdownMenu.elementRef.nativeElement.parentNode.removeChild(this._dropdownMenu);
```
intellix: First property of removeChild should be Node, like so:
```
this._dropdownMenu.elementRef.nativeElement.parentNode.removeChild(this._dropdownMenu.elementRef.nativeElement);
```

src/dropdown/dropdown.ts
intellix: Actually, this has even larger performance implications. If you put a console.log(1); inside the subscribe, you'll see they happen more and more until it crashes. I think too many subscribes are happening. Is this supposed to complete on the first result? It seems like they're building up endlessly. I had 50x 1, then 604x 1 then 80,000x 1 until the browser crashed

adamk33n3r:
Is there something we can do to get this ball rolling again? Muchly needed feature :smile:

ASHFAQPATWARI:
This is a much needed feature. There doesn't seem to be any updates on this thread since Jan.

lugovsky:
@wesleycho @pkozlowski-opensource could you guys please provide any estimate on when this could be completed as we're trying to get an idea if we need to do a form to fix that or if this would be available in some nearest releases.

Let me also know if any help needed here

Thanks!

wesleycho:
I will probably revisit this and file a new PR sometime in the next two weeks (unless someone wants to beat me to it, that's fine too :) ). I have been very busy and relatively inactive in open source the past year or so, but I am inspired to be involved heavily again.

pkozlowski-opensource:
So this PR didn't see update in several months and got terribly outdated at this point :-/ We still need to have the append to body support but with all the changes in Bootstrap 4 markup and our dropdown implementation this work need to be started from scratch, I'm afraid...

PR #1020, opened 2016-11-09
demo: use lazy loading

mlaval:
This PR introduces lazy loading for the demo application.
It was done with the following setup which could be changed:
- 18 chunks: main page, the "Getting started" page and one for each component demo
- no pre-loading, chunks are downloaded only when a navigation happens

Om my machine, the size of the main bundle goes from 2.09MB to 605kB, and the site loads around twice faster! 

Note that some code (the shared modules) is duplicated in most chunks for some reason. It can be improved later.

### DIFF
--- a/demo/src/app/app.module.ts
+++ b/demo/src/app/app.module.ts
@@ -1,29 +1,21 @@
 import {NgModule} from '@angular/core';
 import {BrowserModule} from '@angular/platform-browser';
 
-import {NgbModule} from '@ng-bootstrap/ng-bootstrap';
-
 import {DefaultComponent} from './default';
-import {GettingStarted} from './getting-started';
 import {AppComponent} from './app.component';
 import {routing} from './app.routing';
-import {NgbdDemoModule} from './components';
 import {NgbdSharedModule} from './shared';
 
 @NgModule({
   declarations: [
     AppComponent,
-    DefaultComponent,
-    GettingStarted
+    DefaultComponent
   ],
   imports: [
     BrowserModule,
     routing,
-    NgbModule.forRoot(),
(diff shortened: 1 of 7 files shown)


pkozlowski-opensource:
@mlaval thnx for the PR, it is amazing to see those things working in practice. I just got your PR for a test drive and I can see ~20% reduction of time to `DOMContentLoaded` on my machine (~480ms => ~380ms) which is good progress. But as discussed inspecting generated bundles revealed that:
* main (app) bundle has entries that probably shouldn't be there (http, forms, ng-bootstrap) => dropping those would save 200+kB (30% for the home page) which could be interesting in terms of timing
* each demo component bundle duplicates:
** shared (api-docs and example box components)
** api-docs data themesevels

The duplication of the shared stuff brings ~100kb of code for each components and it starts to represent a lot for all components. So I think that we should see how could we minimize / remove this duplication. As discussed, path to explore would be:
* commons chunk plugin
* different bundle split (one for home page + getting started and another one for all component demos).

I'm going to keep this PR open so others can review / test it but I will hold merging of it till we can explore different options for removing code duplication among bundles.

mlaval:
I couldn't find a way to solve the dupes issue with so many bundles.
But here is another solution with 3 chunks only: main page, getting started and components (all of them together).

It would be nicer with some preloading, but this feature is only available in 2.1.0. For now, the big components chunk is loaded when the user accesses one of the demo.

pkozlowski-opensource:
@mlaval based on your experiments with the closure compiler I think that we should go down that path instead. Let's wait for 4.0 to become stable and then let's switch to the closure compiler! 

Closing this PR based on the above.

Issue #465, opened 2016-07-23
Accordion / Tabset: adding close button in a tab header


pkozlowski-opensource:
It is quite common to have a little close button on each tab of a tabset to close a given tab. Unfortunately we can't easily add such a button today since content of the `ngTabTitle` tab is added _inside_ header's link (`<a>` element). This setup limits what users can put in a tab header.

Potential solutions:
- have users provide the whole link as part of the `ngTabTitle` template - the downside here is that users would have to manage the `active` and `disabled` states as well as a selection logic;
- add a new directive (`ngTabHeader`) - if provided its content would be added _after_ a link. The downside here is that there might be someone coming and saying that I want to insert sth _before_ a link...

Not sure what is the best course of action here, to be discussed.

tsm91:
This is not valid html, but works fine.

```html
<a class="nav-link active" href="#">
  Active
  <button class="close" type="button">×</button>
</a>
```

Another solution would be (if nesting templates is possible in ng2):

```html
<ngb-tabset>
  <ngb-tab>
    <template ngbTabHeader>
      <!-- prepend something -->
      <template ngbTabTitle><b>Fancy</b> title</template>
      <!-- append something -->  
    </template>
    <template ngbTabContent>content</template>
  </ngb-tab>
</ngb-tabset>
```

pkozlowski-opensource:
Ref: https://github.com/twbs/bootstrap/issues/518

pkozlowski-opensource:
Another use-case to take into account: http://stackoverflow.com/questions/41767571/how-to-give-style-to-ngb-bootstrap-tab

sh977218:
Is this issue/improvement/suggestion being resolving or still in discussing?

sklipnoty:
Any updates? This is relatively easy in "normal" bootstrap => Yet I'm struggling with this ...

AbakumovAlexandr:
Instead of adding a Close `button` as a child of an `a` element, you can consider adding a styled `span` instead. A basic example is in this [Stackblitz](https://stackblitz.com/edit/ng-bootstrap-tab-set-issues).

maxokorokov:
Since version `5.2.0` there is a new `NgbNav` set of directives that fix this issue. `NgbNav` is meant to replace the `NgbTabset` component (completely deprecated in version `6.0.0`) as a more flexible alternative.

I suggest you migrate to it → https://ng-bootstrap.github.io/#/components/nav/overview

PR #1023, opened 2016-11-09
Toggle accordion using entire header

cmckni3:


pkozlowski-opensource:
@cmckni3 tnx for the PR. before I review it I would love to know more about motivations for this change. Could you tell us more about your use case / problem and how this PR helps?

For the reference we were previously discussing different options of changing clicks handling in accordion / tabs headers, see #465 and #717. Is your use-case similar to what was discussed in the mentioned issues? The thing is that #465 and #717 are aiming at providing generic solution that _could_ cover your use-case as well. 

Care to elaborate?

cmckni3:
Sort of similar to #465 except I want the entire header to be clickable (it's a bigger target area for the user). I'll admit this was my quick hack to get it working and in my application.

I think a better solution would be to remove the `a` and only have the template ref for the panel header. You could then provide the `a` and a different template as a default.

cmckni3:
Thoughts @pkozlowski-opensource?

PR #1032, opened 2016-11-11
feat(typeahead): do not trap keydown enter event

### LINKED ISSUE #877: Feature (Typeahead): Keydown.Enter without focused item to trigger selectItem event

### LINKED ISSUE #980: Selecting type-ahead item with keyboard.enter fires both selectItem event and keyup.enter on input field
Please remember, the issues forum is **NOT** for support requests. It is for bugs and feature requests only.
### Bug description:

I have an input field where the user needs to be able to enter an id to start a search.  They press enter on the field to start the search.  But I want them to be able to select from list of type-ahead entries as well.  This works fine in the following cases:
1. No type-ahead entries available, and user types in the id and presses Enter key
2. Type-ahead entries exist and user selects one using mouse

However, in the case where type-ahead entries exist and user navigates to desired type-ahead entry and presses enter key, then both the 'selectItem' event and the 'keyup.enter' event get fired.  This results in the search getting run twice, which is not a good (or efficient) thing.  In my code (not the Plunker), I've put a boolean in called 'searchAlreadyInProgress' that will be set to true once the search starts, then false, once search completes.  If that boolean is true when the search method is entered, I just return.  However, this is not a good solution, because error handling may be done further upstream.

Rather, it seems that this type ahead directive ought to provide a way to stop the propagation of the 'keyup.enter' event if the 'selectItem' event is fired....

Please note, that in my plunker, I demonstrate the issue by just logging a message. When you select a type-ahead entry via the keyboard, you will see these two messages in the console:

onSelectItem: item selected = Alabama
keyup.enter: id=Alabama

Whereas, when you select a type-ahead item using the mouse, you will only see this console message:

onSelectItem: item selected = Alabama

http://plnkr.co/edit/Xp1SldfD27rshGZtWJMv?p=preview

Angular:
2.1.0
ng-bootstrap:
1.0.0-alpha.9
Bootstrap:
4.0.0-alpha.4

deeg:
> * selectItem should not be fired when nothing is selected
>* we do not do preventDefault() on keydown.enter when there is nothing selected
>* we do preventDefault() and stopPropagation() when there are things to select.

Closes #958
Closes #877
Closes #980

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -28,8 +28,9 @@ enum Key {
 }
 
 function createKeyDownEvent(key: number) {
-  const event = {which: key, preventDefault: () => {}};
+  const event = {which: key, preventDefault: () => {}, stopPropagation: () => {}};
   spyOn(event, 'preventDefault');
+  spyOn(event, 'stopPropagation');
   return event;
 }
 
@@ -231,6 +232,7 @@ describe('ngb-typeahead', () => {
            expectInputValue(compiled, 'one');
            expect(fixture.componentInstance.model).toBe('one');
            expect(event.preventDefault).toHaveBeenCalled();
+           expect(event.stopPropagation).toHaveBeenCalled();
          });
        }));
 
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
Merged, thnx!

PR #1033, opened 2016-11-11
fix(timepicker): display 12 PM/AM when merdian is true

### LINKED ISSUE #1031: Timepicker: 12 PM is displayed as 00 AM when meridian option is true.
### Bug description:
Visit the demo page for timepicker: [https://ng-bootstrap.github.io/#/components/timepicker](https://ng-bootstrap.github.io/#/components/timepicker)
If the meridian option is true, timepicker will display 12:mm pm as 00:mm AM. The underlying hour value itself is actually correct at 12 , as displayed in the json pipe below.

yehudag:
fix #1031

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -521,6 +521,36 @@ describe('ngb-timepicker', () => {
              });
        }));
 
+    it('should render 12 PM/AM as 12:mm and meridian button with proper value', async(() => {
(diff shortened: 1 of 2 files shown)


deeg:
@yehudag, looks like there are some lint errors. Would you mind cleaning those up?

yehudag:
Sorry about that, I fixed the errors

pkozlowski-opensource:
Merged, thnx!

PR #1041, opened 2016-11-15
fix(typeahead): catch click select

### LINKED ISSUE #1040: Typeahead click inside modal causes modal to close
### Bug description:
When having a typeahead inside a modal, when I click a selection the  modal closes. 

## Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/y0QWWHxKfn0x5UXjUN6a?p=preview

Plunker is build in the 2 examples from git.

Angular: 2.1.2 (Also tried angular 2.2.0)

ng-bootstrap: alpha 13 (it was not happening in alpha 11 but now it seems to be doing that as well)

Bootstrap: alpha 5

deeg:
Figured it was fine to just catch the click event which prevents the modal from closing when a typeahead is inside.

Closes #1040

### DIFF
--- a/src/typeahead/typeahead-window.spec.ts
+++ b/src/typeahead/typeahead-window.spec.ts
@@ -138,9 +138,12 @@ describe('ngb-typeahead-window', () => {
 
       expectResults(fixture.nativeElement, ['+bar', 'baz']);
 
-      links[1].triggerEventHandler('click', {});
+      const event = jasmine.createSpyObj('event', ['preventDefault', 'stopPropagation']);
+      links[1].triggerEventHandler('click', event);
       fixture.detectChanges();
       expect(fixture.componentInstance.selected).toBe('baz');
+      expect(event.preventDefault).toHaveBeenCalled();
+      expect(event.stopPropagation).toHaveBeenCalled();
     });
 
     it('should return selected row via getActive()', () => {
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -28,7 +28,7 @@ export interface ResultTemplateContext {
     <template ngFor [ngForOf]="results" let-result let-idx="index">
       <button type="button" class="dropdown-item" [class.active]="idx === activeIdx" 
         (mouseenter)="markActive(idx)" 
-        (click)="select(result)">
+        (click)="select($event, result)">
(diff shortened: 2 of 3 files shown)


wesleycho:
Do we need to assert that stopPropagation and preventDefault has been called each time?

Otherwise, LGTM

deeg:
> Do we need to assert that stopPropagation and preventDefault has been called each time?

Do you mean in the tests? We at least need to pass through an event object which has those functions so no errors are thrown. You cannot just pass in an empty object. I figured adding the assertions were easy and didn't harm anything.

LMK if you want me to remove them.

pkozlowski-opensource:
@deeg I'm not sure I understand the root cause of the issue and this is why I'm a bit cautious about the proposed fix... Why is a backdrop receiving a click in this situation? Is this because it bubbles up and modal window is "within" backdrop? If so maybe we should re-consider my initial fix for dropdowns? Maybe we should just test if a backdrop is clicked instead of using `contains` here: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal-window.ts#L48
This issue with this proposal is that pretty much anything in a modal seems to be effected now and it will include our directives and user directives.... 

I'm traveling right now and can't do much of the investigation but after a quick look this one looks fishy to me. Could you please check if removing `contains` from https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal-window.ts#L48 wouldn't be a better option?

deeg:
@pkozlowski-opensource, I get what you are saying makes sense. I will investigate.

deeg:
@pkozlowski-opensource, it looks like that line is the cause of the issue.

The reason the contains check is returning false is because by the time the check is running, the dom element of the typeahead of datepicker has already been removed from the dom.

It is hard at the current state to tell if you are clicking on the backdrop because the ngb-modal-window element has the same position as the backdrop (fixed, covering entire screen) but a higher z-index. I will try to think of a nice solution, any other input is welcomed!

pkozlowski-opensource:
@deeg could you please have a look at the alternative approach in #1067? If it looks reasonable to you I will add tests.

pkozlowski-opensource:
I'm going to close this one in favor of #1067 as I believe that we should rather fix modal instead of adjusting each and every directive that could be used in a modal.

PR #1046, opened 2016-11-15
fix(datepicker): catch datepicker click

### LINKED ISSUE #1040: Typeahead click inside modal causes modal to close

### LINKED ISSUE #1042: When having a date-picker inside a modal, when I select a date the modal closes.
When having a date-picker inside a modal, when I select a date the modal closes.

Example to reproduce issue:
http://plnkr.co/edit/ATyV4MJ2vvoDgCc9W5Wn?p=preview

deeg:
I took the same approach as to fix #1040 with #1041 

Closes #1042

### DIFF
--- a/src/datepicker/datepicker-month-view.spec.ts
+++ b/src/datepicker/datepicker-month-view.spec.ts
@@ -1,5 +1,6 @@
 import {TestBed, ComponentFixture} from '@angular/core/testing';
 import {createGenericTestComponent, isBrowser} from '../test/common';
+import {By} from '@angular/platform-browser';
 
 import {Component} from '@angular/core';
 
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker-month-view.spec.ts
```diff
@@ -1,5 +1,6 @@
 import {TestBed, ComponentFixture} from '@angular/core/testing';
 import {createGenericTestComponent, isBrowser} from '../test/common';
+import {By} from '@angular/platform-browser';
```
deeg: Is there a convention on which import block this one goes with? Or are the line breaks here arbitrary?

pkozlowski-opensource:
See https://github.com/ng-bootstrap/ng-bootstrap/pull/1041#issuecomment-260758547 - I'm concerned that this might not be the "right" fix...

pkozlowski-opensource:
@deeg could you please have a look at the alternative approach in #1067? If it looks reasonable to you I will add tests.

pkozlowski-opensource:
I'm going to close this one in favor of #1067 as I believe that we should rather fix modal instead of adjusting each and every directive that could be used in a modal.

PR #1049, opened 2016-11-16
feat(pagination): custom boundary/directional text - duplicate of PR #1055

### LINKED ISSUE #899: Custom pagination boundary and directional link text

``` html
</a>
```

- first-text
- last-text
- next-text
- previous-text

angular: 2.0.1
ng-bootstrap: 1.0.0-alpha.7
bootstrap: 4.0.0-alpha.4

alex7olaru:
Closes #899 

Added supported for setting custom text for the boundary and directional links. If the custom text is configured for either of these 4 links, that text will be applied to both the regular link(s) and the screen reader link(s). If these new properties are not set, the old text values are displayed. This also provides the benefit of setting the text in other languages for situations where the default screen reader English text might not be desired.

A new section is added for the pagination demo page that shows an example of how to use these 4 new properties.


deeg:
It looks like your branch has conflicts. Would you mind fixing those, and squashing into one commit and someone will take a look and review?

wesleycho:
This also needs merge commits removed and squashed into one.

alex7olaru:
Thank you for the guidance. I addressed the issues mentioned above with a new PR: #1055. I wanted to have the PR for these changes linked to an issue branch in my forked repo instead of the master branch. I recommend closing this PR and tracking the proposed changes going forward with the new PR. I apologize for the inconvenience.

PR #1055, opened 2016-11-17
feat(pagination): custom boundary/directional text 

### LINKED ISSUE #899: Custom pagination boundary and directional link text

alex7olaru:
Closes #899

Added supported for setting custom text for the boundary and directional links. If the custom text is configured for either of these 4 links, that text will be applied to both the regular link(s) and the screen reader link(s). If these new properties are not set, the old text values are displayed. This also provides the benefit of setting the text in other languages for situations where the default screen reader English text might not be desired.

A new section is added for the pagination demo page that shows an example of how to use these 4 new properties.

This PR replaces PR #1049:
- squashed commits into one
- removed the usage of innerHtml with interpolated strings for binding, if this would work

Please note that I could not get the single/double arrows to bind correctly when their values were expressed as encoded html equivalents (the values were not being decoded at display time). As such, I included in the code the actual unencoded characters for double angle quotation marks.

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -51,8 +51,11 @@ function expectSameValues(pagination: NgbPagination, config: NgbPaginationConfig
   expect(pagination.boundaryLinks).toBe(config.boundaryLinks);
   expect(pagination.directionLinks).toBe(config.directionLinks);
   expect(pagination.ellipses).toBe(config.ellipses);
+  expect(pagination.lastText).toBe(config.lastText);
   expect(pagination.maxSize).toBe(config.maxSize);
+  expect(pagination.nextText).toBe(config.nextText);
   expect(pagination.pageSize).toBe(config.pageSize);
+  expect(pagination.previousText).toBe(config.previousText);
   expect(pagination.rotate).toBe(config.rotate);
   expect(pagination.size).toBe(config.size);
 }
@@ -519,6 +522,22 @@ describe('ngb-pagination', () => {
       expectPages(fixture.nativeElement, ['« Previous', '1', '+2', '-» Next']);
     });
 
+    it('should render and respond to direction links changes', () => {
+      const html = `<ngb-pagination [collectionSize]="collectionSize" [page]="1" [firstText]="firstText" 
+      [previousText]="previousText" [nextText]="nextText" [lastText]="lastText" 
+      [boundaryLinks]="boundaryLinks"></ngb-pagination>`;
+      const fixture = createTestComponent(html);
+
+      fixture.componentInstance.boundaryLinks = true;
+      fixture.componentInstance.collectionSize = 20;
+      fixture.componentInstance.firstText = 'TF';
+      fixture.componentInstance.previousText = 'TP';
+      fixture.componentInstance.nextText = 'TN';
+      fixture.componentInstance.lastText = 'TL';
+      fixture.detectChanges();
+      expectPages(fixture.nativeElement, ['-TF TF', '-TP TP', '+1', '2', 'TN TN', 'TL TL']);
+    });
+
     it('should not emit "pageChange" for incorrect input values', fakeAsync(() => {
          const html = `<ngb-pagination [collectionSize]="collectionSize" [pageSize]="pageSize" [maxSize]="maxSize" 
         (pageChange)="onPageChange($event)"></ngb-pagination>`;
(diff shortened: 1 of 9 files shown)


src/pagination/pagination.ts
```diff
@@ -33,22 +33,32 @@ import {NgbPaginationConfig} from './pagination-config';
...
+            <span aria-hidden="true">{{lastText}}</span>
+            <span class="sr-only">{{_lastTextSr}}</span>
           </a>                
         </li>        
       </ul>
     </nav>
   `
 })
 export class NgbPagination implements OnChanges {
+  private _firstText;
```
wesleycho: I don't think you should use private - these are all to be accessed by the template, it doesn't seem like a private property.

src/pagination/pagination.ts
```diff
@@ -68,6 +78,19 @@ export class NgbPagination implements OnChanges {
   @Input() ellipses: boolean;
 
   /**
+   * Text to display for the first page button.
+   */
+  @Input()
+  set firstText(value: string) {
+    this._firstText = value;
+    if (value !== this._defaultConfig.firstText) {
+      this._firstTextSr = value;
```
wesleycho: Why the difference between the screenreader text and the regular text? Why not just use the regular text?

src/pagination/pagination.spec.ts
```diff
@@ -519,6 +522,22 @@ describe('ngb-pagination', () => {
       expectPages(fixture.nativeElement, ['« Previous', '1', '+2', '-» Next']);
     });
 
+    it('should render and respond to direction links changes', () => {
+      const html = '<ngb-pagination [collectionSize]="collectionSize" [page]="1" [firstText]="firstText"' +
+          ' [previousText]="previousText" [nextText]="nextText" [lastText]="lastText"' +
+          ' [boundaryLinks]="boundaryLinks"></ngb-pagination>';
```
wesleycho: Should use a template string here, makes it easier to read.

src/pagination/pagination.ts
alex7olaru: In its default configuration, the pagination shows the text (First, Previous, etc.) for screen-readers and the arrows for the other types of displays. This info is not transparent though in the API docs - maybe a potential for enhancement here. 

This difference between sr and non-sr displays was there before and it feels to me like it makes sense to continue to keep it.

src/pagination/pagination.ts
alex7olaru: I think the non-SR (screen reader) properties need to be private because if they become public (by removing the underscore and making the properties public) we would get into an infinite loop in their corresponding setter methods. The template does not actually use the private property reference, it uses the public one: e.g. firstText.

It is true though that the template uses the private SR properties (e.g. _firstTextSr). I thought about making the SR properties public as well but then it felt to me like they would need the Input decorator and that they would need to be included in the public API and the docs. Adding 4 more properties to the UI felt too much to me at this point. Without the 4 additional public SR properties it is true that we won't have the flexibility of allowing the SR and non-SR values to be set to different values once the default configuration is overridden.

Maybe a good compromise for now is to change the description for the non-SR properties (and hence the API docs in the demo) to say something like this?

Text to display for the first page button. _This value will be used for both screen readers and non-screen reader displays. Default value: «« (for non-screen reader displays); First for screen-reader displays._

If users will end up needing in the future the ability to set the SR text to values different than the non-SR text we could expose the SR properties in the public properties when that need arises. 

On a slightly different note, I think I saw an API docs issue talking about the extraction of default values to include in the API docs. I hope that hard coding the default values in the property's comment won't interfere with the resolution of that issue.

src/pagination/pagination.spec.ts
alex7olaru: It does not seem to me like you are talking here about moving the template to a template file. I am not exactly sure what you mean by a template string different than the one already included. Could you please point me to an example of a template string that I could use to model my change after?

alex7olaru: Got it. Thank you - I see that being used too in other templates in the spec file. I will do a commit tomorrow for the template string change and the change for the more detailed description for the public properties. 

Given the additional context I provided would you still like me to proceed with applying the other 2 changes you have suggested (getting rid of the private properties and combining the non-SR and SR properties into one set) ?

alex7olaru: I included in the original commit only the changes for the template string and the more detailed description for the public properties. I can make the rest of the requested changes once you would have a chance to provide additional feedback.

pkozlowski-opensource:
Thnx for the suggestion @alex7olaru but we are going to implement an alternative approach where one be able to change the entire markup for a given cell and not only text. Even if your PR is not merged we are still grateful for it since it allowed us to consider and explore different options. Thnx!

alex7olaru:
Great to hear about the plan. I am glad that the PR was at least useful in this way.

PR #1063, opened 2016-11-18
fix(datepicker): add default month to prevent infinite loop

### LINKED ISSUE #1062: NgbDatepicker: infinite loop when using Date object with datepicker toggle
### Bug description:

Browser crash/full cpu/infinite loop when setting model as new Date() and toggling the datepicker popup.

CAREFUL : See plunkr and click the button for the toggle, will have to restart script/tab/browser.

Checked on Firefox and Chrome.

http://plnkr.co/edit/z4FuG5CXzalen5BFV9Cc?p=preview

Angular: 2.2.1

ng-bootstrap: 1.0.0-alpha.13

Bootstrap: v4-alpha

CrustyJew:
Without a default month, this causes an infinite loop in datepicker.ts `navigateTo` function looking for the first day, of the last week, of the previous month to display. `_calendar.getNext` returns `NaN` values causing an infinite loop.

Solution was to set a default month of 1 following the same pattern as the default day of 1.

Closes #1062

### DIFF
--- a/src/datepicker/ngb-date.spec.ts
+++ b/src/datepicker/ngb-date.spec.ts
@@ -60,4 +60,13 @@ describe('ngb-date', () => {
       expect(date.after(new NgbDate(2017, 8, 18))).toBeFalsy();
     });
   });
+
+  describe('from', () => {
+    it('should return null for invalid object', () => { expect(NgbDate.from(null)).toBeNull(); });
+    it('should return null for dates that are not convertable to javascript Date objects', () => {
+      expect(NgbDate.from({year: 275760, month: 10, day: 1})).toBeNull();
+      expect(NgbDate.from({year: 275759, month: 1, day: 11}).equals(new NgbDate(275759, 1, 11))).toBeTruthy();
+      expect(NgbDate.from({year: -275760, day: 1})).toBeNull();
+    });
+  });
 });
--- a/src/datepicker/ngb-date.ts
+++ b/src/datepicker/ngb-date.ts
@@ -1,6 +1,15 @@
 export class NgbDate {
-  static from(date: {year: number, month: number, day?: number}) {
-    return date ? new NgbDate(date.year, date.month ? date.month : 1, date.day ? date.day : 1) : null;
+  static from(date: {year: number, month?: number, day?: number}) {
+    let nDate = date ? new NgbDate(date.year, date.month ? date.month : 1, date.day ? date.day : 1) : null;
+    if (nDate) {
+      // add 1 to year to prevent any date in last valid javascript Date being picked and causing issues around
+      // September
+      let jDate = new Date(nDate.year + 1, nDate.month - 1, nDate.day);
+      if (isNaN(jDate.getTime())) {
+        nDate = null;
+      }
+    }
+    return nDate;
   }
 
   constructor(public year: number, public month: number, public day: number) {}
--- a/src/datepicker/ngb-date.ts
+++ b/src/datepicker/ngb-date.ts
@@ -1,6 +1,6 @@
 export class NgbDate {
   static from(date: {year: number, month: number, day?: number}) {
-    return date ? new NgbDate(date.year, date.month, date.day ? date.day : 1) : null;
+    return date ? new NgbDate(date.year, date.month ? date.month : 1, date.day ? date.day : 1) : null;
   }
 
   constructor(public year: number, public month: number, public day: number) {}


pkozlowski-opensource:
I don't see how this change is related to the original issue. I'm not saying that this is _not_ a good fix - just saying that without a test and a bug analysis we are tabbing in the dark. From my quick look it seems that the service used to generate data model (https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-service.ts) is at fault and a fix + tests likely belong there. 

It would be -1 from me when it comes to merging it in the current form as it seems for me that we are dealing with symptoms rather than the root issue here.

CrustyJew:
@pkozlowski-opensource The service would be another place to fix it, but the service comes in expecting to have a valid date object. When an invalid date gets passed in (`{year:NaN, month: NaN, day: NaN}` caused by new NgbDate() with invalid / null data) , the service gets stuck in an infinite loop. The problem used to exist for days as well until someone added in the default day value of 1 hence why I fixed months in the same way.

It may be a better approach to define what the datepicker should do if given an invalid date in the input element and expanding the datepicker. Options could include erasing the bad value in the input element and default to minimum date / today . Or it could default the values like this PR and hope for the best. Or it could prevent you expanding the date picker at all with invalid data there.

I didn't include a test case as the error is an infinite loop which I personally don't know if you can make a test case for. A test case that could be made is if the `NgbDate` is an error object / NaN though.

pkozlowski-opensource:
> It may be a better approach to define what the datepicker should do if given an invalid date in the input element and expanding the datepicker. Options could include erasing the bad value in the input element and default to minimum date / today . Or it could default the values like this PR and hope for the best. Or it could prevent you expanding the date picker at all with invalid data there.

I would argue is that in this case we should treat an invalid date as `null` and default to today. It should be rather straightforward to add a test for this case. Would you like to take a stab at it?

CrustyJew:
I can take a stab at it sure.  I wasn't sure if that's a big enough design
decision that it shouldn't be done in a pull request which is why i went
minimal change to fix the issue.

I'll see what i can do though

CrustyJew:
@pkozlowski-opensource so the main issue was invalid javascript date objects being passed into the model, specifically from the input box. I added checks for valid dates in the `NgbDate.from` function to check for this and return null if it's invalid. I also added test cases for this.

This doesn't prevent code passing in an invalid date to the datepicker service itself, which may need to be improved on as well. But this does prevent it from passing in from the input.

src/datepicker/ngb-date.spec.ts
```diff
@@ -60,4 +60,13 @@ describe('ngb-date', () => {
       expect(date.after(new NgbDate(2017, 8, 18))).toBeFalsy();
     });
   });
+
+  describe('from', () => {
+    it('should return null for invalid object', () => { expect(NgbDate.from(null)).toBeNull(); });
+    it('should return null for dates that are not convertable to javascript Date objects', () => {
+      expect(NgbDate.from({year: 275760, month: 10, day: 1})).toBeNull();
```
pkozlowski-opensource: Why this would be an invalid year?

src/datepicker/ngb-date.ts
```diff
@@ -1,6 +1,15 @@
 export class NgbDate {
-  static from(date: {year: number, month: number, day?: number}) {
-    return date ? new NgbDate(date.year, date.month ? date.month : 1, date.day ? date.day : 1) : null;
+  static from(date: {year: number, month?: number, day?: number}) {
+    let nDate = date ? new NgbDate(date.year, date.month ? date.month : 1, date.day ? date.day : 1) : null;
+    if (nDate) {
+      // add 1 to year to prevent any date in last valid javascript Date being picked and causing issues around
+      // September
+      let jDate = new Date(nDate.year + 1, nDate.month - 1, nDate.day);
```
pkozlowski-opensource: Do we really care if JS `Date` object is happy here or not? I don't think we are using `Date` objects anywhere else in a datepicker and different calendars might have a different idea about date object validity.

CrustyJew: Due to how NgbCalendar works for the Gregorian Calendar implementation, yeah we have to. [It converts to a JS Date](https://github.com/ng-bootstrap/ng-bootstrap/blob/3af832995eaf2c85b516b465b8e9c03092b787fe/src/datepicker/ngb-calendar.ts#L37) in `getNext` , `getWeekday` and `getWeekNumber`. The invalid JS Date object is what was causing the infinite loop in the DatepickerService

src/datepicker/ngb-date.spec.ts
CrustyJew: JS Dates max out in September of 275760

src/datepicker/ngb-date.ts
pkozlowski-opensource: Yeh, but this is specific to Gregorian Calendar only while `NgbDate` is calendar independent...

CrustyJew: I'd argue there aren't any real world use cases that would be redefining dates outside of what can be represented in javascript, and the risk of not implementing it at the `NgbDate` level would be that any other calendar implementation would fall into the same infinite loop trap. If at some point the service was hardened up and wouldn't cause an infinite loop, than it could be safely moved inside the specific calendar.

That being said, I can see where you are coming from, but in that case I think the only solution would be to extend the Calendar interface to include a method to validate a given NgbDate and call that before attaching the NgbDate to the model. Otherwise you could bind an invalid date by typing one in, but not pick it in the popup area, and,  in fact, cause errors generating that calendar view.

pkozlowski-opensource:
I've re-discussed this PR with @maxokorokov and we decided that the right fix is to add date check on the calendar level. We don't want to have any calendar-specific logic in the `NgbDate` to keep our options open.

@maxokorokov will submit a PR with a fix. 

Thnx for the contribution - even if your PR is not merged we really _do_ appreciate all PRs and yours triggered discussion that pointed us in the right direction. Thnx again.

CrustyJew:
No problem, also I do have updates to my branch that moves the logic to more appropriate places including a validator addition to `datepicker-input` . It doesn't directly solve the infinite loop issue, but it stops the infinite loop from happening via the input box that I was running in to. 

Should I open a new pull request for that? It's more of a feature than a direct fix

pkozlowski-opensource:
> Should I open a new pull request for that? It's more of a feature than a direct fix

You could check #1092 which aims at solving the original issue.

PR #1064, opened 2016-11-19
feat(popover): Support 'outsideClick' close trigger

### LINKED ISSUE #933: Close popover on click outside
Hi, feature request, quite simple case - possibility to close popover when user clicked outside one.

spongessuck:

Add support for a special `outsideClick` close trigger to close popovers when a click occurs outside of it. Closes #933.

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -40,8 +40,13 @@ export class NgbPopoverWindow {
 /**
  * A lightweight, extensible directive for fancy popover creation.
  */
-@Directive({selector: '[ngbPopover]', exportAs: 'ngbPopover'})
-export class NgbPopover implements OnInit, OnDestroy {
+@Directive({
+  selector: '[ngbPopover]',
+  exportAs: 'ngbPopover',
+  host: {'(document:click)': '_handleClickOutsidePopover($event)'}
+})
+export class NgbPopover implements OnInit,
+    OnDestroy {
   /**
    * Content to be displayed as popover.
    */
@@ -77,6 +82,12 @@ export class NgbPopover implements OnInit, OnDestroy {
   private _unregisterListenersFn;
   private _zoneSubscription: any;
 
+  /**
+   * Set to `true` after the popover opens; used to cancel the document click
+   * handler so it doesn't immediately close the popover.
+   */
+  private _triggered: boolean;
+
   constructor(
       private _elementRef: ElementRef, private _renderer: Renderer, injector: Injector,
       componentFactoryResolver: ComponentFactoryResolver, viewContainerRef: ViewContainerRef, config: NgbPopoverConfig,
@@ -96,6 +107,36 @@ export class NgbPopover implements OnInit, OnDestroy {
     });
   }
 
+  private _handleClickOutsidePopover(e) {
+    if (this._triggered) {
+      return this._triggered = false;
+    }
+
+    if (!this._windowRef) {
+      return;
+    }
+
+    /* Make sure click didn't happen inside popover */
+    if (e.target !== this._windowRef.location.nativeElement &&
+        !this._windowRef.location.nativeElement.contains(e.target)) {
+      /**
+       * Trigger event called 'outsideClick' so that the `listenToTriggers` utility function automatically handles the
+       * close action when the close trigger is set to 'outsideClick'.
+       */
+      let ocEvent = document.createEvent('Event');
+      if (ocEvent.initEvent) {
+        /* IE-compatible event initialization */
+        ocEvent.initEvent('outsideClick', true, true);
+      } else {
+        /**
+         * IE-compatible solution is deprecated. Use Event constructor instead if `initEvent` becomes unavailable.
+         */
+        ocEvent = new Event('outsideClick');
+      }
+      this._elementRef.nativeElement.dispatchEvent(ocEvent);
+    }
+  }
+
   /**
    * Opens an element’s popover. This is considered a “manual” triggering of the popover.
    */
@@ -108,6 +149,8 @@ export class NgbPopover implements OnInit, OnDestroy {
       if (this.container === 'body') {
         window.document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
       }
+
+      this._triggered = true;
 
       // we need to manually invoke change detection since events registered via
       // Renderer::listen() are not picked up by change detection with the OnPush strategy
@@ -124,6 +167,7 @@ export class NgbPopover implements OnInit, OnDestroy {
       this._popupService.close();
       this._windowRef = null;
       this.hidden.emit();
+      this._triggered = false;
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/popover/popover.ts
```diff
@@ -64,26 +65,41 @@ export class NgbPopover implements OnInit, OnDestroy {
...
    */
   @Output() shown = new EventEmitter();
   /**
    * Emits an event when the popover is hidden
    */
   @Output() hidden = new EventEmitter();
+  /**
+   * Emits an event when a click occurs outside of the popover element.
+   */
+  @Output() clickOutside = new EventEmitter();
```
spongessuck: The popover wouldn't necessarily be closed when a user clicks outside of it; that's just how I have it set up in the demo, and was the idea behind creating it.

I've actually been thinking that (clickOutside) should be a more generic directive not specific to the popover. I'm not sure if it would be useful for other NGB components, though.

I'm OK with getting rid of it. I just thought it would be a more flexible solution than [closeOnClickOutside].

spongessuck: @pkozlowski-opensource What do you think? Should we just keep [closeOnClickOutside]?

spongessuck:
@pkozlowski-opensource Let me know what you want me to do with this.

colinmorelli:
Just to voice another opinion here:

I'm +1 to @pkozlowski-opensource comments about removing the clickOutside event. For the reasons mentioned, as well as it feels weird for a component to expose an event that is triggered specifically when something happens _outside_ of that component. That said, I can understand (and also would benefit from) the ability to auto close on outside click.

The other thing I'm not convinced about is a separate property for closeOnOutsideClick. Isn't this really just another `trigger`? Granted for closing vs opening. Perhaps the more "robust" solution is to have separate `openTrigger` and `closeTrigger` properties (or, do what ui-bootstrap does and just have an `outsideClick` option as a `trigger`). Just my 2 cents.

spongessuck:
@colinmorelli I agree with your point about `closeOnOutsideClick`; I always thought it was a bit clumsy, but I couldn't think of a more obvious API.

The `clickOutside` trigger from ui-bootstrap seems like it does what this PR is supposed to do. The name doesn't really inform what it does though- it opens with a click, but closes with a click outside?

Despite the confusion, it does seem like a more elegant API.

colinmorelli:
Agreed @spongessuck - that's why I was suggesting potentially having separate `openTrigger` and `closeTrigger` inputs, or potentially `trigger="{ open: 'click', close: ['click', 'outsideClick'] }"`, though with default well-known string variants for `trigger="click"`. This makes a lot more sense to me, since there's actually almost no practical situation in which the same events that open a popover should close it. `click` should be closed, most logically, by `click` or "outsideclick"; `mouseenter` by `mouseleave`; etc. That said, it's probably outside the scope of this PR.

spongessuck:
I'll wait for input from @pkozlowski-opensource before I submit another PR so I don't waste time implementing something that won't get merged.

spongessuck:
It was easy to add support for an `outsideClick` close trigger. If this method is preferable to the others, I'll get rid of the attribute and event features.

spongessuck:
I'm confident this is the cleanest solution. Should be ready to go.

bolota:
@spongessuck 

Hi and many thanks for this trigger, how much time do I need to wait for this merge?

spongessuck:
Not sure, @bolota, it's up to the maintainers.

Chalkin:
Would be happy to see this merge happening soon. We urgently need this asap.

@bolota Some thoughts on this?

src/popover/popover.ts
```diff
@@ -96,6 +103,33 @@ export class NgbPopover implements OnInit, OnDestroy {
     });
   }
 
+  @HostListener('document:click', ['$event'])
+  /**
+   * Determines if a click occurs outside of the popover.
+   */
+  private _docClickListener(e) {
```
pkozlowski-opensource: We should try to name this method in a descriptive way instead of adding a comment on top of it.

src/popover/popover.ts
```diff
@@ -3,6 +3,7 @@ import {
   Directive,
   Input,
   Output,
+  HostListener,
```
pkozlowski-opensource: Could you please use the `host` property of `@Directive()` instead?

pkozlowski-opensource:
@spongessuck I had another pass on this PR and I must say that I'm rather confused about how it should / was imagined to work. I've left comments in the PR. In any case it is currently not in the state where we could consider it for merging.

src/popover/popover.ts
spongessuck: Sure.

src/popover/popover.ts
```diff
@@ -96,6 +103,33 @@ export class NgbPopover implements OnInit, OnDestroy {
...
+        !this._windowRef.location.nativeElement.contains(e.target)) {
+      let ocEvent = document.createEvent('Event');
+      if (ocEvent.initEvent) {
+        /* For IE */
+        ocEvent.initEvent('outsideClick', true, true);
+      } else {
+        /* Avoid deprecated API */
+        ocEvent = new Event('outsideClick');
+      }
+      this._elementRef.nativeElement.dispatchEvent(ocEvent);
```
spongessuck: It's dispatching an event with the name 'clickOutside' so that the listeners already set up to check for events passed to the `triggers` attribute handle the close action.

If I were to call `close()` on the popover from here, I'd have to use the trigger parsing utility to check if the close trigger name is equal to 'clickOutside.' Since the listener logic is already doing this, I figured dispatching an event would be the more logical solution.

src/popover/popover.ts
spongessuck: Sure, how about `_handleClickOutsidePopover`?

src/popover/popover.ts
spongessuck: Isn't HostListener best practice, though? It's what they use in the Angular docs: https://angular.io/docs/ts/latest/guide/attribute-directives.html

spongessuck:
I've made the suggested changes. Please see my comment for the rationale behind dispatching an event vs. closing the popover in the click handler.

The best way to see how I imagined it to work is to check the demo I added.

spongessuck:
It should be noted that the implementation of this feature is completely different than how I originally imagined it.

pkozlowski-opensource:
@spongessuck I'm really sorry but after thinking about this feature some more I believe that this PR is not implementing API that we want to have in the end. I've provided details in https://github.com/ng-bootstrap/ng-bootstrap/issues/933#issuecomment-269779770 and subsequent comments. Would be good to have your opinion / thoughts.

Moreover I  think that we need a single, performant mechanism that would cover all directives that open popups today (datepicker, dropdown, popover, tooltip, typeahead...) - otherwise we are going to name inputs differently, have different capabilities for each directive etc. As such I'm going to close this PR and work on a more generic solution. Still I would love to hear your thoughts on https://github.com/ng-bootstrap/ng-bootstrap/issues/933#issuecomment-269779770

Even if your PR is not merged it triggered deeper thoughts on the outside clicks handling and as such was very valuable. Thank you!

PR #1065, opened 2016-11-19
chore: mention adding css stylesheet in README

ajzawawi:

Update README to mention adding stylesheet in the index's head section. This is more user friendly for those starting out with ng-bootstrap and angular in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ import {NgbModule} from '@ng-bootstrap/ng-bootstrap';
 
 @NgModule({
   declarations: [AppComponent, ...],
-  imports: [NgbModule.forRoot(), ...],  
+  imports: [NgbModule.forRoot(), ...],
   bootstrap: [AppComponent]
 })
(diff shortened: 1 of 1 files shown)


ajzawawi:
@Foxandxss That's a valid point, I agree. How about a note at least to mention the stylesheet? I help junior devs at work and not knowing that a stylesheet needs to be included somewhere is one of the most common issues I get for ng-bootstrap. They're usually confused why the module is not working.

Foxandxss:
I agree with that, but I don't see how is that our duty.

Perhaps a "Don't forget to load your bootstrap CSS" could be fine.

Let's see what the other think.

wesleycho:
I don't think this is really our responsibility either - users load Bootstrap in various ways. For example, how we load it at my workplace is through a common repository that builds stylesheets for import with modifications to color variables and such to keep the company theme consistent across projects.

I agree with @Foxandxss here.

PR #1092, opened 2016-11-28
fix(datepicker): protect against invalid min, max and start dates

### LINKED ISSUE #1062: NgbDatepicker: infinite loop when using Date object with datepicker toggle

maxokorokov:
This will protect against invalid inputs for `minDate`, `maxDate` and `startDate` as well as `navigateTo(date)` and set them to default values

Fixes #1062

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -1,6 +1,6 @@
 import {TestBed, ComponentFixture, fakeAsync, tick} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
-import {createGenericTestComponent} from '../test/common';
+import {createGenericTestComponent, isBrowser} from '../test/common';
 
 import {Component} from '@angular/core';
 import {FormsModule} from '@angular/forms';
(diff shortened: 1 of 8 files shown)


CrustyJew:
I still get an infinite loop issue in certain cases.

Steps to reproduce:

1. running this PR locally with `gulp demo-server` 

2. navigating to "http://localhost:9090/#/components/datepicker".

3. scroll to "Datepicker in a popup"

4. enter `999999-01-01` in the input. 

5. click calendar icon

maxokorokov:
@CrustyJew ok, probably you're right. Will accept date structures that can safely be converted to JS date object (as that's what is used internally by the Gregorian calendar implementation)

Amended. Plus added some checks for the `datepicker-input` text values

maxokorokov:
Some date tests are failing in Safari and I can't easily reproduce them locally for now. Apparently Safari 9 thinks `3000000-1-1` is a valid date and works fine with it while Safari 10 and any other browser doesn't anymore...

maxokorokov:
Ok, issue resolved. Apparently in Safari 9:
- `let d = new Date(300000, 1, 1)` → returns an invalid date
- but `d.setFullYear(300000, 1,  1)` → returns a valid date

PR #1097, opened 2016-11-29
feat(rating): enable form integration

### LINKED ISSUE #1087: Can't use reactive form with rating
For some reason I can't make the ratings work with reactive form (and the `formControlName`).

I've tried [the following](http://plnkr.co/edit/Pg6HcQyRAdARy5xfrnT1?p=preview) without success. Is it a bug or should we use reactive forms differently with this component (in this case we should add to the documentation)? 

HTML:
```
<ngb-rating [(rate)]="currentRate" formControlName='answer'>
  <template let-fill="fill">
    <span class="star" [class.filled]="fill === 100">&#9733;</span>
  </template>
</ngb-rating>
```

maxokorokov:
Enables form integration for rating:

- both template driven and reactive forms tests
- `rate` event emitter becomes asynchronous (same story as pagination)
- ability to use both `[(rate)]` and `[(ngModel)]`
- form integration demo with reactive

Fixes #1087

### DIFF
--- a/demo/src/app/components/rating/demos/form/rating-form.html
+++ b/demo/src/app/components/rating/demos/form/rating-form.html
@@ -1,0 +1,16 @@
+<p>NgModel and reactive forms can be used without the 'rate' binding</p>
+
+<div class="form-group" [class.has-success]="ctrl.valid" [class.has-danger]="ctrl.invalid">
+  <ngb-rating [formControl]="ctrl"></ngb-rating>
+  <div class="form-control-feedback">
+    <div *ngIf="ctrl.valid">Thanks!</div>
+    <div *ngIf="ctrl.errors">Please rate us</div>
+  </div>
+</div>
+
+<hr>
+<pre>Model: <b>{{ ctrl.value }}</b></pre>
+<button class="btn btn-sm btn-outline-{{ ctrl.disabled ? 'danger' : 'success'}}" (click)="toggle()">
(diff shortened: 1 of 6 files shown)


demo/src/app/components/rating/demos/form/rating-form.html
```diff
@@ -0,0 +1,16 @@
+<p>You can use both ngModel and reactive forms. You don't have to use 'rate' binding in this case</p>
```
wesleycho: Probably would read better as
```html
<p>ngModel and reactive forms can be used without using the 'rate' binding</p>
```

maxokorokov: Thanks. Amended

maxokorokov:
Found an issue with readonly rating, will have to add some more tests...

maxokorokov:
It's OK now, but will add proper `mouseenter` and `mouseleave` tests in a separate PR, as we're missing them currently

natcohen:
Any plan to merge this for the next release?

pkozlowski-opensource:
LGTM generally speaking. Could you just update the commit message to have reference to the issue this commit is fixing? Also this one has a breaking change, no? (sync -> async delivery of events?)

maxokorokov:
Resolved conflicts and updated commit message with breaking changes. Can merge when green then.

PR #1114, opened 2016-12-05
fix(pagination): prevent focus of disabled selectors

### LINKED ISSUE #1108: Pagination: prevent focusing disabled selectors
When a page selector is disabled, i.e current page is first or last, user should not be able to focus the disabled element. It should have attribute `tabindex` set to -1.

[Plunker](http://plnkr.co/edit/ZEM3hswj9YhMvFKp69yr?p=preview)
[Bootstrap](http://v4-alpha.getbootstrap.com/components/pagination/#disabled-and-active-states)

deeg:
Fixes #1108

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -27,10 +27,16 @@ function expectPages(nativeEl: HTMLElement, pagesDef: string[]): void {
       expect(pages[i]).not.toHaveCssClass('active');
       expect(pages[i]).toHaveCssClass('disabled');
       expect(normalizeText(pages[i].textContent)).toEqual(pageDef.substr(1));
+      if (normalizeText(pages[i].textContent) !== '...') {
+        expect(pages[i].querySelector('a').getAttribute('tabindex')).toEqual('-1');
+      }
     } else {
       expect(pages[i]).not.toHaveCssClass('active');
       expect(pages[i]).not.toHaveCssClass('disabled');
       expect(normalizeText(pages[i].textContent)).toEqual(pageDef);
+      if (normalizeText(pages[i].textContent) !== '...') {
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.spec.ts
```diff
@@ -27,10 +27,16 @@ function expectPages(nativeEl: HTMLElement, pagesDef: string[]): void {
       expect(pages[i]).not.toHaveCssClass('active');
       expect(pages[i]).toHaveCssClass('disabled');
       expect(normalizeText(pages[i].textContent)).toEqual(pageDef.substr(1));
+      if (normalizeText(pages[i].textContent) !== '...') {
```
deeg: I did this to put the check in for all checkPages() calls. If you want me to break this out into a separate test I can.

src/pagination/pagination.spec.ts
```diff
@@ -27,10 +27,16 @@ function expectPages(nativeEl: HTMLElement, pagesDef: string[]): void {
...
+        expect(pages[i].querySelector('a').getAttribute('tabindex')).toEqual('-1');
+      }
     } else {
       expect(pages[i]).not.toHaveCssClass('active');
       expect(pages[i]).not.toHaveCssClass('disabled');
       expect(normalizeText(pages[i].textContent)).toEqual(pageDef);
+      if (normalizeText(pages[i].textContent) !== '...') {
+        expect(pages[i].querySelector('a').getAttribute('tabindex')).not.toEqual('-1');
```
maxokorokov: maybe `expect(pages[i].querySelector('a').hasAttribute('tabindex')).toBeFalsy()`? 
I think we're expecting that attribute should not be there at all

deeg: Sounds good, will do.

deeg:
Not sure why travis failed. Maybe it needs to be restarted? I think a saucelabs issue.

PR #1128, opened 2016-12-09
feat(datepicker): Add option to keep invalid input

ajafff:

Add input property keepInvalidInput. When set to true, the input field is
not cleared, when input cannot be parsed by the Parser.

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -83,6 +83,17 @@ describe('NgbInputDatepicker', () => {
       expect(fixture.componentInstance.date).toBeNull();
     });
 
+    it('should keep invalid input when a user enters invalid date with keepInvalidInput', fakeAsync(() => {
+      const fixture = createTestCmpt(`<input ngbDatepicker [(ngModel)]="date" [keepInvalidInput]="true">`);
+      const inputDebugEl = fixture.debugElement.query(By.css('input'));
+      inputDebugEl.nativeElement.value = 'aaa';
+
+      inputDebugEl.triggerEventHandler('change', {target: {value: 'aaa'}});
+      tick();
+      fixture.detectChanges();
+      expect(inputDebugEl.nativeElement.value).toEqual('aaa');
+    }));
+
     it('should propagate disabled state', fakeAsync(() => {
          const fixture = createTestCmpt(`
         <input ngbDatepicker [(ngModel)]="date" #d="ngbDatepicker" [disabled]="isDisabled">
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -84,15 +84,15 @@ describe('NgbInputDatepicker', () => {
     });
 
     it('should keep invalid input when a user enters invalid date with keepInvalidInput', fakeAsync(() => {
-      const fixture = createTestCmpt(`<input ngbDatepicker [(ngModel)]="date" [keepInvalidInput]="true">`);
-      const inputDebugEl = fixture.debugElement.query(By.css('input'));
-      inputDebugEl.nativeElement.value = 'aaa';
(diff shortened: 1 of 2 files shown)


ajafff:
Seems like I need help to investigate, why 0 tests are executed

deeg:
This is happening on more than just your PR:

```src/rating/rating.spec.ts(31,58): error TS2365: Operator '!==' cannot be applied to types '"(*)"' and '"( )"'.```

Not sure exactly what is going on yet.

maxokorokov:
@deeg that was fixed by #1131. However there were saucelabs issues on top of that... Now it should work. Don't know how to relaunch travis except for pushing to PR though.

pkozlowski-opensource:
@ajafff what us the use-case / justification for this change. Without knowing more context my first knee-jerk reaction is to not merge it. Happy to change my mind if you can defend a real-life use-case.

ajafff:
@pkozlowski-opensource it's all about user experience. The intention of this option is to give the user a chance to correct his faulty input. If you hit the wrong key and type '2016-12-32', your input would be completely gone.
With `keepInvalidInput` the app can show a nice error message to notifiy the user about the error.

There is one other thing to consider: Should this directive implement the `Validator` interface and provide itself as `NG_VALIDATORS` to report the invalid input (maybe configurable through another input property?) or should that be left to the "user" of the directive?


About the failed CI: I try to get rebasing and fixing the test done in the next few days.

pkozlowski-opensource:
@ajafff if I understand you correctly it is more about proper validation messages rather than keeping invalid date in the model, right? As such I think that we should rather explore proper validation support rather than introduce an option proposed in this PR. WDYT?

ajafff:
Actually it really is about keeping the invalid date in the model.
Maybe I lost myself thinking ahead with validation, but that should not be
subject of this PR.

I want the invalid date to remain in the input field, to give users a
chance to recognize and correct their faulty input.

Am 13.12.2016 18:41 schrieb "Pawel Kozlowski" <notifications@github.com>:

@ajafff <https://github.com/ajafff> if I understand you correctly it is
more about proper validation messages rather than keeping invalid date in
the model, right? As such I think that we should rather explore proper
validation support rather than introduce an option proposed in this PR.
WDYT?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/ng-bootstrap/ng-bootstrap/pull/1128#issuecomment-266807995>,
or mute the thread

ajafff:
I thought about this again. I think I get your point now. The problem with
my solution is, that ngModel would return NgbDateStruct for valid dates,
null for empty input and string for invalid dates. That's not desirable and
right now I have no idea how to work around that.

Seems like "proper validation" would be the better option. But then you
face similar issues: validation is triggered after the model value changes.
So the invalid date is cleared, leaving you with `null` for validation. Now
you have to guess if the user intentionally removed his input or if the
date was invalid.

Looking forward for your ideas regarding this matter.

Am 13.12.2016 18:56 schrieb "Klaus Meinhardt" <klaus.meinhardt1@gmail.com>:

pkozlowski-opensource:
> I want the invalid date to remain in the input field, to give users a
chance to recognize and correct their faulty input.

That is fine. I'm talking about sth else - that is - if invalid date should be propagated to the _model_ (without discussing how it effect input fields / UI display). My answer would be _no_ - there is no point in polluting application data (model) with invalid values.

I hope we both can agree with the following:
* we should be able to provide users with clear validation messages when they enter invalid date
* entering an invalid date shouldn't clear input fields so users can correct their mistake.

If so I'm just suggesting that this PR is not how I would like the above problems to be solved in this library. We should rather have a validator that sets an error for invalid dates. This way you could distinguish between empty and invalid dates and drive UI appropriately.  Does it make sense?

ajafff:
@pkozlowski-opensource I just notices, that I mixed model and UI in one of my posts above. I totally agree with you. It doesn't make sense to populate invalid values to the model.

I'm closing this PR. Maybe I can come up with a better solution if I have time for it.

Thank you for your patience.

PR #1131, opened 2016-12-09
chore: freeze typescript version

maxokorokov:
`typescript: ^2.0.6` installs the latest `2.1.4` version that breaks the build. 
So to unblock for now:

- freezing to latest `2.0.10` (angular doesn't work with `2.1` yet, and our ngc build fails)
- future-proofing: removing the useless check that breaks `2.1` in tests


pkozlowski-opensource:
LGTM. Go ahead and merge it @maxokorokov. Thnx!

PR #1139, opened 2016-12-13
fixed case if label wasn't provided

Le0Michine:

since label is an optional dependency it should be checked for undefined

### DIFF
--- a/src/buttons/radio.ts
+++ b/src/buttons/radio.ts
@@ -142,7 +142,9 @@ export class NgbRadio implements OnDestroy {
   update(value, isDisabled) {
     this._checked = (this.value === value && value !== null);
     this._disabled = isDisabled;
-    this._label.active = this._checked;
-    this._label.disabled = this._disabled;
+    if (this._label) {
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@Le0Michine could you please add a test for this case? Unfortunately we can't merge code changes without associated tests.

pkozlowski-opensource:
@Le0Michine I just did some changes to buttons in cc88ed2 and now I think that the only situation where we can get into troubles is when using radio buttons with `ngbRadioGroup` and _without_ the `<label>` element. But I don't think that this is the case we want to support.

I'm going to close this PR as it is missing tests and doesn't merge cleanly any more. Please _do_ open a new PR if you still think that we've got legitimate cases where we can run into troubles. But please make sure that such PR has an associated test so we don't introduce regressions in the future. Thnx!

Le0Michine:
@pkozlowski-opensource I agree with your point regarding this case, but in this case label shouldn't be an optional dependency. Also in this property

set focused(isFocused: boolean)

_label is checked for udefined

Issue #779, opened 2016-09-21
Export instances of all directives


pkozlowski-opensource:
Based on https://github.com/ng-bootstrap/ng-bootstrap/issues/778#issuecomment-248655998

If people want to query for directive instances they can't. We need to evaluate for which directives it makes sense to have query and export those directives.

colinmorelli:
Is there a reason you wouldn't want to export all directives that the module exports? IMO if it's something the angular module exports, then it's part of the public API and the TS module should export it as well.

wesleycho:
Some may be meant purely for internal use with no external exposure. It also avoids the problem of users doing not great things and trying to pressure the library to support a subpar api as a result instead of solving a problem the right way in the library.

pkozlowski-opensource:
I'm going to close this one till a real need surfaces.

colinmorelli:
> Some may be meant purely for internal use with no external exposure. It also avoids the problem of users doing not great things and trying to pressure the library to support a subpar api as a result instead of solving a problem the right way in the library.

I'm only referring to exporting types for components that are already being exported in the angular module. If they're internal only, the angular module shouldn't be exporting them and it's a non issue.

Anything that is being discussed here is already possible (and even encouraged, in many cases, on the docs site). Not exposing type information is not preventing people from doing anything (or at least not preventing them from doing much), it's just ensuring their code is less-safe when they do so, by forcing users to type cast to `any` and throwing away the value of TypeScript for these cases.

Given that we can assume people are calling methods on the component instances within templates (the docs actively encourage this in many places), it stands to reason that they'll also be called from within user-defined components. I think it's fairly easy to find a whole host of cases in which this is useful to have, and I see no reason to _not_ export the primary components of the library (i.e. all the components that are documented on the doc site and are expected to be used).

pkozlowski-opensource:
OK, got convinced by @colinmorelli - let's export public (documented) directives.

CSchulz:
I want to import NgbAccordion to be able to trigger the toggle via component.
```
import {NgbAccordion} from '@ng-bootstrap/ng-bootstrap';
```

The panel is disabled at first and the user should click on next. This will trigger a method in the component, which checks the form and set the property to enable the tab. Furthermore it should toggle the panel.

pkozlowski-opensource:
@CSchulz could you put together a plunker with what you are trying to do? It sounds like it _should_ be possible but I'm not sure I understand your exact use-case.

[...]

PR #1145, opened 2016-12-15
Context in popovers

isaacplmann:
I added the ability to pass context into popover templates.

### DIFF
--- a/demo/src/app/components/popover/demos/tplwithcontext/popover-tplwithcontext.ts
+++ b/demo/src/app/components/popover/demos/tplwithcontext/popover-tplwithcontext.ts
@@ -1,0 +1,22 @@
+import {NgbPopover} from '@ng-bootstrap/ng-bootstrap';
+import {Component, ViewChild} from '@angular/core';
+
+@Component({
+  selector: 'ngbd-popover-tplwithcontext',
+  templateUrl: './popover-tplwithcontext.html'
+})
+export class NgbdPopoverTplwithcontext {
+  greeting = {};
+  name = 'World';
+
+  @ViewChild('p') public popover: NgbPopover;
+
+  public changeGreeting(greeting: any): void {
+    const isOpen = this.popover.isOpen();
+    this.popover.close();
+    if (greeting !== this.greeting || !isOpen) {
+      this.greeting = greeting;
+      this.popover.open(greeting);
+    }
+  }
+}
--- a/demo/src/app/components/tooltip/demos/tplwithcontext/tooltip-tplwithcontext.ts
+++ b/demo/src/app/components/tooltip/demos/tplwithcontext/tooltip-tplwithcontext.ts
@@ -1,0 +1,22 @@
+import {NgbTooltip} from '@ng-bootstrap/ng-bootstrap';
+import {Component, ViewChild} from '@angular/core';
+
+@Component({
+  selector: 'ngbd-tooltip-tplwithcontext',
+  templateUrl: './tooltip-tplwithcontext.html'
+})
+export class NgbdTooltipTplwithcontext {
+  greeting = {};
+  name = 'World';
+
+  @ViewChild('t') public tooltip: NgbTooltip;
+
+  public changeGreeting(greeting: any): void {
+    const isOpen = this.tooltip.isOpen();
+    this.tooltip.close();
+    if (greeting !== this.greeting || !isOpen) {
+      this.greeting = greeting;
+      this.tooltip.open(greeting);
+    }
+  }
+}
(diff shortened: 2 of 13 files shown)


pkozlowski-opensource:
Could you please share motivation for this change? Also please note that we can't merge any PRs without associated tests so adding those would be mandatory.

But yeh, before speaking about tests I would like to hear more about a use-case...

isaacplmann:
I'm writing a product tour library - [ng2-tour](https://isaacplmann.github.io/ng2-tour/)

I need to be able to pass in a template for the popover content that can have the context values inserted when the popover is shown.  You can see in the demo page that the same template is used for all the tour steps, but the content text is different.

The template looks like this:
```html
    <template #tourStep let-step="step">
      <p class="tour-step-content">{{step?.content}}</p>
      <div class="tour-step-navigation">
          // buttons, etc, go here...
      </div>
    </template>
```

That `let-step` directive will never get a value supplied to it unless a context is associated with the popover when it is created.  Angular has the ability built in, you just weren't surfacing the API.  Hence, the PR.

pkozlowski-opensource:
I see, make sense. For this PR to go forward it would need:
* tests;
* tooltips update (we are trying to make both tooltip and popovers aligned but can't share more code till we switch to Angular 2.3.x);
* documentation / demo update.

Are you willing to put more work into it to push it forward?

isaacplmann:
Working on it now.  Can you point me to the tooltip/documentation files that need updated?  I figure you know the organization better than I do.

pkozlowski-opensource:
> Can you point me to the tooltip/documentation files that need updated? I figure you know the organization better than I do.

Documentation is in tooltip / popover files (jsdoc). This is where you would have to document new `@Input()`s. Demos reside in https://github.com/ng-bootstrap/ng-bootstrap/tree/master/demo/src/app/components/popover and https://github.com/ng-bootstrap/ng-bootstrap/tree/master/demo/src/app/components/tooltip

Thnx!

isaacplmann:
I added tests and docs.  There weren't any new @Inputs, just an optional new parameter to the open() function.  The popover and tooltip pages now document the new optional context parameter and there's a new demo section on each page.

isaacplmann:
Any feedback on the tests/docs?

demo/src/app/components/popover/demos/tplwithcontext/popover-tplwithcontext.ts
```diff
@@ -0,0 +1,22 @@
+import { NgbPopover } from './../../../../../../../src/popover/popover';
```
pkozlowski-opensource: Wrong import

demo/src/app/components/tooltip/demos/tplwithcontext/tooltip-tplwithcontext.ts
```diff
@@ -0,0 +1,22 @@
+import { NgbTooltip } from './../../../../../../../src/tooltip/tooltip';
```
pkozlowski-opensource: Wrong import

pkozlowski-opensource:
@isaacplmann could you please:
* fix imports (I've left comments in the code)
* squash all the commits into one
* amend the commit message to follow our conventions?

isaacplmann:
Changed the imports to import from '@ng-bootstrap/ng-bootstrap'.  And squashed the commits into one.

pkozlowski-opensource:
@isaacplmann I was trying to pull in this change but the demo site fails with the following errors (there are similar ones for popovers):

> ERROR in ./demo/src/app/components/tooltip/demos/tplwithcontext/tooltip-tplwithcontext.ts
(1,9): error TS2305: Module '"/home/pk/work/gitrepos/gh/pkozlowski-opensource/core/src/index"' has no exported member 'NgbTooltip'.

> ERROR in ./~/prismjs-loader?lang=typescript!./demo/src/app/components/tooltip/demos/tplwithcontext/tooltip-tplwithcontext.ts
(1,9): error TS2305: Module '"/home/pk/work/gitrepos/gh/pkozlowski-opensource/core/src/index"' has no exported member 'NgbTooltip'.

I guess that the root cause is #779 so let me prepare a PR that exports all the directive instances. Then you will be able to rebase on top of it and things should work. Unless you want to take a stab at #779?

isaacplmann:
It looks like you're already fixing this in #1154 .  If you merge that into master, I'll rebase this PR.

pkozlowski-opensource:
@isaacplmann #1154 landed, please rebase.

isaacplmann:
Rebased.  And the demo now works.

pkozlowski-opensource:
Merged, thnx!

PR #1148, opened 2016-12-16
fix(buttons): disabled state of group interactions

### LINKED ISSUE #1143: Disabled buttons corner cases
While investigating button-related issues I could spot some corner cases around the `disabled` state handling:
* it is possible to enable an individual button even if the entire group is disabled
* switching group from `disabled` => `enabled` re-enables all buttons, even the ones that are explicitly disabled.

We should probably re-think `disabled` state propagation from a group to individual buttons (keep both buttons and group state and calculate final states based on the 2).

alex321:
Closes #1143

### DIFF
--- a/src/buttons/radio.spec.ts
+++ b/src/buttons/radio.spec.ts
@@ -398,6 +398,62 @@ describe('ngbRadioGroup', () => {
            });
      }));
 
+  it('disable all radio buttons when group is disabled regardless of button disabled state', async(() => {
+       const html = `
+      <form>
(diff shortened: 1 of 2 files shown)


src/buttons/radio.spec.ts
```diff
@@ -398,6 +398,62 @@ describe('ngbRadioGroup', () => {
            });
      }));
 
+  it('disable all radio buttons when group is disabled regardless of button disabled state', async(() => {
```
pkozlowski-opensource: Start with 'should ....'

alex321: Sorry about that.

PR #1158, opened 2016-12-23
feat(datepicker): allow outsideDays for any number of displayMonths

alexluecke:
# Motivation

I would expect the outside days to be present on any value of `displayMonths`. Foremost, the `outsideDays` for multiple months can be explicitly set to hidden to achieve the hidden outside days, so it does not make sense to inhibit potentially useful functionality. In addition, day display can be controlled through the `dayTemplate` input.

# Issues:

This is obviously a breaking change. However, since the project is still in alpha I would prefer nipping this one in the bud now to allow for more customizeable calendars in the future.

----

### DIFF
--- a/src/datepicker/datepicker.spec.ts
+++ b/src/datepicker/datepicker.spec.ts
@@ -259,18 +259,29 @@ describe('ngb-datepicker', () => {
     expect(fixture.debugElement.query(By.directive(NgbDatepickerNavigation))).toBeNull();
   });
 
-  it('should override outside days to "hidden" if there are multiple months displayed', () => {
-    const fixture = createTestComponent(
-        `<ngb-datepicker [displayMonths]="displayMonths" [outsideDays]="'collapsed'"></ngb-datepicker>`);
+  it('should set outside days to "visible" if not provided when there is a single month displayed', () => {
+    const fixture = createTestComponent(`<ngb-datepicker [displayMonths]="1"></ngb-datepicker>`);
+    let [month] = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
+    expect(month.componentInstance.outsideDays).toBe('visible');
+  });
 
+  it('should set outside days to "hidden" if provided when there is a single month displayed', () => {
+    const fixture = createTestComponent(`<ngb-datepicker [displayMonths]="1" outsideDays="hidden"></ngb-datepicker>`);
+    let [month] = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
+    expect(month.componentInstance.outsideDays).toBe('hidden');
+  });
 
-    let months = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
-    expect(months[0].componentInstance.outsideDays).toBe('collapsed');
+  it('should set outside days to "hidden" if not provided when there are multiple months displayed', () => {
+    const fixture = createTestComponent(`<ngb-datepicker [displayMonths]="2"></ngb-datepicker>`);
+    let [month] = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
+    expect(month.componentInstance.outsideDays).toBe('hidden');
+  });
 
-    fixture.componentInstance.displayMonths = 2;
-    fixture.detectChanges();
-    months = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
-    expect(months[0].componentInstance.outsideDays).toBe('hidden');
+  it('should set outside days if provided when there are multiple months displayed', () => {
+    const fixture =
+        createTestComponent(`<ngb-datepicker [displayMonths]="2" [outsideDays]="'visible'"></ngb-datepicker>`);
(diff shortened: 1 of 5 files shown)


maxokorokov:
Hey, @alexluecke. Good suggestion, it was just simpler to implement this way in the beginning. I have some remarks though:

1. We have 3 values for `outsideDays` and we have decide on behaviour for multiple months:

- `hidden` → works now
- `collapsed` → would work the same as for 1 month I suppose
- `visible` → some days would be displayed twice in the view - selection have to be fixed, as now both of them would be highlighted. Plus what will happen when you click on them?

2. I think I personally would prefer to have `visible` as default for 1 month view and `hidden` for 1+ month view

3. Please implement corresponding unit tests, especially for the `visible` case, as we won't be able to merge the PR without them.

alexluecke:
So I've implemented the changes and added tests. For now I think it might be acceptable to have both days selected in the multi-calendar view.

Initially, I thought it might be a good idea to only mark days as selected if they aren't disabled, which would allow the user to disable the multi-date selection issue. However, when this is done the outside days won't be selectable and show up as greyed out days, so I don't think much is gained from just setting the `outsideDays` to `hidden` (as was the default). Being the case, the multi-selection would be my preferred path forward for now.

src/datepicker/datepicker.ts
```diff
@@ -84,7 +84,7 @@ export interface NgbDatepickerNavigateEvent {
             [showWeekdays]="showWeekdays"
             [showWeekNumbers]="showWeekNumbers"
             [disabled]="disabled"
-            [outsideDays]="displayMonths === 1 ? outsideDays : 'hidden'"
+            [outsideDays]="outsideDays || displayMonths === 1 ? 'visible' : 'hidden'"
```
miller-time: I think with operator precedence you need some parens here:

```
[outsideDays]="outsideDays || (displayMonths === 1 ? 'visible' : 'hidden')"
```

Otherwise you'll be modifying the value of the month view's `outsideDays` to always be "visible" any time it is explicitly set on the outer datepicker.

alexluecke: If this is true, then I am probably missing a test as well. Let me follow up.

alexluecke: Good catch. Added test and fixed up the code.

src/datepicker/datepicker.spec.ts
```diff
@@ -259,20 +259,31 @@ describe('ngb-datepicker', () => {
...
+    expect(months[0].componentInstance.outsideDays).toBe('hidden');
+  });
 
-    fixture.componentInstance.displayMonths = 2;
-    fixture.detectChanges();
-    months = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
+  it('should set outside days to "hidden" if not provided when there are multiple months displayed', () => {
+    const fixture = createTestComponent(`<ngb-datepicker [displayMonths]="2"></ngb-datepicker>`);
+    let months = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
     expect(months[0].componentInstance.outsideDays).toBe('hidden');
```
wesleycho: Just a minor nitpick if you want to make the adjustment, you could do
```ts
let [month] = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
expect(month.componentInstance.outsideDays).toBe('hidden');
```
or
```ts
let [{componentInstance: {outsideDays}}] = fixture.debugElement.queryAll(By.directive(NgbDatepickerMonthView));
expect(outsideDays).toBe('hidden');
```
Whatever is preferred is I guess whatever makes it more readable, just food for thought.

alexluecke: I think the first version is much cleaner, thanks for the suggestion. I will make the changes.

maxokorokov:
I'm sorry, I can't agree to merge as is, because this particular case doesn't look good to me at all:

screen shot 2017-04-13 at 18 31 22

With `[outsideDates]='visible'` and multiple months you'll get currently selected day displayed in double

Maybe in this case (multiple months, outsideDays == visible) we should simply 
- collapse all the days in between August and September
- leave July and October days visible

?

alexluecke:
@maxokorokov 

Could you clarify your asks? By "collapse" do you mean give the "collapse" view for outsideDays (instead of purely "visible")? For the use case, I just need to have all of the days for a given month (see: https://en.wikipedia.org/wiki/Broadcast_calendar for a calendar where this is applicable).

maxokorokov:
Yeah, maybe it wasn't clear, sorry. 

My suggestion is:
* red ones should be always forced to `collapsed` in case of multiple months
* yellow ones should depend on the actual `[outsideDays]` value

febe15c0-2077-11e7-8aff-e5cc8f860746

pkozlowski-opensource:
@alexluecke we haven't heard back from you for a while so I wonder if we should close this one as stale? @maxokorokov WDYT?

alexluecke:
Yeah, I got a working solution that I was really not happy with. I will close ticket and open a new PR if I ever come up with a better solution.

Issue #1178, opened 2017-01-07
Support bootstrap 4.0.0-alpha.6


giggio:
### Bug description:

Since the release of alpha 6 today (1/6/17) some components stopped working. I noticed that modal does not work anymore.

### Link to minimally-working plunker that reproduces the issue:

You can try your own plunker at:

https://ng-bootstrap.github.io/app/components/modal/demos/component/plnkr.html

It is already not working as it links to the dist folder from bootstrap.

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 2.0.0

ng-bootstrap: 1.0.0-alpha.16

Bootstrap: 4.0.0-alpha.6

richardsengers:
Dropwdowns stopped working as well

aL3891:
Also having issues with dropdowns and possibly navbar, though that might be due to the dropdowns

pkozlowski-opensource:
For anyone arriving / commenting on this issue: we are well aware of all the Bootstrap 6 compatibility breaking changes and got all the directives working again with `4.0.0-alpha.6` in the [bootstrap-alpha.6-upgrade branch](https://github.com/ng-bootstrap/ng-bootstrap/tree/bootstrap-alpha.6-upgrade).

We just have to do some cosmetic changes to the demo site (the exact list of changes in the comment below) and will be releasing beta.18 with full Bootstrap `4.0.0-alpha.6` compatibility today / tomorrow the latest.

pkozlowski-opensource:
Changes to the demo page before we can release:
* [ ] - progressbar demo looks wired (lack of vertical space between bars)
* [ ] - update IE9 compatibility info

pkozlowski-opensource:
We've just released 1.0.0-alpha.18 with full support for Bootstrap 4.0.0-alpha.6. Enjoy!

Just a note: 1.0.0-alpha.18 is _not_ backward compatible with Bootstrap 4.0.0-alpha.5. So, as soon as you go with 1.0.0-alpha.18 you need to update your full project to Bootstrap 4.0.0-alpha.6!

PR #1189, opened 2017-01-07
chore(popover): upgrade popover to v4-alpha.6

jnizet:
Not really linked to alpha-6, but since the migration to alpha-6 isn't backward-compatible, keeping the caret for compatibility with alpha.4 isn't needed anymore.

refs #1178


pkozlowski-opensource:
Merged as 60fd5d954cff7aa982dd328845689c039251ffe2

PR #1192, opened 2017-01-07
chore: upgrade documentation to bootstrap-4.0.0-alpha.6

jnizet:
refs #1178

BREAKING CHANGE: This version is compatible with bootstrap-4.0.0-alpha.6, and is not backward-compatible with previous versions of bootstrap. If you choose to upgrade to this version of ng-bootstrap, you will thus also have to migrate to bootstrap-4.0.0-alpha.6.


pkozlowski-opensource:
Merged as 449992756ede4d6bb61140bfce11171bf4373232

Issue #1106, opened 2016-12-02
NgbTimepicker - When meridian is true: error Cannot read property 'hour' of undefined


jchewitt:
Under timepicker.ts you'll see that **model** isn't initialized, only declared. Model is essentially undefined until angular2 initiates the binding with ControlValueAccessor and writeValue() is called.

```
export class NgbTimepicker implements ControlValueAccessor,
    OnChanges {
  disabled: boolean;
  model: NgbTime;
```

I'm not sure why it's working in your demo, however I can only assume that something has changed between versions and this fix will be needed as I'm on the latest (2.2.0) Also doesn't work with (2.0.2).

If you'll notice below that we don't hold model accountable when trying to access parameters from it until we get to checking if meridian is true.

```
            <input type="text" class="form-control" maxlength="2" size="2" placeholder="MM"
              [value]="formatMinSec(model?.minute)" (change)="updateMinute($event.target.value)" 
              [readonly]="readonlyInputs" [disabled]="disabled">
          </td>
          <template [ngIf]="seconds">
            <td>&nbsp;:&nbsp;</td>
            <input type="text" class="form-control" maxlength="2" size="2" placeholder="SS"
              [value]="formatMinSec(model?.second)" (change)="updateSecond($event.target.value)" 
              [readonly]="readonlyInputs" [disabled]="disabled">
          </template>
          <template [ngIf]="meridian">
            <td>&nbsp;&nbsp;</td>
            <td>
              <button type="button" class="btn btn-outline-primary" (click)="toggleMeridian()">{{model.hour >= 12 ? 'PM' : 'AM'}}</button>
```

I believe we need to soak up the undefined reference by treating the model as possibly undefined if the meridian is set to true. Currently we're expecting there to be an hour key on a possibly defined object and that doesn't work since model isn't set on class initiation, only when a change to the model is made and detected through controlvalueaccessor.

alex321:
@jchewitt  Could you please submit a minimal plnkr demonstrating what is going wrong. It is hard to replicate and understand what exactly the problem is. I believe that `writeValue` is called by angular on initialisation as well.

pkozlowski-opensource:
@jchewitt as commented by @alex321 please provide a plunker with a minimal reproduction scenario, otherwise this one is hardly actionable. Here is why: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/README.md#you-think-youve-found-a-bug

dudleycodes:
[...]

PR #1195, opened 2017-01-09
fix(timepicker): handle undefined state when meridian is true

shibulijack:
Fixes #[1106](https://github.com/ng-bootstrap/ng-bootstrap/issues/1106)

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -101,7 +101,7 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
           <template [ngIf]="meridian">
             <td>&nbsp;&nbsp;</td>
             <td>
-              <button type="button" class="btn btn-outline-primary" (click)="toggleMeridian()">{{model.hour >= 12 ? 'PM' : 'AM'}}</button>
+              <button type="button" class="btn btn-outline-primary" (click)="toggleMeridian()">{{model?.hour >= 12 ? 'PM' : 'AM'}}</button>
             </td>
           </template>
         </tr>


jedijashwa:
I downloaded and built this change and it fixed the issue on my machine and did not break any other functionality.

pkozlowski-opensource:
I would _really_ like to see a plunker with a reproduce scanario before we move forward on this. Without a reproduce scenario we don't understand what is going on exactly so we can't add a test (and this PR is missing a test so there is zero gurantee that we are not going to re-introduce regressions in the future). 

@shibulijack @jedijashwa since you are seeing this pb on your side could you please distill it in a plunker and add a test to this PR?

jedijashwa:
@pkozlowski-opensource it will be a while but I can if @shibulijack cannot.

shibulijack:
@pkozlowski-opensource I tried to reproduce the scenario in a Plunker, but couldn't. My project is scaffolded using Angular Cli, so it uses Webpack instead of System JS. How can I replicate this using Plunker?

jedijashwa:
My project is using system js but I pulled out the timepicker stuff, hence it taking so long for me to set up a plunker

pkozlowski-opensource:
@shibulijack I've just spent 1/2h trying to reproduce the problem that this PR is supposed to solve but couldn't - was trying all kind of crazy input combinations and couldn't break it. As such I'm going to close this PR since it doesn't merge cleanly, we don't have, reproduce scenario etc.

More than happy to re-visit as soon as someone can reproduce the original issue in a plunker. Thnx!

PR #1205, opened 2017-01-10
feat(datepicker): improve default look and feel

### LINKED ISSUE #706: datepicker: visual improvements

maxokorokov:
Updating default datepicker look and feel as a part of #862 :

* Moving to flex layout instead of table layout
* More space and border around the datepicker
* Less pronounced weekdays
* Visual separation of the header from the body
* Buttons same as in the timepicker
* Prefixing css classes with `ngb-dp-`

Before:
screen shot 2017-01-10 at 11 38 46

After:
screen shot 2017-01-10 at 11 39 17

screen shot 2017-01-10 at 11 41 47

P.S. Will also add full month names with a separate request later

BREAKING CHANGES: switched to flex layout instead of table-based (drop IE9 support) and prefixed datepicker-related css classes with `ngb-dp-`

Fixes #706

### DIFF
--- a/src/datepicker/datepicker-day-view.ts
+++ b/src/datepicker/datepicker-day-view.ts
@@ -4,16 +4,22 @@ import {NgbDateStruct} from './ngb-date-struct';
 @Component({
   selector: '[ngbDatepickerDayView]',
   styles: [`
-    :host {      
+    :host {
       text-align: center;
-      padding: 0.185rem 0.25rem;      
+      width: 2rem;
+      height: 2rem;
+      line-height: 2rem;      
       border-radius: 0.25rem;
+    }
+    :host.outside {
+      opacity: 0.5;
     }
   `],
   host: {
     '[class.bg-primary]': 'selected',
     '[class.text-white]': 'selected',
     '[class.text-muted]': 'isMuted()',
+    '[class.outside]': 'isMuted()',
     '[class.btn-secondary]': '!disabled'
   },
   template: `{{ date.day }}`
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -47,36 +47,49 @@ export interface NgbDatepickerNavigateEvent {
 @Component({
   exportAs: 'ngbDatepicker',
   selector: 'ngb-datepicker',
-  host: {'class': 'd-inline-block'},
+  host: {'class': 'd-inline-block rounded'},
   styles: [`
-    .month:first-child {
-      padding-left: 0 !important;
+    :host {
+      border: 1px solid rgba(0, 0, 0, 0.125);
     }
+    .ngb-dp-header {
+      border-bottom: 1px solid rgba(0, 0, 0, 0.125);
+    }
+    .ngb-dp-month:first-child {
+      margin-left: 0 !important;
+    }    
+    .ngb-dp-month-name {
+      font-size: larger;
+      height: 2rem;
+      line-height: 2rem;
+    }    
   `],
   template: `
     <template #dt let-date="date" let-currentMonth="currentMonth" let-selected="selected" let-disabled="disabled">
        <div ngbDatepickerDayView [date]="date" [currentMonth]="currentMonth" [selected]="selected" [disabled]="disabled"></div>
     </template>
+    
+    <div class="ngb-dp-header bg-faded pt-1 rounded-top" [style.height.rem]="getHeaderHeight()" 
+      [style.marginBottom.rem]="-getHeaderMargin()">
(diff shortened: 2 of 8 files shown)


src/datepicker/datepicker-day-view.ts
```diff
@@ -4,16 +4,22 @@ import {NgbDateStruct} from './ngb-date-struct';
...
     }
+    :host.outside {
+      opacity: 0.5;
+    }
   `],
   host: {
     '[class.bg-primary]': 'selected',
     '[class.text-white]': 'selected',
     '[class.text-muted]': 'isMuted()',
+    '[class.outside]': 'isMuted()',
```
deeg: What is the reason for the different classes if they go off the same exact check? Is it possible to have text-muted without outside, or vice versa?

src/datepicker/datepicker.ts
```diff
@@ -47,36 +47,63 @@ export interface NgbDatepickerNavigateEvent {
...
-          {{ i18n.getMonthName(month.number) }} {{ month.year }}
-        </td>
-      </tr>
-      <tr>
-        <td *ngFor="let month of months" class="pl-1 month">
+    
+    <div class="ngb-dp-navigation bg-faded">
+      <ngb-datepicker-navigation *ngIf="navigation !== 'none'"
+        [date]="months[0]?.firstDate"
+        [minDate]="_minDate"
```
deeg: Should private variables be used in templates? Not sure if these should be public or not. Maybe they are, just going by the name of the var here.

src/datepicker/datepicker.ts
```diff
@@ -47,36 +47,63 @@ export interface NgbDatepickerNavigateEvent {
...
+        [disabled]="disabled"
+        [showWeekNumbers]="showWeekNumbers"
+        [showSelect]="navigation === 'select'"
+        (navigate)="onNavigateEvent($event)"
+        (select)="onNavigateDateSelect($event)">
+      </ngb-datepicker-navigation>
+    </div>
+
+    <div class="ngb-dp-months d-flex">
+      <div class="ngb-dp-month-spacer-sm bg-faded" [style.height.rem]="navigation !== 'select' || displayMonths > 1 ? 4 : 2"></div>
```
deeg: I don't mind it that much, but might it be easier to read if you move this logic to a function? Or is the standard to keep simple stuff like this int he HTML?

src/datepicker/datepicker-day-view.ts
maxokorokov: `text-muted` is the bootstrap class. I wanted to add 50% opacity on top of it, so two options:

- `[style.opacity]="isMuted() ? 0.5 : 1"`
- add the css class

I picked the last one, not sure which one is better though

src/datepicker/datepicker.ts
maxokorokov: Yeah, this is not a private variable... the thing is there is an `@Input() minDate: NgbDateStruct` and an internal data model `_minDate: NgbDate`

Not sure about this whole `NgbDateStruct` vs `NgbDate` thing, but it's a whole another story :)

src/datepicker/datepicker.ts
maxokorokov: I've just simplified layout significantly and got rid of spacers altogether. Done with css now. Thanks for pointing this

maxokorokov:
Amended:

- simplified layout even more → removed spacers and did them with css
- fixed and added tests with various `[showWeekdays]`, `[navigation]` and `[displayMonths]` values
- fixed TS < 2.0 compatibility

P.S. Tested in IE 10, 11, Edge. Seems to look fine.

Matmo10:
Hi @maxokorokov, thanks for the visual refresh on this.   Do you happen to know what version this will land / landed in? I just installed `1.0.0-alpha.20` but I still see the old calendar. I'm just slightly confused as the demo website already shows the new design, and says that `1.0.0-alpha.20` is the latest release - but I can't see the new calendar after installing `1.0.0-alpha.20`.

maxokorokov:
@Matmo10 it landed in [alpha.19](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CHANGELOG.md#100-alpha19-2017-01-25).

I've just upgraded to `alpha.20` on one of my projects and it worked fine. 
Did the same, also deleted `node_modules/@ng-bootstrap` before installing just to be sure too.

PR #1208, opened 2017-01-11
feat(modal): remove container onDestroy.

aitboudad:


pkozlowski-opensource:
@aitboudad thnx for this PR. Before I review it I would like to know more about the motivation for this PR. Did you bump into any real-life scenarios where the container needs to be destroyed (we are not creating it programatically so...).

Also please note that in the very near future we are going to drop the need for a container altogether so I'm not keen on merging this PR without understanding impact on real-life applications. Could you elaborate? Thnx!

aitboudad:
I used `NgbModal` on a child module and when it's destroyed I noticed the `ModalContainer` remains alive in `ModalStack`
So if you are going to drop the need for a container then you can close this PR!

pkozlowski-opensource:
Yeh, I prefer to just drop the need for `ModalContainer`. Thnx for the PR though!

PR #1210, opened 2017-01-12
feat(tabset): tabs horizontal alignment

ktriek:
# Tabs Horizontal alignment
Add the horizontal alignment support to the tabs

- Start (Default)
- Center
- End

http://v4-alpha.getbootstrap.com/components/navs/#horizontal-alignment

PR #1216, opened 2017-01-13
feat(pagination): add horizontal alignment config option

tmeneau:
This is similar to pull request #1210 -- except applied to the pagination component -- so I used the same configuration name (`justify`) and configurable values for consistency: 

* 'start' (default)
* 'center'
* 'end'

### DIFF
--- a/src/pagination/pagination-config.ts
+++ b/src/pagination/pagination-config.ts
@@ -15,4 +15,5 @@ export class NgbPaginationConfig {
   pageSize = 10;
   rotate = false;
   size: 'sm' | 'lg';
+  justify: 'start' | 'center' | 'end' = 'start';
 }
--- a/src/pagination/pagination-config.ts
+++ b/src/pagination/pagination-config.ts
@@ -7,13 +7,13 @@ import {Injectable} from '@angular/core';
  */
 @Injectable()
 export class NgbPaginationConfig {
-  disabled = false;
   boundaryLinks = false;
   directionLinks = true;
+  disabled = false;
   ellipses = true;
+  justify: 'start' | 'center' | 'end' = 'start';
   maxSize = 0;
   pageSize = 10;
   rotate = false;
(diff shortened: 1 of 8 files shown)


src/pagination/pagination-config.ts
```diff
@@ -15,4 +15,5 @@ export class NgbPaginationConfig {
   pageSize = 10;
   rotate = false;
   size: 'sm' | 'lg';
+  justify: 'start' | 'center' | 'end' = 'start';
```
maxokorokov: Please order alphabetically

src/pagination/pagination.ts
```diff
@@ -109,6 +109,11 @@ export class NgbPagination implements OnChanges {
    */
   @Input() size: 'sm' | 'lg';
 
+  /**
+   * Horizontal positioning with Flex justification: start, center, or end
+   */
+  @Input() justify: 'start' | 'center' | 'end';
```
maxokorokov: Please order alphabetically

tmeneau:
Thanks for the review! I went ahead and corrected the ordering in the:

* component members
* configuration members
* spec references and test classes

pkozlowski-opensource:
@tmeneau do you have a real-life use-case where just applying Bootstrap utility classes isn't enough? Asking since I can already position pagination control without any changes to this library: http://plnkr.co/edit/vIzZoSE6auPfrTBm4rxG?p=preview

IMO pagination use-case is different as compared to the tabset one (in a tabset we want to justify tabs to which we don't have separate access but in pagination we can take the entire control and move it around...).

tmeneau:
@pkozlowski-opensource Well, I'm a moron; thanks for taking the time to put together the plunker, I'd somehow missed that the pagination container isn't a flexbox container by default. You're completely right, I can't imagine any use case not already covered as far as pagination alignment is concerned.

Sorry about that! I'll close this pull request.

PR #1224, opened 2017-01-17
fix(datepicker): propagate null for invalid dates entered manually

### LINKED ISSUE #1222: Datepicker popup: invalid maxDate will not invalidate form
### Bug description:
Datepicker popup: If a date after maxDate is entered manually, the form will be valid by mistake. 

http://plnkr.co/edit/5sb1RN6AoJRbdiV32nke?p=preview

### Version of ng-bootstrap and Bootstrap:

ng-bootstrap: ng-bootstrap@1.0.0-alpha.18
Bootstrap: 4.0.0-alpha.6

pkozlowski-opensource:
This is part of #1222. To fix #1222 we also need to add implementation of the `Validator` interface (this will be done in a separate PR).

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -116,6 +116,46 @@ describe('NgbInputDatepicker', () => {
       expect(fixture.componentInstance.date).toBeNull();
     });
 
+    it('should propagate null to model when a user date outside of specified range', () => {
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
Actually this is wrong - what seems to happen with built-in value accessors is that invalid model values _are_ propagated back to the model (more testing needed here). What we are missing today is the implementation of the `Validator` interface.

Issue #1242, opened 2017-01-23
Typeahead modifies Observable prototype


achimha:
### Bug description:

The current code modifies the prototype of Observable in the typeahead component using statements like here:

https://github.com/ng-bootstrap/ng-bootstrap/blob/cfbc24a020d17b17181e684575df915d82e44a26/src/typeahead/typeahead.ts#L21

`import 'rxjs/add/observable/fromEvent';`

This is bad for several reasons. The consuming application is not aware of what is imported where and under which conditions. Also it constitutes a side effect which is generally to be avoided. The right way to do it is to import the operator as a function and use it standalone (e.g. `map.call(myObservable, ...)`), passing the Observable as an argument. An example can be found in the Angular code base:

https://github.com/angular/angular/blob/master/modules/%40angular/router/src/router.ts#L20

https://github.com/angular/angular/blob/master/modules/%40angular/router/src/router.ts#L685

pkozlowski-opensource:
@achimha thnx for reporting. While I agree on the principle I would like to hear more about the exact practical problems that this is causing in your case (to properly determine priority). Would you mind elaborating.

Also I'm not sure changing this will change much in practice. 

> The consuming application is not aware of what is imported where and under which conditions.

This won't change even if we change code as you suggest...

> Also it constitutes a side effect which is generally to be avoided.

Once again, agree on the principle but currently _any_ operator in RxJS works this way.

I'm trying to do a "good" thing here but would like to make sure that we are solving a real, practical problem.

achimha:
I had a very unpleasant bug in my application where it would mysteriously throw a runtime exception when run on IE11 due to a missing operator. It turned out that I had forgotten to import the RxJS operator but a 3rd party library did, although not under all circumstances! This is just an example of a side-effect that can bite you.

> Once again, agree on the principle but currently any operator in RxJS works this way.

I went through all 3rd party libraries I depend on to look for how widespread the problem is and while it is a common problem, only some libraries do it. Angular itself is very careful to not do it. RxJS offers each operator in a prototype-changing-add-fashion and as a separate function in the typical ES6 import fashion.

I do believe it is a practical problem, a library should not have side-effects, it defeats the purpose of the whole module/ES6 import paradigm.

The fix is very simple and ng-bootstrap only has it in one component so if you prefer I could try my luck with a PR...

DzmitryShylovich:
[...]

PR #1244, opened 2017-01-23
fix(typeahead): Resubscribe for value changes on blur, esc, enter

### LINKED ISSUE #723: typeahead: blur / esc / enter doesn't un-subscribe
There is a bug in the current implementation of the typeahead which might be exposed using a (relatively) long-running async operation. To reproduce go to our WIKIPEDIA search demo and:
- type any letter
- immediately after (before results arrive) press TAB

You will see that the result window gets opened even if a user is not interested in those any more (moved focus to a different field already).

dmytroyarmak:
Fix #723

By using `BehaviorSubject` and `switchMap` now we are doing unsubscription from current not-finished typeahead functions on following events:
- input blur
- press Escape when a popup is opened
- select option from a popup (by keyboard or mouse)

As a result of this unsubscription we have:
- typeahead function can do cleanup when a user doesn't want to see new results anymore (abort XHR request)
- a popup is not opened when new results are received but user already went to next input, selected option from current results or closed popup by Escape

### DIFF
--- a/demo/src/app/components/typeahead/demos/http/typeahead-http.ts
+++ b/demo/src/app/components/typeahead/demos/http/typeahead-http.ts
@@ -8,6 +8,7 @@ import 'rxjs/add/operator/distinctUntilChanged';
 import 'rxjs/add/operator/do';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/switchMap';
+import 'rxjs/add/operator/merge';
 
 @Injectable()
 export class WikipediaService {
@@ -41,6 +42,7 @@ export class NgbdTypeaheadHttp {
   model: any;
   searching = false;
   searchFailed = false;
+  hideSearchingWhenUnsubscribed = new Observable(() => () => this.searching = false);
 
   constructor(private _service: WikipediaService) {}
 
@@ -56,5 +58,6 @@ export class NgbdTypeaheadHttp {
               this.searchFailed = true;
               return Observable.of([]);
             }))
-      .do(() => this.searching = false);
+      .do(() => this.searching = false)
+      .merge(this.hideSearchingWhenUnsubscribed);
 }
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -19,6 +19,9 @@ const createTestComponent = (html: string) =>
 const createOnPushTestComponent = (html: string) =>
     createGenericTestComponent(html, TestOnPushComponent) as ComponentFixture<TestOnPushComponent>;
 
+const createAsyncTestComponent = (html: string) =>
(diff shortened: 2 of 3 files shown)


demo/src/app/components/typeahead/demos/http/typeahead-http.ts
```diff
@@ -41,6 +42,7 @@ export class NgbdTypeaheadHttp {
   model: any;
   searching = false;
   searchFailed = false;
+  hideSearchingWhenUnsubsctibed = new Observable(() => () => this.searching = false);
```
dmytroyarmak: I'm not sure if it's possible to better handle unsubscription of `search` observable

dmytroyarmak:
I've rebased PR after #1242 changes and remove some "refactoring" (that was previously in this PR) to make it easier to review.

demo/src/app/components/typeahead/demos/http/typeahead-http.ts
deeg: Looks to be a typo here.

dmytroyarmak: Fixed

dmytroyarmak:
@pkozlowski-opensource @maxokorokov what do you think about this PR?

pkozlowski-opensource:
@dmytroyarmak it looks good after a quick look and we are really grateful for all your contributions! But this is kind of a bigger change so it will take a bit of time to review. On top of this I'm planning some bigger changes around typeahead so I want to make sure that your PR doesn't prevent us from handling use-cases I want to support (ex. search on focus). 

Once again, we do appreciate the PR, we just need to ask you for a bit of patience. Thnx!

dmytroyarmak:
@pkozlowski-opensource no problem :)
Search on focus is the PR that I thought to work on next. 
It shouldn't be difficult to implement on top of the current code.
The `inputValues$` stream that we are passing to `ngbTextahead` function should re-emit current value on additional events (focus) instead of just on value changes: https://github.com/ng-bootstrap/ng-bootstrap/pull/1244/files#diff-4b999ef7b06784feb23d420f3d909d1aR162

deeg:
@pkozlowski-opensource, have you had any more thoughts on this PR? Often when using my search bar with an autocomplete, if I decide to do a search instead of using an autocomplete I see the results pop up later, even when I expect them never to show up.

saikatbhadra:
Any update on this? My team has hacked over the component to fix this but would love to have this

pkozlowski-opensource:
@dmytroyarmak this is an awesome PR, just merged it, thnx! Sorry it took so long - been busy with other components. 

Once again, thnx. Keep PRs lke this one coming :-)

PR #1249, opened 2017-01-24
demo: remove console.log polyfill

pkozlowski-opensource:
Bootstrap 4 dropped IE9 support so we don't need
console.log polyfill any more


dmytroyarmak:
@pkozlowski-opensource make sense 👍

PR #1256, opened 2017-01-25
fix(typeahead): doesn't modify Observable.prototype

### LINKED ISSUE #1242: Typeahead modifies Observable prototype

DzmitryShylovich:
Closes #1242

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -18,9 +18,9 @@ import {
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
 import {Observable} from 'rxjs/Observable';
 import {Subscription} from 'rxjs/Subscription';
-import 'rxjs/add/observable/fromEvent';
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
Merged, thnx!

PR #1258, opened 2017-01-25
feat(timepicker): add input and meridian button sizing

### LINKED ISSUE #1094: Timepicker feature request: allow for setting the underlying input control size
### Bug description:
This is a feature request. I am working on a form where smaller form controls would be preferable.  Currently there is now way to set the height size of the Timepicker inputs. 

In BS3 we were able to control the sizing of controls by adding the class `form-group-lg/sm`  on the parent element. Bootstrap 4, removed that option: we must now add the class `form-control-lg/sm` on the form-control itself. [https://v4-alpha.getbootstrap.com/components/forms/#control-sizing](https://v4-alpha.getbootstrap.com/components/forms/#control-sizing)

Just a link to the demo site showing that there is no such option. 
[https://ng-bootstrap.github.io/#/components/timepicker](https://ng-bootstrap.github.io/#/components/timepicker)

Angular:
NA
ng-bootstrap:
1.0.0-alpha.14
Bootstrap:
4.0.0-alpha.5

yehudag:
closes #1094

### DIFF
--- a/src/timepicker/timepicker-config.spec.ts
+++ b/src/timepicker/timepicker-config.spec.ts
@@ -12,5 +12,6 @@ describe('ngb-timepicker-config', () => {
     expect(config.secondStep).toBe(1);
     expect(config.disabled).toBe(false);
     expect(config.readonlyInputs).toBe(false);
+    expect(config.size).toBe('medium');
   });
 });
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
Thnx @yehudag - I've merged it since it looks about fine. One issue I could see, thought, is that chevrons look disproportionately small for `size="large"`. Could you follow up with another PR that make those bigger when we use `size="large"`?

I'm thinking if it wouldn't be easier to turn those chevrons into SVG so it is easier to scale them...

PR #1273, opened 2017-01-30
feat(datepicker): add keyboard navigation

divdavem:
The following keyboard shortcuts are now available:
Arrow left/right: previous/next day
Arrow up/down: previous/next week
Page up/down: previous/next month
Shift+page up/down: previous/next year
Home/end: beginning/end of the currently displayed month(s)
Shift+home/end: min/max selectable date

Note that the previous/next month and previous/next year shortcuts should probably keep the current day in the previous/next month and the current month in the previous/next year. I think this should be implemented by a change in the behavior of getNext/getPrev in ngb-calendar. This probably has to be discussed, especially with @maxokorokov who designed this API. I have opened another PR about it: #1277

### DIFF
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -7,10 +7,12 @@ import {
   OnInit,
   SimpleChanges,
   EventEmitter,
-  Output
+  Output,
+  ElementRef,
+  HostListener
 } from '@angular/core';
 import {NG_VALUE_ACCESSOR, ControlValueAccessor} from '@angular/forms';
-import {NgbCalendar} from './ngb-calendar';
+import {NgbCalendar, NgbPeriod} from './ngb-calendar';
 import {NgbDate} from './ngb-date';
 import {NgbDatepickerService} from './datepicker-service';
 import {MonthViewModel, NavigationEvent} from './datepicker-view-model';
@@ -47,7 +49,7 @@ export interface NgbDatepickerNavigateEvent {
 @Component({
   exportAs: 'ngbDatepicker',
   selector: 'ngb-datepicker',
-  host: {'class': 'd-inline-block rounded'},
+  host: {'class': 'd-inline-block rounded', '[attr.tabindex]': 'disabled ? undefined : "0"'},
   styles: [`
     :host {
       border: 1px solid rgba(0, 0, 0, 0.125);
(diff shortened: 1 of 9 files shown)


src/datepicker/datepicker.ts
```diff
@@ -271,16 +290,103 @@ export class NgbDatepicker implements OnChanges,
...
   }
 
+  @HostListener('keydown', ['$event'])
+  onKeyDown(event: KeyboardEvent) {
+    let focusedDate = this.focusedDate;
+    if (!focusedDate) {
+      return;
+    }
+    switch (event.keyCode) {
+      case 33 /* page up */:
```
maxokorokov: In `rating` and  `typeahead` we have an `enum Keys` that contains key codes (and ideally should be shared), so switch reads like:  `case Key.PageUp:`

Not sure which one I prefer though...

src/datepicker/datepicker.ts
```diff
@@ -271,16 +290,103 @@ export class NgbDatepicker implements OnChanges,
...
+  @HostListener('focus', ['$event'])
+  onFocus(event: FocusEvent) {
+    const firstDate = this.getFirstDisplayedDate();
+    const lastDate = this.getLastDisplayedDate();
+    const model = this.model;
+    this.focusedDate = (!model || model.before(firstDate) || model.after(lastDate)) ? firstDate : model;
+  }
+
+  @HostListener('blur', ['$event'])
+  onBlur(event: FocusEvent) {
```
maxokorokov: could be `'(blur)': 'focusedDate = null'` I suppose

src/datepicker/datepicker.ts
```diff
@@ -271,16 +290,103 @@ export class NgbDatepicker implements OnChanges,
...
+        this._setRelativeFocusedDate('m', 1);
         break;
     }
 
     this._updateData();
   }
 
+  @HostListener('keydown', ['$event'])
+  onKeyDown(event: KeyboardEvent) {
+    let focusedDate = this.focusedDate;
```
maxokorokov: could be `const`, or just use `this.focusedDate` altogether?

src/datepicker/datepicker.ts
```diff
@@ -310,6 +416,46 @@ export class NgbDatepicker implements OnChanges,
     }
   }
 
+  private getFirstDisplayedDate() { return this.months[0].firstDate; }
```
maxokorokov: method name should start with _

src/datepicker/datepicker.ts
```diff
@@ -310,6 +416,46 @@ export class NgbDatepicker implements OnChanges,
...
+
+  private getLastDisplayedDate() {
```
maxokorokov: method name should start with _

divdavem:
@maxokorokov Thank you for your review. I have pushed some changes following your comments.

Regarding the idea of implementing keyboard navigation as a separate service, I had actually asked the question to @pkozlowski-opensource but he had answered to keep it simple in the same file for now.
We can do it later for sure to allow customization of navigation shortcuts, but, however, I don't see the point of actually disabling keyboard navigation. Cases such as range-selection can also benefit from keyboard navigation.

divdavem:
@maxokorokov I have rebased this PR and resolved the conflict with the master branch. Do you think it could be merged soon?

maxokorokov:
As we've discussed @divdavem, we'll try to refactor this quite a bit:
- merge navigation and focusing logic, because it's essentially the same thing
- isolate it into something like `NgbDatepickerNavigationService` and remove from `NgbDatepicker` itself

maxokorokov:
Moved everything from from here in the #1551, closing this one.

PR #1277, opened 2017-01-31
feat(datepicker): keep day/month when adding/removing month/year in ngb-calendar

divdavem:
This pull request contains a change I am suggesting: I think `calendar.getNext()`/`calendar.getPrev()` with the period parameter set as `'m'`/`'y'` should not automatically return the first day of the month / the first day of the year.
I think the day in the month and the month in the year of the input date should be kept as much as possible in the returned value. In case the destination month is is not long enough to contain a day with the same number, the last day of the month should be used.

For example:
`calendar.getNext(new NgbDate(2016, 7, 22), 'm'))`
previously returned: `new NgbDate(2016, 8, 1))`
with this PR, it returns: `new NgbDate(2016, 8, 22))`

`calendar.getNext(new NgbDate(2016, 7, 22), 'y'))`
previously returned: `new NgbDate(2017, 1, 1))`
with this PR, it returns: `new NgbDate(2017, 7, 22))`

`calendar.getNext(new NgbDate(2016, 1, 30), 'm'))`
previously returned: `new NgbDate(2016, 2, 1))`
with this PR, it returns: `new NgbDate(2016, 2, 29))`
 (because 2016-02-30 is not a valid date)

Especially, I think this feature is useful for #1273 when pressing page up/page down to change the month or shift+page up/shift+page down to change the year.

@maxokorokov: What do you think?

### DIFF
--- a/src/datepicker/ngb-calendar.spec.ts
+++ b/src/datepicker/ngb-calendar.spec.ts
@@ -44,25 +44,41 @@ describe('ngb-calendar-gregorian', () => {
   });
 
   it('should add months to date', () => {
-    expect(calendar.getNext(new NgbDate(2016, 7, 22), 'm')).toEqual(new NgbDate(2016, 8, 1));
+    expect(calendar.getNext(new NgbDate(2016, 7, 22), 'm')).toEqual(new NgbDate(2016, 8, 22));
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@maxokorokov @divdavem what is the status of this one? Is it still relevant?

maxokorokov:
This commit was cherrypicked and merged in #3355

PR #1281, opened 2017-02-01
feat(datepicker): add accessibility (work in progress)

divdavem:
This PR is built on top of #1273 
It adds ARIA attributes to make the `ngb-datepicker` usable with screen readers.

I checked with the latest version of [NVDA](https://www.nvaccess.org/) on the latest version of Chrome and it works.

**This is a work in progress**
Here is the remaining work to do:
- [ ] Add some tests
- [ ] Check on other combinations of browsers / screen readers


pkozlowski-opensource:
@divdavem what is the status of this one, especially given #2165?
cc @maxokorokov @ExFlo

maxokorokov:
Will close this one as @ExFlo used it as a starting point. Thanks!

PR #1283, opened 2017-02-01
fix(time picker): Inputs too narrow for values

jpwiddy:
Widened the inputs to `size=“3”` because at 2 they were too narrow for
the minutes placeholder `MM` as well as some digit values.

Tested on my local machine after changes were made. Please see
screenshots ~

Current:
screen shot 2017-02-01 at 4 20 15 pm

New:
screen shot 2017-02-01 at 4 21 05 pm

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -86,20 +86,20 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
         </tr>
         <tr>
           <td>
-            <input type="text" class="form-control" [ngClass]="setFormControlSize()" maxlength="2" size="2" placeholder="HH"
+            <input type="text" class="form-control" [ngClass]="setFormControlSize()" maxlength="2" size="3" placeholder="HH"
               [value]="formatHour(model?.hour)" (change)="updateHour($event.target.value)"
               [readonly]="readonlyInputs" [disabled]="disabled">
           </td>
           <td>&nbsp;:&nbsp;</td>
           <td>
-            <input type="text" class="form-control" [ngClass]="setFormControlSize()" maxlength="2" size="2" placeholder="MM"
+            <input type="text" class="form-control" [ngClass]="setFormControlSize()" maxlength="2" size="3" placeholder="MM"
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@jpwiddy what is the setup (browser, OS, screen resolution) on which you are testing? I'm not seeing the problem on my end (Ubuntu, FFox / Chrome) - see the screenshot
timepicker_size
.

jpwiddy:
@pkozlowski-opensource  I'm using an HD 1920x1080 monitor, Chrome, and OSX. Thinking it's a browser issue?

pkozlowski-opensource:
> Thinking it's a browser issue?

Not sure. Let me test on more devices / OS combinations when I'm back at home.

lazarljubenovic:
From [MDN on input's `size` attribute](https://developer.mozilla.org/en/docs/Web/HTML/Element/input#attr-size):

> HTML5 simply states "the user agent should ensure that at least that many characters are visible", but different characters can have different widths in certain fonts. In some browsers, a certain string with x characters will not be entirely visible even if size is defined to at least x.

Here's a somewhat extreme example using **Linux Biolinum Keyboard O** font with the inputs (changed through dev tools, not tested).

image

I guess it gets stretched too far because font is [known for having some large glyphs](https://i.stack.imgur.com/wawu7.png). So the browser has to leave enough space even for space (ba-dum).

I have no idea _how_ the browser calculates this space, though. My guess is that it reads some data given by the font itself, but I'm surprised that Mac's default browser font has this kind of obvious issue.

wesleycho:
I would avoid `em` or `rem` since if the end user is using the opposite one, that probably will break styles.

@pkozlowski-opensource thoughts? This looks to be a pretty innocuous change to me.

lazarljubenovic:
But using pixels would break it for everyone who tries to change the font size, right?
This is use-case for `em`, I believe: It should scale along with the font-size given to the element itself.

pkozlowski-opensource:
We've merged another fix that doesn't keeps width at `2`, see: #2460
Thnx for the PR, though, we appreciate all help!

PR #1294, opened 2017-02-06
feat(pagination): Added aria label input for accessibility

thbt:

### DIFF
--- a/demo/src/app/components/pagination/demos/basic/pagination-basic.html
+++ b/demo/src/app/components/pagination/demos/basic/pagination-basic.html
@@ -1,5 +1,5 @@
 <div>Default pagination</div>
-<ngb-pagination [collectionSize]="70" [(page)]="page"></ngb-pagination>
+<ngb-pagination [collectionSize]="70" [(page)]="page" aria-label="Default pagination"></ngb-pagination>
 
 <div>directionLinks = false</div>
 <ngb-pagination [collectionSize]="70" [(page)]="page" [directionLinks]="false"></ngb-pagination>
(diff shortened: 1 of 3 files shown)


src/pagination/pagination-config.ts
```diff
@@ -15,4 +15,5 @@ export class NgbPaginationConfig {
   pageSize = 10;
   rotate = false;
   size: 'sm' | 'lg';
+  ariaLabel: string = 'Page navigation';
```
maxokorokov: no need in `:string` type

thbt:
Thanks for the review!
1. Makes sense, I removed it from all demo except config.
2. It might not change a lot for the screen readers, but we wouldn't be in line with bootstrap markup anymore if we applied the attribute on the host instead of the inner `<nav>`.

pkozlowski-opensource:
> Maybe I'm talking nonsense, but can't we simply stick [attr.aria-label] on the host element by the way, not on the inner <nav>. They won't read it properly?

@maxokorokov has a point and tbh I would go even a step further - if we would to simply forward a proerty binding to an attribute on a host... why do we need to do things at all? I mean, why not show people to simply use `<ngb-pagination aria-label="...">`?

Could we test it with screen readers and check of we need to do anything here at all?

pkozlowski-opensource:
I've just checked with ChromeVox and it seems like sticking `aria-label` on `<ngb-pagination>` has no effect... So we need to introduce a new input. Was just wondering if we can have it as `aria-label` instead of `ariaLabel` (probably not, but worth trying)?

Other thing: this PR doesn't add `<span class="sr-only">(current)</span></a>` for an active item as done in http://v4-alpha.getbootstrap.com/components/pagination/#disabled-and-active-states - should be added as part of the accessibility work for pagination.

pkozlowski-opensource:
@thbt could you please add support for the `<span class="sr-only">(current)</span>` to this PR as commented above? Ref: http://v4-alpha.getbootstrap.com/components/pagination/#disabled-and-active-states

You might also find this article useful: http://www.a11ymatters.com/pattern/pagination/

pkozlowski-opensource:
LGTM. Let's wait for the TravisCI build and if it is green I'm going to merge this one. Thnx!

PR #1299, opened 2017-02-07
docs(buttons): Add demo of using button group with Reactive forms

### LINKED ISSUE #869: Document how to use NgButtonGroup with ReactiveFormsModule
The documentation of the buttons ( **https://ng-bootstrap.github.io/#/components/buttons** ) shows how to use the radio button group with ngModel; however, Reactive Forms are quite common in Angular 2, in this case the formControlName directive is used instead of the ngModel one. The issue is that it is not clear how to set the "active" status of the buttons when using the formControlName directive.
Could you please provide an example?

dmytroyarmak:
Fix #869


pkozlowski-opensource:
Merged, thnx!

PR #1306, opened 2017-02-09
fix(rating): prevent rateChange to be fired when element is pristine

### LINKED ISSUE #1282: rateChange fires on ngbRating when element is pristine
### Bug description:
when you using the rating the rateChange event fires when the element is still pristime

[link](http://plnkr.co/edit/6SvNvxza3opm1PL8STeq?p=preview)

Angular: 2.4.5
ng-bootstrap: latest
Bootstrap: 4.0.0-alpha.6

fbasso:
Close #1282

### DIFF
--- a/src/rating/rating.spec.ts
+++ b/src/rating/rating.spec.ts
@@ -85,6 +85,33 @@ describe('ngb-rating', () => {
     expect(stars.length).toBe(3);
   });
 
+  it('handles the change event', fakeAsync(() => {
+       const fixture =
+           createTestComponent('<ngb-rating [(rate)]="currentRate" (rateChange)="onRateChange($event)"></ngb-rating>');
+       const el = fixture.debugElement;
+       const rating = el.query(By.directive(NgbRating)).children[0];
+
+       spyOn(fixture.componentInstance, 'onRateChange');
+
+       tick();
+
+       expect(fixture.componentInstance.onRateChange).not.toHaveBeenCalled();
+
+       // enter 2 -> 2/5, rate = 3
+       getDbgStar(el, 2).triggerEventHandler('mouseenter', {});
+       fixture.detectChanges();
+       tick();
+
+       expect(fixture.componentInstance.onRateChange).not.toHaveBeenCalled();
+
+       getStar(el.nativeElement, 2).click();
+       fixture.detectChanges();
+       tick();
+
+       expect(fixture.componentInstance.onRateChange).toHaveBeenCalled();
+
+     }));
+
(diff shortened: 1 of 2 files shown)


src/rating/rating.spec.ts
```diff
@@ -85,6 +85,33 @@ describe('ngb-rating', () => {
     expect(stars.length).toBe(3);
   });
 
+  it('handles the change event', fakeAsync(() => {
+       const fixture =
+           createTestComponent('<ngb-rating [(rate)]="currentRate" (rateChange)="onRateChange($event)"></ngb-rating>');
```
pkozlowski-opensource: `[(rate)]="currentRate"` is a shortcut for `[rate]="currentRate" (rateChange)="currentRate = $event"`. As such I don't think that we want to have 2 `rateChange` event handlers in this test.

More generally, IMO the simplest possible test for the issue we are trying to solve here could be:

```typescript
  it('should now fire change event initially',  fakeAsync(() => {
    const fixture = createTestComponent('<ngb-rating [rate]="3" (rateChange)="changed = true"></ngb-rating>');
    tick();
    expect(fixture.componentInstance.changed).toBeFalsy();
  }));
```

pkozlowski-opensource:
@fbasso thnx for the PR - I left one comment about a test but it actually took me long time to understand the logic of the current rating control and the fix. When I finally got what is going on I think that we could a simpler fix.

Your solution fixes the issue and does so in little code change which is good. But at the same time it introduces new state variable (`_initDone`) which is one more thing to worry about in the future.

After going through the code of the rating directive I think that the main issue is that we are using the `rate` variable to hold _both_ "current" state and "next" state (potential one, based on mouseover). This means that we can't use `rate` to do easy comparisons (since it can mean 2 different things at different times).

Given all the above I would like to propose a refactoring where we introduce explicit state for hovered value and don't change `rate` till the selection is done. This means that we can reliably use `rate` in change comparisons. I've got an alternative PR that fixes #1282 as well but without introducing new state (no need for `_initDone`). I believe that the proposed alternative is slightly simpler as it doesn't require new state and just refactors the existing code instead of introducing new code.

I've just opened #1357 with an alternative implementation - please have a look and let me know what you think and which one you would prefer to be merged.

fbasso:
@pkozlowski-opensource Agree. I focused only on the issue but it's right that using the rate attribute for both current value and mouse over create a mess. The code you proposed render the rate component easier to understand (and fix this issue at the same time).

So let's go for your changes !

PR #1310, opened 2017-02-10
added addional modal size supported by Bootstrap

squadwuschel:
added the 'md' modal size supported by Bootstrap

 - [x] added/updated any applicable tests. => not tests found for size check
 - [] added/updated any applicable API documentation. => can't find where you add it to the documentation
 - [] added/updated any applicable demos. => there are no size demos

### DIFF
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -21,7 +21,7 @@ export interface NgbModalOptions {
   /**
    * Size of a new modal window.
    */
-  size?: 'sm' | 'lg' | 'md';
+  size?: 'sm' | 'md' | 'lg';
 
   /**
    * Custom class to append to the modal window
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -21,7 +21,7 @@ export interface NgbModalOptions {
   /**
    * Size of a new modal window.
    */
-  size?: 'sm' | 'lg';
+  size?: 'sm' | 'lg' | 'md';
 
   /**
    * Custom class to append to the modal window


src/modal/modal.ts
```diff
@@ -21,7 +21,7 @@ export interface NgbModalOptions {
   /**
    * Size of a new modal window.
    */
-  size?: 'sm' | 'lg';
+  size?: 'sm' | 'lg' | 'md';
```
jnizet: It doesn't matter technically, but I would find it clearer in the documentation if the sizes were sorted: 

    'sm' | 'md' | 'lg'

But most importantly, I don't see the 'md' size documented anywhere in bootstrap. [The documentation](https://v4-alpha.getbootstrap.com/components/modal/#optional-sizes) says:

> Modals have two optional sizes, available via modifier classes to be placed on a .modal-dialog

Does the md size really exist?

squadwuschel: yes when you look in the _variabled.scss there is the size defined

squadwuschel: https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L795

jnizet: The size exists in the variables, but if I understand correctly, it isn't supposed to be used in the markup. The generated CSS is

```
@media (min-width: 576px) {
  .modal-dialog {
    max-width: 500px;
    margin: 30px auto;
  }
  .modal-sm {
    max-width: 300px;
  }
}

@media (min-width: 992px) {
  .modal-lg {
    max-width: 800px;
  }
}
```

IMHO the correct way to have the "normal" md size is to avoid specifying the size in the NgbModalOptions. Specifying an 'md' size would be equivalent in not specifying the size, in terms of the size of the rendered modal, but it would add a useless `.modal-md` CSS class to the markup.

I understand that it could be more convenient to specify an alternate size rather than not specifying it at all (or passing null or undefined), but adding a useless CSS class that isn't defined by bootstrap doesn't look right to me.

@pkozlowski-opensource what do you think?

squadwuschel: whats the best way to tell someone that there is a "third" size aviable, I would prefer to see the size in the definition, then its "clear" that there is a third size.

alex321: I also agree with **jnizet** that adding "md" is unnecessary. It's how everything works in general. Eg. all controls have a default size but you can add "lg" or "sm" classes. However, there is no "md" setting as logically that is the default.

squadwuschel: then it would be goot if you could extend the api to tell that the default value is not "sm" or "lg" that the default value is "md"

pkozlowski-opensource:
@squadwuschel thnx for the PR but I'm with @jnizet on this one - Bootstrap 4 doesn't have the `md` size but rather have a default size when none is specified. This is exactly how our API is shaped and IMO it should stay this way. Based on this we will pass on this PR. But hey, even if this PR is not merged we _do_ appreciate all contributions!

PR #1321, opened 2017-02-17
feat(typeahead) : accessibility support

thbt:
Fixes focus management and speech when using a screen reader

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -20,14 +20,14 @@ export interface ResultTemplateContext {
 @Component({
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
-  host: {'class': 'dropdown-menu', 'style': 'display: block', 'role': 'listbox'},
+  host: {'class': 'dropdown-menu', 'style': 'display: block', 'role': 'listbox', '[id]': 'id'},
   template: `
     <template #rt let-result="result" let-term="term" let-formatter="formatter">
       <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>
     </template>
     <template ngFor [ngForOf]="results" let-result let-idx="index">
       <button type="button" class="dropdown-item" role="option"
-        [attr.id]="id + '-' + idx"
+        [id]="id + '-' + idx"
         [class.active]="idx === activeIdx"
         (mouseenter)="markActive(idx)"
         (click)="select(result)">
@@ -44,7 +44,7 @@ export class NgbTypeaheadWindow implements OnInit {
    *  An optional id for the typeahead. The id should be unique.
    *  If not provided, it will be auto-generated.
    */
-  @HostBinding('attr.id') @Input() id: string = undefined;
+  @Input() id: string;
 
   /**
    * Flag indicating if the first row should be active initially
@@ -83,7 +83,7 @@ export class NgbTypeaheadWindow implements OnInit {
 
   markActive(activeIdx: number) {
     this.activeIdx = activeIdx;
-    this.activeChangedEvent.emit(this.id + '-' + this.activeIdx);
+    this._activeChanged();
   }
 
   next() {
@@ -92,7 +92,7 @@ export class NgbTypeaheadWindow implements OnInit {
     } else {
       this.activeIdx++;
     }
-    this.activeChangedEvent.emit(this.id + '-' + this.activeIdx);
+    this._activeChanged();
   }
 
   prev() {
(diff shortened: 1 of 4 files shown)


src/typeahead/typeahead-window.ts
```diff
@@ -39,6 +41,12 @@ export class NgbTypeaheadWindow implements OnInit {
   activeIdx = 0;
 
   /**
+   *  An optional id for the typeahead. The id should be unique.
+   *  If not provided, it will be auto-generated.
+   */
+  @HostBinding('attr.id') @Input() id: string = undefined;
```
pkozlowski-opensource: Could you please use 'host' instead of `@HostBinding` so all the host bindings are in one place only?

src/typeahead/typeahead-window.ts
```diff
@@ -20,16 +20,18 @@ export interface ResultTemplateContext {
...
+  host: {'class': 'dropdown-menu', 'style': 'display: block', 'role': 'listbox'},
   template: `
     <template #rt let-result="result" let-term="term" let-formatter="formatter">
       <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>
     </template>
     <template ngFor [ngForOf]="results" let-result let-idx="index">
-      <button type="button" class="dropdown-item" [class.active]="idx === activeIdx" 
-        (mouseenter)="markActive(idx)" 
+      <button type="button" class="dropdown-item" role="option"
+        [attr.id]="id + '-' + idx"
```
pkozlowski-opensource: I think that you can use `[id]="..."` instead of `[attr.id]` as buttons should have the `id` property.

src/typeahead/typeahead-window.ts
pkozlowski-opensource: There is no need for ` = undefined` - it is initialized like this by default.

src/typeahead/typeahead-window.ts
```diff
@@ -69,16 +77,22 @@ export class NgbTypeaheadWindow implements OnInit {
...
+    this.activeChangedEvent.emit(this.id + '-' + this.activeIdx);
+  }
 
   next() {
     if (this.activeIdx === this.results.length - 1) {
       this.activeIdx = this.focusFirst ? (this.activeIdx + 1) % this.results.length : -1;
     } else {
       this.activeIdx++;
     }
+    this.activeChangedEvent.emit(this.id + '-' + this.activeIdx);
```
pkozlowski-opensource: You've got this `this.activeChangedEvent.emit(this.id + '-' + this.activeIdx)` logic repeated in several places, would be good to extract it. Also, shouldn't `-1` by handled as a special case?

src/typeahead/typeahead-window.ts
pkozlowski-opensource: Also the comment seems to be invalid as we are not auto-generating id when one is not provided.

src/typeahead/typeahead.ts
```diff
@@ -124,6 +129,10 @@ export class NgbTypeahead implements ControlValueAccessor,
    */
   @Output() selectItem = new EventEmitter<NgbTypeaheadSelectItemEvent>();
 
+  @HostBinding('attr.aria-activedescendant') activeDescendant: string = undefined;
```
pkozlowski-opensource: `activeDescendant: string` is enough, no need for ` = undefined`

src/typeahead/typeahead.ts
```diff
@@ -124,6 +129,10 @@ export class NgbTypeahead implements ControlValueAccessor,
...
+  @HostBinding('attr.aria-owns') popupId: string = `ngb-typeahead-${nextId++}`;
```
pkozlowski-opensource: Please use `host` instead of `@HostBinding`

pkozlowski-opensource: No need for type (`: string`) as TS will infer it

src/typeahead/typeahead.ts
```diff
@@ -124,6 +129,10 @@ export class NgbTypeahead implements ControlValueAccessor,
...
+  @HostBinding('attr.aria-expanded') isOpen: boolean = false;
```
pkozlowski-opensource: No need for type (`: boolean`) as TS will infer it

pkozlowski-opensource:
@thbt thnx for the PR - I did the first pass on it and left some comments. Please address those and then I will have another look. Thnx!

src/typeahead/typeahead-window.ts
thbt: The `id` is actually generated in the typeahead.ts file and set when the popup is opened (see `_openPopup` method).
It's actually kinda hackish but I don't know how else I can set the `id` of the popup. If I generate it in the typeahead-window.ts file, a new `id` will be generated each time the popup is opened (since a new instance is created each time)

I applied your other comments.

PR #1343, opened 2017-02-28
refactor(carousel): remove `template[ngbSlide]` (ng4 deprecated) in favor of `ngb-slide`

somombo:
Refer to Issue https://github.com/ng-bootstrap/ng-bootstrap/issues/1344

In the above spirit, I have put together a refactor of the carousel component with the above things in mind. I've removed `<template ngbSlide>` in favor of  `<ngb-slide>`  and gotten rid of the `template` throughout the carousel code. 

It is working (demo looks fine and gulp tests have passed)

Since it is refactored in a way that works with both angular 2 and 4, my hope is that (even though this is a breaking change) we may discuss considering this (or a revised version) for PR without having to wait for stable angular 4 to go live. 

I thought I'd try PR with just the carousel component before trying to potentially contribute code for refactoring some of the other components (removing `<template>`).

I'm open to all suggestions.
What do y'all think?
Thanks!

### DIFF
--- a/demo/src/app/components/carousel/demos/basic/carousel-basic.html
+++ b/demo/src/app/components/carousel/demos/basic/carousel-basic.html
@@ -1,23 +1,23 @@
 <ngb-carousel>
-  <template ngbSlide>
-    <img src="http://lorempixel.com/900/500?r=1" alt="Random first slide">
+  <ngb-slide>
+    <img src="http://lorempixel.com/900/500?r=1" alt="Random first ngb-slide">
     <div class="carousel-caption">
-      <h3>First slide label</h3>
+      <h3>First ngb-slide label</h3>
       <p>Nulla vitae elit libero, a pharetra augue mollis interdum.</p>
     </div>
-  </template>
-  <template ngbSlide>
-    <img src="http://lorempixel.com/900/500?r=2" alt="Random second slide">
+  </ngb-slide>
+  <ngb-slide>
+    <img src="http://lorempixel.com/900/500?r=2" alt="Random second ngb-slide">
     <div class="carousel-caption">
-      <h3>Second slide label</h3>
+      <h3>Second ngb-slide label</h3>
       <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p>
     </div>
-  </template>
-  <template ngbSlide>
-    <img src="http://lorempixel.com/900/500?r=3" alt="Random third slide">
+  </ngb-slide>
+  <ngb-slide>
+    <img src="http://lorempixel.com/900/500?r=3" alt="Random third ngb-slide">
     <div class="carousel-caption">
-      <h3>Third slide label</h3>
+      <h3>Third ngb-slide label</h3>
       <p>Praesent commodo cursus magna, vel scelerisque nisl consectetur.</p>
     </div>
-  </template>
+  </ngb-slide>
 </ngb-carousel>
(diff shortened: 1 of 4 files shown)


intellix:
I personally think <ngb-slide> style should only be used by Components but this is still a Directive. I imagine it would be much easier to just rename stuff `<template ngbSlide>` to `<ng template ngbSlide>`

somombo:
@intellix I personally dont mind changing the selector  `ng-slide` to just `[ngbSlide]`. The thing I do take issue with specifically, is having the selector be `template[ngbSlide]`. This forces the use of the NgbSlide Directive only on `template` tags which I believe is problematic. 

(@pkozlowski-opensource regarding https://github.com/ng-bootstrap/ng-bootstrap/issues/1344:)
One problem with forcing the usage of template with NgbSlide is it breaks the default emulated view encapsulation, so that for example if you try to style `.carousel-item` it wont work (unless e.g.  you explicitly set encapsulation to none). As a result, if a user tries to style certain aspects of thier ng-carousel manually such as:

```typescript
@Component({
  template: `
    <ngb-carousel>
      <template ngbSlide>
        <!-- stuff-->
      </template>

    </ngb-carousel>
  `,
    //encapsulation: ViewEncapsulation.None,
  styles: [`
    .carousel-item {
        border: 3px dashed red;
        width: 75%;
        margin: 0 auto;
    }  
    .carousel-item img {
        width: 100%;
    }
  `]
})
```
This _**will not work**_ (unless encapsulation is off).  See [plunker reproducing this](http://plnkr.co/edit/QlgucVg6A2cqVQOHtVk4).

However with this PR, adding styles should work as expected even with the default view encapsulation.
Here is a screenshot of this PR working as expected when you use this PR's version of NgbSlide instead (I tried example above on the repo demo):

expected

pkozlowski-opensource:
I'm going to close this one based on the discussion in https://github.com/ng-bootstrap/ng-bootstrap/issues/1344

@somombo regarding custom classes and view encapsulation - your PR "works" only because you are using a directive in a place where a component would be more appropriate. More generally customization is an altogether different topic and we want to address it across the board.

I appreciate your point of view and contributions but at this point we are not ready to make the jump you are suggesting in this PR.

somombo:
@pkozlowski-opensource  kindly clarify yourself. *Why* is a component more appropriate?
Is a component more performant in some way? is it more secure?? Kindly clarify what you mean. By the way, my understanding is a component is simply a directive with a view. There was no need for a view here so I got it to work without using a component.

I am not seeing what big cost this PR is coming at when you dismiss it as "customization that is a different topic". Do we really need to wait till you all are in the "customization" phase of this project just so we are able to do some basic css overriding of the defaults.. e.g. changing the height of a carousel!?

It would be ideal if perhaps there was an attribute on ng-carousel that allowed changing the height (if this is what you mean by 'customization') .... but in the meantime people should just be able to do it by overriding the css, which as I've pointed out is currently impossible.

Please note that I don't mind my PR not being accepted. I mean I wont complain if its truly is for the Greater Good. I'd like to see this project succeed... Perhaps there is another PR someone else can come up with to fix this issue. But it doesn't make any sense to me that even something that's still in alfa stage should have such restrictions when the benefits are not clear.

I mean you are the leaders of the project, so if you simply dont want to fix the issue https://github.com/ng-bootstrap/ng-bootstrap/pull/1343#issuecomment-283206861 I pointed out, then that's perfectly fine.. Of course you are under no obligations.

The ngx-bootrap https://github.com/valor-software/ngx-bootstrap library is a competing to do the exact same as this one. Yet, while they haven't yet achieved every customization either, they atleast allow overriding things with css.

PR #1348, opened 2017-03-01
Move hardcoded strings in config

fbasso:


pkozlowski-opensource:
So this PR touches 2 components:
- alert;
- pagination.

It is never a good idea to do multiple changes in one PR as it makes change set harder to review and slower to merge. Anyway, the main point of this PR is to pull strings out of templates to make those configurable. While I understand the motivation I don't think this is the way to go: we should rather take advantage of the built-in Angular translations. Otherwise we are using bindings for the translations / strings configuration and this would have negative impact on the performance.

I'm going to close this PR based on the above - we want to promote Angular translations that are performant instead of ad-hoc translations through pipes / bindings.

PR #1357, opened 2017-03-03
fix(rating): don't fire change event on init

### LINKED ISSUE #1282: rateChange fires on ngbRating when element is pristine

pkozlowski-opensource:
Fixes #1282

Alternative to #1306

### DIFF
--- a/src/rating/rating.spec.ts
+++ b/src/rating/rating.spec.ts
@@ -111,6 +111,12 @@ describe('ngb-rating', () => {
     expect(getState(compiled)).toEqual([true, true, true, true, true]);
   });
 
+  it('should now fire change event initially', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


src/rating/rating.spec.ts
```diff
@@ -111,6 +111,12 @@ describe('ngb-rating', () => {
     expect(getState(compiled)).toEqual([true, true, true, true, true]);
   });
 
+  it('should now fire change event initially', fakeAsync(() => {
```
maxokorokov: now → not

PR #1362, opened 2017-03-09
Pagination: Boundary and Direction CSS Classes and Custom Text 

dopry:
I ran into an issue on a project where I needed to replace the text and style the boundary and direction controls separately from the page links, so I made a few modifications to the ngbPagination component. I've tried to adhere to the guidelines as far as I understand them. I hope this is useful, let me know if I should make any changes to get this merged. 

This PR expands on the existing markup adding additional classes to help select the direction and boundary controls from CSS. 

* .boundary-link, . first, .last
* .direction-link, .previous, .next

in addition it adds inputs to override the default text displayed for boundary and direction links.

* boundaryLinksFirst
* boundaryLinksLast
* directionLinksFirst
* directionLinksLast

 - [*] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) guide.
 - [*] built and tested the changes locally.
 - [*] added/updated any applicable tests.
 - [*] added/updated any applicable API documentation.
 - [*] added/updated any applicable demos.

### DIFF
--- a/demo/src/app/components/pagination/demos/advanced/pagination-advanced.html
+++ b/demo/src/app/components/pagination/demos/advanced/pagination-advanced.html
@@ -5,7 +5,7 @@
 <ngb-pagination [collectionSize]="120" [(page)]="page" [maxSize]="5" [rotate]="true" [boundaryLinks]="true"></ngb-pagination>
 
 <div>maxSize = 5, rotate = true, ellipses = false</div>
-<ngb-pagination [collectionSize]="120" [(page)]="page" [maxSize]="5" [rotate]="true" [ellipses]="false" [boundaryLinks]="true"></ngb-pagination>
+<ngb-pagination [collectionSize]="120" [(page)]="page" [maxSize]="5" [rotate]="true" [ellipses]="false" [boundaryLinks]="true" boundaryLinksFirst="&lt;&lt;" boundaryLinksLast="&gt;&gt;" directionLinksPrev="&lt;" directionLinksNext="&gt;" ></ngb-pagination>
 
 <hr>
 
--- a/demo/src/app/components/pagination/demos/controls-text/pagination-controls-text.html
+++ b/demo/src/app/components/pagination/demos/controls-text/pagination-controls-text.html
@@ -1,0 +1,6 @@
+<div>boundaryLinksFirst = &lt;&lt;,  boundaryLinksLast = &gt;&gt;,  directionLinksPrevious = &lt;, directionLinksNext = &gt;</div>
+<ngb-pagination  [collectionSize]="70" [(page)]="page" [boundaryLinks]="true" boundaryLinksFirst="&lt;&lt;" boundaryLinksLast="&gt;&gt;" directionLinksPrevious="&lt;" directionLinksNext="&gt;" ></ngb-pagination>
+
+<hr>
+
+<pre>Current page: {{page}}</pre>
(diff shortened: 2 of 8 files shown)


dopry:
@icfantv,  As a contributor I don't find this feedback useful or actionable. How should I respond?

icfantv:
@dopry that's fair and I probably shouldn't have done this so late at night.  I should have given more context.  One of the problems we ran into with the UIBS library (the ng1 version of this library) was that developers clamor for more functionality and configurability and as that gets added, the library becomes unmaintainable - even with the proper tests in place.

Remember that most OSS devs are unpaid volunteers and the projects usually stem from a work-related need.  Beyond that, maintaining an increasingly growing library becomes an ever larger time-sink.  Only the lucky few seem to have careers that let them get paid for doing OSS work.  The rest have to strike a balance and the OSS side seems to lose out in the end should no compromise arise.

In your case, you found a need to override default behaviors which is totally normal.  When I say extensibility, we should structure our classes so that you can extend them to add the functionality you need.  If they are not structured thus, please let us know.

I hope that was more clear.  Please let me know if you have any questions.  To be clear, I'm not suggesting what you're doing is wrong.  I'm just circling back to this project after a hiatus and perhaps we're now ok with extra configuration.  I will check.

dopry:
I can absolutely see it becoming an options bloat. I was a bit wary of that as I was implementing this particular PR, but didn't see any other recommendations for altering this type of functionality. 

In the case of the ng-bootstrap pagination control, I've limited my configuration to the strings/characters that are hard coded into the stock templates and the addition of a couple of css classes.  I feel this is a reasonable level of configuration to make available to the consumers of this component.  

Let me know if this is something the project maintainers would have any interested in merging or if there is a change to the implementation that might make more sense. I totally understand if it's a direction that ng-bootstrap doesn't want to go.

pkozlowski-opensource:
Thnx for the PR - we do appreciate all contributions although we are not going to merge this particular PR. 

The main point of this PR is to pull strings out of templates to make those configurable. While I understand the motivation I don't think this is the way to go: we should rather take advantage of the built-in Angular translations. Otherwise we are using bindings for the translations / strings configuration and this would have negative impact on the performance.

Once again, thnx for taking time to contribute!

dopry:
@pkozlowski-opensource, no worries. 

1. you would prefer to see a PR with i18n tags instead for text instead?
2. are the additional boundary and direction css classes acceptable for styling?

If the two above are true I can modify the PR to reflect how you'd want to see this implemented.

pkozlowski-opensource:
@dopry not sure what you mean by "you would prefer to see a PR with i18n tags for text instead of variables". Maybe an example could help clarify your intention here?

> are the additional boundary and direction css classes acceptable for styling

To be frank, not at this point. We want to think through / design the extensibility / customizability before moving on / accepting PRs. You are of course most welcomed to start brainstorming / design doc.

dopry:
@pkozlowski-opensource 

1. in order to use the built in i18n support you mentioned, the i18n directive needs to be added to the  elements wrapping the text in the component. see: https://angular.io/docs/ts/latest/cookbook/i18n.html 

2. currently there are no css classes that allow you to specifically select direction and boundry links in the pagination control. so if you want different colors, font sizes for the different links it is not possible.

I understand. Unfortunately I don't have the bandwidth for that kind of communal specification process, but good luck.

PR #1366, opened 2017-03-10
fix(datepicker): allow datepicker calendar override

maxokorokov:
- `NgbCalendar` export was missing
- `NgbDatepickerService` should not be global, but per-datepicker as it is using `NgbCalendar` itself


pkozlowski-opensource:
Could we add a test where we have a component that uses a custom calendar + i18n? It could be a separate test file with bigger / integration tests.

maxokorokov:
@pkozlowski-opensource added `datepicker-integration.spec.ts`. Something like this?

pkozlowski-opensource:
LGTM

alexluecke:
This fix breaks the ability to have custom providers for the DatepickerService. Was this intentional?

maxokorokov:
There was no intention to expose publicly the `NgbDatepickerService` for now. I don't think it is exported by `NgbDatepickerModule` at all.

alexluecke:
How would I generate a custom view for something like "don't show weeks that contain the 14th day of the month" ? Seems like it would be nice to be able to modify what days actually gets rendered by a particular month.

Issue #1161, opened 2016-12-25
[ngbTabset] add option to hide inactive tabs instead of destroying them


Eugeny:
Feature proposal: add a NgbTabset subclass where the template would bind to `visibility` instead of `*ngIf` so that inactive tab contents won't get destroyed when switching.

This makes sense for tabs that contain components which are expensive to re-instantiate or contain state that gets lost, for example a video player.

mattlewis92:
Cont from #1167 - sorry about the duplicate!

@pkozlowski-opensource how come you're on the fence about this feature request? Are there any better solutions you can see to allow routing in tabs?

thbt:
It also makes sense regarding accessibility, according to the WAI-ARIA spec : 
> When a tabbed interface is initialized, one tab panel is displayed and its associated tab is styled to indicate that it is active. When the user activates one of the other tab elements, the previously displayed tab panel is hidden, the tab panel associated with the activated tab becomes visible, and the tab is considered "active".

jw61:
Definite up-vote for this feature! I have a requirement for users to enter different types of data on different tabs. If they change tab they lose everything they've entered, but I can't force the user to correct any validation errors before they're allowed to change tabs.

Thommas:
We also need this feature.

How can we help ?

jiayihu:
I'd like to make a PR if there is confirm that this behaviour is something acceptable for the team of `ng-bootstrap`. Destroying components/DOM is causing some issues whenever the content of the tab/panel is something not trivial, such as a form, because you loose the state.

AlexKhymenko:
Super useful. I have use case when you should add green or red circle depending if form inside of tab valid or not. Can't do it on initial load if there is no component :-)

Issue #1167, opened 2016-12-29
Cannot use the router outlet inside of a tab


mattlewis92:
### Bug description:
Placing a router-outlet inside of an ngbTab works on the first load, but then if you change the tab and then switch the tab back again the outlet content is gone.

The issue appears to be related to the ngIf used [here](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/tabset/tabset.ts#L92) which by design is incompatible with the router: https://github.com/angular/angular/issues/11100

The only obvious solution I can see to this is an option to use `[hidden]="true"` instead of the ngIf i.e. by default the ngIf will still be used to preserve backwards compatibility (and keep the performance benefit), but the new option would only hide the content e.g.

`*ngIf="!keepInDom && tab.id === activeId" [hidden]="keepInDom && tab.id !== activeId"`

The `[keepInDom]="true"` option could probably use some bikeshedding.

If you're happy with the implementation, I will send a PR that implements it.

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/7uYe1ymKOGJ4OHV3JpJa?p=preview

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 2.4.1

ng-bootstrap: 1.0.0-alpha.15

Bootstrap: 4

pkozlowski-opensource:
This is roughly duplicate of #1161 (that is, #1161 is about hiding DOM of panels instead of destroying).

TBH I'm on the fence when it comes to this feature request - for sure it would require design, Let's centralize discussion in  #1161

PR #1332, opened 2017-02-23
fix(accordion): comply with bootstrap markup

thbt:
fix #1050

Removed similar IDs from demos

Before submitting a pull request, please make sure you have at least performed the following:

 - [x] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) guide.
 - [x] built and tested the changes locally.
 - [x] added/updated any applicable tests.
 - [x] added/updated any applicable API documentation.
 - [x] added/updated any applicable demos.

PR #1370, opened 2017-03-11
feat(accordion): support nodes preservation

jiayihu:

Fixes https://github.com/ng-bootstrap/ng-bootstrap/issues/1161 and https://github.com/ng-bootstrap/ng-bootstrap/issues/1167 
In this case I left `hidden` because there are no examples of `accordion` component in Bootstrap 4 examples for `Card` and `active` would require writing custom CSS to hide the panel

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -104,18 +104,22 @@ export interface NgbPanelChangeEvent {
     <template ngFor let-panel [ngForOf]="panels">
       <div role="tab" id="{{panel.id}}-header" [attr.aria-selected]="panel.focused"
         [class]="'card-header ' + (panel.type ? 'card-'+panel.type: type ? 'card-'+type : '')" [class.active]="isOpen(panel.id)">
-        <a href (click)="!!toggle(panel.id)" (focus)="panel.focused = true" 
-          (blur)="panel.focused = false" [class.text-muted]="panel.disabled" 
+        <a href (click)="!!toggle(panel.id)" (focus)="panel.focused = true"
+          (blur)="panel.focused = false" [class.text-muted]="panel.disabled"
           [attr.aria-expanded]="isOpen(panel.id)" [attr.aria-controls]="panel.id">
           {{panel.title}}<template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></template>
         </a>
       </div>
-      <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'" class="card-block" *ngIf="isOpen(panel.id)">
+      <div id="{{panel.id}}" role="tabpanel"
+        *ngIf="!destroyOnHide || isOpen(panel.id)"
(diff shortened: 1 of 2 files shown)


src/accordion/accordion.ts
```diff
@@ -199,6 +207,14 @@ export class NgbAccordion implements AfterContentChecked {
    */
   isOpen(panelId: string): boolean { return this._states.get(panelId); }
 
+  /**
+   * @internal
```
pkozlowski-opensource: Using `@internal` will make it fail in the AoT mode, please remove. I think that it would be better to just in-line this logic in a template and IMO we should just remove this method altogether.

pkozlowski-opensource:
@jiayihu thnx - it looks good, generally speaking but there is one change I need you to do before we can merge this one - please see comments in the code.

src/accordion/accordion.ts
jiayihu: Are you sure about `@internal`? I believe AOT has problems with `private` methods/fields used in the template. Note then that `isOpen` also has `@internal` comment.

About inline the logic I completely agree but there's currenctly a strange behaviour where `null` is converted into `false` and the attribute is the added `hidden="false"`, which is selected by Boostrap CSS selector `[hidden] {...}`.
I found a similar [issue](https://github.com/angular/angular/issues/8960).

pkozlowski-opensource: Was pretty sure about the `@internal` but then noticed that it is used on `isOpen`... Need to investigate.

In any case let's try to in-line this method (and yes, you need to return null if you want to remove the attribute altogether).

jiayihu: > In any case let's try to in-line this method (and yes, you need to return null if you want to remove the attribute altogether).

The problem is that with inline `null` it becomes `false` and the attribute is not removed from DOM, it's added instead. That's the reason of a method. Personally I don't understand why result is different from methods vs inline.

pkozlowski-opensource: @jiayihu once again, the main problem with the `hidden` attribute is that ... you shouldn't use it at all!!! Please use appropriate Bootstrap 4 CSS classes instead, following Bootstrap's markup as documented officially: https://v4-alpha.getbootstrap.com/components/collapse/#accordion-example

BTW, we might need to merge #1332 first.

jiayihu: > In this case I left hidden because there are no examples of accordion component in Bootstrap 4 examples for Card and active would require writing custom CSS to hide the panel

I didn't find the example :P Anyway I'll make the proper changes then, thanks for pointing the example.

jiayihu:
@pkozlowski-opensource I replaced `hidden` with `show` but it won't work without `collapse` CSS class added by #1332 .

pkozlowski-opensource:
Merged, thnx! Sorry it took so long, been busy with preparing 1.0 release.

Once again, thnx, appreciate all the PRs!

PR #760, opened 2016-09-17
feat(datepicker): initial version of Hijri calendar

AhmedMustafa:
The initial version of Hijri calendar support.

refs #656

Issue #656, opened 2016-09-01
Add Hijri & Hebrew Calendar support to datepicker


icfantv:
The datepicker needs to have Hijri Calendar support.  We should not support the other three Arabic calendars.
- The Hijri Calendar has days, weeks, months, and years.
- The new year comes 11 days earlier every year compared to the Gregorian Calendar so there's a shift over time.
- There are the same number of months as the Gregorian Calendar
- The Higri Calendar reads RTL with the Arabic equivalent of Saturday being the first day of the week.
- We need abbreviated days of the week.
- We need abbreviated months of the year.

icfantv:
@AhmedMustafa, I know this was a request of yours for UIBS and we recognize the importance of getting it into the ng2 port so we would love your feedback and assistance here.  Can you please review the issue bullets and add, via comments, anything that would help correct/clarify?  Thank you.  E.g.,

Is it more proper to utilize Arabic abbreviations of days of the week and month names or use abbreviations of transliterated names?

wesleycho:
Is there a reason we don't want to support the other calendars? Is there something special about them?

icfantv:
@wesleycho the way our i18n (Localization) department explained it to me was that the bulk of the Arab world watches one channel (station?) called MBC and this is the calendar format that they (the TV station) use.  If anyone else chimes in with a _legitimate_ reason to be supporting the other calendars, it may be worth revisiting.  But for now, I say just stick with the Hijri calendar.

AhmedMustafa:
Hijri is a general term used to describe the Islamic lunar-based calendar. There is no specific Islamic calendar named Hijri (all are Hijri calendar).
However, there are many types of Hijri/Islamic calendars used in the Islamic countries ([Calendar Types](http://cldr.unicode.org/development/development-process/design-proposals/islamic-calendar-types) ). All have the following characteristics:
- All have days, weeks, months, and years.
- All have a fixed 12 months per year
- Each month has 29/30 days (no more, no less)
- Islamic year is about 354/355 days
- The first day of the week differs from country to another. It is a country-related not calendar-related property.

**Islamic(arithmetic), Islamic-civil, Islamic-tabular & Islamic-umalqura** all are different types of Hijri calendar. Many of them are implemented in a plenty of technologies such as Java, Dojo, ibm-js, jQuery, Globalize,..etc. Each calendar is using a different type of calculations (Astronomical, Table-Based,...). 
[...]

PR #1374, opened 2017-03-13
feat(datepicker): add Hijri calendar support

maxokorokov:
As discussed with @AhmedMustafa it is built on top of #760, which is all his work

I just:
- did a couple of minor fixes and made it work with injection
- added calendar demo

Part of #656


maxokorokov:
Edge looks very unstable. Had to relaunch build a couple of times, and now its green even though not all the tests were executed...

maxokorokov:
Hey, @icfantv It's nothing special yet, just different numbers, as only calendar implementation was replaced.

screen shot 2017-03-14 at 10 13 55

PR #1379, opened 2017-03-15
Demo AoT build fixes

pkozlowski-opensource:
While releasing alpha.21 I've bumped into AoT errors with the freshly added datepicker calendar demo. After the investigation it turned out to be a missing `export` and it was easy to fix. The real problem, thought, was that there was nothing that warned us about the fact that our demo page was broken in the AoT mode.

This PR fixes the AoT issue and adds demo build process to our CI so we can catch those errors as early as possible. Additionally it fixes an issue where the `test` task was exiting process too early without waiting for test coverage reports to be generated. This was causing issue with any tasks executed after `test` - those tasks were executed too early, while tests were still running.

@maxokorokov since you've added the datepicker demo _and_ were playing with Karma setup, could you please review this one? Ping me if you've got any questions.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,3 +1,4 @@
+var asyncDone = require('async-done');
 var gulp = require('gulp');
 var gutil = require('gulp-util');
 var ddescribeIit = require('gulp-ddescribe-iit');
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -283,3 +285,5 @@ gulp.task(
     'deploy-demo', function(done) { runSequence('clean:demo', 'build:demo', 'demo-push', 'clean:demo-cache', done); });
 
 gulp.task('default', function(done) { runSequence('lint', 'enforce-format', 'ddescribe-iit', 'test', done); });
+
+gulp.task('ci', function(done) { runSequence('default', 'build:demo', done); });
```
maxokorokov: So you'll have to tell travis to run this one afterwards, right?

pkozlowski-opensource: Oh man, it should be part of this PR. Good catch, thnx. Will add and merge if Travis gets green.

PR #1390, opened 2017-03-20
chore: angular 4 compatibility

eddy-geek:

run `npm test` but I got " Executed 0 of 0 " (maybe because I don't have SauceLabs?)

Maybe not mergeable but could still be useful to someone else ;-)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,19 +23,19 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/core": "2.3.1",
-    "@angular/common": "2.3.1",
-    "@angular/forms": "2.3.1"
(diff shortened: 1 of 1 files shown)


wesleycho:
It should be noted that the package.json for the distribution bundle is generated, so modifying this package.json will do nothing for Angular 4 compatibility.

That said, ng-bootstrap uses peerDependencies, so this PR is invalid as far as user consumption is concerned. If you are on npm 2 & are encountering this problem, then you must upgrade to npm 3+. Angular itself does not support npm 2 from what I understand.

pkozlowski-opensource:
Agree with @wesleycho - on top of this we don't want to make ng-bootstrap incompatible with Angular 2.x, especially that Angular 4.x is not yet out. But then again, we've got just peerDependencies in our package.json so people can freely choose if they want to use 2.x or 4.x.

eddy-geek thnx for the PR but it is invalid based on the above comments.

PR #1400, opened 2017-03-23
fix(rating): clarify priorities of custom star templates

pkozlowski-opensource:
This also clarifies behaviour for Angular 4.0.0, see:
https://github.com/angular/angular/issues/15417

### DIFF
--- a/src/rating/rating.spec.ts
+++ b/src/rating/rating.spec.ts
@@ -274,6 +274,17 @@ describe('ngb-rating', () => {
     expect(getStateText(compiled)).toEqual(['x', 'x', 'o', 'o']);
   });
 
+  it('should give priority to a custom star template as property binding when both are present', () => {
+    const fixture = createTestComponent(`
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@maxokorokov could you please LGTM this one for me? The main motivation for this change is that Angular 4.x introduces subtle changes that might break our code so it is better to clarify what are priorities of various ways of providing a custom template.

See more info in https://github.com/angular/angular/issues/15417 and https://github.com/angular/angular/pull/15429

src/rating/rating.spec.ts
```diff
@@ -274,6 +274,17 @@ describe('ngb-rating', () => {
     expect(getStateText(compiled)).toEqual(['x', 'x', 'o', 'o']);
   });
 
+  it('should give priority to a custom star template as property binding when both are present', () => {
```
maxokorokov: There is already a test doing this. The one just below

pkozlowski-opensource:
Seems like the issue on the angular side was fixed via https://github.com/angular/angular/pull/15429 so we can drop this PR for now.

Issue #1356, opened 2017-03-03
ngb-carousel Slide event


shabbir-dhangot:
### Bug description:
ngb-carousel Slide event. I am trying to get it with 
`<ngb-carousel [interval]="slideInterval" (on-carousel-change)="onSlideChanged()">` But not working

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular:
2.3.1
ng-bootstrap:
1 alpha 20
Bootstrap:
4 beta 6

pkozlowski-opensource:
See #1331

Issue #1331, opened 2017-02-23
carousel expose an event of index change


byunru:
suggest expose the carousel event for carousel when been navigated to a new image

pkozlowski-opensource:
Could you share your real-life use-case where such event would be helpful? This would allow us to properly prioritize this feature request.

pkozlowski-opensource:
I'm going to close this one as we haven't heard back from @byunru and I'm not keen on extending API surface without a concrete use-case.

byunru:
pkozlowski, sorry, I didn't get notice of this.
the scenario is when I click left/right arrow button,  the content outside of the carousel need to updated accordingly based on which image currently show.
I guess that expose an event only need few lines of code + document.

Thanks.

pkozlowski-opensource:
@byunru feel free to send a PR if you care about adding this functionality. Thnx!

lauraiurb:
I have a situation that I need that functionality, I'm using carousel with charts.
The first chart is ok but the other charts does not show if I don't reload it when is in the active slide. So I need that event for refresh the charts.
I will use ngx-bootstrap because I found th way to get the slide index with "activeSlide"

pkozlowski-opensource:
@lauraiurb huh? We've added the `(slide)` event in #1406 like... almost a year ago... Check documentation for `NgbSlideEvent` and note that you can access all information about the current and next slide: https://ng-bootstrap.github.io/#/components/carousel/api

lauraiurb:
@pkozlowski-opensource 
Ohhh, I understand now, I read it before but i didn't get it. Thanks a lot! I will use it

mafia99:
Hi I am new to angular so I have no Idea how to use this slide event. would be great if I can see a working example.

Krunal-Gosalia:
Even I would like to see working example of it. If someone can post - it will be really helpful.

4i-helpynet:
Hi all

EXAMPLE:

1: in .html:  
<ngb-carousel *ngIf="images" (slide)="onSlide($event)"> 

2: in .ts
public onSlide(slideData) {
	  console.log('SLIDE:',slideData);
// your JS here
}

KDSGYT:
Thank you so much for the help....

PR #1406, opened 2017-03-25
feat(carousel): add event on carousel slide with direction info

giampierobono:

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -7,7 +7,9 @@ import {
   Input,
   OnDestroy,
   AfterContentChecked,
-  OnInit
+  OnInit,
+  Output,
+  EventEmitter
 } from '@angular/core';
 import {NgbCarouselConfig} from './carousel-config';
 
@@ -43,7 +45,8 @@ export class NgbSlide {
   },
   template: `
     <ol class="carousel-indicators">
-      <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId" (click)="cycleToSelected(slide.id)"></li>
+      <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId" 
+          (click)="cycleToSelected(slide.id, _getSlideEventDirection(activeId, slide.id))"></li>
     </ol>
     <div class="carousel-inner">
       <div *ngFor="let slide of slides" class="carousel-item" [class.active]="slide.id === activeId">
@@ -85,6 +88,12 @@ export class NgbCarousel implements AfterContentChecked,
    */
   @Input() activeId: string;
 
+  /**
+   * A carousel slide event fired when the slide transition is completed.
+   * See NgbSlideCarouselEvent for payload details
+   */
+  @Output() slide = new EventEmitter<NgbSlideCarouselEvent>();
+
   constructor(config: NgbCarouselConfig) {
     this.interval = config.interval;
     this.wrap = config.wrap;
@@ -104,7 +113,7 @@ export class NgbCarousel implements AfterContentChecked,
    * Navigate to a slide with the specified identifier.
    */
   select(slideId: string) {
-    this.cycleToSelected(slideId);
+    this.cycleToSelected(slideId, this._getSlideEventDirection(this.activeId, slideId));
     this._restartTimer();
   }
 
@@ -134,13 +143,16 @@ export class NgbCarousel implements AfterContentChecked,
    */
   cycle() { this._startTimer(); }
 
-  cycleToNext() { this.cycleToSelected(this._getNextSlide(this.activeId)); }
+  cycleToNext() { this.cycleToSelected(this._getNextSlide(this.activeId), NgbSlideCarouselEventDirection.LEFT); }
 
-  cycleToPrev() { this.cycleToSelected(this._getPrevSlide(this.activeId)); }
+  cycleToPrev() { this.cycleToSelected(this._getPrevSlide(this.activeId), NgbSlideCarouselEventDirection.RIGHT); }
 
-  cycleToSelected(slideIdx: string) {
+  cycleToSelected(slideIdx: string, direction: NgbSlideCarouselEventDirection) {
     let selectedSlide = this._getSlideById(slideIdx);
     if (selectedSlide) {
+      if (selectedSlide.id !== this.activeId) {
+        this.slide.emit({prev: this.activeId, current: selectedSlide.id, direction: direction});
+      }
       this.activeId = selectedSlide.id;
     }
   }
@@ -196,6 +208,43 @@ export class NgbCarousel implements AfterContentChecked,
     return isFirstSlide ? (this.wrap ? slideArr[slideArr.length - 1].id : slideArr[0].id) :
                           slideArr[currentSlideIdx - 1].id;
   }
+
+  private _getSlideEventDirection(currentActiveSlideId: string, nextActiveSlideId: string):
+      NgbSlideCarouselEventDirection {
+    const currentActiveSlideIdx = this._getSlideIdxById(currentActiveSlideId);
+    const nextActiveSlideIdx = this._getSlideIdxById(nextActiveSlideId);
+
+    return currentActiveSlideIdx > nextActiveSlideIdx ? NgbSlideCarouselEventDirection.RIGHT :
+                                                        NgbSlideCarouselEventDirection.LEFT;
+  }
+}
+
+/**
+* The payload of the slide event fired when the slide transition is completed
+*/
+export interface NgbSlideCarouselEvent {
+  /**
+   * Previous slide id
+   */
+  prev: string;
+
+  /**
+   * New slide ids
+   */
+  current: string;
+
+  /**
+   * Slide event direction
+   */
+  direction: NgbSlideCarouselEventDirection;
+}
+
+/**
+ * Enum to define the carousel slide event direction
+ */
+export enum NgbSlideCarouselEventDirection {
+  LEFT = <any>'left',
+  RIGHT = <any>'right'
 }
 
 export const NGB_CAROUSEL_DIRECTIVES = [NgbCarousel, NgbSlide];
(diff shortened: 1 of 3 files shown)


giampierobono:
Don't understand why the build fails...in local it works

src/carousel/carousel.ts
```diff
@@ -196,6 +212,48 @@ export class NgbCarousel implements AfterContentChecked,
...
+   */
+  direction: NgbSlideCarouselEventDirection;
+}
+
+/**
+ * Enum to define the carousel slide event direction
+ */
+export enum NgbSlideCarouselEventDirection {
+  LEFT = <any>'left',
+  RIGHT = <any>'right'
```
wesleycho: Why type these to `any` and not make use of the implicit string typing?

wesleycho:
The build failure is some flakiness with Edge testing in Sauce Labs - that is to be looked into.

This PR looks mostly good to me.

src/carousel/carousel.ts
giampierobono: As far as I know, you cannot assign directly a string to an enum value. In fact if I remove <any>, typescript says: type string is no assignable to type NgbSlideCarouselEventDirection.
Is there another way to do it?

wesleycho: Looks like you should just be able to do
```ts
export enum NgbSlideCarouselEventDirection {
  left,
  right
}
```
and use it like `NgbSlideCarouselEventDirection.left` to represent the string value `left`

src/carousel/carousel.ts
```diff
@@ -141,6 +149,14 @@ export class NgbCarousel implements AfterContentChecked,
   cycleToSelected(slideIdx: string) {
     let selectedSlide = this._getSlideById(slideIdx);
     if (selectedSlide) {
+      if (selectedSlide.id !== this.activeId) {
+        this.onCarouselSlide.emit({
+          type: 'slide',
+          prevSlideId: this.activeId,
+          newSlideId: selectedSlide.id,
```
wesleycho: Maybe these can be shortened to `prev` and `current`.

src/carousel/carousel.ts
giampierobono: but doing like this NgbSlideCarouselEventDirection.left and NgbSlideCarouselEventDirection.right values won't be 0 and 1 instead of 'left' and 'right'?

src/carousel/carousel.ts
giampierobono: Totally agree. I will change them. Thanks for the suggestion :)

src/carousel/carousel.ts
```diff
@@ -196,6 +212,48 @@ export class NgbCarousel implements AfterContentChecked,
     return isFirstSlide ? (this.wrap ? slideArr[slideArr.length - 1].id : slideArr[0].id) :
                           slideArr[currentSlideIdx - 1].id;
   }
+
+  private _getSlideEventDirection(currentActiveSlideId: string, nextActiveSlideId: string):
+      NgbSlideCarouselEventDirection {
+    const currentActiveSlideIdx = this._getSlideIdxById(currentActiveSlideId);
+    const nextActiveSlideIdx = this._getSlideIdxById(nextActiveSlideId);
+
+    return currentActiveSlideIdx > nextActiveSlideIdx ? NgbSlideCarouselEventDirection.RIGHT :
```
mdoing: I would expect that when you are on the first slide and click the back arrow that the direction would be RIGHT. However, when you are on the first slide with this logic it will always return LEFT, regardless of which direction you are moving in. The same thing applies if you are on the last slide, the direction will always be RIGHT, regardless of which direction you move in.

giampierobono: You are totally right! Thanks a lot! 

I have just updated my PR. Now the logic is simpler, basically when you click on prev is always RIGHT and when you click on next is always LEFT. The computation based on the index is now done only clicking on the indicators.

src/carousel/carousel.ts
```diff
@@ -43,7 +45,7 @@ export class NgbSlide {
   },
   template: `
     <ol class="carousel-indicators">
-      <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId" (click)="cycleToSelected(slide.id)"></li>
+      <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId" (click)="select(slide.id)"></li>
```
pkozlowski-opensource: This changes the behaviour in subtle way: now we would reset timer on click. Was it intentional? If so we should add a test for it.

src/carousel/carousel.ts
```diff
@@ -85,6 +87,12 @@ export class NgbCarousel implements AfterContentChecked,
    */
   @Input() activeId: string;
 
+  /**
+   * A carousel slide event fired when the slide transition is completed.
+   * See NgbSlideCarouselEvent for payload details
+   */
+  @Output() onCarouselSlide = new EventEmitter<NgbSlideCarouselEvent>();
```
pkozlowski-opensource: can we name it just `slide`? We are in the carousel context so no need to add the `Carousel` in the name. Similarly we know that this is an event so no need for `on` (we are not adding it for other directives / events).

package.json
```diff
@@ -27,8 +27,8 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/core": "2.3.1",
     "@angular/common": "2.3.1",
+    "@angular/core": "2.3.1",
```
pkozlowski-opensource: Why this change? While it doesn't hurt it always best idea to avoid unrelated changes in a single PR.

demo/src/app/components/carousel/carousel.component.ts
```diff
@@ -8,6 +8,7 @@ import {DEMO_SNIPPETS} from './demos';
       <ngbd-api-docs directive="NgbCarousel"></ngbd-api-docs>
       <ngbd-api-docs directive="NgbSlide"></ngbd-api-docs>
       <ngbd-api-docs-config type="NgbCarouselConfig"></ngbd-api-docs-config>
+      <ngbd-api-docs-class type="NgbSlideCarouselEvent"></ngbd-api-docs-class>
```
pkozlowski-opensource: I don't think this new functionality warrants a separate demo - 2 carousel side by side already look wired on our demo page so I would prefer adding more demos unless those really add a lot of value.

pkozlowski-opensource:
@giampierobono thnx, it looks good generally speaking but I've left some comments that I would love to see addressed before we merge this one. Thnx again!

demo/src/app/components/carousel/carousel.component.ts
giampierobono: Isn't it the way to update the documentation adding specific information about this new event?

pkozlowski-opensource: Documentation is generated automatically from the source code - as long as you've got JS doc on the `@Output` it will make it to the demo site.

src/carousel/carousel.ts
thbt: What about using an union type for the value (`'left' || 'right'`) instead of an enum ?

pkozlowski-opensource:
@giampierobono are you planning to do further work on this PR? If so could you please remove the demo part and address all the remaining comments so I could merge this PR? Thnx!

giampierobono:
@pkozlowski-opensource I have just reverted the demo changes and I have already applied all the other changes in comments. I was waiting for your reply to my comment about the documentation. 

The PR should be ok now. Let me know if something is still missing.

Thanks again.

src/carousel/carousel.ts
```diff
@@ -196,6 +208,48 @@ export class NgbCarousel implements AfterContentChecked,
...
+}
+
+/**
+* The payload of the slide event fired when the slide transition is completed
+*/
+export interface NgbSlideCarouselEvent {
+  /**
+   * Event type
+   */
+  type: 'slide';
```
pkozlowski-opensource: Why do we need `type`? Do we have other types and we need to distinguish between those? Isn't interface enough?

giampierobono: When I have done this the first time, I have checked the bootstrap carousel doc and there are 2 events: slide and slid (the first is for transition start and the second for transition end if I remember correctly) and the initial plan was to reproduce the same behavior, that's why we have type. 
Need to add the slid event but I had not time till now, so I have pushed this. I can remove it if you want, I agree that since by now there is only one event it's not necessary.

pkozlowski-opensource: Yeh, let's remove it. We should rather have 2 different event interfaces rather than trying to discriminate those by a string.

giampierobono: of course 2 different events and not only one discriminated by strings :) it's the bs event structure having the 'type' prop so I have reproduced it. I am going to update the PR

pkozlowski-opensource:
@giampierobono I was about to merge this PR but then I had a closer look and started to ask myself the following question: what is the use-case for exposing slide direction in the new event? Do you have a specific need? AFAIK it wasn't explicitly asked for in any of the prior issues (#1331, #1356).

If you don't have specific use-case I would prefer that we leave off this functionality as it adds a bit of code and currently people can't even visually see slide direction as we don't support animations (!)

giampierobono:
@pkozlowski-opensource sorry I have just seen your message.

For me one use-case is keep in sync the carousel and some panels below it (with text adding more details to each carousel image), that will be displayed or not, following the event direction without relying only on the image id. Or another use case could be bind to the event and perform an action only if the direction is "right" or "left" and not both.

pkozlowski-opensource:
@giampierobono sure, I can also _imagine_ use cases - I was rather asking if you, or any users that you know, already _has_ a specific use-case :-)

giampierobono:
For something I am working on, the second use-case is the one I would need.

kesarion:
Is something still blocking this pull from being merged?

jspizziri:
@pkozlowski-opensource,

Is there further work on this PR before it get's merged in? I have a couple of use-cases that this would be greatly helpful for.

Thanks!

PR #1408, opened 2017-03-26
refactor(modal): setting NgbModalRef as a generic type

jayoungers:
Modal implementation code goes from:

        let modal = this.modalService.open(MyModalComponent);
        (modal.componentInstance as MyModalComponent).sampleInput= 'Sample Value';

to:

        let modal = this.modalService.open(MyModalComponent);
        modal.componentInstance.sampleInput= 'Sample Value';

### DIFF
--- a/src/modal/modal-ref.ts
+++ b/src/modal/modal-ref.ts
@@ -1,4 +1,4 @@
-import {Injectable, ComponentRef} from '@angular/core';
+﻿import {Injectable, ComponentRef} from '@angular/core';
 
 import {NgbModalBackdrop} from './modal-backdrop';
 import {NgbModalWindow} from './modal-window';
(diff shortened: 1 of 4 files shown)


wesleycho:
I do not think this is correct - it should be noted that the `open` method takes more than just components. It also takes in strings of plain text and a template ref as well I believe, so any typing with templates needs to take this into account.

jayoungers:
I agree it's a little awkward since the method accepts and can do so many things. It may make sense to make a specific 'openComponent()' method

wesleycho:
I think separating opening for the various types might be a good idea - @pkozlowski-opensource thoughts?

pkozlowski-opensource:
@wesleycho is right, we need to account for all the other argument types (TemplateRef, string). And no, I don't think we are ready to create separate `open` methods at this point. As such I'm going to close this PR. Having said this I would love to see better type flowing here, so if you could poke around and come up with a solution that covers all the existing arguments, please let us know!

Thnx for taking time to contribute - even if this PR is not merged we do appreciate all interactions and this one brought our attention the the pb I've missed previously....

PR #1421, opened 2017-03-30
fix(pagination): swapped out disabled anchor for span

thbt:
disabled pagination is not focusable by tab anymore
fix a bug where it would be possible to select a page when the pagination was disabled
improve accessibility 


pkozlowski-opensource:
Let's keep it consistent with accordions and tabs where we take different approach to marking disabled links. We can review this decision after the first pass on validation.

PR #1425, opened 2017-04-02
feat(datepicker): initial version of Jalali calendar

mehrabisajad:
Add jalali calendar for datepicker.

The Jalali calendar is a solar calendar that was used in Persia, variants of which today are still in use in Iran as well as Afghanistan.
Site : https://en.wikipedia.org/wiki/Jalali_calendar


src/datepicker/jalali/ngb-calendar-persian.ts
```diff
@@ -0,0 +1,263 @@
+/**
+ * Created by mehrabisajad on 2017/03/28.
+ */
+import {NgbDate} from '../ngb-date';
+import {NgbPeriod} from '../ngb-calendar';
+import {Injectable} from '@angular/core';
+import {NgbCalendarJalali} from './ngb-calendar-jalali';
+
+function mod(a: number, b: number): number {
```
wesleycho: Not sure I like the function name `mod` for this - in mathematics, `mod` is typically used to refer to the modulus operation, for which the closest analog in JS is the `%` operator.

src/datepicker/jalali/ngb-calendar-jalali.ts
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+    date.day = day;
+    return date;
+  }
+
+  setMonth(date: NgbDate, month: number): NgbDate {
+    month = +month;
+    date.year = date.year + Math.floor((month - 1) / 12);
+    date.month = Math.floor(((month - 1) % 12 + 12) % 12) + 1;
```
wesleycho: Couldn't you move the `+ 12` inside the first parenthetical so that you only would need one execution of `% 12`?

mehrabisajad: i think it needs to be like this to calculate negative "month". although my reference to this code is from : 
/src/datepicker/hijri/ngb-calendar-hijri.ts#L39-L44

maxokorokov:
@mehrabisajad thanks for the PR, unfortunately calendars were not a priority for a long time.

I didn't want to force-push to your fork's `master`, so I opened a new PR #2572 with cherry-picking and rebasing your commits.

Also the changes from your PR:
- I put all static functions in `jalali.ts` removing `NgbJalaliCalendar` as it is not really required currently from what I see
- I put `NgbDatepickerI18nPersian` in the demo only for now, not exporting it publicly from the library. We'll do it later if necessary.

Thanks again for contributing!

PR #1452, opened 2017-04-07
feat(demo): add fragment links to doc

thbt:


pkozlowski-opensource:
@thbt this needs a bit more work:
* CI build is failing on lint, needs fixing: https://travis-ci.org/ng-bootstrap/ng-bootstrap/builds/219690818
* there is quite a bit of code duplication, could you think of minimising it?

thbt:
Hi @pkozlowski-opensource I made changes that should fix the build and reduce code duplication (I could reduce it further with inheritance if you want)

I also added an svg icon file distributed under CC-BY license ([link to project](https://dribbble.com/shots/1934932-77-Essential-Icons-Free-Download)), but I don't know where I should add the credits to the author, which file should I modify ?

demo/src/app/components/shared/api-docs/api-docs-class.component.ts
```diff
@@ -19,13 +22,30 @@ import {Analytics} from '../../../shared/analytics/analytics';
...
+      display: inline-block;
+    }
     `
   ]
 })
-export class NgbdApiDocsClass {
+export class NgbdApiDocsClass implements AfterViewInit, OnDestroy {
   apiDocs: ClassDesc;
+  routeFragmentSubscription: Subscription;
+  scrollToFragment = scrollToFragment;
```
pkozlowski-opensource: Is this assignment necessary? Could you just call the imported function?

thbt: The `click` event calling it was causing an error in the template file, but it was actually unneeded.

thbt: For some reason clicking on the link was not working correctly (first click worked, then it changed the route but would not scroll, or it wouldnt scroll on init), I suppose moving the subscription to ngAfterViewInit fixed it

wesleycho:
One way to get around the duplication is to create a component just for the `h2` element and its children. As far as I can tell, this can be successfully extracted into a component that has the appropriate inputs and outputs (output mainly for the ability to track clicks to the source code for the file).

PR #1454, opened 2017-04-07
fix(typeahead): fix ARIA attributes

thbt:

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -72,7 +72,8 @@ let nextWindowId = 0;
     'autocapitalize': 'off',
     'autocorrect': 'off',
     'role': 'combobox',
-    'aria-autocomplete': 'list',
+    'aria-multiline': 'false',
+    '[attr.aria-autocomplete]': 'showHint ? "both" : "list"',
     '[attr.aria-activedescendant]': 'activeDescendant',
     '[attr.aria-owns]': 'isPopupOpen() ? popupId : null',
     '[attr.aria-expanded]': 'isPopupOpen()'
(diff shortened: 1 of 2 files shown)


wesleycho:
Setting the role to `textbox` seems weird here - can you explain the reasoning for this? Since it's an input, I would think that it naturally has the correct accessibility settings by default for user input, so having to set the role to textbox and setting `aria-multiline` appears to be an unnecessary workaround.

src/typeahead/typeahead.ts
```diff
@@ -72,7 +72,8 @@ let nextWindowId = 0;
     'autocapitalize': 'off',
     'autocorrect': 'off',
     'role': 'combobox',
-    'aria-autocomplete': 'list',
+    'aria-multiline': 'false',
```
pkozlowski-opensource: We don't seem to have a test for the `aria-multiline`. Could you please add one?

thbt:
@wesleycho These changes were made to reflect [the combobox spec](https://w3c.github.io/aria/aria/aria.html#combobox), but actually it doesn't make this component behave any better in NVDA and VoiceOver.

I rolled back the changes on `combobox` vs `textbox` because it works better in NVDA. We could add `<div role="combobox">` around the input in the examples to show good practices, but in the end, it doesn't add much value (VoiceOver is barely supported : it allows to navigate the options, but the component is not announced as a combo box). However, I'm keeping `aria-multiline` as it is required by the spec.

I'm watching what other library are doing and it seems we are on the same track : we should watch [this issue from material](https://github.com/angular/material2/issues/3544), as they are having the same issues with VoiceOver

pkozlowski-opensource:
LGTM as far as my limited knowledge allows me to say...

PR #1463, opened 2017-04-10
chore(docs): add ability to configure port

wesleycho:
- Add ability to configure port used for demo server

This allows users to override the configuration locally in the event the user already uses the default port for other servers (i.e. using webpack-dev-server for local development of their angular application).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ npm-debug.log
 .publish
 .sw*
 .idea
+.vscode
 coverage
 demo/src/api-docs.ts
 demo/src/public/**/plnkr.html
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
Why not. We should probably start thinking about switching to CLI but merging this won't hurt so I'm going to land it.

wesleycho:
Should I open an issue for migrating to the cli for the demo server?

pkozlowski-opensource:
> Should I open an issue for migrating to the cli for the demo server

Won't hurt. But not the top priority IMO - I really want to release 1.0 soooo badly :-)

PR #1465, opened 2017-04-10
feat(modal): add aria-labelledby/modal to modal component

giampierobono:

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -1,20 +1,25 @@
 import {
+  AfterViewInit,
   Component,
-  Output,
+  ContentChild,
+  ContentChildren,
+  ElementRef,
   EventEmitter,
   Input,
-  ElementRef,
-  Renderer,
+  OnDestroy,
   OnInit,
-  AfterViewInit,
-  OnDestroy
+  Output,
+  Renderer
 } from '@angular/core';
 
 import {ModalDismissReasons} from './modal-dismiss-reasons';
 
+let nextId = 0;
+
 @Component({
   selector: 'ngb-modal-window',
   host: {
+    '[attr.aria-modal]': 'true',
     '[class]': '"modal fade show" + (windowClass ? " " + windowClass : "")',
     'role': 'dialog',
     'tabindex': '-1',
@@ -64,6 +69,7 @@ export class NgbModalWindow implements OnInit,
     if (!this._elRef.nativeElement.contains(document.activeElement)) {
       this._renderer.invokeElementMethod(this._elRef.nativeElement, 'focus', []);
     }
+    this._setAriaLabelledByToModal();
   }
 
   ngOnDestroy() {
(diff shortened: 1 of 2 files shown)


src/modal/modal-window.ts
```diff
@@ -1,20 +1,25 @@
...
 } from '@angular/core';
 
 import {ModalDismissReasons} from './modal-dismiss-reasons';
 
+let nextId = 0;
+
 @Component({
   selector: 'ngb-modal-window',
   host: {
+    '[attr.aria-modal]': 'true',
```
pkozlowski-opensource: I don't see this attribute being added in the Bootstrap 4 documentation: https://v4-alpha.getbootstrap.com/components/modal/

Could you please open an issue in the Bootstrap repository first to clarify if this attribute is needed or not (when opening an issue please reference any specs / articles that made you add this attribute).

src/modal/modal-window.ts
```diff
@@ -76,4 +82,15 @@ export class NgbModalWindow implements OnInit,
...
+  private _setAriaLabelledByToModal(): void {
+    const modalHeader: HTMLHeadingElement = this._elRef.nativeElement.getElementsByClassName('modal-title')[0];
+    if (modalHeader) {
+      const modalHeaderId: string = modalHeader.getAttribute('id') === null || modalHeader.getAttribute('id') === '' ?
+          `modal-header-${nextId++}` :
+          modalHeader.getAttribute('id');
+      this._renderer.setElementAttribute(modalHeader, 'id', modalHeaderId);
+      this._renderer.setElementAttribute(this._elRef.nativeElement, 'aria-labelledby', modalHeaderId);
+    }
+  }
```
pkozlowski-opensource: Hmm, I see what you are doing here but it is super, super hackish... And it doesn't have tests... 

I'm not saying that I've got a better solution atm but there is way too much low-level DOM manipulation going on here... As such I would like to think some more about it. 

In the meantime, could you please open an issue describing this missing functionality? Thnx!

src/modal/modal-window.ts
giampierobono: Hello Pawel,
   I know that this attribute is not added into Bootstrap modals, but I am taking this https://github.com/w3c/aria-practices/blob/master/examples/dialog-modal/dialog.html as reference. 
I found this after our phone call about the aria-selected, it was stored as bookmark not in my work google account. Just for information I will share same links here also.

Kind of specifications about accessible modals can be found here: https://cdn.rawgit.com/w3c/aria-practices/master/aria-practices.html#dialog_modal 
and details about this aria attribute here: http://w3c.github.io/aria/aria/aria.html#aria-modal

By the way I will open an issue in the Bootstrap repo to see if this has to be added or not. 

Thanks for the review.

src/modal/modal-window.ts
giampierobono: Personally I don't like this kind of DOM manipulations but, since the modal content can be everything I had no better ideas.
Please let me know if there will be any update / ideas on this cause I am really interested in this. 

Thanks.

giampierobono: Issue on this repo here: https://github.com/ng-bootstrap/ng-bootstrap/issues/1477

src/modal/modal-window.ts
giampierobono: Here the link to the issue opened on the twbs repo: https://github.com/twbs/bootstrap/issues/22429

giampierobono: Since it's part of the newest ARIA specs and not supported yet by assistive technologies, they say for now there is no real benefit in adding it. 

So it can be removed.

src/modal/modal-window.ts
wesleycho: If one wanted to do direct DOM manipulation like this, it would be better done on the user side by manually walking up the DOM tree until it reached the parent to set the id via the renderer.

Another option is to add a host binding for `[attr.aria-labelledby]` on the modal window component, and let the user set the id on the window component instance on the NgbModalRef that is returned from `modal.open`. This would also let us autogenerate an id in the component that the user can read for use in binding on the title...which also opens up the potential for us adding a lightweight NgbModalTitleDirective whose purpose is to add the id attribute to the element for use with the aria labelled-by attribute.

changLiuUNSW:
Any update on this PR

pkozlowski-opensource:
We are going to land #2049 instead

Issue #1467, opened 2017-04-10
Clearing input model on blur does not work if results list is present


deeg:
### Bug description:
* Trying to clear the input model programmatically on blur event does not work if the result window is present and results are showing

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/XSJLmqnPei2ZObg3KVIT?p=preview

* In search type `lkjhsdfsdf` and blur, notice the text clears correctly
* In search type 'ma' and notice results are present. Blur the input and notice the text stays in the input

The same issue happens if you try to set the value of the input directly as well.

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.0.1

ng-bootstrap: 1.0.0-alpha.22

Bootstrap: 4.0.0-alpha.6

deeg:
[It has to do with this line](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/typeahead/typeahead.ts#L194).

What is happening is after the user clears the model on blur event of the input, the dismissPopup function is setting it to the old user value. This code is in place due to the showHint option.

I already have a patch which would fix #1276, but this bug would still exist if the user chose not to use the editable=false option and tried to control it themselves.

@pkozlowski-opensource, what are your thoughts on this? Should I put up the review anyways and we can discuss if this bug would matter after that, or should we aim to fix it so the user could control it themselves too?

deeg:
I ended up putting up the review for discussion purposes.

[This line in the review](https://github.com/ng-bootstrap/ng-bootstrap/pull/1468/files#diff-4b999ef7b06784feb23d420f3d909d1aR202) makes it work when etiable=false, by clearing the private _userInput.

Not sure if we need to fix how dismissPopup works, or if we should just go with the review I have out and close this bug as won't fix.

dimitriy-k:
what is the status? is it fixed?

fredgate:
There is a pull request but it is pending, and has now some branch conflicts.

khodgkins:
One issue I see to the fix that deeg put for review is that it sets the input to null instead of the `[(ngModel)]` if one was previously selected.


```
onFocus = () => {
    //clear the model
    this.model = new Model();
}
onBlur = () => { 
    //set to current selected model
    this.model = this.previousModel; 
}
onSelectItem = (item: Model) => {
    this.model = item;
    this.previousModel = this.model;
    this.selectItem.emit(this.model);
}
```

[...]

PR #1468, opened 2017-04-10
fix(typeahead): clear input text on blur when editable is false

### LINKED ISSUE #1276: clear input (on blur) when editable false
feature request on follow-up on  #829:
- on editable false clear input on blur
- on not found clear input

deeg:
Fixes #1276

I am putting this up for discussion purposes on #1276 and #1467.

I will add unit tests after deciding on the correct approach.

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -197,7 +197,14 @@ export class NgbTypeahead implements ControlValueAccessor,
 
   isPopupOpen() { return this._windowRef != null; }
 
-  handleBlur() { this._onTouched(); }
+  handleBlur() {
+    if (!this.editable) {
+      this._userInput = null;
+      this._onChange(undefined);
+      this.writeValue(undefined);
+    }
+    this._onTouched();
+  }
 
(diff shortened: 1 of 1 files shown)


khodgkins:
One issue I see to the fix  is that it sets the input to null instead of the `[(ngModel)]` if one was previously selected.

Below is a sample of how I am using the typeahead:
```
onFocus = () => {
    //clear the model
    this.model = new Model();
}
onBlur = () => { 
    //set to current selected model
    this.model = this.previousModel; 
}
onSelectItem = (item: Model) => {
    this.model = item;
    this.previousModel = this.model;
    this.selectItem.emit(this.model);
}
```

Perhaps modifying the dismissPopup function would be a better option:

```
dismissPopup() {
    if (this.isPopupOpen()) {
        this._closePopup();
        this._writeInputValue(this._userInput); //not sure the purpose of writing the input value
    }
}
```

jrmlstf:
Will this PR, even after some rework, be merged ?

pkozlowski-opensource:
This PR doesn't merge cleanly any more and the solution needs more thinking / design so I'm going to close it for now.

PR #1469, opened 2017-04-10
demo: text -> icons on the home page

pkozlowski-opensource:
This is still WIP but opening an early PR so we can discuss content (icons, titles and text) as well as the design.


demo/src/app/default/default.component.html
```diff
@@ -2,8 +2,9 @@
   <div class="container">
     <img class="bd-booticon" src="img/logo.svg"/>
     <p class="lead hidden-md-down">
-      Bootstrap 4 components, powered by Angular and written by the angular-ui Team.
+      Bootstrap 4 components, powered by Angular
     </p>
+    <p>Currently at v{{version}}</p>
```
deeg: Might be nice to link this to the changelog.

demo/src/app/default/default.component.html
```diff
@@ -12,25 +13,77 @@
...
+  </div>
+
+  <div class="row mt-5">
+    <div class="col-2">
+      <img src="/img/test.svg">
+    </div>
+    <div class="col-4">
+      <h4>Quality</h4>
+      <p class="text-muted">
+        All code is tested (almost 100% test coverage) and all changes reviewed. We are not cutting corners.
```
deeg: Might not need `(almost 100% test coverage)`. May be better to link to the actual test coverage or something, where the latest results will be. That way if we ever do hit 100 you don't need to come back and change this.

demo/src/app/default/default.component.html
```diff
@@ -12,25 +13,77 @@
...
+    </div>
+
+    <div class="col-2">
+      <img src="/img/escalator.svg">
+    </div>
+    <div class="col-4">
+      <h4>Accessible</h4>
+      <p class="text-muted">
+        All the widgets are accessible. We use proper HTML elements and required aria attributes.
+        Keyboard navigation and focus management is taken care of.
```
deeg: What information are we trying to convey with `Keyboard navigation and focus management is taken care of.`

Seems like the sentence could use a little re-work.

demo/src/app/default/default.component.html
```diff
@@ -12,25 +13,77 @@
...
+      </p>
+    </div>
+
+     <div class="col-2">
+      <img src="/img/world.svg">
+    </div>
+    <div class="col-4">
+      <h4>Community</h4>
+      <p class="text-muted">
+        We give back to the community be contributting to projects we build upon (Angular, Bootstrap). We help people on StackOverflow.
```
deeg: Again not great at English, but I feel this needs a bit of a touch up:

`We help people on StackOverflow.`

Maybe a link to the stack page with our tag with something like: `Get help from our great members at StackOverflow`?

deeg:
I like it. The images might look better with some color, but not sure.

For people to see what it looks like without downloading the patch:

screen shot 2017-04-10 at 1 32 55 pm

demo/src/app/default/default.component.html
```diff
@@ -12,25 +13,77 @@
...
+  </div>
+
+  <div class="row mt-5">
+    <div class="col-2">
+      <img src="/img/team.svg" class="img-fluid">
+    </div>
+    <div class="col-4">
+      <h4>Team</h4>
+      <p class="text-muted">
+        We've created angular-ui/bootstrap and spent several years doing widget development. Several team members are core Angular contributtors.
```
Foxandxss: contributors, only one t.

demo/src/app/default/default.component.html
pkozlowski-opensource: Yeh, maybe. But TBH I'm not sure how to do it since we don't have any obvious mapping between version no and GH tags in the changelog (those will look like https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CHANGELOG.md#100-alpha23-2017-04-12) <- notice the date

Given that this PR doesn't change this part (just moves it around) and the technical difficulty involved I would rather tackle this in a separate PR / issue.

demo/src/app/default/default.component.html
pkozlowski-opensource: @deeg agree, but:
* we don't have any test coverage reports exposed publicly 
* I'm not sure how links would look here

I'm not opposed to the idea but it would require more setup, work etc. I'm trying to make baby steps here :-)

demo/src/app/default/default.component.html
deeg: Sounds good, I just meant the file as a whole as though.

pkozlowski-opensource:
Thnx @deeg and @Foxandxss for the review so far - I've taken into account the feedback (at lest where I knew how to do it :-)) and pushed a new version. Here is how it looks right now:

ngb-icons

Could you please have another look?

Foxandxss:
Yes, work for me

PR #1473, opened 2017-04-11
chore: reactivate vendor bundle

pkozlowski-opensource:
This change makes it possible to load more JS
code in parallel and shave off few hundreds ms
from the first full pain on mobile.

It is also closer to the angular-cli way of
doing things.

### DIFF
--- a/demo/src/main.ts
+++ b/demo/src/main.ts
@@ -1,6 +1,3 @@
-import 'prismjs/themes/prism.css';
-import 'bootstrap/dist/css/bootstrap.css';
-
 import {enableProdMode} from '@angular/core';
 import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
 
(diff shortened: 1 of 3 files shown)


Foxandxss:
Will review at home.

wesleycho:
Is the merging in order to reduce the number of files the end user has to download?

pkozlowski-opensource:
> Is the merging in order to reduce the number of files the end user has to download?

huh? We are not changing no of files (this stay the same). It is more about balancing _size_ of those files so more download happens in parallel.

Foxandxss:
I made two changes.

One: I created a vendorStyles bundles so we can remove all the import lines for CSS on the main.ts. It is clearer in my point of view

Second: I created an inline bundle (the manifest from webpack) so it doesn't end on the vendor bundle. This allow a way better caching solution since the vendor bundle won't need to be downloaded again after **any** kind of change. That also allows us to remove the custom chunk sort.

PR #1476, opened 2017-04-11
fix(buttons): provide constant in forRoot

### LINKED ISSUE #1125: NgbModule.forRoot() breaks 'traditional' style radio buttons
### Bug description:

I am attempting to use 'traditional' style radio buttons, i.e the radio's you can see [here](http://v4-alpha.getbootstrap.com/components/forms/#radios)

The plunker below shows that these radio's do not work when NgbModule.forRoot() has been called.

You can still interact with the radios, however they do not show the initial value of the model.

Angular: 2.0.0

ng-bootstrap: 1.0.0-alpha.14

Bootstrap: 4.0.0-alpha.5

wesleycho:
- Provide constant in `module.forRoot` in order for it to be available

See [Plunker](http://plnkr.co/edit/Zk74D6KrgIJVpW48JnCl?p=preview) for fix in action.

This should fix #1125.

### DIFF
--- a/src/buttons/radio.module.ts
+++ b/src/buttons/radio.module.ts
@@ -1,5 +1,5 @@
 import {NgModule, ModuleWithProviders} from '@angular/core';
-import {NgbRadio, NgbActiveLabel, NgbRadioGroup} from './radio';
+import {NgbRadio, NgbActiveLabel, NgbRadioGroup, NGB_RADIO_VALUE_ACCESSOR} from './radio';
 
 export {NgbRadio, NgbActiveLabel, NgbRadioGroup} from './radio';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@wesleycho as discussed yesterday I had a quick look at this PR but I must say that I don't understand what is the root cause of this issue and why the proposed code change fixes the problem. As a minimum we need a test that would fail without your fix.

IMO it needs more investigation + test.

pkozlowski-opensource:
Trying to get the PR list under control so closing this one for now. @wesleycho feel free to re-open it when you've got time to resume work.

PR #1487, opened 2017-04-12
demo: use lorempixel over https

pkozlowski-opensource:
lorempixel now supports https and without https
we are getting warnings in the console.


deeg:
Can the browser ever be http? I think if you try to go to http it will forward you to https anyways.

pkozlowski-opensource:
@wesleycho @thbt given that "protocol-relative URLs" are not "the thing" any more I would prefer to leave it as-is. More info on the topic: https://jeremywagner.me/blog/stop-using-the-protocol-relative-url

wesleycho:
That's fair - I just thought it unproductive to have to make changes like this for multiple assets whenever they come up. I didn't feel strongly about it either way.

PR #1490, opened 2017-04-13
demo: API docs script adjustements

### LINKED ISSUE #927: API extraction script adjustements
2 issues spotted recently:
- [ ] - undocumented `@Input`s are showing up while they should not: https://ng-bootstrap.github.io/#/components/buttons
- [ ] - getters are documented properly, see: https://github.com/ng-bootstrap/ng-bootstrap/pull/903#discussion_r83921381

pkozlowski-opensource:
Fixes #927

### DIFF
--- a/misc/api-doc.js
+++ b/misc/api-doc.js
@@ -13,21 +13,34 @@ const ANGULAR_LIFECYCLE_METHODS = [
   'ngAfterViewInit', 'ngAfterViewChecked', 'writeValue', 'registerOnChange', 'registerOnTouched', 'setDisabledState'
 ];
 
+function hasNoJSDoc(member) {
+  if (!member.symbol) {
+    return true;
+  }
+
+  const jsDoc = ts.displayPartsToString(member.symbol.getDocumentationComment());
+  return jsDoc.trim().length === 0;
+}
+
 function isInternalMember(member) {
   if (!member.symbol) {
     return true;
   }
 
   const jsDoc = ts.displayPartsToString(member.symbol.getDocumentationComment());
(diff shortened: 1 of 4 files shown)


misc/api-doc.js
```diff
@@ -13,21 +13,34 @@ const ANGULAR_LIFECYCLE_METHODS = [
...
+}
+
 function isInternalMember(member) {
   if (!member.symbol) {
     return true;
   }
 
   const jsDoc = ts.displayPartsToString(member.symbol.getDocumentationComment());
-  return jsDoc.trim().length === 0 || jsDoc.indexOf('@internal') > -1;
+  return jsDoc.indexOf('@internal') > -1;
```
wesleycho: If we mandate using Node 6, we could just do
```js
return jsDoc.includes('@internal');
```
Not a big deal either way

misc/api-doc.js
```diff
@@ -13,21 +13,34 @@ const ANGULAR_LIFECYCLE_METHODS = [
...
 }
 
 function isAngularLifecycleHook(methodName) {
   return ANGULAR_LIFECYCLE_METHODS.indexOf(methodName) >= 0;
 }
 
+function isPrivate(member) {
+  return (member.flags & ts.NodeFlags.Private) !== 0;
```
wesleycho: What are the values possible for these items, booleans?

pkozlowski-opensource: Those are bit-flags. Not sure if there is any documentation for this besides TS source code. But for the record: I'm not changing this code, just moving it around.

wesleycho: I noticed that just afterwards, was just curious what data structures those were.

PR #1507, opened 2017-04-20
fix(dropdown): fixed zone for closeFromOutsideClick

### LINKED ISSUE #1506: Dropdowns not auto-closing due to wrong zone
### Bug description:

`autoClose` does not work immediately because the `closeFromOutsideClick` handler runs in the `<root>` zone. This is reproducible at https://ng-bootstrap.github.io/#/components/dropdown

Eugeny:
The `document:` listener runs in the root zone.

This fixes #1506


Eugeny:
I wasn't sure how to properly type the Zone dependency, since it's typings are broken: https://github.com/angular/zone.js/issues/297

Eugeny:
Was not a real bug after all

PR #1524, opened 2017-04-27
chore: add more accordion tests

Foxandxss:
I reviewed the entire Accordion and I ended adding a few tests to have a 100% coverage.

The logic is a bit meh, but that is for another day.

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -19,6 +19,10 @@ function getPanelsContent(element: HTMLElement): HTMLDivElement[] {
   return <HTMLDivElement[]>Array.from(element.querySelectorAll('div .card-block'));
 }
 
+function getPanelsTitle(element: HTMLElement): HTMLDivElement[] {
+  return <HTMLDivElement[]>Array.from(element.querySelectorAll('div .card-header a'));
+}
+
 function getButton(element: HTMLElement, index: number): HTMLButtonElement {
   return <HTMLButtonElement>element.querySelectorAll('button')[index];
 }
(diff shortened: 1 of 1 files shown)


src/accordion/accordion.spec.ts~RFff54cea.TMP
```diff
@@ -0,0 +1,538 @@
+import {TestBed, ComponentFixture, inject} from '@angular/core/testing';
```
pkozlowski-opensource: Looks like this file made it to this PR by accident, could you please remove it?

pkozlowski-opensource:
@Foxandxss thnx for this, it LGTM apart from some temp file that made it to this PR by accident it seems. Could you please remove it?

src/accordion/accordion.spec.ts~RFff54cea.TMP
Foxandxss: Weird, I Will remove it when I get back to the computer.

src/accordion/accordion.ts
```diff
@@ -105,7 +105,7 @@ export interface NgbPanelChangeEvent {
           {{panel.title}}<ng-template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></ng-template>
         </a>
       </div>
-      <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'" class="card-block" *ngIf="isOpen(panel.id)">
+      <div [id]="panel.id" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'" class="card-block" *ngIf="isOpen(panel.id)">
```
pkozlowski-opensource: I would rather leave it as-is since the id="{{panel.id}}" form will make sure that we convert whatever is passed to string. Not super-important though.

Foxandxss:
Fixed.

PR #1512, opened 2017-04-23
feat: bump up Angular dependency to 4.0.3

pkozlowski-opensource:
Closes #1337
Closes #1415
Closes #1433
Closes #1439

BREAKING CHANGE:

ng-bootstrap requires a minimal version of Angular 4.0.3

PR #1528, opened 2017-05-01
feat: bump up Renderer to Renderer2

wKoza:
For complete the PR #1512 , Renderer should be replaced by Renderer2.

### DIFF
--- a/src/popover/popover.spec.ts
+++ b/src/popover/popover.spec.ts
@@ -76,7 +76,7 @@ describe('ngb-popover', () => {
       directive.triggerEventHandler('click', {});
       fixture.detectChanges();
       expect(getWindow(fixture.nativeElement)).toBeNull();
-      expect(directive.nativeElement.getAttribute('aria-describedby')).toBeNull();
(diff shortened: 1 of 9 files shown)


pkozlowski-opensource:
Awesome, thnx @wKoza - I was planning to do this one!

src/popover/popover.spec.ts
```diff
@@ -76,7 +76,7 @@ describe('ngb-popover', () => {
       directive.triggerEventHandler('click', {});
       fixture.detectChanges();
       expect(getWindow(fixture.nativeElement)).toBeNull();
-      expect(directive.nativeElement.getAttribute('aria-describedby')).toBeNull();
+      expect(directive.nativeElement.getAttribute('aria-describedby')).toBe('null');
```
pkozlowski-opensource: Hmm, this doesn't sound right. It seems like here we would have `aria-describedby="null"` while what we want is no attribute at all. Was this change necessary after migration to the new renderer?

wKoza: Outch, Ok I see. I fix this

pkozlowski-opensource:
LGTM, thnx!

Issue #890, opened 2016-10-14
Button groups does not support focus customization


vicb:
### Bug description:

Button groups do not support styling the `:focus` state
### Link to minimally-working plunker that reproduces the issue:

https://ng-bootstrap.github.io/#/components/buttons

Click on any "button" and press tab, the next button show no focus decoration
### Root cause

Compare the ng-bs markup:

``` html
<div class="btn-group" data-toggle="buttons">
  <label class="btn btn-primary" [class.active]="model.left">
    <input type="checkbox" [(ngModel)]="model.left"> Left (pre-checked)
  </label>
</div>
```

with the [bs4 markup](http://v4-alpha.getbootstrap.com/components/button-group/)

``` html
<div class="btn-group" role="group" aria-label="Basic example">
  <button type="button" class="btn btn-secondary">Left</button>
</div>
```

bs4  uses the focus style of the `button` - I'm not sure there is anything we can do with the current example in ngbs ?
### Version of Angular, ng-bootstrap, and Bootstrap:

ng: 2.0.0
bs: 4.0.0-alpha.4
ng-bs: 1.0.0-alpha.7

wesleycho:
I wonder if Bootstrap 4 changed its approach to button group markup at some point...I know Bootstrap 3 went the route of using the input element. We should probably considering changing it too to better match the HTML for styling reasons.

pkozlowski-opensource:
@wesleycho @vicb there are 2 _different_ styling for "buttons" being used:
* regular button groups, ex.: http://v4-alpha.getbootstrap.com/components/button-group/#basic-example
* radio / checkbox buttons: http://v4-alpha.getbootstrap.com/components/buttons/#checkbox-and-radio-buttons

The reason for existence of those 2 different markups is accessibility. Indeed, radio / check buttons are semantically different from regular buttons and different HTML elements need to be used for those 2 different use-cases.

Now, coming back to the issue reported by @vicb - it is a valid issue - we are not adding properly the `focus` CSS class on a label when an input gets focused. This should be an easy fix though. Thnx for reporting.

alex321:
@pkozlowski-opensource Let me know if there is anything you want to improve in your commit 23d412b. It seems good to me.

alex321:
Sorry for the noise. I can see that your commit has already gone in. What else needs to be done?

thbt:
[...]

PR #1530, opened 2017-05-02
fix(buttons): properly handle focus on checkbox buttons 

wKoza:
Part of #890

To properly support check-boxes we should introduce a dedicated directive for them.
Checkbox do not support styling the :focus state


pkozlowski-opensource:
This was fixed in #1688. Thnx for the PR though!

PR #1534, opened 2017-05-03
perf(datepicker): prevent memory leak caused by ngZone subscription

Ikody:
The datepicker component cannot be correctly garbage collected if any subscription remains after the onDestroy call.

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -12,6 +12,7 @@ import {
   EventEmitter,
   Output,
   OnChanges,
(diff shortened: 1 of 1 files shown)


phgnx:
Thank you I've been wasting way too much time too investigate memory leaks, this fixes my issue :)

pkozlowski-opensource:
Wow, thnx @Ikody - I'm not sure how we've missed this one as all the other directives opening popup got proper `OnDestroy` hooks :-/

LGTM, merging. Thnx again.

PR #1536, opened 2017-05-03
fix(docs): template to ng-template

aleeeftw:
Since **template** has been deprecated, use in the documentation **ng-template** like the code example does.

Thanks!


pkozlowski-opensource:
Good catch, thnx!

PR #1538, opened 2017-05-04
chore: bump up zone.js version

pkozlowski-opensource:
zone.js 0.8.10 comes with a 4x boot-time
speedup and no breaking changes so no reason
to wait with the upgrade :-)

### DIFF
--- a/demo/src/public/index.html
+++ b/demo/src/public/index.html
@@ -34,6 +34,13 @@
       ga('create', 'UA-37467169-2', 'auto');
       ga('send', 'pageview');
       var _gaq = window._gaq || null; var ga = window.ga || null;
+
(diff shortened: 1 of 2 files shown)


demo/src/public/index.html
```diff
@@ -34,6 +34,13 @@
       ga('create', 'UA-37467169-2', 'auto');
       ga('send', 'pageview');
       var _gaq = window._gaq || null; var ga = window.ga || null;
+
+      __Zone_disable_Error = true;
```
wesleycho: Do we want to prefix on window?

pkozlowski-opensource: Why not... Just not sure if there is any advantage to it...

PR #1542, opened 2017-05-07
chore: remove as string

Foxandxss:
A friend of mine (@pablonete) told me about this feature and I remembered using it on the accordion.

WDYT?

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -178,7 +178,7 @@ export class NgbAccordion implements AfterContentChecked {
   ngAfterContentChecked() {
     // active id updates
     if (isString(this.activeIds)) {
-      this.activeIds = (this.activeIds as string).split(/\s*,\s*/);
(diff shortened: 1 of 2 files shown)


src/accordion/accordion.ts
```diff
@@ -178,7 +178,7 @@ export class NgbAccordion implements AfterContentChecked {
   ngAfterContentChecked() {
     // active id updates
     if (isString(this.activeIds)) {
-      this.activeIds = (this.activeIds as string).split(/\s*,\s*/);
+      this.activeIds = (this.activeIds).split(/\s*,\s*/);
```
wesleycho: You can probably remove the parentheses here

wesleycho:
Looks like this was introduced in TS 1.6.0-beta.

Details on user defined type guards: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

I'm ok with this.

Foxandxss:
Ready for review again.

PR #1545, opened 2017-05-08
feat(datepicker): allow separate style for year with .ngb-dp-year

christarczon:

This is a niche thing that probably won't be widely used but it sure would help me out if it's ok!

### DIFF
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -94,7 +94,7 @@ export interface NgbDatepickerNavigateEvent {
       <ng-template ngFor let-month [ngForOf]="months" let-i="index">
         <div class="ngb-dp-month d-block ml-3">            
           <div *ngIf="navigation !== 'select' || displayMonths > 1" class="ngb-dp-month-name text-center">
-            {{ i18n.getMonthFullName(month.number) }} {{ month.year }}
+            {{ i18n.getMonthFullName(month.number) }} <span class="ngb-dp-year">{{ month.year }}</span>
           </div>
           <ngb-datepicker-month-view
             [month]="month"


src/datepicker/datepicker.ts
```diff
@@ -94,7 +94,7 @@ export interface NgbDatepickerNavigateEvent {
       <ng-template ngFor let-month [ngForOf]="months" let-i="index">
         <div class="ngb-dp-month d-block ml-3">            
           <div *ngIf="navigation !== 'select' || displayMonths > 1" class="ngb-dp-month-name text-center">
-            {{ i18n.getMonthFullName(month.number) }} {{ month.year }}
+            {{ i18n.getMonthFullName(month.number) }} <span class="ngb-dp-year">{{ month.year }}</span>
```
wesleycho: If we go this route, wouldn't it be worthwhile to also have a span element around the month name with an appropriate class too?

christarczon: You can style the div however and then override the span as necessary.

But yes, if I was starting from scratch I would move the ngb-dp-month-name down to a span around the month name. But that could break existing styles and I didn't want to push it. If a new class is added I don't know what to name it, there's already ngb-dp-month and ngb-dp-month-name!

You could reuse a class and use the tag in the selector to choose but that's not very clean.

The cleanest ignoring compatibility would be to move ngb-dp-month-name down to the span and replace it with ngb-dp-month-year-text.

pkozlowski-opensource:
@christarczon what is the use-case that you are trying to cover with this PR? It looks like it might be related to the customisation needs, but I would like to confirm.

If this happens to be about customisation, I don't think that we are in the position to start accepting PRs like this on. For ones the internal datepicker markup is still subject to change as we work more on accessibility, animations etc. Then we are also not convinced that customisation through CSS classes sprinkled here and there is the way to go. If you would push this approach to the extreme you would end up with custom classes on each and every element. This would mean that our entire markup becomes API - and it would be dead of the project as we couldn't do any more changes.

We want to have a proper customisation story but I would like to do be properly designed instead of just "happening".

christarczon:
I understand. Just to show you, this is the design I was given:
datepicker month-year

I'm accomplishing now by hiding the label and making my own hooked up to (navigate). Not a huge deal and I was able to style the rest with no problems. If only CSS had :first-word !

pkozlowski-opensource:
@christarczon you can always override names of months by providing your implementation of the `NgbDatepickerI18n`: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-i18n.ts

I'm going to close this PR based on the previous discussion.

PR #1546, opened 2017-05-09
docs(README): Add instructions for modular builds to README

MattMorrisDev:
I spent a ton of time trying to figure this out for myself, so I figured I'd save others some time by adding this information to the README. 

This [stack overflow link](http://stackoverflow.com/questions/43429897/webpack2-angular2-ng-bootstrap-tree-shaking/) appears to be the only information online that mention "tree shaking" or modular builds in relation to ng-bootstrap. It would be good to document this information in the README, which this PR accomplishes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,48 @@ export class OtherModule {
 }
 ```
 
+### Including only what you need
+For production scenarios, you will likely only want to include the parts of ng-bootstrap that you are using in your application.
+To do so, you can import only the specific modules that you need. For example:
+
+ ```js
+ import {NgbModalModule, NgbDropdownModule, NgbRatingModule} from '@ng-bootstrap/ng-bootstrap';
+ 
+ @NgModule({
+   declarations: [...],
+   imports: [
+      NgbModalModule.forRoot(),
+      NgbDropdownModule.forRoot(),
+      NgbRatingModule.forRoot(),
+      ...
+    ], 
+ })
+ export class AppModule {
+ }
+ ```
+ 
+ Any ng-bootstrap modules that you import in lazily loaded feature modules must also be imported in your main "AppModule" using the `.forRoot()` method.
+ Also note that components in your lazily loaded modules should use deep imports to avoid bringing in unused code. For example,
+ one common way to open ng-bootstrap modals is by doing the following:
+ 
+ ```js
+ constructor(private modalService: NgbModal) {
+    
+ }
+ 
+   
+ 
+ openModal() {
+   this.modalService.open(SomeModalComponent);
+ }
+ ```
+ 
+ If this were a component in a lazily loaded module, you would want to import `NgbModal` in the following way:
+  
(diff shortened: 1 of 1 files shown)


christarczon:
Having just been through this myself, the important part of that link is that modules need to be imported directly from their module.ts file for this to work.

This:
`import {NgbModalModule, NgbDropdownModule, NgbRatingModule} from '@ng-bootstrap/ng-bootstrap';`

Should be this:
`import { NgbModalModule } from '@ng-bootstrap/ng-bootstrap/modal/modal.module';`
`import { NgbDropdownModule } from '@ng-bootstrap/ng-bootstrap/dropdown/dropdown.module';`
`import { NgbRatingModule } from '@ng-bootstrap/ng-bootstrap/rating/rating.module';`

MattMorrisDev:
I agree, that is probably the safest way. But for some reason, I can import my modules from `'@ng-bootstrap/ng-bootstrap'` without adding unused code to my bundles. It's only once I import from there in my components that belong to lazily loaded modules that unused code ends up in my lazily loaded bundles. 

Anyways, I will make that change momentarily.

christarczon:
Interesting. Splitting the imports was the last piece of the puzzle for me. I'm not doing lazy loading yet.

MattMorrisDev:
Just as an example, in case anyone with more webpack knowledge wants to chime in, here is an simple example repo demonstrating the issue in my last comment:

https://github.com/MattMorrisDev/angular-experiments/tree/bootstrap-bundling

The code in question can be seen [here.](https://github.com/MattMorrisDev/angular-experiments/blob/bootstrap-bundling/src/app/modules/feature-a/feature-a.component.ts#L3).

- Install (`npm install` or `yarn install`)
- Run (`npm start` or `yarn start`)
- View bundles (`npm run sme` or `yarn sme`) (Won't work on windows unless you modify the script)

README.md
```diff
@@ -57,6 +57,52 @@ export class OtherModule {
 }
 ```
 
+### Including only what you need
+For production scenarios, you will likely only want to include the parts of ng-bootstrap that you are using in your application.
+To do so, you can import only the specific modules that you need. For example:
+
+ ```js
+ import { NgbModalModule } from '@ng-bootstrap/ng-bootstrap/modal/modal.module';
```
wesleycho: I don't think this is necessary if you're using a tool like Webpack 2 or Rollup that can tree shake unused code.

The only example I can think of off the top of my head where it matters where you import from is `rxjs`, and that package is special due to how the additional operators add to the prototype of the `Observable` class.

MattMorrisDev: Like I mentioned above, I personally don't have this problem when importing the modules. But for importing things like `NgbModal` while in lazily loaded modules, I do have to use the deep imports, otherwise the rest of ng-bootstrap gets pulled in.

wesleycho: This is maybe more important for lazy loading - it might be worth noting for that specific scenario.

PR #1551, opened 2017-05-15
feat(datepicker): add keyboard navigation

maxokorokov:
This essentially contains two things:
- All the work @divdavem has done for adding keyboard navigation (#1273)
- Complete refactoring or the datepicker internals

Problems with the datepicker before refactoring:
- All the logic is inside of the `NgbDatepicker` component - became a huge mess after adding of the keyboard handling
- Very little unit tests for the logic because of the above

Refactoring:
- Public API unchanged
- `NgbDatepicker` component is as dumb as possible, just displays `DatepickerViewModel` and propagates `@Inputs` to the service
- `NgbDatepickerService` generates the `DatepickerViewModel` with a very simple API
- `NgbDatepickerKeyMappingService` handles key event mapping to actions 
- Many more unit tests for the `service`
- `OnPush` everywhere for the datepicker components

So now it looks something like this:
dp design

P.S. It might seem a lot, but 2/3 of the PR are unit tests
P.P.S. Tried lots of different solutions based on observables, but for now they add only boilerplate code...

### DIFF
--- a/src/datepicker/datepicker-navigation-select.ts
+++ b/src/datepicker/datepicker-navigation-select.ts
@@ -18,11 +18,20 @@ import {NgbCalendar} from './ngb-calendar';
     }
   `],
   template: `
-    <select [disabled]="disabled" class="custom-select d-inline-block" [value]="date?.month" (change)="changeMonth($event.target.value)">
-      <option *ngFor="let m of months" [value]="m">{{ i18n.getMonthShortName(m) }}</option>
-    </select>` +
-      `<select [disabled]="disabled" class="custom-select d-inline-block" [value]="date?.year" (change)="changeYear($event.target.value)">
-      <option *ngFor="let y of years" [value]="y">{{ y }}</option>
+    <select
+      [disabled]="disabled"
+      class="custom-select d-inline-block"
+      [value]="date?.month"
+      (change)="changeMonth($event.target.value)"
+      tabindex="-1">
+        <option *ngFor="let m of months" [value]="m">{{ i18n.getMonthShortName(m) }}</option>
+    </select><select
+      [disabled]="disabled"
+      class="custom-select d-inline-block"
+      [value]="date?.year"
+      (change)="changeYear($event.target.value)"
+      tabindex="-1">
+        <option *ngFor="let y of years" [value]="y">{{ y }}</option>
     </select> 
   `  // template needs to be formatted in a certain way so we don't add empty text nodes
 })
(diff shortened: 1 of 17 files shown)


maxokorokov:
Thanks, @divdavem!

Amended fixes for all your comments. 
Also added the `BREAKING CHANGE` label because of the `OnPush` introduction

maxokorokov:
Aligned with master, resolved conflicts

pkozlowski-opensource:
@maxokorokov I don't see TravisCI being run on those commits, I'm not sure it is because of getting commits from multiple people. But I want to see "green tick" before merging this one :-)

src/datepicker/datepicker-navigation-select.ts
```diff
@@ -18,13 +19,22 @@ import {NgbCalendar} from './ngb-calendar';
...
-      <option *ngFor="let m of months" [value]="m">{{ i18n.getMonthShortName(m) }}</option>
-    </select>` +
-      `<select [disabled]="disabled" class="custom-select d-inline-block" [value]="date?.year" (change)="changeYear($event.target.value)">
-      <option *ngFor="let y of years" [value]="y">{{ y }}</option>
+    <select
+      [disabled]="disabled"
+      class="custom-select d-inline-block"
+      [value]="date?.month"
+      (change)="changeMonth($event.target.value)"
+      tabindex="-1">
```
pkozlowski-opensource: Why `tabindex="-1"`?

src/datepicker/datepicker.spec.ts
```diff
@@ -541,6 +572,143 @@ describe('ngb-datepicker', () => {
       expect(getYearSelect(fixture.nativeElement).value).toBe(`${today.getFullYear()}`);
     });
 
+    it('should correctly navigate with keyboard', () => {
```
pkozlowski-opensource: This test is quite big, and if one condition fails we would have hard time figuring out what failed exactly. 

Could you split it into multiple small tests? (ideally one test for one navigation key)?

pkozlowski-opensource:
Great stuff @divdavem @maxokorokov ! I've left some comments. The diff is big so I can't really comment on every line of code, but I haven't seen anything scary :-)

@maxokorokov could you please expand on the "BREAKING CHANGES" (that is, take it from the user's point of view?)

src/datepicker/datepicker-navigation-select.ts
maxokorokov: In the current implementation its the `ngb-datepicker` itself that captures all events and prevents default behaviour to avoid page scrolling when hitting arrows, space, etc.

Now that I think of, it might be better in future to focus only a specific date instead of everything, so the default navigation behaviour with keyboard select boxes and arrows would continue working as before

src/datepicker/datepicker.spec.ts
maxokorokov: Think of it of as a big integration test :) 
key-mapping service has unit tests of its own, but yes, we should split this one

maxokorokov:
Split the `datepicker.spec.ts` test into smaller pieces

PR #1558, opened 2017-05-19
feat(allowInsideClick): Add 'closeFromInsideClick' to dropdown

### LINKED ISSUE #1022: Dropdown - autoclose on outside click option
### Feature request

It would be a useful addition to have another autoClose option: `onOutsideClick` (or another meaningful name), so we can keep the dropdown open as long as we click somewhere inside it.

danielronnkvist:
This adds an option so that `closeFromOutsideClick` ignores clicks inside the dropdown. This option will be ignored when `autoClose` is set to `false`.

Closes #1022

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -29,6 +29,12 @@ export class NgbDropdown {
   @Input() autoClose: boolean;
 
   /**
+   * Indicates that dropdown should be closed from a click within the dropdown. If autoClose is false, then this is not
+   * taken into account.
+   */
+  @Input() closeFromInsideClick: boolean;
+
+  /**
    *  Defines whether or not the dropdown-menu is open initially.
    */
   @Input('open') _open = false;
@@ -39,9 +45,10 @@ export class NgbDropdown {
    */
   @Output() openChange = new EventEmitter();
 
-  constructor(config: NgbDropdownConfig) {
+  constructor(config: NgbDropdownConfig, private elementRef: ElementRef) {
     this.up = config.up;
     this.autoClose = config.autoClose;
+    this.closeFromInsideClick = config.closeFromInsideClick;
   }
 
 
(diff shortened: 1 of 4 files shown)


wesleycho:
This feels wrong to me - I feel like what we really want is the ability to support events from the popover component for closing.

src/dropdown/dropdown.ts
```diff
@@ -82,7 +89,8 @@ export class NgbDropdown {
   }
 
   closeFromOutsideClick($event) {
-    if (this.autoClose && $event.button !== 2 && !this._isEventFromToggle($event)) {
+    if (this.autoClose && this._shouldcloseFromInsideClick($event) && $event.button !== 2 &&
```
pkozlowski-opensource: I don't understand this logic... Why is the `_shouldcloseFromInsideClick` impacting logic of outside clicks?

danielronnkvist: @pkozlowski-opensource Hm, this methods name is a bit misleading since it doesn't take into account if the click is from outside the dropdown. One solution is to implement the `!this.elementRef.nativeElement.contains($event.target)` check in the if statement to exclude inside clicks, thus making this method only responding to outside clicks.

Another solution just to rename it to `closeFromClick`.

danielronnkvist:
@wesleycho Could you elaborate? I don't really follow you. You can still allow close events from the dropdown by creating explicit actions that call either the `close` or `toggle` method.

wesleycho:
My comment was in reference to events from inside the popover to trigger the close in general and not just a click. The api design of this feels wrong.

pkozlowski-opensource:
@danielronnkvist woooops, it seems like I've fixed this one via bb975afcd929ce00c400c74b64d3fd2a48ed5106 and didn't take your PR into account :-/ I'm really sorry for this... We really _do_ value all the PRs - this was just a hiccup on my side... 

I truly hope that you will find time to go over bb975afcd929ce00c400c74b64d3fd2a48ed5106 and verify that it covers your use-case(s). And of course I do hope to see more PRs from you in the future!

Issue #1098, opened 2016-11-30
Carousel does not support swipe


rolandoldengarm:
### Bug description:
The carousel does not support swipe
https://ng-bootstrap.github.io/#/components/carousel

On a mobile device, the carousel cannot be swiped to left & right.

pkozlowski-opensource:
Yeh, we should investigate / add proper mobile support....

divdavem:
If [hammer.js](https://github.com/hammerjs/hammer.js) is no longer actively supported, maybe we could use [swiper](http://idangero.us/swiper/) which is used in ionic.

kobvel:
Any updates on this one?

Babelfish112:
Any updates? Also looking for a solution here.

egrassinode:
Any update on this issue?

rodrigoGA:
Any update on this issue?

angelvalverde21:
Aun no funciona

ksivaganesh99:
Any update on this functionality...This became a blocker for us.

PR #1574, opened 2017-05-24
feat(carousel): add ability to swipe slides

wesleycho:
- Add ability to swipe carousel slides on mobile

This should address #1098


src/carousel/carousel.ts
```diff
@@ -157,6 +157,14 @@ export class NgbCarousel implements AfterContentChecked,
     }
   }
 
+  swipeCycle(event) {
```
pkozlowski-opensource: `event` -> `$event`

pkozlowski-opensource:
Cool @wesleycho ! Could you please add test for this as well?

wesleycho:
Good thing you asked for a test - turned out I put the swipe on the wrong element!

wesleycho:
There is one unfortunate side effect of the addition of the swipe - it seems to require hammer.js to be loaded. How do you feel about this @pkozlowski-opensource ?

wesleycho:
I just opened https://github.com/angular/angular/issues/16992 since I think this is important for Angular to support.

Only1MrAnderson:
any traction on this?

pkozlowski-opensource:
So I looked into the topic a bit and it seems like `hammer.js` is not actively supported any more. As such I wouldn't like to tie ng-bootstrap to it. We will need to find another approach in #1098

Issue #1163, opened 2016-12-27
Carousel Cycle Interval alive


aralroca:
Is not an issue, is a doubt.

In Carousel. I want that the cycle interval will alive if the mouse is entered... How can I implement this 😅  ?

Thank you so much! 🙂

pkozlowski-opensource:
We don't have such option currently but I can mark it as a feature request.

aralroca:
Oh cool!  Is it possible to contribute via PR? Maybe I can help 🙂

pkozlowski-opensource:
@aralroca PRs are always welcomed provided that those contain tests, documentation update etc. Thnx!

wesleycho:
Question about this - how should this be implemented? One would have to have a feature flag I think since this would be a breaking change. Is there any other ways we could implement this that would minimize the amount of configuration needed?

pkozlowski-opensource:
Yep, this would have to be a new option / input.

jodhman:
Is there no temporary fix?
I can't seem to find a viable option.

DheerajVislavath:
Any temporary fix to disable mouse events ?

PR #1577, opened 2017-05-25
feat(carousel): add support for trigger on mouseenter

wesleycho:
- Add optional support for triggering the carousel slide cycling on `mouseenter` via `playOnMouseenter` binding

This should address #1163.

Let me know if there is a better name for this binding @pkozlowski-opensource - I picked it only for lack of imagination.


erzaehlsalex:
What is the current status of this? This is a very important feature.

pkozlowski-opensource:
@wesleycho I'm going to merge #2436 as your PR doesn't merge cleanly anymore....

PR #1555, opened 2017-05-18
feat(typeahead): expose the NgbHighlight component

alexandrepayette:
Enables users to reuse the same highlight behavior as the default one when using custom result templates.

Closes #1553

PR #1579, opened 2017-05-25
feat(typeahead): expose the NgbHighlight component

### LINKED ISSUE #1553: [NgbTypeahead] Support term highlighting in result templates
Hi,

I wanted to create a custom result template for an instance of `ngb-typeahead` in my project but I also wanted to keep the default highlighting behavior. I looked though the default template and it uses a component called `ngb-highlight` to handling this behavior. Unfortunately, it is not exported by the typeahead module so there is no way to use it directly without duplicating the code into my own project. Is it possible to export this component in the module as well?

This is a small change so I could easily do a PR for it if needed.

alexandrepayette:
Enables users to reuse the same highlight behavior as the default one when using custom result templates.

Closes #1553
Follows #1555 that was closed a few days ago

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -69,7 +69,8 @@ export {
   NgbTypeaheadModule,
   NgbTypeaheadConfig,
   NgbTypeaheadSelectItemEvent,
-  NgbTypeahead
+  NgbTypeahead,
+  NgbHighlight
 } from './typeahead/typeahead.module';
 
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@alexandrepayette #1553 was already solved by a89cc97 (which is your commit plus changes I did to export things from index.ts). All the changes will be part of the next release.

Issue #1225, opened 2017-01-18
NgbInputDatepicker highjacks change detection on input field


phil-cap:
The NgbInputDatepicker directive appears to highjack change detection capabilities of Angular 2 on input elements.
Once the directive is added to an input element, i do not receive any change events on keyup when changing the date directly in the input field. 
The changes are however detected once i click somewhere outside the input field.

pkozlowski-opensource:
Could you please provide a minimal plunker that demonstrates a difference with and without the datepicker directive. AFAIK the `change` event is executed on blur only so won't execute on keyup events - you might be looking for the `input` event (?).

What are you trying to achieve, exactly?

phil-cap:
@pkozlowski-opensource apologies for the lack of information. I was frustrated of multiple hours of failed attempts at resolving the issue and lacked the necessary patience.
Hiere is the plunk: [plunker](http://plnkr.co/edit/DIBD1Te1cXCj2RqssNOy?p=preview)

Essentially form validation, which is triggered by `keyup` events, stops working as expected once the `ngbDatePicker ` directive is applied. The event then switches to `on blur` from what i can tell.

I currently do not know how to enforce input validation to ensure content is valid before form submission.

pkozlowski-opensource:
@phil-cap thnx for the plunk. I think I see what you mean. My first knee-jerk reaction is that it works as expected. Looking at the validation constraints you've got in your plunk (`required`, `minlength` and `maxlength`) I would argue that:
* `required` should get valid only if a valid date is entered - making it valid on any character doesn't make sense IMO since for a datepicker an atomic entry is entire date;
`minlength`  and `maxlength` doesn't make sense at all in the context of picking date so I would not expect those validation constraints to have any effect. What was your intended behavior here?

Having said all the above I can see from where you are coming from. I'm planning to work on another issue around date validation this / next week so I will look into several use case. In the meantime I would ask you to:
* share more of your thoughts on `required`, `minlength`  and `maxlength`  validation constraints given my above comments;
* have a look at #1222 which also talks about date validation - maybe we should merge the 2 issues?

phil-cap:
@pkozlowski-opensource thanks for your speedy feedback. I must admit that the `minlength` and `maxlength` validations are unrealistic in this `date` context. I was using them as regular expression placeholders. I had trouble with Angular 2 concerning regexes before on a separate issue and hence picked something simpler.

[...]

PR #1605, opened 2017-06-03
fix(datepicker): change trigger event for NgbInputDatepicker

PerfectPixel:
Changes the trigger event in NgbInputDatepicker for manual date change from `change` to `input`as discussed in #1225.

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -74,7 +74,7 @@ describe('NgbInputDatepicker', () => {
       const fixture = createTestCmpt(`<input ngbDatepicker [(ngModel)]="date">`);
       const inputDebugEl = fixture.debugElement.query(By.css('input'));
 
-      inputDebugEl.triggerEventHandler('change', {target: {value: '2016-09-10'}});
+      inputDebugEl.triggerEventHandler('input', {target: {value: '2016-09-10'}});
       expect(fixture.componentInstance.date).toEqual({year: 2016, month: 9, day: 10});
     });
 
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
@PerfectPixel thnx for the PR, it looks good. One question though - did you test it on IE10/11 and / or Edge? I recall that there are some issues with how the `input` event is handled in those browsers and I would like to verify that we are not breaking code on those browsers (existing unit tests won't catch those issues, unfortunately...)

PerfectPixel:
@pkozlowski-opensource ok there is a rub. [This line](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-input.ts#L190) causes the following behaviour:

1. User types "2017"
2. "2" is not a valid `NgbDate` -> `this._model = null`, therefore, `_onChange = '2'` and the input is overwritten with `null` by `_writeModelValue(null)`
3. "0" becomes the first number again as the input was reset with null.
4. and so on...

Here is a gif where I am typing numbers randomly.
untitled

The easiest solution is to uncomment [this line](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-input.ts#L190). However, there are a few changes in behaviour:

1. Typing a year + month and press enter no longer autocompletes the input to year + month + 1st of month (although the model stays the same).
2. It is now possible to enter letters (a-zA-Z etc.) into the input.

PerfectPixel:
@pkozlowski-opensource have a look at my last commit. This fixes the input. Also, it mitigates the first issue as I introduced a change event that updates the view with the model enabling autocomplete of the day once again. The 2nd and 3rd issues are also happening in the original version just on a different event.

pkozlowski-opensource:
@PerfectPixel it all looks good to me! We've got a big PR for date picker in-flight (keyboard navigation) so I'm going to wait with merging till the end of the week (so other PR can land first).

Thnx!

pkozlowski-opensource:
@PerfectPixel the other PR I've mentioned just landed so we can get in this one. Would you mind rebasing so we can land your PR?

PerfectPixel:
Sure, I'll have a look at it tonight

PerfectPixel:
@pkozlowski-opensource all done :-)

PR #1607, opened 2017-06-04
Introduced destroyOnHide like with tabset

achimha:
Setting destroyOnHide to false will create all panels and hide the collapsed ones instead of destroying them.

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -105,7 +105,8 @@ export interface NgbPanelChangeEvent {
           {{panel.title}}<ng-template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></ng-template>
         </a>
       </div>
-      <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'" class="card-block" *ngIf="isOpen(panel.id)">
+      <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'" class="card-block"
+        *ngIf="!destroyOnHide || isOpen(panel.id)" [hidden]="!destroyOnHide && !isOpen(panel.id)">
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
Hi there! Thnx for the PR. As noted in https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md we can't merge any PRs without the associated test.

Could you please add relevant tests? You could use  https://github.com/ng-bootstrap/ng-bootstrap/commit/79dd101cf73ce21e5ea40b2b315e8799cd07f219 as inspiration

achimha:
Yes, of course, I was just trying to get feedback if you're open to this change.

I still don't like the `destroyOnHide` name -- here and in the tabset. Setting it to `false` will not only not destroy it on hide but also create all invisible components when the tabset/accordion is created. That is a side effect that is not always intended. I believe the behavior should better be separated into two properties.

What do you think?

pkozlowski-opensource:
@achimha seems like we've got already a PR that is doing the same + has tests and was opened earlier on. Could you please have a look at #1370 and let me know what you think?

achimha:
Yes, #1370 achieves the goal.

Still, there is a flaw with `destroyOnHide` for accordion and tabset. It not only won't destroy nodes on hide, it also causes all nodes to be instantly created when the accordion/tabset is created. I believe in many (most?) cases this is not desirable.

pkozlowski-opensource:
> Yes, #1370 achieves the goal.

Good to hear. I'm going to close this one in favour of #1370 then, as it has tests etc.

> Still, there is a flaw with destroyOnHide for accordion and tabset. It not only won't destroy nodes on hide, it also causes all nodes to be instantly created when the accordion/tabset is created. I believe in many (most?) cases this is not desirable.

@achimha could you please open a separate issue for this? Thnx!

PR #1612, opened 2017-06-09
fix: support closure-compiler

mlaval:
This PR adds support for building ng-bootstrap with closure-compiler, both the library and the demo.
The changes are about properties access and RxJS imports.

See https://github.com/mlaval/ng-bootstrap-closure-compiler to build the full demo with closure-compiler


alexeagle:
How hard is it to add some testing so regressions would be reported?
I have a couple setups, let me know if I can help

pkozlowski-opensource:
@alexeagle we can definitively add another build step to verify that closure stuff is not broken. If you've got any playgrounds / sample setups, this would help!

alexeagle:
https://github.com/angular/closure-demo has a very simple test setup:
- runs with a docker image (angular/ngcontainer) -  this means the build
script doesn't have to install stuff
- using CircleCI (I think you could do it with travis instead, haven't
tried)
- builds with closure (see "build" and "closure" scripts in package.json)
- has a simple protractor test (see the "pretest" and "test" scripts in
package.json)

Issue #1302, opened 2017-02-08
ngb-tabset no longer stacks (not responsive) on small widths. 


robgorman:
### Bug description:

Ngb-tabset no longer stacks on small width windows. 

### Link to minimally-working plunker that reproduces the issue:

Plunker that illustrates the issue  [here](http://plnkr.co/edit/bASTGZ1FOW4IeWuRcM27) . Plunker forked directly from the current angular samples docs. 

Stacking behavior of tabs used to work perfectly. But in some recent release, or due to a dependency release, I am not sure which, tab stacking stopped working. 

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 2.3.1 (router 3.3.1)

ng-bootstrap: 1.0.0 alpha 20

Bootstrap: 4.0.0 alpha 6

pkozlowski-opensource:
This seem to be linked to a change between Bootstrap alpha.5 and alpha.6 and not something we can fix in this project. You can confirm that the issue is on the CSS side by looking at [this plunk](http://plnkr.co/edit/ONwIEyo90A7vvarwrsvJ?p=preview) and observing that:
* pure Bootstrap 4 example doesn't stack either;
* going back to Bootstrap alpha.5 makes tabs stack

Going to close it here as Bootstrap 4 - please open a relevant issue directly in the Bootstrap repository.

jesben:
Add this: "flex-column flex-md-row" to make it stack again

`<ul class="nav nav-tabs flex-column flex-md-row" role="tablist">`

dannyhchan:
Jesben, how did you inject the classes in the UL?  Do you have a code sample?

Thanks,

jesben:
dannyhchan, I just created an pull request: [Link](https://github.com/ng-bootstrap/ng-bootstrap/pull/1621)

For my prod. build right now, I added it in:
node_modules\@ng-bootstrap\ng-bootstrap\tabset\tabset.js

Before: (line 113)
[class]=\"'nav nav-' + type + '  justify-content-' + justify\"

After: (line 113)
[class]=\"'nav nav-' + type + ' flex-column flex-md-row justify-content-' + justify\"

dannyhchan:
Thank you.

geetee24:
[...]

PR #1621, opened 2017-06-18
Making tabs responsive again

jesben:
Problem came in Bootstrap v4.0.0-alpha.6, after they changed everything to flexbox.

### DIFF
--- a/src/tabset/tabset.ts
+++ b/src/tabset/tabset.ts
@@ -79,7 +79,7 @@ export interface NgbTabChangeEvent {
   selector: 'ngb-tabset',
   exportAs: 'ngbTabset',
   template: `
-    <ul [class]="'nav nav-' + type + ' justify-content-' + justify" role="tablist">
+    <ul [class]="'nav nav-' + type + ' flex-column flex-md-row justify-content-' + justify" role="tablist">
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@jesben thnx for the PR - I just had a look at it and I'm worried about the fact that it makes somehow arbitrary decisions and doesn't expose full flexibility available to the regular Bootstrap users: https://getbootstrap.com/docs/4.0/components/navs/#working-with-flex-utilities

I understand that giving people full control _and_ nice abstraction _is_ hard but I feel like we need to explore design / APIs of tabs some more. Could you think of alternative approaches? 

Also please note that I've just merged 73ba757fedb34be79809d6acf242232a011b050b which needs to be taken into account as well.

I would be willing to work with you / anyone else on improving tabs flexibility API but I don't think that this PR is ready to be merged as-is.

benouat:
I just checked on the [plunkr](http://plnkr.co/edit/bASTGZ1FOW4IeWuRcM27?p=preview) referenced in #1302 and tested with both bootstrap@4.0.0 and bootstrap@4.1.1.

Tabs are properly stacking.

image


Hence I am closing this PR as it is not needed anymore.

PR #1634, opened 2017-06-26
Demo fragments processing using NavigationEnd router event

benouat:
Each fragment url we use in the demo used to be relying on a `NgbdFragment` directive which is subscribing to the router's fragment observable for each instance.
I simply moved this logic to an high order component (`app.component.ts`) in which I just subscribe one time to the router's observable: `router.events`.
The directive still remains but is very simplistic, and just add an id on the host element that corresponds to the fragment name itself.

I also had to tweak a bit the corresponding css code to have them **not** to be using `display: none`


pkozlowski-opensource:
LGTM. Restarting TravisCI to get it green and will merge just after. Thnx!

PR #1643, opened 2017-06-29
fix(tooltip): Tooltips not positioned properly

mschoudry:
fix for bug 1371


pkozlowski-opensource:
LGTM, thnx!

pkozlowski-opensource:
@mschoudry thnx for this one. Could you also please send a similar PR for popovers (https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/popover/popover.ts#L105-L123)?  Thnx!

PR #1645, opened 2017-06-29
feat(demo): Component page using a tabset

benouat:
The main idea behind is that it basically prevents the user to have to scroll down the entire page when you want just see a demo.

I updated the demo Component page to actually use a `nab-tabset` to separate *api* from *examples*.
The router part, and thus the url, is now reflecting this new level of indirection.

screen shot 2017-06-29 at 16 26 37

What do you think ?

### DIFF
--- a/demo/src/app/app.routing.ts
+++ b/demo/src/app/app.routing.ts
@@ -21,27 +21,114 @@ import {
   NgbdTypeahead
 } from './components';
 
+const DEFAULT_API_PATH = {path: '', pathMatch: 'full', redirectTo: 'api'};
+
+const componentRoutes = [{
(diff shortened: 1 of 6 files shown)


benouat:
My bad, build is not ok. I'll have a look at it.

benouat:
I simply forgot about the 'static' aspect of the router configuration to be able to built in AOT mode.
I rolled-back my initial change that was using a method call, with 'hardcoded' static definitions.

pkozlowski-opensource:
Discussed this and the general consensus in the team is that we like it 👍 

I'm going to merge it, thnx @benouat !

PR #1646, opened 2017-06-30
feat(NgbTime): add compare methods

irossimoline:
Add 'before' and 'after' methods to NgbTime, that can be used to compare times. These methods can be used in a further implementation of time ranges, or in validation for other specific purposes.
There is not a way to compare Times on current NgbTime class.

### DIFF
--- a/src/timepicker/ngb-time.spec.ts
+++ b/src/timepicker/ngb-time.spec.ts
@@ -219,4 +219,37 @@ describe('NgbTime', () => {
 
   it('should have a validity flag with optional seconds checking',
      () => { expect(new NgbTime(11, 0).isValid(false)).toBeTruthy(); });
+
+  it('should compare times, if seconds not specified in any compare only by hour and minutes', () => {
+    const t = new NgbTime(11, 15, 30);
+    expect(t.before(t)).toBeFalsy();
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
> These methods can be used in a further implementation of time ranges, or in validation for other specific purposes. There is not a way to compare Times on current NgbTime class.

@irossimoline do you have more specific info on those "further implementation"s ? The thing is that if we don't introduce changes you are hinting at that this is pretty much dead code that contributes to the library size but doesn't serve any useful purpose.

I see how those methods can be useful when adding more functionality to timepicker. But again, without this additional functionality this is just dead code... As such I'm not keen on merging it right now. On the other hand if you would be willing to elaborate and work on more timepicker features, than we could introduce methods you are proposing here.

Care to elaborate?

irossimoline:
@pkozlowski-opensource I am willing to implement at least a min and max time configuration in timepicker. That would be a useful feature by it's own, in order to prevent to change the time beyond of those boundaries.

My goal with this is to implement time ranges, and i think that the implementation of this feature is a good first approach, and useful for the current use of timepicker. On a next step, I'll propose the implementation of a timepicker inpup (following the same style as datepicker). In this case, I consider a popup very useful to make a group of time pickers more readables.

PR #1670, opened 2017-07-10
feat(datepicker): using DatePipe for generating months and weekdays

Fredx87:
Translations of weekdays and months are generated automatically by
NgbDatepickerI18nDefault class, using the default Date pipe of
Angular 2+, instead of forcing the user to provide translations
for each locale that he wants to support.
The customization of the translations is still available by
extending the base class.

Note: is not possible to show a demo of the datepicker translated with a LOCALE_ID different from en-US, because LOCALE_ID can be set only on BrowserModule: https://github.com/angular/angular/issues/16960. So the only way to show a demo maybe is an external plunker.

### DIFF
--- a/demo/src/app/components/datepicker/demos/calendars/datepicker-calendars.ts
+++ b/demo/src/app/components/datepicker/demos/calendars/datepicker-calendars.ts
@@ -1,7 +1,7 @@
 import {Component, Injectable} from '@angular/core';
 import {NgbDateStruct, NgbCalendar, NgbCalendarIslamicCivil, NgbDatepickerI18n} from '@ng-bootstrap/ng-bootstrap';
 
-const WEEKDAYS_SHORT = ['Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa', 'Su'];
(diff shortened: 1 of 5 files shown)


Fredx87:
About the errors on CI: 

- Safari 9 and IE 10 don't support Intl API, so for them the polyfill is required (see https://github.com/angular/angular/issues/10809)
- There is an issue on Safari 10 that return wrong months and day of week names, I can't make tests because a don't have a Mac, maybe it is related with this issue: https://github.com/angular/angular/issues/17576

Anyway I think that having months and days of week translated automatically can be a useful feature, and the only way to do it without using external libraries is using the Date Pipe (other libraries uses moment, but it a new dependency).

pkozlowski-opensource:
Thnx for the suggestion but the things is that Intl pipes are being totally re-worked on the Angular side as we speak. In the (hopefully near) future we are going to have a service that will allow us to pull i18n settings and data and use those in [our  impl](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-i18n.ts)

Given the above I'm going to pass on this PR since merging it would mean that we have to rework code and introduce breaking changes (current pipes from Angular are rather buggy and the new ones will change behaviour in a subtle way).

Thnx for the PR though!

Issue #1596, opened 2017-05-29
Build with Closure compiler 


pkozlowski-opensource:
This issue is here to track all the necessary changes in this library to make it work with the closure compiler with the least pain possible. 

My current understanding is that require changes, so far, boil down to "stringifying" property access in some objects.

dmitrysteblyuk:
I'm currently using ng-bootstrap with closure compiling. I found some issues with Modal, Dropdown, Datepicker components and util/positioning.
I had to add some externs because of those issues (like accessing components inputs or object properties using quotes syntax whereas those properties where defined without quotes). Here they are:

var NgbModalRef = function () {};
NgbModalRef.prototype.componentInstance;

var NgbDropdown = function () {};
NgbDropdown.prototype.toggleElement;

// util/positioning.js

var shiftWidth = function () {};
shiftWidth.prototype.left;
shiftWidth.prototype.center;
shiftWidth.prototype.right;

var shiftHeight = function () {};
shiftWidth.prototype.top;
shiftWidth.prototype.center;
shiftWidth.prototype.bottom;

var NgbDatepicker = function () {};
NgbDatepicker.prototype.dayTemplate;
NgbDatepicker.prototype.displayMonths;
NgbDatepicker.prototype.firstDayOfWeek;
NgbDatepicker.prototype.markDisabled;
NgbDatepicker.prototype.minDate;
NgbDatepicker.prototype.maxDate;
NgbDatepicker.prototype.navigation;
NgbDatepicker.prototype.outsideDays;
NgbDatepicker.prototype.showNavigation;
NgbDatepicker.prototype.showWeekdays;
NgbDatepicker.prototype.showWeekNumbers;

var NgbDatepickerInput = function () {};
NgbDatepickerInput.prototype.dayTemplate;
NgbDatepickerInput.prototype.displayMonths;
NgbDatepickerInput.prototype.firstDayOfWeek;
NgbDatepickerInput.prototype.markDisabled;
NgbDatepickerInput.prototype.minDate;
NgbDatepickerInput.prototype.maxDate;
NgbDatepickerInput.prototype.navigation;
NgbDatepickerInput.prototype.outsideDays;
NgbDatepickerInput.prototype.showNavigation;
NgbDatepickerInput.prototype.showWeekdays;
NgbDatepickerInput.prototype.showWeekNumbers;

dmitrysteblyuk:
[...]

PR #1674, opened 2017-07-11
fix(typeahead): avoid unecessery re-creation of DOM nodes

### LINKED ISSUE #1659: TypeAhead is broken - cant select a row in search result if click on the actual text

### Bug description:
TypeAhead is broken - cant select a row in search result if click on the actual text.  If however you click on the non text area in the blue row, it will actually select the row and generate the select event. see it also broken in the demo. See attached JPG showing issue.
img0000

Please note that we can _not

_ act on bug reports without a minimal reproduction scenario in plunker. Here is why:

Angular: 4.

ng-bootstrap: latest

Bootstrap:

pkozlowski-opensource:
#1612 introduced regression (see #1659 for more details). At the end of the day the proper fix need to come from the Angular itself, we are going to track closure support in #1596)

Reverts part of #1612
Fixes #1659

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -32,7 +32,7 @@ export interface ResultTemplateContext {
         (mouseenter)="markActive(idx)"
         (click)="select(result)">
           <ng-template [ngTemplateOutlet]="resultTemplate || rt"
-          [ngOutletContext]="_getResultContext(result)"></ng-template>
+          [ngOutletContext]="{result: result, term: term, formatter: formatter}"></ng-template>
       </button>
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@mlaval would you mind taking a look?

pkozlowski-opensource:
Additional note: https://github.com/angular/angular/issues/14292 is fixed on the Angular side so the work-around I'm reverting here should not be needed any more.

pkozlowski-opensource:
LGTM here: https://github.com/ng-bootstrap/ng-bootstrap/issues/1659#issuecomment-314390602

PR #1675, opened 2017-07-11
demo(datepicker): range selection demo for inline datepicker

maxokorokov:
A simple demo for the inline datepicker that shows how one to do range selection with existing API

range


pkozlowski-opensource:
LGTM, thnx!

Matmo10:
I suppose this is mainly for small ranges? I would imagine you can't drag from 2017 to 2019, for example.

maxokorokov:
@Matmo10 check the working [example](https://ng-bootstrap.github.io/#/components/datepicker/examples) on the demo site, you definitely can. Plus demo is just a starting point if you want to implement it differently.

PR #1678, opened 2017-07-12
fix(buttons): make buttons selectors more specific

### LINKED ISSUE #1125: NgbModule.forRoot() breaks 'traditional' style radio buttons

pkozlowski-opensource:
BREAKING CHANGE:

Selectors for radio buttons related directives were changed and now both label
and input require ng-bootstrap specific attributes as selectors.

Before:

```
<div [(ngModel)]="model" ngbRadioGroup>
  <label class="btn">
    <input type="radio" name="radio" [value]="values[0]"/> {{ values[0] }}
  </label>
  <label class="btn">
    <input type="radio" name="radio" [value]="values[1]"/> {{ values[1] }}
  </label>
</div>
```

After:

```
<div [(ngModel)]="model" ngbRadioGroup>
  <label ngbButtonLabel>
    <input ngbButton type="radio" name="radio" [value]="values[0]"/> {{ values[0] }}
  </label>
  <label ngbButtonLabel>
    <input ngbButton type="radio" name="radio" [value]="values[1]"/> {{ values[1] }}
  </label>
</div>
```

Notice new `ngbButtonLabel` and `ngbButton` attributes that act as new selectors.

Fixes #1125

### DIFF
--- a/src/buttons/radio.ts
+++ b/src/buttons/radio.ts
@@ -1,4 +1,4 @@
-import {Directive, forwardRef, Optional, Input, Renderer2, ElementRef, OnDestroy} from '@angular/core';
+import {Directive, forwardRef, Input, Renderer2, ElementRef, OnDestroy} from '@angular/core';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
 
 const NGB_RADIO_VALUE_ACCESSOR = {
@@ -57,8 +57,8 @@ export class NgbRadioGroup implements ControlValueAccessor {
 }
 
 
-@Directive({selector: 'label.btn'})
-export class NgbActiveLabel {
+@Directive({selector: '[ngbButtonLabel]', host: {'[class.btn]': 'true'}})
+export class NgbButtonLabel {
   constructor(private _renderer: Renderer2, private _elRef: ElementRef) {}
 
   set active(isActive: boolean) {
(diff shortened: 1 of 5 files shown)


src/buttons/radio.ts
```diff
@@ -57,8 +57,8 @@ export class NgbRadioGroup implements ControlValueAccessor {
 }
 
 
-@Directive({selector: 'label.btn'})
-export class NgbActiveLabel {
+@Directive({selector: '[ngbButtonLabel]', host: {'[class.btn]': 'true'}})
```
maxokorokov: `class: btn`

pkozlowski-opensource: This would potentially overwrite classes that people would like to set on the `<label>` element, so I would prefer to leave it as-is.

pkozlowski-opensource:
@maxokorokov  thnx for the review - I've completely forgotten about the demo part!!! Pushed amended commit with the demo. 

Regarding renames - I'm going to open another PR with the checkbox buttons where (hopefully) the existing names will make more sense. Hold a sec.

PR #1684, opened 2017-07-15
fix(timepicker): switch to PM when hour > 12

### LINKED ISSUE #1680: Timepicker does not recognize AM/PM change (Meridian)
### Bug description:
Using a  timepicker with `meridian` set to true, I expect that entering "22" as the hour results in an update to "10" and a switch from "AM" to "PM" and vice versa. This does not happen in the current version.
Using the spinner to update the time, the switch from AM to PM (and vice versa) works.

See demo page: https://ng-bootstrap.github.io/#/components/timepicker/examples

Angular: 4.2.6

ng-bootstrap: 1.0.0-alpha.28

Bootstrap: 4.0.0-alpha.6

dmytroyarmak:
Fixes #1680

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -224,7 +224,11 @@ export class NgbTimepicker implements ControlValueAccessor,
   updateHour(newVal: string) {
     const isPM = this.model.hour >= 12;
     const enteredHour = toInteger(newVal);
-    this.model.updateHour((this.meridian ? enteredHour % 12 : enteredHour) + (this.meridian && isPM ? 12 : 0));
+    if (this.meridian && (isPM && enteredHour < 12 || !isPM && enteredHour === 12)) {
+      this.model.updateHour(enteredHour + 12);
+    } else {
+      this.model.updateHour(enteredHour);
+    }
     this.propagateModelChange();
   }
 
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -578,6 +578,172 @@ describe('ngb-timepicker', () => {
                expect(meridianButton.innerHTML).toBe('AM');
              });
        }));
+
+
+    it('should respect meridian when propagating model (PM)', async(() => {
+         const html = `<ngb-timepicker [(ngModel)]="model" [meridian]="true"></ngb-timepicker>`;
+
+         const fixture = createTestComponent(html);
+         fixture.componentInstance.model = {hour: 14, minute: 30};
+         fixture.detectChanges();
+
+         const inputs = fixture.debugElement.queryAll(By.css('input'));
+
+         fixture.whenStable()
+             .then(() => {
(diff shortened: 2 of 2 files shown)


src/timepicker/timepicker.ts
```diff
@@ -224,7 +224,11 @@ export class NgbTimepicker implements ControlValueAccessor,
   updateHour(newVal: string) {
     const isPM = this.model.hour >= 12;
     const enteredHour = toInteger(newVal);
-    this.model.updateHour((this.meridian ? enteredHour % 12 : enteredHour) + (this.meridian && isPM ? 12 : 0));
+    if (this.meridian && (isPM && enteredHour < 12 || !isPM && enteredHour === 12)) {
+      this.model.updateHour(enteredHour + 12);
```
pkozlowski-opensource: What would happen if I enter, say 30? I don't think that this logic here is enough...

pkozlowski-opensource:
@dmytroyarmak thnx for the PR. I can see how it would solve the existing issue, but IMO if we want to fix it one and for all we need to decide what happens if a user enters _any_ number > 12 for the meridian on and AM / PM being selected. As such we should have tests for all those cases.

Would you be willing to work more on this PR and add more tests that prove proper behaviour in all corner cases?

src/timepicker/timepicker.ts
dmytroyarmak: if you enter `30` it will try to set it as an hour without any corrections and `updateHour` will do `% 24`, so the result will be `6`, so it will switch to AM.

dmytroyarmak:
@pkozlowski-opensource yeah, I can do more work on this PR
So, how it works in my PR is:
```
if current mode is AM then
  if 0 <= entered value < 12 then
    save value as is in 24h format (no correction needed)
  else if entered value is 12 then
    store it as 0 in 24h format (in current implementation we add 12 to it and then `mod 24` transform it to 0)
  else if entered value > 12 then
    store it as is in 24h format by doing `mod 24`

else if current mode is PM then
  if 0 <= entered value < 12 then
    add 12 and save
  else if  entered value > 12
    store it as is in 24h format by doing `mod 24`
```

So, in general, if value is < 12 (and equal to 12 for AM) we store it as a correct value in current meridian by doing needed corrections to transform it to 24h format. But if value is > 12, we just save its remainder of division by 24 as value in 24h format.

@pkozlowski-opensource If this logic is correct I can cover not-covered already cases with unit tests.

pkozlowski-opensource:
@dmytroyarmak Yup, your reasoning looks fine! If you could please add more tests for the listed cases it would be great. I will merge your PR as soon as we've got tests covering all the cases.

Thnx so much for helping it, this is much appreciated! 👍

dmytroyarmak:
@pkozlowski-opensource I found that there were already some test about handling different values in meridian, I've moved them to 'meridian' section and added missed cases

PR #1685, opened 2017-07-15
feat(modal): add injector option

matt-duch:
I have an application whose components depend heavily on scoped providers, which makes it difficult to reuse those components within a modal. This has also come up on stack overflow (we do something similar with a reflective injector):
https://stackoverflow.com/questions/42568336/how-to-open-a-component-in-ng-bootstrap-modal-using-a-specific-injector

We made a small change: add a property 'injector: Injector' to NgbModalOptions and use that within NgbModalStack.open
this._getContentRef(moduleCFR, contentInjector, content, activeModal);
-->
this._getContentRef(moduleCFR, options.injector || contentInjector, content, activeModal);

### DIFF
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -37,7 +37,7 @@ export class NgbModalStack {
     }
 
     const activeModal = new NgbActiveModal();
-    const contentRef = this._getContentRef(moduleCFR, contentInjector, content, activeModal);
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
LGTM. This change makes sense, I wanted to add it at some point but wanted to wait till someone comes up with a real-life use-case.

Merging, thnx!

PR #1687, opened 2017-07-17
demo: fixes not working 'getting-started' url

maxokorokov:
Bug was because of the following redirect chain → `/getting-started` → `../api` → nothing matches → `''` → `home`

Content-wrapper seems to be used both as a wrapper for components and for normal pages → so if there no `component` input, doesn't make sense to listen to router events and trying to redirect to `api` by default

I guess this should be refactored further as we add more normal pages...

### DIFF
--- a/demo/src/app/components/accordion/accordion.component.ts
+++ b/demo/src/app/components/accordion/accordion.component.ts
@@ -4,7 +4,7 @@ import {DEMO_SNIPPETS} from './demos';
 @Component({
   selector: 'ngbd-accordion',
   template: `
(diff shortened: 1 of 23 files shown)


maxokorokov:
Actually give me 5 min, I'll split it properly into `page-wrapper` and `component-wrapper`

maxokorokov:
So I replaced `content-wrapper` by:
- `component-wrapper` that has `api/examples` tabs, jumbotron and component nav
- `page-wrapper` without tabs, but with jumbotron and component nav

PR #1688, opened 2017-07-18
feat(buttons): add support for checkbox buttons

### LINKED ISSUE #890: Button groups does not support focus customization

pkozlowski-opensource:
Closes #890


pkozlowski-opensource:
@maxokorokov could you PTAL?

src/buttons/checkbox.ts
```diff
@@ -0,0 +1,79 @@
...
+    '(focus)': 'focused = true',
+    '(blur)': 'focused = false'
+  },
+  providers: [NGB_CHECKBOX_VALUE_ACCESSOR]
+})
+export class NgbCheckBox implements ControlValueAccessor {
+  checked;
+
+  /**
+   * A flag indicating if a given radio button is disabled.
```
maxokorokov: radio button → checkbox

src/buttons/label.ts
```diff
@@ -0,0 +1,30 @@
+import {Directive, Renderer2, ElementRef} from '@angular/core';
+
+@Directive({selector: '[ngbButtonLabel]', host: {'[class.btn]': 'true'}})
+export class NgbButtonLabel {
+  constructor(private _renderer: Renderer2, private _elRef: ElementRef) {}
```
maxokorokov: True that we could have done it with 3 fields and host bindings, don't really need Renderer, right?

src/buttons/checkbox.ts
```diff
@@ -0,0 +1,79 @@
...
+  multi: true
+};
+
+
+/**
+ * Easily create Bootstrap-style checkbox buttons. A value of a checked button is bound to a variable
+ * specified via ngModel.
+ */
+@Directive({
+  selector: '[ngbButton][type=checkbox]',
```
maxokorokov: Ha, that's a much better solution!

src/index.ts
```diff
@@ -27,7 +27,7 @@ export {
   NgbPanelContent
 } from './accordion/accordion.module';
 export {NgbAlertModule, NgbAlertConfig, NgbAlert} from './alert/alert.module';
-export {NgbButtonsModule, NgbRadioGroup} from './buttons/radio.module';
+export {NgbButtonsModule, NgbRadioGroup} from './buttons/buttons.module';
```
maxokorokov: Seems strange, that we have only `NgbRadioGroup` here

src/buttons/label.ts
pkozlowski-opensource: Yep, will send a follow up PR to refactor it. Just didn't want to make too many changes in one PR.

src/index.ts
pkozlowski-opensource: This is so people can query for it. We could add the new checkbox button here as well.

PR #1694, opened 2017-07-20
added safe nav operator

RonNewcomb:
line 129


pkozlowski-opensource:
Closing in favour of #1699 as it has a well identified root cause and the associated test.

PR #1697, opened 2017-07-24
feat(tooltip&popover): Automatic positioning

ababakov:
Would like to help with automatic positioning of tooltip&popover

Have a few odd moments with ngZone - moved class generation to ngZone.onStable, but not pretty sure i'm right. Need an advice if it's ok


pkozlowski-opensource:
@ababakov thnx for the PR but I've decided to merge #1782 as it had tests, documentation and type updates. Even if not merged we _do_ appreciate all PRs so please keep them coming!

Issue #1712, opened 2017-07-30
datepicker i18n demo is broken


jnizet:
The demo is broken for two reasons:

- the breaking changes brought to ngbRadio haven't been applied in this demo
- the OnPush strategy prevents the i18ned labels to change

Regarding the first problem, I will submit a PR.

The second problem is trickier: you need to interact with the datepicker and its navigation to actually see the language change become effective.

One might argue that the use-case of the demo is not very realistic (change the language on the fly, in a page displaying a datepicker), especially given that native angular i18n requires one separate application per locale, but the demo feels broken, and I don't have a clear idea on how best to fix it.

maxokorokov:
Hey, @jnizet!
I've seen this also and would agree with your last passage. Maybe just use the french example from the start without any switching? I can't come up with any simple solution for this example either.

jnizet:
OK. I'll do that tomorrow

PR #1713, opened 2017-07-30
demo(datepicker): fix usage of ngbRadioGroup in i18n demo

jnizet:
partial fix for #1712


pkozlowski-opensource:
I think that we rather should do https://github.com/ng-bootstrap/ng-bootstrap/issues/1712#issuecomment-319004240 - that is, remove radios altogether.

Would you be up to sending another PR that removes the switch and just keeps the French version?

jnizet:
Yes, I can do that. Will do tomorrow.

PR #1717, opened 2017-07-31
fix(datepicker): focus datepicker after opening it inside the popup

### LINKED ISSUE #1708: Datepicker navigation by keyboard
Currently, datepicker doesn't support navigation by press arrows and selecting a date by press enter.

maxokorokov:
- The datepicker will de focused when opened in the popup, so keyboard navigation would work right from the start without clicking on anything first
- Introduce `focus()` method to programmatically focus datepicker

For this to work synchronously the default `tabindex: 0` is required on the datepicker

Fixes #1708


pkozlowski-opensource:
LGTM

Issue #1714, opened 2017-07-30
Reactive form demos of checkbox and radio buttons don't open in plnkr


jnizet:
The two reactive forms demos of the buttons module don't open in plunkr. They result in a 404.

The datepicker range demo opens in plunkr, but throws an exception.

maxokorokov:
```typescript
export const DEMO_SNIPPETS = {
  'radioReactive': {
    'code': require('!!prismjs-loader?lang=typescript!./radio-reactive/buttons-radio-reactive'),
    'markup': require('!!prismjs-loader?lang=markup!./radio-reactive/buttons-radio-reactive.html')}
};
```

Haven't been digging into how Plnkr generator works, but in the example above to avoid 404 → `radioReacive` key should become `radio-reactive`. However in this case in the plunkr it will try to use `NgbdButtonsRadio-reactive` component, because too much dashes in the name (?)

So my guess is the solution would be either fixing plnkr generation or renaming demos to `buttons-radioreactive` and `buttons-checkboxreactive`

PR #1718, opened 2017-07-31
demo(datepicker): range demo opens in Plnkr without exceptions

maxokorokov:
There were two issues:
- external `tools.ts` was not opening obviously - so inlined it into component
- `ngbd-datepicker-range-inline` had one slash too much for the parser, so renamed it to `ngbd-datepicker-range`

Part of #1714


pkozlowski-opensource:
LGTM

PR #1726, opened 2017-08-02
demo(datepicker): use buttons for calendar icon

pkozlowski-opensource:
`<div>`s are not focusable with keyboard so let's use
`<buttons>`s instead.


pkozlowski-opensource:
Good catch @maxokorokov , thnx! Will add the `type` and merge this one.

PR #1731, opened 2017-08-05
feat(typeahead): add support for the 'container' input

### LINKED ISSUE #1105: Typeahead: support like 'append-to-body'
### Bug description: Feature Request

I hope to support like `typeahead-append-to-body` of ui-bootstrap(for Angular1)
because typeahead window in modal window is clipped.

Angular: 2.2.4

ng-bootstrap: 1.0.0.beta-14

pkozlowski-opensource:
Closes #1105

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -91,6 +91,12 @@ export class NgbTypeahead implements ControlValueAccessor,
 
 
   /**
+   * A selector specifying the element the tooltip should be appended to.
+   * Currently only supports "body".
+   */
+  @Input() container: string;
+
+  /**
    * A flag indicating if model values should be restricted to the ones selected from the popup only.
    */
   @Input() editable: boolean;
@@ -142,6 +148,7 @@ export class NgbTypeahead implements ControlValueAccessor,
       private _elementRef: ElementRef, private _viewContainerRef: ViewContainerRef, private _renderer: Renderer2,
       private _injector: Injector, componentFactoryResolver: ComponentFactoryResolver, config: NgbTypeaheadConfig,
       ngZone: NgZone) {
+    this.container = config.container;
     this.editable = config.editable;
     this.focusFirst = config.focusFirst;
     this.showHint = config.showHint;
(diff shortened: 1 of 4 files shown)


src/typeahead/typeahead.ts
```diff
@@ -91,6 +91,12 @@ export class NgbTypeahead implements ControlValueAccessor,
 
 
   /**
+   * A selector specifying the element the tooltip should be appended to.
+   * Currently only supports "body".
+   */
+  @Input() container: string;
```
maxokorokov: 1. Can't we limit it to `@Input() container: 'body';` in this case?
2. You should probably add it to the `NgbTypeaheadConfig`

src/typeahead/typeahead.ts
```diff
@@ -241,6 +250,10 @@ export class NgbTypeahead implements ControlValueAccessor,
       this._windowRef.instance.id = this.popupId;
       this._windowRef.instance.selectEvent.subscribe((result: any) => this._selectResultClosePopup(result));
       this._windowRef.instance.activeChangeEvent.subscribe((activeId: string) => this.activeDescendant = activeId);
+
+      if (this.container === 'body') {
+        window.document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
```
wesleycho: We probably want to inject DOCUMENT and Renderer2 and make use of those

pkozlowski-opensource: Current `Renderer2` interface doesn't have any method(s) to query nodes and / or call an arbitrary method on a DOM node: https://github.com/angular/angular/blob/10897d64738484036443f6614ad3cbef095da0d5/packages/core/src/render/api.ts#L147-L191

src/typeahead/typeahead.ts
pkozlowski-opensource: > Can't we limit it to @Input() container: 'body'; in this case?

I've modeled the proposed API after all the other components that have the `container` option (tooltip, popover etc.). We can change it across the board in a separate PR.

> You should probably add it to the NgbTypeaheadConfig

Yup, good idea 👍

PR #1743, opened 2017-08-11
feat(dropdown): distinguish inside and outside clicks for autoClose

pkozlowski-opensource:
Closes #1022

BREAKING CHANGE:

Dropdown menu now requires usage of the new `ngbDropdownMenu` directive.

Before:

```html
<div ngbDropdown>
  <button ngbDropdownToggle>Toggle dropdown</button>
  <div class="dropdown-menu">
    <a class="dropdown-item">Action</a>
  </div>
</div>
```

After (notice **ngbDropdownMenu**):

```html
<div ngbDropdown>
  <button ngbDropdownToggle>Toggle dropdown</button>
  <div ngbDropdownMenu>
    <a class="dropdown-item">Action</a>
  </div>
</div>
```

PR #1745, opened 2017-08-11
Bootstrap4 beta dropdowns

pkozlowski-opensource:
This builds on top of #1743 and adds Bootstrap 4.beta specific markup changes (the `show` class needs to be added on the menu now).

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -308,11 +308,11 @@ describe('ngb-dropdown-toggle', () => {
     expect(dropdownEl).toHaveCssClass('show');
   });
 
-  it('should close on item click if autoClose is set to false', () => {
+  it('should not close on item click if autoClose is set to false', () => {
     const html = `
       <div ngbDropdown [open]="true" [autoClose]="false">
           <button ngbDropdownToggle>Toggle dropdown</button>
-          <div class="dropdown-menu" aria-labelledby="dropdownMenu1">
+          <div ngbDropdownMenu>
             <a class="dropdown-item">Action</a>
           </div>
       </div>`;
@@ -330,11 +330,11 @@ describe('ngb-dropdown-toggle', () => {
     expect(dropdownEl).toHaveCssClass('show');
   });
 
-  it('should close on item click', () => {
+  it('should close on item click by default', () => {
     const html = `
       <div ngbDropdown [open]="true">
           <button ngbDropdownToggle>Toggle dropdown</button>
-          <div class="dropdown-menu" aria-labelledby="dropdownMenu1">
+          <div ngbDropdownMenu aria-labelledby="dropdownMenu1">
             <a class="dropdown-item">Action</a>
           </div>
       </div>`;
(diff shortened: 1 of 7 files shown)


src/dropdown/dropdown.ts
```diff
@@ -26,7 +74,7 @@ export class NgbDropdown {
   /**
    * Indicates that dropdown should be closed when selecting one of dropdown items (click) or pressing ESC.
    */
-  @Input() autoClose: boolean;
+  @Input() autoClose: boolean | 'outside' | 'inside';
```
maxokorokov: Should update documentation for 2 new values

src/dropdown/dropdown.ts
```diff
@@ -1,7 +1,53 @@
...
+} from '@angular/core';
 import {NgbDropdownConfig} from './dropdown-config';
 
 /**
+ */
+@Directive({selector: '[ngbDropdownMenu]', host: {'[class.dropdown-menu]': 'true', '[class.show]': 'isOpen'}})
+export class NgbDropdownMenu {
+  isOpen = false;
+
+  constructor(private _elementRef: ElementRef) {}
```
maxokorokov: Can't you inject the `NgbDropdown` as you do in the `NgbDropDownToggle` and remove internal `isOpen` state which is duplicated now?

Would it work if you just use `[class.show]: 'dropdown.isOpen()'`?
Can't see why you want to explicitly have `open` and `show` methods

src/dropdown/dropdown.ts
```diff
@@ -93,30 +162,19 @@ export class NgbDropdown {
...
-})
-export class NgbDropdownToggle {
-  constructor(public dropdown: NgbDropdown, elementRef: ElementRef) {
-    dropdown.toggleElement = elementRef.nativeElement;
+    return false;
   }
 
-  toggleOpen() { this.dropdown.toggle(); }
+  private _isEventFromMenu($event) {
+    if (this._menu) {
```
maxokorokov: can be a ternary one-liner, but don't know if like it :)

src/dropdown/dropdown.spec.ts
```diff
@@ -15,8 +15,41 @@ function getDropdownEl(tc) {
...
+        };
+      }
+    };
+  }
+};
+
 describe('ngb-dropdown', () => {
+
   beforeEach(() => {
+    jasmine.addMatchers(jasmineMatchers);
```
maxokorokov: I think it could be done once in `beforeAll`

src/dropdown/dropdown.ts
pkozlowski-opensource: Yeh, changing to ternary

src/dropdown/dropdown.ts
pkozlowski-opensource: Great catch, it is much simpler indeed! Changing.

src/dropdown/dropdown.spec.ts
pkozlowski-opensource: Will open a separate issue / PR for this as I've noticed other things in this test that need attention.

pkozlowski-opensource:
> The dropup is not working. What would be the plan to fix it?

I will open a separate issue for this as a proper fix will require usage of the positioning service and we need to design it properly.

pkozlowski-opensource:
Thnx for the review @maxokorokov - great feedback. Waiting for Travis to get green and merging it. There are 2 more points needed attention:
* dropups
* test improvements
but let's tackle those separately - will open issues.

pkozlowski-opensource:
Merged into the `bootstrap4_beta` branch

Issue #1171, opened 2017-01-03
Dropdown menu cropped and not automatically adjusted to left/right 


sinedied:
### Bug description:

When using the predecessor `ui-bootstrap` dropdown component, the dropdown menu would automatically switch to being on the left or right if needed to avoid being cropped.

This is not the case with the current `ng-bootstrap` dropdown component, as seen in the provided plunker: when the dropdown is floated the right, the menu is cropped and goes off-screen instead of being switch to the right.

This is problematic in responsive scenario, like when using a right-floated dropdown in a responsive navbar with a collapse: the dropdown menu should be attached to the right when in "desktop mode" and the dropdown is floated to the right, and it should be on the left when the dropdown is not floated (in the collapsed menu). The scenario was working without additional code in Bootstrap 3 with ui-bootstrap.

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/Jb2MzMSjOXt2M6UJkwq9?p=preview

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 2.4.1

ng-bootstrap: 1.0.0-alpha.15

Bootstrap: 4.0.0-alpha.5

nixallover:
This might be related to an issue that I'm seeing as well. I just updated from ng-bootstrap 1.0.0-alpha.13 to 1.0.0-alpha.18 and it seems that the dropdown is no longer visible, which also seems like it's the case in this plunkr.

salacis:
Thats still a thing in alpha 6.

pkozlowski-opensource:
This will be fixed by #1814

PR #1752, opened 2017-08-16
fix(dropdown): restore support for dropups

### LINKED ISSUE #1747: fix dropups and add positioning
dropdowns are not dropping up with Bootstrap 4.beta. The fix is not trivial as we need to use the positioning service.

pkozlowski-opensource:
Fixes #1747
Part of #1171
Part of #1012

BREAKING CHANGE:

The `up` input is no longer supported by you can use more flexible
`placement` setting now.

Before:

```html
<div ngbDropdown [up]="true">
```

After:

```html
<div ngbDropdown placement="top-right">
```

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -53,13 +53,6 @@ describe('ngb-dropdown', () => {
     TestBed.configureTestingModule({declarations: [TestComponent], imports: [NgbDropdownModule.forRoot()]});
   });
 
-  it('should initialize inputs with provided config', () => {
-    const defaultConfig = new NgbDropdownConfig();
-    const dropdown = new NgbDropdown(defaultConfig);
-    expect(dropdown.up).toBe(defaultConfig.up);
-    expect(dropdown.autoClose).toBe(defaultConfig.autoClose);
-  });
-
   it('should be closed and down by default', () => {
     const html = `<div ngbDropdown><div ngbDropdownMenu></div></div>`;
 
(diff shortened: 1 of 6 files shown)


src/dropdown/dropdown.spec.ts
```diff
@@ -70,8 +63,8 @@ describe('ngb-dropdown', () => {
     expect(compiled).not.toBeShown();
   });
 
-  it('should be up if up input is true', () => {
-    const html = `<div ngbDropdown [up]="true"></div>`;
+  it('should have dropup CSS class is placed on top', () => {
```
maxokorokov: don't really understand what you mean. is → if ?

src/dropdown/dropdown.ts
```diff
@@ -37,7 +39,11 @@ export class NgbDropdownMenu {
   }
 })
 export class NgbDropdownToggle {
-  constructor(@Inject(forwardRef(() => NgbDropdown)) public dropdown, private _elementRef: ElementRef) {}
+  anachorEl;
```
maxokorokov: typo → anchor

src/dropdown/dropdown.ts
```diff
@@ -83,16 +86,26 @@ export class NgbDropdown {
   @Input('open') _open = false;
 
   /**
+   * Placement of a dropdown. Use "top-right" for dropups.
```
maxokorokov: Again, I think you should either list the acceptable values in the doc or type it appropriately

src/dropdown/dropdown.ts
pkozlowski-opensource: Fixed.

src/dropdown/dropdown.spec.ts
pkozlowski-opensource: Yep, fixed.

src/dropdown/dropdown.ts
pkozlowski-opensource: Defintivelly. We are going to add appropriate type in a PR that introduces more positioning options (including `auto`).

PR #1757, opened 2017-08-17
docs: add warning about jQuery / Bootstrap JS

### LINKED ISSUE #1749: Documentation: Add "do not use plain Bootstrap JS files"
The Getting Started page gives all the info needed to use ng-bootstrap from a fresh app, but some people may only start using it on a work-in-progress application, and have Bootstrap 3 running.

Could just use a warning there on that page (and/or on the Radio Button page, which is where I ran into this issue), to *NOT* use the Bootstrap proper's JS files, as it results in certain conflicts and things not working as they should (in the case of radio buttons, flowing only to the template, but not back to the model).

I've been working on an internal application that already had the Bootstrap JS files imported, and wasn't getting any errors in the console with ng-bootstrap's radio controls, but they weren't working, and I've spent about 2 days trying to figure out why.

pkozlowski-opensource:
Closes #1749

### DIFF
--- a/demo/src/app/getting-started/getting-started.component.html
+++ b/demo/src/app/getting-started/getting-started.component.html
@@ -15,6 +15,12 @@
     </li>
   </ul>
 
+  <ngb-alert type="warning" [dismissible]="false">
+    <p>Should I add <b>bootstrap.js</b> or <b>bootstrap.min.js</b> to my project?</p>
+
+    No, the goal of ng-bootstrap is to <i>completely replace</i> JavaScript implementation for components. Nor should you include other dependencies like jQuery or popper.js. It is not necessary and might interfere with ng-bootstrap code.
+  </ngb-alert>
+
   <h3>
     Supported browsers
   </h3>


maxokorokov:
Your version:
screen shot 2017-08-21 at 22 42 48

My suggestion for wording and colours:
screen shot 2017-08-21 at 22 44 58

Here is the snippet if you want to take something from it:

```html
<ngb-alert type="warning" [dismissible]="false">
    <p>Should I add <b>bootstrap.js</b> or <b>bootstrap.min.js</b> to my project?</p>

      NO, the goal of ng-bootstrap is to <i>completely replace</i> JavaScript implementation for components. Nor should you include other dependencies like jQuery or popper.js. It is not necessary and might interfere with ng-bootstrap code.
</ngb-alert>
```

Foxandxss:
I like it, but we should say "No" and not "NO"

pkozlowski-opensource:
Thnx for all the inputs @maxokorokov and @Foxandxss - updated the PR with your suggestions.

jofran73:
Is mixing native bootstrap with ng-bootstrap a scenario that cannot be safely done?
I'm asking because I have a bootstrap site that has a common layout and navigation master. Only one page contains an Angular app as content of the master page, so I basically have both versions of bootstrap loaded. This is working so far but I don't know what future conflicts to expect.

I'm not sure if you are saying I should not load bootstrap.js (because it's unnecessary) or I must not load bootstrap.js (because it conflicts).
If there is a way to work around arising issues (e.g. with some kind of namespacing), I would very much appreciate any efforts/documentation to help support this scenario.

Issue #295, opened 2016-06-29
animations: initial push for animations


pkozlowski-opensource:
Some of the BS components need animations beyond simple CSS3 transitions (those are taken care of by BS classes). Components that need non-trivial animations are:
- [ ] Accordion
- [x] Alert => #1800
- [ ] Carousel
- [ ] Collapse
- [ ] Popover
- [ ] Modal
- [x] Tabset => #1802
- [ ] Tooltip

Foxandxss:
Animations that needs the @host can't be done as today until https://github.com/angular/angular/issues/9044 is fixed.

pkozlowski-opensource:
> Animations that needs the @host can't be done as today

Yeh, run into this one very quickly after playing with animations for a sec... Marking it as blocked on the upstream issues...

pkozlowski-opensource:
Blocked on https://github.com/angular/angular/issues/9044

aciccarello:
FYI, angular/angular#9044 was closed on July 14th

pkozlowski-opensource:
@aciccarello unfortunately  angular/angular#9044 added support for animations in `@Component` host but not on `@Directive`, sadly...

aciccarello:
Ah, so you must be waiting on angular/angular#9947 then.

pkozlowski-opensource:
> Ah, so you must be waiting on angular/angular#9947 then.

Yep.

rolandoldengarm:
Ah I thought they would do that in RC5?

Can you guys add the carousel to the list as well?

wesleycho:
@StanFaas please read the discussion & links provided.

An update will be posted if the status changes - consider silence to mean that this is still blocked.

colinmorelli:
Correct me if I'm wrong here, but I don't believe all of these items are blocked by the angular issue. Modals, for example, are based off of `@Component` which does have rich animation support. We should be able to support at least a stock animation for modal windows based on what already exists, no?

pkozlowski-opensource:
@colinmorelli you are totally right, not all animations / widgets are blocked. In fact I had a chat with Matias recently and IMO we are only blocked on:
* accordion, carousel and tabs that need `query` support
* collapse that need support for animations in directive hosts

Certain basic fade / slide animations should be doable right now so in theory it should possible to cover: Alert, Modal, Popover / Tooltip. It is just that till now we were focusing on stabilizing APIs to minimize breaking changes in the future. 

[...]

PR #1765, opened 2017-08-22
feat(modal): add animation

mattlewis92:

Any comments on how to improve this would be more than welcome! 😄 

Preview:

### DIFF
--- a/src/modal/modal-backdrop.ts
+++ b/src/modal/modal-backdrop.ts
@@ -1,5 +1,20 @@
 import {Component} from '@angular/core';
+import {trigger, transition, useAnimation} from '@angular/animations';
+import {fadeIn, fadeOut} from '../animations/fade';
+import {AnimationsHelper} from '../util/animations-helper';
 
-@Component({selector: 'ngb-modal-backdrop', template: '', host: {'class': 'modal-backdrop fade show'}})
+@Component({
+  selector: 'ngb-modal-backdrop',
+  template: '',
+  host: {
+    'class': 'modal-backdrop fade show',
+    '[@modalBackdrop]': 'animationsHelper.state',
+    '(@modalBackdrop.done)': 'animationsHelper.stateChanges.next($event)'
+  },
+  animations: [trigger(
+      'modalBackdrop',
+      [transition('void => enter', useAnimation(fadeIn)), transition('enter => exit', useAnimation(fadeOut))])]
+})
 export class NgbModalBackdrop {
+  animationsHelper = new AnimationsHelper();
 }
--- a/src/modal/modal-backdrop.spec.ts
+++ b/src/modal/modal-backdrop.spec.ts
@@ -1,10 +1,12 @@
 import {TestBed} from '@angular/core/testing';
+import {NoopAnimationsModule} from '@angular/platform-browser/animations';
 
 import {NgbModalBackdrop} from './modal-backdrop';
 
 describe('ngb-modal-backdrop', () => {
 
-  beforeEach(() => { TestBed.configureTestingModule({declarations: [NgbModalBackdrop]}); });
+  beforeEach(
+      () => { TestBed.configureTestingModule({declarations: [NgbModalBackdrop], imports: [NoopAnimationsModule]}); });
 
   it('should render backdrop with required CSS classes', () => {
     const fixture = TestBed.createComponent(NgbModalBackdrop);
--- a/src/modal/modal-ref.ts
+++ b/src/modal/modal-ref.ts
@@ -1,5 +1,6 @@
 import {Injectable, ComponentRef} from '@angular/core';
 
+
 import {NgbModalBackdrop} from './modal-backdrop';
 import {NgbModalWindow} from './modal-window';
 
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -48,17 +48,18 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/core": "4.0.3",
-    "@angular/common": "4.0.3",
-    "@angular/forms": "4.0.3"
+    "@angular/animations": "^4.3.5",
```
pkozlowski-opensource: This would be a breaking change for ng-bootstrap (bumping up minimal version of the dependency). Do you need `4.3.5`, specifically? What would be the earliest version of Angular that would work here?

Also, please don't use `^`.

src/animations/fade.ts
```diff
@@ -0,0 +1,7 @@
+import {animation, style, animate, AnimationReferenceMetadata} from '@angular/animations';
+
+export const fadeIn: AnimationReferenceMetadata =
+    animation([style({opacity: 0}), animate('.15s linear', style({opacity: 0.5}))]);
+
+export const fadeOut: AnimationReferenceMetadata =
+    animation([style({opacity: 0.5}), animate('.15s linear', style({opacity: 0}))]);
```
pkozlowski-opensource: Good move to extract those animations 👍

package.json
mattlewis92: I noted the breaking change in a separate commit: https://github.com/ng-bootstrap/ng-bootstrap/pull/1765/commits/61109dd45a0da30eba171d44333afd25d5484ca6 (I kept it this way so it could be cherry-picked to another branch and used to add additional animations in a further PR).

Looking at the change log, a minimum of 4.2.0 is required, however there have been a lot of bug fixes in future versions so maybe that's higher in reality. I will try it out with 4.2.0 and see if it works ok.

pkozlowski-opensource:
Not sure about usage of observables here as we are essentially communicating one async value here ("done"). Any particular reason to choose observables over promises here?

mattlewis92:
The original implementation did use promises, but I changed it in an attempt to make the tests not need the async changes (although that didn't work). I basically copied the implementation from material which also uses observables: https://github.com/angular/material2/blob/master/src/lib/dialog/dialog-container.ts#L94

pkozlowski-opensource:
> The original implementation did use promises, but I changed it in an attempt to make the tests not need the async changes (although that didn't work).

Yeh, saw that the tests still seem to need async so this argument is void. 

it seems to me that using promises would result in simpler code _and_ wouldn't force us to pull more dependencies from RxJS.

mattlewis92:
OK I will update the PR to just use promises then 😄

src/modal/modal-backdrop.ts
```diff
@@ -1,5 +1,29 @@
...
+  selector: 'ngb-modal-backdrop',
+  template: '',
+  host: {
+    'class': 'modal-backdrop fade show',
+    '[@modalBackdrop]': 'animationState',
+    '(@modalBackdrop.done)': 'animationStateChanged.next($event)'
+  },
+  animations: [trigger(
+      'modalBackdrop',
+      [transition('void => enter', useAnimation(fadeIn)), transition('enter => exit', useAnimation(fadeOut))])]
```
intellix: why not use `:enter` and `:leave`?

mattlewis92: You can't use `:leave` as you need to wait for the animation to complete before removing the element from the DOM, otherwise you won't see the exit animation. `:enter` could maybe be used in place of `void => enter` but they're subtly different as `:enter` is an alias for `void => *` so it would also match `void => exit` (that's not currently a valid state, but I think it's better to be explicit here for future maintenance)

mattlewis92:
@pkozlowski-opensource I've made the changes you requested, in the process I extracted the animation exit logic out into its own helper class that will make popover and tooltip animations trivial. The tab and alert animations should also now be very simple to add 🎉 I will work on follow up PRs once this is accepted and merged 😄

pkozlowski-opensource:
@mattlewis92 awesome, thnx for all the updates in this PR. After going over the code it all looks good to me and I don't see any red flags. Having said this I would prefer that we take small steps here and do the following:
* create a separate branch for animations 
* merge a PR that adds animations infrastrucure (angular animate etc.)
* add a couple of simple animations for places where the dynamic element creation / deletion is not needed (the simpler the better, ex.: alert)  
*  add a couple of simple animations for places where dynamic element creation / destruction is involved (ex. tooltip / popover)

Going through those basic steps would assure that we establish basic patterns so other people can jump into in help with the effort using the existing examples. In my view modal is one of the most complex (if not the most complex) use-cases when it comes to animations so I would prefer that we do it as _last_ rather than _first_.

How does it sound? Would you be up to extracting basic infrastructure piece from this PR into another PR (I could do it on my end by simply picking faf52d7)? Would you be up to working on other simpler animations before we merge this one?

Once again, thnx for the PR, much appreciate the effort!

mattlewis92:
That makes a lot of sense - would you be able to make a new branch off master on the main repo for these new animations to land in? 

I actually deliberately picked the modal animation first as it was the most complex - as I went along I made all the appropriate parts re-usable so that it will be trivial to use them for other animations. I guess I can pull those parts (fadeIn animations, animations-helper class) into a separate PR to merge into this new branch.

pkozlowski-opensource:
> That makes a lot of sense

Awesome, glad we are on the same page.

> would you be able to make a new branch off master on the main repo for these new animations to land in

Yep, I can create a new branch with faf52d7 or you can create a new PR with that commit + whatever other code you would like to have as a starting point. Let me know what works best for you.

>I guess I can pull those parts (fadeIn animations, animations-helper class) into a separate PR to merge into this new branch.

Yep, smaller PRs are much easier / quicker to review so I believe that we can move faster this way. Plus others could join the effort seeing small examples to follow.

Thnx for stepping in and helping out!

mattlewis92:
@pkozlowski-opensource here you go: https://github.com/ng-bootstrap/ng-bootstrap/pull/1786

kmathy:
Hi,
THanks for the great work!
Do you know when this feature will be released in the master branch? :)

mattlewis92:
I needed this today in my app as well so I made a friendly fork of this project until all the animations PRs get released properly `@mattlewis92/ng-bootstrap`

kukac7:
`npm install --save @ng-bootstrap/ng-bootstrap#animations`

rafaelss95:
@pkozlowski-opensource do you have an idea when this and others PRs related to animations can be merged?

nicco91:
How much do we need to wait more to just have a modal animation? This is so frustrating!!!

david-meza:
Thanks for all the great work. Is there a reason why this PR didn't make it into the master branch? If I can contribute in any way so animations can be used with the latest library version please let me know...

adam-marshall:
What is the status of this please?

kukac7:
@adam-marshall https://github.com/ng-bootstrap/ng-bootstrap/issues/295

PR #1772, opened 2017-08-24
feat(tooltip): Tooltip / popover - support more positioning options

mschoudry:

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -83,7 +83,7 @@ export class NgbPopover implements OnInit, OnDestroy {
    */
   @Input() popoverTitle: string;
   /**
-   * Placement of a popover. Accepts:
+   * Placement of a popover accepts:
    *    "top", "top-left", "top-right", "bottom", "bottom-left", "bottom-right",
    *    "left", "left-top", "left-bottom", "right", "right-top", "right-bottom"
    */
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -18,7 +18,7 @@ import {
 } from '@angular/core';
 
 import {listenToTriggers} from '../util/triggers';
-import {positionElements, PlacementType} from '../util/positioning';
+import {positionElements, Placement} from '../util/positioning';
 import {PopupService} from '../util/popup';
 import {NgbPopoverConfig} from './popover-config';
 
@@ -64,7 +64,7 @@ let nextId = 0;
   `]
 })
 export class NgbPopoverWindow {
-  @Input() placement: PlacementType = 'top';
+  @Input() placement: Placement = 'top';
   @Input() title: string;
   @Input() id: string;
 }
@@ -87,7 +87,7 @@ export class NgbPopover implements OnInit, OnDestroy {
    *    "top", "top-left", "top-right", "bottom", "bottom-left", "bottom-right",
    *    "left", "left-top", "left-bottom", "right", "right-top", "right-bottom"
    */
-  @Input() placement: PlacementType;
+  @Input() placement: Placement;
   /**
    * Specifies events that should trigger. Supports a space separated list of event names.
    */
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -18,7 +18,7 @@ import {
 } from '@angular/core';
 
 import {listenToTriggers} from '../util/triggers';
-import {positionElements} from '../util/positioning';
+import {positionElements, PlacementType} from '../util/positioning';
 import {PopupService} from '../util/popup';
 import {NgbPopoverConfig} from './popover-config';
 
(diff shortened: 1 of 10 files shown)


pkozlowski-opensource:
@mschoudry thnx for the PR. Could you please have a look at the CI failrue

> ERROR in /home/travis/build/ng-bootstrap/ng-bootstrap/src/util/positioning.ts (162,29): Exported type alias 'PlacementType' has or is using private name 'PlacementSimpleType'.
/home/travis/build/ng-bootstrap/ng-bootstrap/src/util/positioning.ts (162,51): Exported type alias 'PlacementType' has or is using private name 'PlacementExtendedType'.
Child demo/src/app/app.component.html:

https://travis-ci.org/ng-bootstrap/ng-bootstrap/builds/267910178

src/popover/popover.ts
```diff
@@ -27,7 +27,11 @@ let nextId = 0;
 @Component({
   selector: 'ngb-popover-window',
   changeDetection: ChangeDetectionStrategy.OnPush,
-  host: {'[class]': '"popover bs-popover-" + placement', 'role': 'tooltip', '[id]': 'id'},
+  host: {
+    '[class]': '"popover bs-popover-" + placement.split("-")[0]+" bs-popover-" + placement',
```
pkozlowski-opensource: This, for cases like `top-right`, would add the following CSS classes: `bs-popover-top` and `bs-popover-top-right`. Is this our intention? If so, why?

src/util/positioning.ts
```diff
@@ -154,3 +154,9 @@ export function positionElements(
   targetElement.style.top = `${pos.top}px`;
   targetElement.style.left = `${pos.left}px`;
 }
+
+type PlacementSimpleType = 'top' | 'bottom' | 'left' | 'right';
+type PlacementExtendedType = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right' | 'left-top' | 'left-bottom' |
+    'right-top' | 'right-bottom';
+
+export type PlacementType = PlacementSimpleType | PlacementExtendedType;
```
pkozlowski-opensource: Did you check how this type renders on our documentation site? Keep in mind that we are generating documentation automatically by walking Typescript AST. We need to make sure that users see a full list of options (`top`, `top-left` etc.) and not just `PlacementType` (or, alternatively, we would have to include documentation for the `PlacementType`)

src/util/positioning.ts
```diff
@@ -154,3 +154,9 @@ export function positionElements(
   targetElement.style.top = `${pos.top}px`;
   targetElement.style.left = `${pos.left}px`;
 }
+
+type PlacementSimpleType = 'top' | 'bottom' | 'left' | 'right';
+type PlacementExtendedType = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right' | 'left-top' | 'left-bottom' |
```
pkozlowski-opensource: What is the added value of splitting `PlacementType` into `PlacementSimpleType` and `PlacementExtendedType`? As far as I can tell we are not using / exporting  `PlacementSimpleType` nor `PlacementExtendedType` anywhere....

pkozlowski-opensource:
@mschoudry this is a good start but it needs more work, please see comments.

src/popover/popover.ts
mschoudry: Yes, You are right the intention is to add a new css-class e.g. bs-popover-top-right to extend the css rules defined by bs-popover-top css-class of bootstrap. The new classes are defined in the styles e.g.
styles: [`
    :host.bs-popover-top .arrow, :host.bs-popover-bottom .arrow {
      left: 50%;
    } 
    ...
  `].

This is same for tooltip.

src/util/positioning.ts
mschoudry: By defining them differently would help to use the same type for other components e.g.  PlacementSimpleType can be used to set position of datepicker, where PlacementExtendedType is not useful. etc.

pkozlowski-opensource: Why are you saying that `PlacementExtendedType` is not not useful for datepicker? It is actually using one of the "extended types": https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-input.ts#L111

Generally speaking, I'm not big fan of adding code that _might_ be used in the future. Let's keep things simple and add more construct as they are needed.

src/popover/popover.ts
pkozlowski-opensource: So, I think that there is a bit of the issue here since for "simple" positioning options (like `top`) we would have `bs-popover-top` class twice here. Could you think of a way of avoiding this?

pkozlowski-opensource:
@mschoudry I left an additional comment in the code.

It all looks good, generally speaking but we need tests asserting that proper classes are added based on "complex" positioning (we need at least one test for tooltip and another one for popover). 

Could you PTAL?

pkozlowski-opensource:
@mschoudry as discussed please rename `PlacementType` to just `Placement`. 

Also it would be good to add documentation for `Placement`  to the API page for both tooltip and popover.

PR #1781, opened 2017-08-28
chore(demo): Examples as default tab with mobile friendly display

benouat:
Via this PR I am fixing this long running "visual issue" I introduced on #1645 that on mobile header title is not properly aligned with the tab-set.

screen shot 2017-08-28 at 11 50 04

To bypass this issue, that is a nightmare to fix with only CSS media queries (thank you Apple with your split screen functionality !!) I propose to simply center both title and tabset on such devices when this is a relatively small viewport.

Here is a capture:

screen shot 2017-08-28 at 11 36 16

As I am tweaking tab-set code, I am also taking the opportunity to swap both tabs (API & Examples) and activate Example as the default one in all environment, not only mobile.

What do you think guys ? cc @pkozlowski-opensource

### DIFF
--- a/demo/src/app/app.routing.ts
+++ b/demo/src/app/app.routing.ts
@@ -20,8 +20,9 @@ import {
   NgbdTooltip,
   NgbdTypeahead
 } from './components';
+import {DEFAULT_TAB} from './shared/component-wrapper/component-wrapper.component';
 
-const DEFAULT_API_PATH = {path: '', pathMatch: 'full', redirectTo: 'api'};
+const DEFAULT_API_PATH = {path: '', pathMatch: 'full', redirectTo: DEFAULT_TAB};
 
 const componentRoutes = [{
     path: 'components/accordion',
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
I like it! @Foxandxss thoughts?

Foxandxss:
I am not sure if I  love the idea of tabs in mobile

icfantv:
I don't have a strong option either way about tabs on mobile.  It doesn't bother me nearly as much as popovers and tooltips.  Also, if you're on a mobile device and looking up API docs, you're either quite desperate or possibly a masochist.  But hey, at least this way we can say we cater to  you.

pkozlowski-opensource:
>  But hey, at least this way we can say we cater to you.

Yep, I don't think we expect people to actually use mobile version of this site on the daily basis. It is more to make sure that things are not broken when someone (for whatever reason) needs to access demo site on mobile. 

> I am not sure if I love the idea of tabs in mobile

Right. I see this PR more like a fix for the immediate issue we've got right now. We can always iterate on it and change the navigation pattern(s) on mobile.

PR #1786, opened 2017-08-29
Feat animations starter

mattlewis92:
Extracted from https://github.com/ng-bootstrap/ng-bootstrap/pull/1765

@pkozlowski-opensource here you go 😄

### DIFF
--- a/karma-test-shim.js
+++ b/karma-test-shim.js
@@ -29,6 +29,9 @@ System.config({
     '@angular/http': 'npm:@angular/http/bundles/http.umd.js',
     '@angular/router': 'npm:@angular/router/bundles/router.umd.js',
     '@angular/forms': 'npm:@angular/forms/bundles/forms.umd.js',
+    '@angular/platform-browser/animations': 'npm:@angular/platform-browser/bundles/platform-browser-animations.umd.js',
+    '@angular/animations/browser': 'npm:@angular/animations/bundles/animations-browser.umd.js',
+    '@angular/animations': 'npm:@angular/animations/bundles/animations.umd.js',
 
     // angular testing umd bundles
     '@angular/core/testing': 'npm:@angular/core/bundles/core-testing.umd.js',
--- a/package.json
+++ b/package.json
@@ -48,17 +48,18 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/core": "4.0.3",
-    "@angular/common": "4.0.3",
-    "@angular/forms": "4.0.3"
+    "@angular/animations": "4.2.0",
+    "@angular/common": "4.2.0",
+    "@angular/core": "4.2.0",
+    "@angular/forms": "4.2.0"
   },
   "devDependencies": {
-    "@angular/compiler": "4.0.3",
-    "@angular/compiler-cli": "4.0.3",
-    "@angular/http": "4.0.3",
-    "@angular/platform-browser": "4.0.3",
-    "@angular/platform-browser-dynamic": "4.0.3",
-    "@angular/router": "4.0.3",
+    "@angular/compiler": "4.2.0",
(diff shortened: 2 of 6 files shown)


pkozlowski-opensource:
@mattlewis92 thnx for all the PRs, I'm starting to review / merge those right now!  Will let you know as soon as a new branch is in place.

mattlewis92:
That's great, thanks! 😄

karma-test-shim.js
```diff
@@ -29,6 +29,9 @@ System.config({
     '@angular/http': 'npm:@angular/http/bundles/http.umd.js',
     '@angular/router': 'npm:@angular/router/bundles/router.umd.js',
     '@angular/forms': 'npm:@angular/forms/bundles/forms.umd.js',
+    '@angular/platform-browser/animations': 'npm:@angular/platform-browser/bundles/platform-browser-animations.umd.js',
+    '@angular/animations/browser': 'npm:@angular/animations/bundles/animations-browser.umd.js',
+    '@angular/animations': 'npm:@angular/animations/bundles/animations.umd.js',
```
pkozlowski-opensource: Do we need all those imports for tests, given that we are using `NoopAnimationsModule`? (Sorry for lame questions, learning about animation-related packages as I'm reviewing this PR :-))

mattlewis92: I believe so, both the `NoopAnimationsModule` and the `BrowserAnimationsModule` are exported from the same file

package.json
```diff
@@ -48,17 +48,18 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/core": "4.0.3",
-    "@angular/common": "4.0.3",
-    "@angular/forms": "4.0.3"
+    "@angular/animations": "4.2.0",
```
pkozlowski-opensource: Do we need to force people to have `@angular/animations` included? I mean, could the `@angular/animations` module be optional (and in this case we would fall back to the current state without animations? Or would it fail miserably?

pkozlowski-opensource:
Shouldn't we add a polyfill for Web Animations API (needed according to https://angular.io/guide/animations)?

package.json
mattlewis92: It would fail due to the new imports from the `@angular/animations` package. Folks that don't want animations can just add the `NoopAnimations` module though. From the users point of view it's a case of `npm install @angular/animations` and adding the module to their root AppModule, so an acceptable breaking change imo.

mattlewis92:
For the demo site + IE / edge etc it will indeed be needed. I can add that to this PR.

package.json
pkozlowski-opensource: Yes, breaking change is fine. I was more worried about payload increase for folks not wanting / needing animations.

pkozlowski-opensource:
> For the demo site + IE / edge etc it will indeed be needed. I can add that to this PR.

Yes, please!

package.json
mattlewis92: Looking at https://unpkg.com/@angular/animations@4.3.6/bundles/ the gzipped size of `animations.umd.min.js` is 1.3kb and `animations-browser.umd.min.js` is 22.9kb. I would expect that to drop though with proper tree shaking 🤞

mattlewis92:
@pkozlowski-opensource done 😄

pkozlowski-opensource:
@mattlewis92 thnx! I've just created the https://github.com/ng-bootstrap/ng-bootstrap/tree/animations branch with (most) of the commits from this PR. I've omitted the 4c0774f since it is not strictly needed yet (going to add it as soon as reviewing / merging first PR that needs it).

The other change I've made is that I've moved `web-animations-js` dependency to `devDependencies` as it should be users responsibility to choose / import a polyfill.

pkozlowski-opensource:
Closing this one, moving on to other PRs :-)

PR #1790, opened 2017-08-29
chore(demo): api page styling improvement

benouat:
From a very old & long standing request from @pkozlowski-opensource,  I finally managed to try improving a bit the visual aspect of the API page.

<details>
  <summary>Here is full size screenshot of the Datepicker API page to give you a glimpse of the changes without having to clone, build and run. Just click to expand</summary>
  
</details>

---

Here is a kind of list of things I changed / ~~removed~~

- blue color for component/directive/class title
- this same title is not a link anymore
- link to Github file has been made clearer, on the right, with an icon, which opacity is changing on over.
- internal link on the left (shown on hover) is still working.
- descriptive titles have been added to both internal link & Github one
- below each entity title a descriptive label has been added: can be one of these: Component / Directive, Class / Service or Service Configuration
- selector & "exported as" are now aligned with all other section titles (selector is nonetheless getting more emphasis than "exported as"
- each entity main description benefits from a subtile font size increase
- lot of spacing has been added a bit everywhere
- input/output types & default values have been swapped with description.
- method signature & return type are now displayed as real code
- all service configuration usually have properties description that refer to the underlying component/directive. this is now a navigable link to the proper thing


demo/src/app/components/shared/api-docs/api-docs-config.component.html
```diff
@@ -1,22 +1,37 @@
...
+    <a
+      class="github-link"
+      href="https://github.com/ng-bootstrap/ng-bootstrap/tree/master/{{apiDocs.fileName}}"
+      target="_blank"
+      title="Link to Github: {{apiDocs.className}}"
+    >
+      <img src="img/github.svg" alt="Link to Github {{apiDocs.className}}"/>
+    </a>
+  </h3>
+  <h5><small class="text-muted">Service Configuration</small></h5>
```
maxokorokov: Configuration Service?

demo/src/app/components/shared/api-docs/api-docs.component.html
```diff
@@ -1,40 +1,51 @@
...
-          <td class="label-cell">Selector</td>
-          <td class="content-cell"><code>{{apiDocs.selector}}</code></td>
-        </tr>
-        <tr *ngIf="apiDocs.exportAs">
-          <td class="label-cell">Exported as</td>
-          <td class="content-cell"><code>{{apiDocs.exportAs}}</code></td>
-        </tr>
-      </tbody>
-    </table>
+  <h5><small class="text-muted">Component / Directive</small></h5>
```
maxokorokov: Would be nice to extract the exact value from the code if later

demo/src/app/components/shared/api-docs/api-docs-class.component.html
```diff
@@ -29,20 +49,22 @@
...
-          <td class="label-cell"><code>{{method.name}}</code></td>
-          <td class="content-cell">
-            <div><i>Signature: </i><code>{{ methodSignature(method) }}</code></div>
-            <div><i>Return type: </i><code>{{ method.returnType }}</code></div>
-            <div style="margin: 10px 0">{{ method.description }}</div>
-          </td>
-        </tr>
+          <tr *ngFor="let method of apiDocs.methods">
+            <td class="label-cell"><code>{{method.name}}</code></td>
+            <td class="content-cell">
```
maxokorokov: I would have added 1 rem bottom padding (see attached screenshot) for readability

maxokorokov:
Another idea (not sure if it's a good one) is to remove method name duplication. Also maybe add return type directly

screen shot 2017-08-29 at 17 55 11

It's true though that for some it might not be as pretty, because of empty space

screen shot 2017-08-29 at 17 55 20

demo/src/app/components/shared/api-docs/api-docs.component.html
pkozlowski-opensource: Yep, we can easily do this in our doc script that walks TS sources.

demo/src/app/components/shared/api-docs/api-docs-config.component.html
pkozlowski-opensource: Or just 'Configuration'? It is not really a service, just an injectable object.

demo/src/app/components/shared/api-docs/api-docs.component.html
pkozlowski-opensource: We could just say `Directive` waiting for the script to be updated.

demo/src/app/components/shared/api-docs/api-docs-config.component.html
benouat: I'll update it to just _Configuration_

benouat:
@maxokorokov 🤔 then if you want to remove the name from the right signature, it really starts to look like what I proposed, which was to display methods not via a table, but just inlined, such as inside this capture

screen shot 2017-08-30 at 09 52 56

maxokorokov:
@benouat I like your last screenshot, though I've no idea if it would fit together with inputs/outputs as a table

benouat:
@maxokorokov: @pkozlowski-opensource told me yesterday that it felt weird to not be aligned with inputs/outputs... it was fine to me.

demo/src/app/components/shared/api-docs/api-docs-class.component.html
benouat: done

pkozlowski-opensource:
Merged, thnx! Those improvements are so nice that I can't look at the old page any more :-) Will try to cut a new release and deploy it ASAP!

PR #1800, opened 2017-08-31
feat(alert): add animations

pkozlowski-opensource:
This PR adds fade out animations to alerts. @mattlewis92 would you mind reviewing this one? 

As you can see I didn't re-use the `fadeOut` animations you've defined as it comes with the initial state forcing `0.5` opacity. Maybe we should parametrize it?

### DIFF
--- a/demo/src/app/components/alert/demos/closeable/alert-closeable.html
+++ b/demo/src/app/components/alert/demos/closeable/alert-closeable.html
@@ -1,6 +1,6 @@
-<p *ngFor="let alert of alerts">
-  <ngb-alert [type]="alert.type" (close)="closeAlert(alert)">{{ alert.message }}</ngb-alert>
-</p>
+<ngb-alert [type]="alert.type" (close)="closeAlert(alert)" *ngFor="let alert of alerts">
+  {{ alert.message }}
+</ngb-alert>
 <p>
   <button type="button" class="btn btn-primary" (click)="reset()">Reset</button>
 </p>
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -1,11 +1,7 @@
-import {
-  Component,
-  Input,
-  Output,
-  EventEmitter,
-  ChangeDetectionStrategy,
-} from '@angular/core';
+import {Component, Input, Output, EventEmitter, ChangeDetectionStrategy} from '@angular/core';
+import {trigger, transition, useAnimation} from '@angular/animations';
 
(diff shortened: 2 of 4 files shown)


src/alert/alert.ts
```diff
@@ -14,14 +10,19 @@ import {NgbAlertConfig} from './alert-config';
 @Component({
   selector: 'ngb-alert',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  animations: [trigger('fadeOut', [transition(':leave', animate('.15s linear', style({opacity: 0})))])],
+  host: {
+    '[class]': '"alert alert-" + type + (dismissible ? " alert-dismissible" : "")',
+    'role': 'alert',
+    '[style.display]': '"block"',
+    '[@fadeOut]': '"shown"'
```
mattlewis92: This can just be `'[@fadeOut]': ''` as the animation doesn't need any states

demo/src/app/components/alert/demos/closeable/alert-closeable.html
```diff
@@ -1,6 +1,6 @@
-<p *ngFor="let alert of alerts">
-  <ngb-alert [type]="alert.type" (close)="closeAlert(alert)">{{ alert.message }}</ngb-alert>
-</p>
+<ngb-alert [type]="alert.type" (close)="closeAlert(alert)" *ngFor="let alert of alerts">
```
mattlewis92: Ah, I struggled when trying to do this one myself, good shout on moving the ngFor onto the alert element itself to get the leave animation to trigger properly. Maybe this should be documented?

pkozlowski-opensource: I'm going to discuss this one with Matias and see what would be the best way of progressing.

mattlewis92: Could you nudge him on https://github.com/angular/angular/issues/18749 as well, as we'll need that for the collapse animation (as it's not currently possible to add animations via directive metadata).

pkozlowski-opensource:
Merged as ece7782, thnx for all the reviews @mattlewis92 - we are going to continue tomorrow with your PRs! I want to merge them ASAP but want to have a deeper look into testing aspect of this all before we proceed.

adaykin:
How can I use this to fade out with the <ngb-alert> ? I am on version 1.0.0-beta.5 for ng-bootstrap.

PR #1803, opened 2017-08-31
feat(accordion): add animation

mattlewis92:
Preview:

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,18 +48,18 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/animations": "4.2.0",
-    "@angular/common": "4.2.0",
-    "@angular/core": "4.2.0",
-    "@angular/forms": "4.2.0"
+    "@angular/animations": "4.3.0",
+    "@angular/common": "4.3.0",
+    "@angular/core": "4.3.0",
+    "@angular/forms": "4.3.0"
   },
   "devDependencies": {
-    "@angular/compiler": "4.2.0",
-    "@angular/compiler-cli": "4.2.0",
-    "@angular/http": "4.2.0",
-    "@angular/platform-browser": "4.2.0",
-    "@angular/platform-browser-dynamic": "4.2.0",
-    "@angular/router": "4.2.0",
+    "@angular/compiler": "4.3.0",
+    "@angular/compiler-cli": "4.3.0",
+    "@angular/http": "4.3.0",
+    "@angular/platform-browser": "4.3.0",
+    "@angular/platform-browser-dynamic": "4.3.0",
+    "@angular/router": "4.3.0",
     "@ngtools/webpack": "1.3.0",
     "@types/jasmine": "2.5.41",
     "@types/node": "^6.0.46",
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -1,11 +1,12 @@
 import {TestBed, ComponentFixture, inject} from '@angular/core/testing';
 import {createGenericTestComponent} from '../test/common';
 
-import {Component} from '@angular/core';
+import {Component, NgZone} from '@angular/core';
 
 import {NgbAccordionModule} from './accordion.module';
 import {NgbAccordionConfig} from './accordion-config';
 import {NgbAccordion} from './accordion';
+import {NoopAnimationsModule} from '@angular/platform-browser/animations';
 
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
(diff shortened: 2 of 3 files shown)


pkozlowski-opensource:
@mattlewis92 I've just pulled this PR and tested locally. The effect I'm getting after building demo page is very janky / not smooth at all (and far from silky animations on the Bootstrap side (https://getbootstrap.com/docs/4.0/components/collapse/#accordion-example).

Is it working fine on your side? What is a browser / version that you are testing with?

pkozlowski-opensource:
The other thing I've noticed that animations run on the initial page display and I don't think this is the effect we would like to have.

mattlewis92:
Oops, sorry I forgot to tag this as WIP. I believe the jankiness is something to do with a CSS rule applied by bootstrap that conflicts with this (or perhaps it's a bug with the lower angular version). I copied the animation code from another component where it works smoothly so I can only assume it's caused by an external factor. I will dig deeper later today, unless you want to take a look, maybe a fresh pair of eyes can spot something obvious I'm missing 😄

pkozlowski-opensource:
>  I will dig deeper later today, unless you want to take a look, maybe a fresh pair of eyes can spot something obvious I'm missing

I will look at your other PRs first and if I've got time left today I will dog into this one.

mattlewis92:
@pkozlowski-opensource this one is ready for review now as well, the reason the animation was janky before is because the element being animated had padding set on it, moving the card-body class inside the animated element fixed it. I also had to bump up the minimum angular version to 4.2.2 as the exit animation wasn't working. LMK what you think 😄

pkozlowski-opensource:
@mattlewis92 thnx for the update, animations are smooth as butter now, which is awesome 👍  
Now, the only remaining thing is that currently we run those animations on both enter (initial page display) and exit (component destroy) while I believe that we should only do so when the actual expand / collapsing occurs. WDYT?

Bumping Angular version is fine.

mattlewis92:
@pkozlowski-opensource that makes sense, to keep the implementation simple I used `@.disabled` which was added in 4.3.0 so required another version bump.

pkozlowski-opensource:
>  to keep the implementation simple I used @.disabled

Hmm, I though that we could define `open` / `close` states and just transition between those 2. I'm not huge fan of the line of code that went into `ngAfterViewInit` :-)

mattlewis92:
The issue is that you need to set the open state after the view has initialised, however as each pane has its own state you cannot do it retrospectively (like I did for the tabs animation), otherwise the slide animation will trigger delayed for open panels, when opening another panel. But maybe there's  a simpler solution that I'm missing 😄  I guess you could move `this.animationsDisabled = false;` into the `toggle` function which would make things a bit cleaner.

chris-allen:
@mattlewis92 So, what's the last mile on this?  It looks quite close.

mattlewis92:
It's just pending review from @pkozlowski-opensource, if you want to use it today you can use my fork of this repo with animations included:
```
npm i @mattlewis92/ng-bootstrap
```

pkozlowski-opensource:
@mattlewis92 sorry for the delay... I'm really trying to see how we could avoid the current trickery in the `ngAfterViewInit`. This code rubs me the wrong way but I can't seem to think of a better solution right now :-/

Give me a couple more days to play with different ideas.

pkozlowski-opensource:
@mattlewis92 OK, so after playing with the different ideas I must say that I can't come up with anything better as compared to your proposal @mattlewis92 - as such I'm going to merge it in a bit.

The only thing to consider is to move state driving disabled animations to a panel level (instead of being on the accordion level). This way a newly added panel wouldn't be animated initially. WDYT?

mattlewis92:
I'm not sure that's possible as panels are directives which can't be animated with todays angular animations library. I think it's better to merge it as is, the underlying implementation can always be changed later if a better way is found. As none of these animations have a public API it would be a seamless experience to refactor later in a non breaking way if a better solution was found 😄

atecey:
Are there any plans to release some of the animations into the master branch?

AlexKhymenko:
Any plans on merging it?

robk9e:
Hello @mattlewis92 any ETA to merge ?

mattlewis92:
@robk9e it's not my project, it's up to @pkozlowski-opensource and co to review and merge when they have the time free.

robk9e:
@mattlewis92 I tried with you repository  @mattlewis92/ng-bootstrap but didnt change. Do I have to set up something?

alvinqle:
Does anyone know how to set up @mattlewis92's fork of the repo for animations to work? I've run `npm i @mattlewis92/ng-bootstrap` and have modified my angular application to pull from that node_modules package, but animations still aren't working for me :(

zolakt:
Any ETA on this?
If it's now breaking, why not just merge it until a better solution is found?

dvose:
@pkozlowski-opensource ETA on merging this?

PR #1806, opened 2017-08-31
demo: add colored badges for items in documentation

maxokorokov:
- Adding type information extraction to the api generation → `Component`, `Directive`, `Class`, `Service`, `Interface`
- Adding colored badges on the demo site
- `Configuration` is a special one set directly in demo components

<details>
  <summary>See the sample screenshot here (omitted inputs/outputs for simplcity)</summary>

</details>

### DIFF
--- a/demo/src/app/components/shared/api-docs/api-docs-badge.component.ts
+++ b/demo/src/app/components/shared/api-docs/api-docs-badge.component.ts
@@ -1,0 +1,27 @@
+import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
+
+const BADGES = {
+  'Directive': 'success',
+  'Component': 'success',
+  'Service': 'primary',
+  'Configuration': 'primary',
+  'Class': 'danger',
+  'Interface': 'danger'
+};
+
+@Component({
+  selector: 'ngbd-api-docs-badge',
+  template: `<h5><span class="badge" [ngClass]="badgeClass">{{text}}</span></h5>`,
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class NgbdApiDocsBadge {
+
+  badgeClass;
+  text;
+
+  @Input()
+  set type(type: string) {
+    this.text = type;
+    this.badgeClass = `badge-${BADGES[type] || 'secondary'}`;
+  }
+}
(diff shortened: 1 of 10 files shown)


benouat:
@maxokorokov I am not that sure anymore about the colour aspect. Seems that it's catching attention a bit too much.

What about trying with only the _greyish_ bootstrap colours ? 
screen shot 2017-09-01 at 10 14 41

benouat:
Let's keep it as your first idea.
I think I were fooled by your first _tweaked_ screenshot. As you removed `Inputs/Ouputs`, it was a bit too much colours in a row.
With the full text, it's better.

maxokorokov:
<details>
  <summary>Here is the full screenshot with inputs/outputs if you want to compare</summary>

</details>

Forgot to tell that I tried to group colours for the same 'type':
- `Directive`, `Component` → green
- `Class`, `Interface` → red
- `Service`, `Configuration` → blue

pkozlowski-opensource:
> I think I were fooled by your first tweaked screenshot. As you removed Inputs/Ouputs, it was a bit too much colours in a row.

Yep, exactly! I've just pulled this PR, tested it locally and it looks good. Just one tiny idea: maybe we could prefix component and directive with `@` so it would become `@directive` \ `@component`?

maxokorokov:
> prefix component and directive with @ so it would become `@directive` \ `@component`?

@pkozlowski-opensource, sorry, can't follow the logic here. What would it bring to the ng-bootstrap user?

demo/src/app/components/shared/api-docs/api-docs-badge.component.ts
```diff
@@ -0,0 +1,27 @@
+import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
+
+const BADGES = {
+  'Directive': 'badge-success',
+  'Component': 'badge-success',
+  'Service': 'badge-primary',
+  'Configuration': 'badge-primary',
+  'Class': 'badge-danger',
+  'Interface': 'badge-danger'
```
pkozlowski-opensource: Maybe don't repeat `badge-` prexif here but move it to the type setter instead?

pkozlowski-opensource:
@maxokorokov  `@` is what users write for component / directive annotations. Also angular.io website is using this symbol in its badges. Anyway, it was just a thought, feel free to ignore.

maxokorokov:
@pkozlowski-opensource fixed your first comment and ignored your `@component` / `@directive` comment :)

Anyway, I can't see where you found it. Here is the the screenshot from the documentation:

screen shot 2017-09-01 at 12 03 48

Will merge when Travis is green

PR #1809, opened 2017-09-01
demo(dropdown): show usage of split and group buttons

### LINKED ISSUE #1785: Dropdowns in .btn-group don't automatically close
### Bug description:
I'd like to have a .btn-group containing a few regular buttons next to a dropdown. I can do this successfully when wrapping the directive in an ng-container instead of a div, but then nothing happens upon any action that would normally close the dropdown (inside or outside).

This may also be a requirement for [split buttons](https://getbootstrap.com/docs/4.0/components/dropdowns/#split-button-dropdowns).

### Demo
https://plnkr.co/edit/AmFw1WXiDXVQQVw7vu6W?p=preview

### Library versions

Angular: 4.3.3

ng-bootstrap: 1.0.0-beta.2

Bootstrap: 4.0.0-beta

pkozlowski-opensource:
Closes #1785

### DIFF
--- a/demo/src/app/components/dropdown/demos/split/dropdown-split.html
+++ b/demo/src/app/components/dropdown/demos/split/dropdown-split.html
@@ -1,0 +1,22 @@
+<p>Bootstrap's split buttons and dropdowns on button groups are supported with the existing dropdown directives.</p>
+
+<div class="btn-group" ngbDropdown>
+  <button type="button" class="btn btn-outline-success">Plain ol' button</button>
+  <button class="btn btn-outline-primary" ngbDropdownToggle>Drop me</button>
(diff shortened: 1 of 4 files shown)


demo/src/app/components/dropdown/demos/split/dropdown-split.html
```diff
@@ -0,0 +1,22 @@
+<p>Bootstrap's split buttons and dropdowns on button groups are supported with the existing dropdown directives.</p>
```
maxokorokov: remove `'s`, just Bootstrap?

pkozlowski-opensource: Changed and merged. Thnx!

PR #1814, opened 2017-09-04
feat(dropdown): auto and array support in placement

mschoudry:
Added support for auto placement and array can be defined inside placement attribute of dropdown.

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -54,17 +54,28 @@ describe('ngb-dropdown', () => {
   });
 
   it('should be closed and down by default', () => {
-    const html = `<div ngbDropdown><div ngbDropdownMenu></div></div>`;
+    const html = `
+      <div ngbDropdown>
+          <div ngbDropdownMenu>
+            <a class="dropdown-item">dropDown item</a>
+            <a class="dropdown-item">dropDown item</a>
+          </div>
+      </div>`;
 
     const fixture = createTestComponent(html);
     const compiled = fixture.nativeElement;
 
-    expect(getDropdownEl(compiled)).toHaveCssClass('dropdown');
     expect(compiled).not.toBeShown();
   });
 
   it('should have dropup CSS class if placed on top', () => {
-    const html = `<div ngbDropdown placement="top"></div>`;
+    const html = `
+      <div ngbDropdown placement="top">
+          <div ngbDropdownMenu>
+            <a class="dropdown-item">dropDown item</a>
+            <a class="dropdown-item">dropDown item</a>
+          </div>
+      </div>`;
 
     const fixture = createTestComponent(html);
     const compiled = fixture.nativeElement;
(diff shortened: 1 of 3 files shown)


src/dropdown/dropdown.spec.ts
```diff
@@ -515,6 +515,42 @@ describe('ngb-dropdown-toggle', () => {
...
+          </div>
+      </div>`;
+
+    const fixture = createTestComponent(html);
+    const compiled = fixture.nativeElement;
+
+    const dropDownEle = getDropdownEl(compiled);
+
+    expect(dropDownEle).toHaveCssClass('dropdown');
+    expect(dropDownEle.getAttribute('ng-reflect-placement')).toEqual('auto');
```
pkozlowski-opensource: We should not use `ng-reflect-` attributes in tests as it is impl-specific and can change at any moment (not to mention the fact that it is not there in production mode).

TBH I'm not sure what you are trying to assert with this test so it is hard for me to suggest changes...

src/dropdown/dropdown.spec.ts
```diff
@@ -515,6 +515,42 @@ describe('ngb-dropdown-toggle', () => {
...
+          </div>
+      </div>`;
+
+    const fixture = createTestComponent(html);
+    const compiled = fixture.nativeElement;
+
+    const dropDownEle = getDropdownEl(compiled);
+
+    expect(dropDownEle).toHaveCssClass('dropdown');
+    expect(dropDownEle.getAttribute('ng-reflect-placement')).toEqual('left-top,top-left,auto');
```
pkozlowski-opensource: Same comment as above...

src/dropdown/dropdown.ts
```diff
@@ -10,20 +10,23 @@ import {
...
 import {NgbDropdownConfig} from './dropdown-config';
-import {positionElements} from '../util/positioning';
+import {positionElements, PlacementArray, Placement} from '../util/positioning';
 
 /**
  */
 @Directive(
     {selector: '[ngbDropdownMenu]', host: {'[class.dropdown-menu]': 'true', '[class.show]': 'dropdown.isOpen()'}})
 export class NgbDropdownMenu {
+  placement: Placement = 'bottom';
```
pkozlowski-opensource: Why not `PlacementArray`?

pkozlowski-opensource: Oh, I see, sorry for the noise.

src/dropdown/dropdown.ts
```diff
@@ -102,9 +108,17 @@ export class NgbDropdown {
     this._zoneSubscription = ngZone.onStable.subscribe(() => { this._positionMenu(); });
   }
 
-  isUp() { return this.placement.indexOf('top') !== -1; }
+  isUp() {
```
pkozlowski-opensource: Maybe we should simplify those methods in the following way:
* `isDown()` is true only when there is `bottom` somewhere in placement
* `isUp()` = `!isDown()`

WDYT?

In any case we should tests for this logic.

jbistis:
Is there a fix for this?

suhailkc:
Is there any fix for this ???

mikelhamer:
A fix would be awesome :)

PR #1819, opened 2017-09-06
fix(positioning): check for secondary available placements

mschoudry:
Positioning was not handling properly secondary placements.


intellix:
got an example of the secondary placements that didn't work?

mschoudry:
The problem is linked to the recent changes to support array inside placement. The secondary position was not taken into consideration while checking for available positions, due to which target element was being placed out of viewport.

PR #1838, opened 2017-09-09
demo: avoid duplicate ids in generated HTML

pkozlowski-opensource:
@benouat could you PTAL?

### DIFF
--- a/demo/src/app/components/shared/example-box/example-box.component.html
+++ b/demo/src/app/components/shared/example-box/example-box.component.html
@@ -13,11 +13,8 @@
     <div class="card-body">
       <ng-content></ng-content>
     </div>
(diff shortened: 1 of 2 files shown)


benouat:
@pkozlowski-opensource I am fine. with this change. I honestly did not think it would be an issue, as we (Angular Template engine) always render one tab at a time, which mean that we never could encounter this situation of having twice the `id="html"` into the generated dom.

Go with it. shipit :shipit: !

pkozlowski-opensource:
@benouat `is`s _are_ duplicated, check https://ng-bootstrap.github.io/#/components/accordion/examples as an example. The reason is that we can have more than one demo on a page.

I'm going to merge it based on the above.

PR #1839, opened 2017-09-11
feat(collapse): add animations

pkozlowski-opensource:
BREAKING CHANGE:

The `ngbCollapse` is a component now (used to be a directive).
Since this change you can't use `ngbCollapse` on other components
(but still can get the same effect by using it on a wrapper element
above).

Also requires Angular 4.3.6 as a minimal version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,18 +48,18 @@
   },
   "homepage": "https://github.com/ng-bootstrap/ng-bootstrap#readme",
   "dependencies": {
-    "@angular/animations": "4.2.0",
-    "@angular/common": "4.2.0",
-    "@angular/core": "4.2.0",
-    "@angular/forms": "4.2.0"
+    "@angular/animations": "4.3.6",
+    "@angular/common": "4.3.6",
+    "@angular/core": "4.3.6",
+    "@angular/forms": "4.3.6"
   },
   "devDependencies": {
-    "@angular/compiler": "4.2.0",
-    "@angular/compiler-cli": "4.2.0",
-    "@angular/http": "4.2.0",
-    "@angular/platform-browser": "4.2.0",
-    "@angular/platform-browser-dynamic": "4.2.0",
-    "@angular/router": "4.2.0",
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
@mattlewis92 I'm still looking into animations and trying to figure out an alternative approach for accordions. To "warm up" I did this PR for collapse directive (thinking here being that accordions are somehow similar).

Would you mind having a look at this PR?

mattlewis92:
Looks like travis is failing as you need to re-run clang format (maybe it could be nice to run this on commit with [lint-staged](https://github.com/okonet/lint-staged) and [husky](https://github.com/typicode/husky)

pkozlowski-opensource:
> Also do we care about what happens when the element with the collapse animation is destroyed? i.e. when used with an ngIf and removing the host from the dom, should the collapse animation be displayed?

I think that that collapse animation should _not_ be run when the underlying DOM element gets removed. This PR should be fine in this respect, right?

> In the future once this bug is fixed it should be possible to go back to a directive instead of using the component hack.

Yup, one more thing to discuss with Matias :-) Added to my list.

> Looks like travis is failing as you need to re-run clang format 

Yeh, just saw it, will re-run. Thnx!

mattlewis92:
> I think that that collapse animation should not be run when the underlying DOM element gets removed. This PR should be fine in this respect, right?
Yup, thats the current behaviour, I just wanted to make sure that's what you expected 😄

collindutter:
Are there plans to merge this sometime soon?

shivarajnaidu:
When these animations will be available for general use(merge)  ?

mbrookson:
Will this be available soon? This library is great but without animations that the main bootstrap library has it just doesn't look polished.

Looking forward to using this library once the animations are all in!

KosrenDQ:
Hey guys, please merge this and made my day 😃

jaredgkraemer:
Please merge this, it would be very nice to have

PR #1843, opened 2017-09-12
refactor(accordion): remove redundant internal state

maxokorokov:
Clean up accordion internal state a little:
- remove redundant `_panelRefs` and `_states`
- instead add `isOpen` flag on the panel itself


pkozlowski-opensource:
LGTM! It is indeed much simpler than the previous version _and_ might help with the animations! 
Thnx so much for this PR, merging!

PR #1844, opened 2017-09-12
fix(accordion): remove double border for inactive panels

maxokorokov:
screen shot 2017-09-12 at 14 55 48

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -111,7 +111,12 @@ export interface NgbPanelChangeEvent {
         </div>
       </div>
     </ng-template>
-  `
+  `,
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
So, there is a discussion of fixing this on the Bootstrap side, see #23836 and #23839. Maybe it would be worth waiting for what comes out of those tickets on the Bootstrap side?

maxokorokov:
Good idea, should have checked Bootstrap before.
Looks like it will also fix rounded corners

pkozlowski-opensource:
Based on the previous discussion, let's allow Bootstrap fix it on their end.

PR #1850, opened 2017-09-14
feat(popover): auto close feature

mschoudry:
support for auto close feature added in popover.

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -21,6 +21,7 @@ import {listenToTriggers} from '../util/triggers';
 import {positionElements, Placement, PlacementArray} from '../util/positioning';
 import {PopupService} from '../util/popup';
 import {NgbPopoverConfig} from './popover-config';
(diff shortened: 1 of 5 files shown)


src/popover/popover.ts
```diff
@@ -114,6 +115,14 @@ export class NgbPopover implements OnInit, OnDestroy {
    */
   @Input() container: string;
   /**
+ * Indicates that dropdown should be closed when selecting one of dropdown items (click) or pressing ESC.
```
pkozlowski-opensource: `dropdown` => `popover`

pkozlowski-opensource:
We are going to implement a different solution based on RxJS, similar to what we did for datepicker already, see: https://github.com/ng-bootstrap/ng-bootstrap/blob/a489cd8152776233f04c069b69479c0757f7e4be/src/datepicker/datepicker-input.ts#L290-L303

Issue #1847, opened 2017-09-14
Dropdown position issue when resize window


ysds:
https://ng-bootstrap.github.io/#/components/dropdown/examples

![modal](https://user-images.githubusercontent.com/4065765/30422456-f54b8cb2-997a-11e7-972a-0be98b0b9eac.gif)

pkozlowski-opensource:
@mschoudry would you mind having a look? I can reproduce it on my side as well.

ysds:
Probably ngbDropdown requires `position: relateive` or `dropdown` class.

pkozlowski-opensource:
> Probably ngbDropdown requires position: relateive or dropdown/dropup class.

Yep! Thnx @ysds

PR #1852, opened 2017-09-14
fix(dropdown): fix for issue 1847

mschoudry:
fix for drop down bug #1847


pkozlowski-opensource:
@mschoudry thnx. Could you please add a test for this case as well?

pkozlowski-opensource:
LGTM, thnx!

Issue #1829, opened 2017-09-07
feature(datepicker): popup disabled state


maxkorz:
### Bug description:
Datepicker popup can't be set to "disabled" state.

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/tXfSAQl13kIPCeV4n8lR?p=preview

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.3.6 

ng-bootstrap: 1.0.0-beta.2

Bootstrap: 4.0.0-beta

pkozlowski-opensource:
`<input>` field and triggering `<button>` are 2 separate controls. You can disable them individually. If you want prevent people from typing into the field, just disable the field. If you want to disable a button - disable a button. If you wan to disable both - disable both: http://plnkr.co/edit/O8sJZ1gp1yBW651e70IO?p=preview

Closing as it works as expected.

maxkorz:
@pkozlowski-opensource I don't want to disable the button, I want to disable `<input>` field and allow to see [disabled calendar](http://plnkr.co/edit/EGirnlWoZJfHPVZN1s4U?p=preview) by clicking on the `<button>`

pkozlowski-opensource:
@maxkorz if you are up to sending a PR with a fix the code change would be around those lines:
* memorize disabled state in an instance variable: https://github.com/ng-bootstrap/ng-bootstrap/blob/0f92c80e7260c65e4464b6e66229804409d439f2/src/datepicker/datepicker-input.ts#L160-L165

* propagate memorized value when a popup gets opened: https://github.com/ng-bootstrap/ng-bootstrap/blob/0f92c80e7260c65e4464b6e66229804409d439f2/src/datepicker/datepicker-input.ts#L275-L284

maxkorz:
@pkozlowski-opensource Thank you, perhaps I will send a PR when I have time (If someone else doesn't send a PR sooner)

maxokorokov:
Do you think we should handle both cases with and without form control for this one?

- `<input ngbDatepicker  [(ngModel)]="text" disabled />` → yes
- `<input ngbDatepicker disabled />` → should we disable datepicker in this case?

@mschoudry and @chenyuzhcy, you might be interested in this question as you've opened same-looking PRs for this one

chenyuzhcy:
@maxokorokov I think we should disable the second case

mschoudry:
[...]

PR #1867, opened 2017-09-22
feat(datepicker): popup disabled state

mschoudry:
Feature for opening datepicker popup in disabled state #1829

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -160,6 +160,10 @@ describe('NgbInputDatepicker', () => {
          expect(input.disabled).toBeTruthy();
          expect(buttonInDatePicker.disabled).toBeTruthy();
 
+         // check if ngb-datepicker popup is disabled
+         const navItem = fixture.nativeElement.querySelector('ngb-datepicker-navigation');
+         expect(navItem.getAttribute('ng-reflect-disabled')).toBeTruthy();
+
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker-input.ts
```diff
@@ -221,6 +221,9 @@ export class NgbInputDatepicker implements OnChanges,
         this.close();
       });
 
+      // forward disabled state
+      this._cRef.instance.setDisabledState(this._elRef.nativeElement.disabled);
```
pkozlowski-opensource: This would mean reading state from the DOM and we don't want to do it. We need to read it from a form control or introduce private state.

src/datepicker/datepicker-input.spec.ts
```diff
@@ -160,6 +160,10 @@ describe('NgbInputDatepicker', () => {
          expect(input.disabled).toBeTruthy();
          expect(buttonInDatePicker.disabled).toBeTruthy();
 
+         // check if ngb-datepicker popup is disabled
+         const navItem = fixture.nativeElement.querySelector('ngb-datepicker-navigation');
+         expect(navItem.getAttribute('ng-reflect-disabled')).toBeTruthy();
```
pkozlowski-opensource: `ng-reflect-*` attributes are Angular internal implementation detail and we should never ever use this info in our implementation / tests.

maxokorokov:
Closing as duplicate of #1887

PR #1883, opened 2017-10-04
Fixed ngDropDown when rendered on server side with ng-universal

Kemaweyan:
Fixed this error when rendered with [ng-universal](https://github.com/angular/universal) on server side:

`TypeError: Cannot read property 'attribs' of undefined`


pkozlowski-opensource:
merged, thnx!

PR #1887, opened 2017-10-07
fix(datepicker): popup disabled state

chenyuzhcy:
#1829

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -160,6 +160,11 @@ describe('NgbInputDatepicker', () => {
          expect(input.disabled).toBeTruthy();
          expect(buttonInDatePicker.disabled).toBeTruthy();
 
+         const dayElements = fixture.nativeElement.querySelectorAll('ngb-datepicker-month-view .ngb-dp-day');
+         expect(dayElements[1]).toHaveCssClass('disabled');
+         expect(dayElements[11]).toHaveCssClass('disabled');
+         expect(dayElements[21]).toHaveCssClass('disabled');
+
          fixture.componentInstance.isDisabled = false;
          fixture.detectChanges();
          tick();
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker-input.ts
```diff
@@ -292,6 +294,7 @@ export class NgbInputDatepicker implements OnChanges,
           }
         });
     datepickerInstance.startDate = this.startDate || this._model;
+    datepickerInstance.setDisabledState(this._disabled);
```
maxokorokov: This is not an input, so I would do it in the `open()` method after the focus handling. Plus you won't need to change the service

src/datepicker/datepicker-service.ts
```diff
@@ -78,7 +78,7 @@ export class NgbDatepickerService {
   }
 
   open(date: NgbDate) {
-    if (!this._state.disabled && this._calendar.isValid(date)) {
+    if (this._calendar.isValid(date)) {
```
maxokorokov: I don't think it should be possible to do anything with the datepicker when it is disabled, so I wouldn't change that. Plus for this PR it is not necessary.

chenyuzhcy: If we don't change this, it won't work if the user actually want to see the disabled date picker popup, because the popup won't open at all.

maxokorokov: I've checked out the code in the PR and it works perfectly fine without this change

`open()` is the `NgbDatepickerService` method that has nothing to do with the popup

The popup works because when you open it, you first

```javascript
// 1. focus()
this._cRef.instance.focus();

// 2. set disabled state
this._setDatepickerDiabledState(this.disabled);
```

it would not work if you set disabled state before calling focus

src/datepicker/datepicker-input.ts
```diff
@@ -141,6 +142,22 @@ export class NgbInputDatepicker implements OnChanges,
...
   @Output() navigate = new EventEmitter<NgbDatepickerNavigateEvent>();
 
+  @Input()
+  get disabled() {
+    return !!this._disabled;
+  }
+  set disabled(value: any) {
+    let isDisabled = value != null && `${value}` !== 'false';
+
+    if (this._disabled !== isDisabled) {
```
maxokorokov: honestly I wouldn't care checking this. I'm sure the `NgbDatepickerService` would not set the same value twice

src/datepicker/datepicker-input.spec.ts
```diff
@@ -202,6 +212,49 @@ describe('NgbInputDatepicker', () => {
        }));
   });
 
+  describe('without form control', () => {
+    it('should propagate disabled state', fakeAsync(() => {
```
maxokorokov: this test should not be asynchronous, you're not using `ngModel` anymore and no need in the additional `describe` I guess

src/datepicker/datepicker-input.ts
```diff
@@ -310,4 +324,6 @@ export class NgbInputDatepicker implements OnChanges,
       this._onTouched();
     }
   }
+
+  private _setDatepickerDiabledState(isDisabled: boolean) { this._cRef.instance.setDisabledState(isDisabled); }
```
maxokorokov: you have a typo in the method name → `Diabled` → `Disabled`

src/datepicker/datepicker-service.ts
chenyuzhcy: yeah it makes sense, I think it wasn't working before because I set disabled before focus

chenyuzhcy:
updated my pr, thanks for the review!

maxokorokov:
Thanks, I've simplified it a little more if you don't mind. Will merge when Travis is green

Issue #1783, opened 2017-08-28
Close date picker modal after selecting an already selected date


timosalm:
### Bug description:
The date picker modal should be closed after selecting/clicking on an already selected date.

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.1

ng-bootstrap: v1.0.0-beta.2

Bootstrap:

PR #1888, opened 2017-10-08
fix(datepicker): Close date picker after selecting the same date

chenyuzhcy:
#1783

### DIFF
--- a/src/datepicker/datepicker-service.ts
+++ b/src/datepicker/datepicker-service.ts
@@ -5,7 +5,6 @@ import {Injectable} from '@angular/core';
 import {isInteger} from '../util/util';
 import {Subject} from 'rxjs/Subject';
 import {buildMonths, checkDateInRange, checkMinBeforeMax, isChangedDate, isDateSelectable} from './datepicker-tools';
-import {NgbDateStruct} from './ngb-date-struct';
 
 import {filter} from 'rxjs/operator/filter';
 import {Observable} from 'rxjs/Observable';
@@ -14,14 +13,16 @@ import {Observable} from 'rxjs/Observable';
 export class NgbDatepickerService {
   private _model$ = new Subject<DatepickerViewModel>();
 
-  private _select$ = new Subject<NgbDateStruct>();
+  private _select$ = new Subject<NgbDate>();
 
   private _state: DatepickerViewModel =
       {disabled: false, displayMonths: 1, firstDayOfWeek: 1, focusVisible: false, months: [], selectedDate: null};
 
-  get model$() { return filter.call(this._model$.asObservable(), model => model.months.length > 0); }
+  get model$(): Observable<DatepickerViewModel> {
+    return filter.call(this._model$.asObservable(), model => model.months.length > 0);
+  }
 
-  get select$(): Observable<NgbDateStruct> { return this._select$.asObservable(); }
+  get select$(): Observable<NgbDate> { return filter.call(this._select$.asObservable(), date => date !== null); }
 
   set disabled(disabled: boolean) {
     if (this._state.disabled !== disabled) {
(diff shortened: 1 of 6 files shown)


src/datepicker/datepicker.ts
```diff
@@ -203,6 +203,11 @@ export class NgbDatepicker implements OnDestroy,
    */
   @Output() navigate = new EventEmitter<NgbDatepickerNavigateEvent>();
 
+  /**
+   * An event fired when select happens.
+   */
+  @Output() onSelect = new EventEmitter<NgbDate>();
```
maxokorokov: Could you please change this to: 
```javascript
@Output() select = new EventEmitter<NgbDateStruct>();
```

select → to keep naming consistent
`NgbDateStruct` → as `NgbDate` is internal implementation

src/datepicker/datepicker.ts
```diff
@@ -222,6 +228,12 @@ export class NgbDatepicker implements OnDestroy,
     this.showWeekNumbers = config.showWeekNumbers;
     this.startDate = config.startDate;
 
+    this._dateSelectedSubscription = _service.dateSelected$.subscribe(selected => {
+      if (selected) {
+        this.onSelect.emit();
```
maxokorokov: it should emit a `NgbDateStruct`

src/datepicker/datepicker.spec.ts
```diff
@@ -442,6 +442,18 @@ describe('ngb-datepicker', () => {
     expect(document.activeElement).toBe(datepicker.nativeElement);
   });
 
+  it('should emit select event when select date', () => {
```
maxokorokov: You should also add the tests for:

- clicking on the same date twice
- pressing Enter key on the same date twice

src/datepicker/datepicker-service.ts
```diff
@@ -12,11 +12,15 @@ import {filter} from 'rxjs/operator/filter';
...
   private _model$ = new Subject<DatepickerViewModel>();
 
+  private _dateSelected$ = new Subject<boolean>();
+
   private _state: DatepickerViewModel =
       {disabled: false, displayMonths: 1, firstDayOfWeek: 1, focusVisible: false, months: [], selectedDate: null};
 
   get model$() { return filter.call(this._model$.asObservable(), model => model.months.length > 0); }
 
+  get dateSelected$() { return filter.call(this._dateSelected$.asObservable(), selected => !!selected); }
```
maxokorokov: please add the following tests to the `datepicker-service.spec.ts`:

- `focus()` → should emit date regardless if it's the same one or not
- `focusSelect()` → should emit date regardless if it's the same one or not

src/datepicker/datepicker-service.ts
```diff
@@ -87,6 +91,7 @@ export class NgbDatepickerService {
     const validDate = this.toValidDate(date, null);
     if (!this._state.disabled && isChangedDate(this._state.selectedDate, validDate)) {
       this._nextState({selectedDate: validDate});
+      this._dateSelected$.next(true);
```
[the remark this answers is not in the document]
chenyuzhcy: have no idea how that check got back there, wasn't there when I tested it;-( thanks for pointing that out.

maxokorokov:
Hey, @chenyuzhcy. Thanks a lot (especially for many tests), I've added a couple of changes on top, if you don't mind:

- squashed your commits for clean history
- renamed `onSelect` event to `select` in public `NgdDatepicker` API
- changed `manuallySelect(date)` to `select(date, options: {emitEvent: boolean})` as they usually do it in Angular APIs and I don't think another entry method is necessary in `NgbDatepickerService`
- found a couple of small bugs and added tests for them

Will merge when Travis is green

Issue #1119, opened 2016-12-07
Typeahead. Ability to leaf through multiple result pages


forik:
It would be nice to add an option for results paging in typeahead component as there are scenarios where might be hundreds of results and it is no good to show them all at once.
Maybe it is possible to add ability to provide custom template for the whole result window popup instead of just a row template.

trumbitta:
Also having the result window popup with internal scrolling would be nice.

Example: https://material.angularjs.org/latest/demo/autocomplete#inside-dialog

Ks89:
I'm agree with @trumbitta.
I think that the scrollable list is the most important, in particular if you have design/UX constraints in your application.

Right now, in my application I have to put a lower bound of chars to prevent very long popups. However this limit is useless, since all my data is already into my client, so it is creating problems to users that are obliged to type 3 or more chars without a real reason.

borgotecnologia:
Yes, it's a problem, I tried to correct it with css, but does not work, please fix this problem....

PendelinP:
One possible workaround is to set the height of the height of the dropdown to a fixed value and enable vertical scrolling (overflow). Unfortunately the up/down-arrow-key navigation doesn't work since the focus gets lost when you select an element outside the visible area: http://plnkr.co/edit/ffZi6LSRkJUhpxA2AyvN

ymeine:
I gave a try at implementing a solution for this issue, and here I come with a proposal.

The features implies two aspects: 

- being able to limit the height of the dropdown and scroll inside
- making sure the active item is entirely visible inside the dropdown when it is being made active

# Limiting the height

For now as this thread suggests, it is already possible to set the style of the class `dropdown-menu` and customize its `max-height` and `overflow`.

However, this might not be very handy and if we want to provide fine customization as well as a feature of automatic scrolling out of the box, I think this would make more sense to specify the maximum height to the directive directly.

Therefore I suggest an optional input property `maxHeight` (a string for maximum flexibility), forcing the dropdown's `max-height` accordingly and its `overflow` to `auto`. Setting this property to a non void value would enable the automatic scrolling feature (see below).

# Automatic scrolling

This is to make sure that the active item — the one highlighted — is always fully visible _inside the dropdown_ (as much as possible, see edge case below).

[...]

PR #1894, opened 2017-10-12
feat(typeahead): allowing proper max-height with results navigation (scrolling) on the dropdown

ymeine:
Related to #1119

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -1,4 +1,4 @@
-import {Component, Input, Output, EventEmitter, TemplateRef, OnInit} from '@angular/core';
+import {Component, Input, Output, EventEmitter, TemplateRef, OnInit, OnDestroy, ViewContainerRef} from '@angular/core';
 
 import {toString} from '../util/util';
 
@@ -20,7 +20,15 @@ export interface ResultTemplateContext {
 @Component({
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
-  host: {'class': 'dropdown-menu', 'style': 'display: block', 'role': 'listbox', '[id]': 'id'},
+  host: {
+    'class': 'dropdown-menu',
+    'style': 'display: block',
+    '[style.max-height]': 'maxHeight == null ? null : maxHeight',
+    '[style.overflow]': 'maxHeight == null ? null : "auto"',
+    'role': 'listbox',
+    '[id]': 'id',
+    '(mouseleave)': 'onMouseLeaveDropdown'
+  },
   template: `
     <ng-template #rt let-result="result" let-term="term" let-formatter="formatter">
       <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>
@@ -29,7 +37,7 @@ export interface ResultTemplateContext {
       <button type="button" class="dropdown-item" role="option"
         [id]="id + '-' + idx"
         [class.active]="idx === activeIdx"
-        (mouseenter)="markActive(idx)"
+        (mouseenter)="onMouseEnterItem($event, idx)"
         (click)="select(result)">
           <ng-template [ngTemplateOutlet]="resultTemplate || rt"
           [ngTemplateOutletContext]="{result: result, term: term, formatter: formatter}"></ng-template>
@@ -37,8 +45,11 @@ export interface ResultTemplateContext {
     </ng-template>
   `
 })
-export class NgbTypeaheadWindow implements OnInit {
+export class NgbTypeaheadWindow implements OnInit,
+    OnDestroy {
   activeIdx = 0;
+  private _latestMousePosition = null;
+  private _mouseMoveListener = null;
 
   /**
    *  The id for the typeahead widnow. The id should be unique and the same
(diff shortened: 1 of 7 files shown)


src/typeahead/typeahead-window.ts
```diff
@@ -20,7 +20,15 @@ export interface ResultTemplateContext {
...
   exportAs: 'ngbTypeaheadWindow',
-  host: {'class': 'dropdown-menu', 'style': 'display: block', 'role': 'listbox', '[id]': 'id'},
+  host: {
+    'class': 'dropdown-menu',
+    'style': 'display: block',
+    '[style.max-height]': 'maxHeight == null ? null : maxHeight',
+    '[style.overflow]': 'maxHeight == null ? null : "auto"',
+    'role': 'listbox',
+    '[id]': 'id',
+    '(mouseleave)': 'onMouseLeaveDropdown'
```
maxokorokov: `onMouseLeaveDropDown()` note the brackets. Should not be called without them?

src/typeahead/typeahead-window.ts
```diff
@@ -73,17 +84,60 @@ export class NgbTypeaheadWindow implements OnInit {
...
+    const y = event.screenY;
+
+    if (latestMousePosition == null || latestMousePosition.x !== x || latestMousePosition.y !== y) {
+      this.markActive(activeIdx);
+    }
+
+    // to keep track of the actual latest mouse position while we keep the mouse inside items
+    if (this._mouseMoveListener == null) {
+      this._mouseMoveListener = (moveEvent) => this._latestMousePosition = {x: moveEvent.screenX, y: moveEvent.screenY};
+      document.addEventListener('mousemove', this._mouseMoveListener);
```
maxokorokov: - This will trigger a huge number of change detections. Just add `ngDoCheck() {}` to see it
- You should use `renderer.listen(...)` to register listeners

src/typeahead/typeahead-window.ts
```diff
@@ -103,17 +157,59 @@ export class NgbTypeaheadWindow implements OnInit {
...
 
   select(item) { this.selectEvent.emit(item); }
 
   ngOnInit() {
     this.activeIdx = this.focusFirst ? 0 : -1;
-    this._activeChanged();
+    this._activeChanged(null);
+  }
+
+  private _activeChanged(direction) {
```
maxokorokov: - can't we use a proper type here for the direction instead of -1 / +1?
- also am I not mistaken that you're not using the actual direction value anywhere?

src/typeahead/typeahead-window.ts
```diff
@@ -73,17 +84,60 @@ export class NgbTypeaheadWindow implements OnInit {
...
+  onMouseEnterItem(event, activeIdx) {
+    // to prevent the case where the mouse cursor stays over the dropdown
+    // while we scroll the content because of a keyboard navigation
+    // scrolling makes a the mouse enter a new element,
+    // and therefore would mark it active otherwise
+    const latestMousePosition = this._latestMousePosition;
+    const x = event.screenX;
+    const y = event.screenY;
+
+    if (latestMousePosition == null || latestMousePosition.x !== x || latestMousePosition.y !== y) {
```
maxokorokov: can't we just have a some kind of flag to distinguish whether it was scrolled with mouse or keyboard or something else potentially?

ymeine:
_A note for later_

After rebasing on latest changes, the tests are broken. Indeed, one of those changes consisted in moving an inline style to the CSS and, for a reason which disconcerts me we are unit testing a GUI library without an essential part of it which: the CSS.

My feature relies on the proper, final display of the DOM, it's not purely based on data model.

From a general standpoint, we should avoid testing on something too far from the reality (another example: calling event handlers instead of dispatching events - even if here dispatching events manually is still far from reality where full user actions occur, but going to protractor is another matter).

As a workaround I could manually add the proper styling again just for the test, but it seems too dirty and not future-proof.

pkozlowski-opensource:
@ymeine I think that you are hitting limits of what what is possible / makes sense with the unit tests. We really do _not_ want to test with CSS and "real" events - this would make existing tests hard to write and slow to execute. This in turn discourage people from writing and running those and bam - the main benefit is gone. 

We must keep existing tests easy to write and lightening-fast to execute. At the same time we don't have to (and never will be able to!) test everything through unit tests. We should introduce other types of tests (Protractor e2e, most probably) for cases like yours. 

Having said all this I've got _real_ concerns with having "scroll" ability for the typeahead in this library, see: https://github.com/ng-bootstrap/ng-bootstrap/issues/1119#issuecomment-377277194

maxokorokov:
I'd say that typeahead and autocomplete widgets don't need this functionality, however we probably should introduce `select / multiple-select` widget where seeing all results at once would make sense. But this requires a study/design doc of what kind of selects are expected, how to filter them, etc.

ymeine:
As long as we don't duplicate too much, be it for the code but also for the API. Yes, I still don't know why we have (why Bootstrap has) Popover and Tooltip. Last time I've checked the code, it was basically the same. And for the features, idem.

My real concern is about the user: if this feature is the only additional thing in the possible new widget, I would always use it and forget about the TypeAhead. If more things change, why not, but let's hope no one will want an all-in-one widget (which makes sense since they would probably be close) which they would eventually implement themselves, dropping the choice of relying on ng-bootstrap.

On another hand, they're all just a combination of an input element with another form of display (popup list, selection list, etc), everything can already be done on the user side after all...

Issue #698, opened 2016-09-07
Typeahed: allow search on focus


pkozlowski-opensource:
In certain use-cases one might want to trigger typeahead search when the input field gets focused. You can see a good example of this on https://www.kayak.fr/flights

It should be trivial to implement this on our side since we should simply listen to the `focus` event, get whatever value we've got in the input and push it down the observable stream. The only tricky part might be taking into account the situation where a given input field has `autofocus` attribute (in this case we shouldn't kick off things).

There should be a flag for this option (I guess false by default).

I'm going to mark this as `medium` difficulty and I think that it is a good issue for ambitious community members.  As always test are essentials.

jnizet:
Another problem to solve is that the usual `distinctUntilChanged()` operator, used to prevent the dropdown to open on two consecutive equal values, would prevent the window to open on focus events.

pkozlowski-opensource:
> Another problem to solve is that the usual distinctUntilChanged() operator, used to prevent the dropdown to open on two consecutive equal values, would prevent the window to open on focus events.

Good point! We could un-subscribe on `blur` (actually - we must since there is a bug in the current implementation - see #723) and re-subscribe on focus.

pkozlowski-opensource:
This is blocked on #723

Ks89:
It could be really useful.
Which is the official schedule for this and #723 ?

GabrielGMartinsBr:
So is it possible to implement or just to suggestion? Sorry for my dumb doubt, i'm trying to understand github yet

bastiankhalil:
@pkozlowski-opensource do you have the implementation that you referred to? how do you mean to push it with the stream?

beaverusiv:
I have a workaround for this for people who want this functionality now.

On the input element that has typeahead add a focus handler:
```
(focus)="onFocus($event)"
```

In the focus handler trigger an input event:
```
public onFocus(e: Event): void {
    e.stopPropagation();
    setTimeout(() => {
        const inputEvent: Event = new Event('input');
        e.target.dispatchEvent(inputEvent);
    }, 0);
}
```

It would be nice to not have to do this, but it works for now.

Yura13:
When will this issue be fixed?

scotschroeder:
@beaverusiv, that works thanks!
However if I focus out and then focus back in it doesn't work the 2nd time.

beaverusiv:
I have created a gist with the entire typeahead wrapper I have:

[...]

PR #1895, opened 2017-10-12

### LINKED ISSUE #1853: TypeAhead's dropdown is closed when clicking on the input
### Bug description

When using the TypeAhead, when the dropdown is open, clicking on the input closes it.

Even though it could be seen as a feature, this is not necessarily intuitive nor wanted. If this behavior is considered acceptable and wanted (as default or not) we should at least provide an alternative where the dropdown doesn't close when clicking on the input.

For reference, at the time of writing those websites keep the dropdown open: 

- [Google](https://www.google.fr/)
- [Booking.com](https://www.booking.com/)
- [Amazon.com](https://www.amazon.com/)
- ...

I just used the first one from the [documentation page](https://ng-bootstrap.github.io/#/components/typeahead/examples): https://ng-bootstrap.github.io/app/components/typeahead/demos/basic/plnkr.html

Here is the frozen plunkr: [Example usage of the typeahead widget from https://ng-bootstrap.github.io - Plunker](http://embed.plnkr.co/yQEbXluwIjbgfaeMMKbw/).

- Angular: 4.0.3
- ng-bootstrap: 1.0.0-beta.4
- Bootstrap: 4.0.0-beta

ymeine:
Related to #698, fixes #1853

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -19,14 +19,14 @@ import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
 import {Observable} from 'rxjs/Observable';
 import {BehaviorSubject} from 'rxjs/BehaviorSubject';
 import {Subscription} from 'rxjs/Subscription';
-import {letProto} from 'rxjs/operator/let';
 import {_do} from 'rxjs/operator/do';
+import {map} from 'rxjs/operator/map';
 import {switchMap} from 'rxjs/operator/switchMap';
 import {fromEvent} from 'rxjs/observable/fromEvent';
 import {positionElements, PlacementArray} from '../util/positioning';
 import {NgbTypeaheadWindow, ResultTemplateContext} from './typeahead-window';
 import {PopupService} from '../util/popup';
-import {toString, isDefined} from '../util/util';
+import {toString, isDefined, isElementOrAncestorOf} from '../util/util';
 import {NgbTypeaheadConfig} from './typeahead-config';
 
 enum Key {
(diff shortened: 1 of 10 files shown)


src/typeahead/typeahead.ts
```diff
@@ -58,6 +58,42 @@ export interface NgbTypeaheadSelectItemEvent {
...
+
+  /**
+   * The current NgbTypeahead instance.
+   */
+  instance: NgbTypeahead;
+
+  /**
+   * The custom context passed to this NgbTypeahead instance.
+   */
+  context: any;
```
maxokorokov: Not sure why we need it for focus handling

src/typeahead/typeahead.ts
```diff
@@ -88,6 +124,9 @@ export class NgbTypeahead implements ControlValueAccessor,
   private _subscription: Subscription;
   private _userInput: string;
   private _valueChanges: Observable<string>;
+  private _input: NgbTypeaheadInitParams['input$'];
```
maxokorokov: Didn't know it worked!
But still would prefer to see the type with my own eyes: `private _input: Observable<...>`
Maybe it's just me though

src/typeahead/typeahead.ts
ymeine: For this feature, we don't need it, it was just a nice addition along with the others whose usefulness - I agree with you - can be discussed.

This was more specifically requested by @divdavem for a valid use case (which can be implemented without this but would be far more complex AFAIU).

src/typeahead/typeahead.ts
ymeine: On one hand I agree with you because indeed in some contexts it's not even easy to go check what `NgbTypeaheadInitParams['input$']` actually means (like here where GitHub displays only a few lines).

On the other hand, the semantics and maintenance complexity would change if I don't use a reference anymore. It wouldn't be linked anymore. It's not a very big deal though in this case.

PS: I would have expected to have a clue from the tools like in Visual Studio Code, for instance something like `NgbTypeaheadInitParams['input$'] => Observable<HTMLInputElementEvent>` in a tooltip over the type reference, but it's not even the case.

src/typeahead/typeahead.ts
```diff
@@ -213,6 +284,29 @@ export class NgbTypeahead implements ControlValueAccessor,
...
+  }
+
+  _isClickInWidget(event) {
+    const target = event.target;
+
+    const input = this._viewContainerRef.element.nativeElement;
+    return this._isElementOrAncestorOf(input, target);
+  }
+
+  _isElementOrAncestorOf(possibleAncestor, reference) {
```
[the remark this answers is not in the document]
ymeine: If you think so, I agree with you! It will just make a bigger change regarding the tests (which don't use any DOM content). Also, `util.ts` seems to deal only with primitive values, providing short, simple functions. Does it fit there? Would providing a `dom-util.ts` (or alike) be an alternative?

src/typeahead/typeahead.ts
```diff
@@ -318,7 +413,7 @@ export class NgbTypeahead implements ControlValueAccessor,
   }
 
   private _writeInputValue(value: string): void {
-    this._renderer.setProperty(this._elementRef.nativeElement, 'value', value);
+    this._renderer.setProperty(this._elementRef.nativeElement, 'value', value == null ? '' : value);
```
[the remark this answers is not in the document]
ymeine: Didn't know about that before you told me, I'll definitely use it!

pkozlowski-opensource:
Suppressed by #1941

Issue #1896, opened 2017-10-12
ngbTooltip placement issue when inside other text


dereklin:
Please remember, the issues forum is __NOT__ for support requests. It is for bugs and feature requests only.
Please read https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md and search
existing issues (both open and closed) prior to opening any new issue and ensure you follow the instructions therein.

### Bug description:
When using ngbTooltip inside other text, the tooltip placement is not stable
### Link to minimally-working plunker that reproduces the issue:
http://plnkr.co/edit/mqliRNfHFvFVi7W160Q7?p=preview

When you resize the window, making it smaller and larger, you might see different placement issues

You can fork a plunker from one of our [demos](https://ng-bootstrap.github.io/#/components) and use it as a starting point.
Please note that we can _not_ act on bug reports without a minimal reproduction scenario in plunker. Here is why:
https://github.com/ng-bootstrap/ng-bootstrap#you-think-youve-found-a-bug

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.3.5 and above

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta

mschoudry:
The problem is with the max-width specified inside .tooltip-inner. remove it and the tooltip will be ok.

Regards,

pkozlowski-opensource:
This is fixed in the latest version of this library (1.0.0) and Bootstrap CSS 4.0.0: http://plnkr.co/edit/3dUpcB0IWH9wBEBUaAEX?p=preview

PR #1899, opened 2017-10-13
fix(positioning): correctly consider the viewport

### LINKED ISSUE #1876: Typeahead availability of placement 'bottom-left' or 'top-left' is dismissed by height, not width
### Bug description:
I believe determining the availability of placement, when using secondary placement of 'left' with the primary placement 'top' or 'bottom', is incorrect. It works as long as the left offset of the host element is smaller than the widow height. Given a rectangular window and a far right position of the host element, the positioning breaks.

I should note that I believe this is a general bug in the Positioning class so I presume it's not limited to the Typeahead component. The method in question is named setSecondaryPlacementForTopBottom:
```
private setSecondaryPlacementForTopBottom(
      hostElemClientRect: ClientRect, targetElemClientRect: ClientRect, primaryPlacement: string,
      availablePlacementArr: Array<string>) {
    let html = document.documentElement;
    // check for left-bottom
    if ((window.innerHeight || html.clientHeight) - hostElemClientRect.left >= targetElemClientRect.width) {
      availablePlacementArr.splice(availablePlacementArr.length, 1, primaryPlacement + '-left');
    }
    if (targetElemClientRect.width <= hostElemClientRect.right) {
      availablePlacementArr.splice(availablePlacementArr.length, 1, primaryPlacement + '-right');
    }
  }  
```
From my understanding of the code, the first if sentence should use the window.innerWidth or html.clientWidth (and not height, as it is), to determine if there's enough space horizontally, since at this point, it was already determined by the primary placement that there's enough space up or down?

As such, when you set the placement of a typeahead field to be limited to ['bottom-left', 'top-left'], neither is available and the last one is used ('top-left'), even when there's enough space to position it under the host element.

For the plunker to work, please ensure the preview IFRAME is wider than it is taller, but still has enough room for the typeahead's drop down menu to be positioned underneath the input. Ensuring the preview IFRAME is around  1100px wide and 730px tall should exhibit the described behaviour. Type 'al' into the input to get the drop down menu to open with 4 results.

http://plnkr.co/edit/gkh3qxwE1pPMuVMZrRCp?p=preview

Angular: 4.0.3
ng-bootstrap: 1.0.0-beta.5
Bootstrap: 4.0.0-beta i presume?

### LINKED ISSUE #1898: "auto" and secondary placement of popovers is incorrectly determined
### Related:
#1896: ngbTooltip placement issue when inside other text
#1876: Typeahead availability of placement 'bottom-left' or 'top-left' is dismissed by height, not width

### Bug description:
[Positioning.getAvailablePositions()](https://github.com/ng-bootstrap/ng-bootstrap/blob/6c7a31b08942532431eeee22f32148f948880ed6/src/util/positioning.ts#L149)'s checks are incorrect, resulting in "auto" placement being resolved to "top" if there is enough space above, even if there is not enough horizontal space. In this case "top-left" or "top-right" would be more appropriate.

[Positioning.setSecondaryPlacementForTopBottom()](https://github.com/ng-bootstrap/ng-bootstrap/blob/6c7a31b08942532431eeee22f32148f948880ed6/src/util/positioning.ts#L215) uses height instead of width to determine if there is enough space the right, as mentioned in #1876. This results in *-left being an available placement, even if there is zero space between the host element and the right edge of the window.

I'm also a bit confused why e.g. "top-left" is "Position the popover on top, and place the little arrow to the left", which results in the popover being placed to the top right. However, I also understand that changing this behaviour would break a lot of code already using these placements, but I don't think this behaviour is reflected in the [API documentiation](https://ng-bootstrap.github.io/#/components/popover/api).

I'm currently working on a pull request for the Positioning service that should fix these problems.

[Old, angular 4.4.4, ng-bootstrap 1.0.0-beta.5](http://plnkr.co/edit/4XSRQHO9MAu3XHI5UHx6?p=preview)
[New angular 6.0.0, ng-bootstrap 2.0.0](http://plnkr.co/edit/bCAmOVzyiO5kYyCe4UtO?p=preview)

Angular: 6.0.0

ng-bootstrap: 2.0.0

Bootstrap: 4.0.0

Edit: Links points to the latest commit (as of this writing) instead of master
Edit2: Update ng-bootstrap 1.0.0-beta.5 -> 2.0.0

burnedram:
Fixes #1898, fixes #1876.
I'm not sure about #1896, as I have not tested this specific use case.

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -243,7 +243,12 @@ const positionService = new Positioning();
 /*
  * Accept the placement array and applies the appropriate placement dependent on the viewport.
  * Returns the applied placement.
- * In case of auto placement, placements are selected in order 'top', 'bottom', 'left', 'right'.
+ * In case of auto placement, placements are selected in order
+ *   'top', 'bottom', 'left', 'right',
+ *   'top-left', 'top-right',
+ *   'bottom-left', 'bottom-right',
+ *   'left-top', 'left-bottom',
+ *   'right-top', 'right-bottom'.
  * */
 export function positionElements(
     hostElement: HTMLElement, targetElement: HTMLElement, placement: string | Placement | PlacementArray,
@@ -253,8 +258,10 @@ export function positionElements(
   // replace auto placement with other placements
   let hasAuto = placementVals.findIndex(val => val === 'auto');
   if (hasAuto >= 0) {
-    ['top', 'right', 'bottom', 'left'].forEach(function(obj) {
-      if (placementVals.find(val => val.search('^' + obj + '|^' + obj + '-') !== -1) == null) {
+    ['top', 'bottom', 'left', 'right', 'top-left', 'top-right', 'bottom-left', 'bottom-right', 'left-top',
+     'left-bottom', 'right-top', 'right-bottom',
+    ].forEach(function(obj) {
+      if (placementVals.find(val => val.search('^' + obj) !== -1) == null) {
         placementVals.splice(hasAuto++, 1, obj as Placement);
       }
     });
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -151,12 +151,17 @@ export class Positioning {
     let hostElemClientRect = hostElement.getBoundingClientRect();
     let targetElemClientRect = targetElement.getBoundingClientRect();
     let html = document.documentElement;
+    let windowHeight = window.innerHeight || html.clientHeight;
+    let windowWidth = window.innerWidth || html.clientWidth;
+    let hostElemClientRectHorCenter = hostElemClientRect.left + hostElemClientRect.width / 2;
+    let hostElemClientRectVerCenter = hostElemClientRect.top + hostElemClientRect.height / 2;
 
     // left: check if target width can be placed between host left and viewport start and also height of target is
     // inside viewport
     if (targetElemClientRect.width < hostElemClientRect.left) {
       // check for left only
-      if ((hostElemClientRect.top + hostElemClientRect.height / 2 - targetElement.offsetHeight / 2) > 0) {
+      if (hostElemClientRectVerCenter > targetElemClientRect.height / 2 &&
+          windowHeight - hostElemClientRectVerCenter > targetElemClientRect.height / 2) {
         availablePlacements.splice(availablePlacements.length, 1, 'left');
       }
       // check for left-top and left-bottom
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@burnedram thnx!
@mschoudry would you mind reviewing?

mschoudry:
@burnedram I was not able to follow the main issue you are targeting. I have tested your PR and it does not seems to be resolving the issue you demonstrated in plunker. Or is it that I misunderstood something?
Regards,

burnedram:
The issue I am tryinc to solve is how "auto" placement is resolved. As [seen](https://imgur.com/YImiyE8) in the [plunkr](http://plnkr.co/edit/4XSRQHO9MAu3XHI5UHx6?p=preview), the button labelled "auto" (which has the attribute "placement='auto'") is placed above the button, even though there is not enough (horizontal) space for the popover.

Another bug can be seen in the plunkr as well. The "left" button has a proper popover, because there is enough horizontal and vertical space. "top" and "bottom" however gets squished by the browser as it is too close to the edge of the viewport. It seems because the popover gets squished, the calculations done in order to horizontally center the popover is no longer correct. It is not my intent to fix this bug, as I don't really know how one should react to there not being enough space in a position strictly chosen by the user.

However, one can mitigate the centering bug by using "auto", but as I've described [here](https://github.com/ng-bootstrap/ng-bootstrap/issues/1898), the "auto" algorithm is broken.

burnedram:
Here is an example of what my pull requests fixes.

[Current ng-bootstrap](https://imgur.com/YrrEztJ), popover is (almost) always positioned above.
[With this pull request applied](https://imgur.com/q3p42QZ), popover is positioned where there is enough space.

burnedram:
@pkozlowski-opensource @mschoudry 

Sorry for bumping, just wondering if there is anything else I should change or clarify :smiley:

burnedram:
@mschoudry If you meant "when the popup width is bigger than the viewport width", then yes, that is the issue I'm trying to solve. (By "width" I mean both horizontal position and width)

The reasoning behind refactoring the code was to improve readability. I have essentially moved all conditions inside if-statements and given them a name. This also has a bonus in that if-statements that share conditions doesn't need code repetition, which reduces the chance of missing to implement a change in all if-statements.

Now each placement's reasoning can be understood by reading the code, instead of having to think about what all the expressions means.

I also opted to remove the `setSecondaryPlacementFor` as it was [broken](https://github.com/ng-bootstrap/ng-bootstrap/issues/1876).

Ofcourse, the bug could probably be solved by just changing the code around those two lines, but I can't help to think that the bug occured due to the code being difficult to follow.

burnedram:
I realise that refactoring isn't always the best option in open-source projects, as it might break other pull requests and general code history, and it is of course up to you if you believe it is acceptable in this case.

pkozlowski-opensource:
@burnedram seems like some of the issues referenced in this PR are already fixed:
* #1876 was fixed by 0df398c
* #1896 is confirmed as fixed in the latest version: http://plnkr.co/edit/3dUpcB0IWH9wBEBUaAEX?p=preview 

Could you please:
* check if your fix is still relevant / valid?
* rebase your PR?

Thnx!

bluecaret:
Anyone still working on this? This is still very much a bug.

burnedram:
@pkozlowski-opensource 
I have update the issue, and also rebased the pull-request.
I decided to forgo the entire rewrite of the affected function, and instead only add the missing checks.

I also changed "auto" to include all secondary placements, as there are situations where no primary placement is available. I also think that "auto" should consider all possible placements.

[New plnkr that shows how broken "auto" is](http://plnkr.co/edit/bCAmOVzyiO5kYyCe4UtO?p=preview)
[With this pull request, all popovers behaves nicely](https://user-images.githubusercontent.com/1688771/39725947-c02b9bc6-524d-11e8-8527-bbb058a53098.png)

burnedram:
Note on the failed CI build: [master is failing](https://travis-ci.org/ng-bootstrap/ng-bootstrap/branches)
Seems to be because of saucelabs.

I've run the tests locally, and they passed (except for two relating to "ngbFocusTrap" and "ngb-typeahead", which also fails on master)

burnedram:
@pkozlowski-opensource 
Rebased onto lastest master commit, now tests are okay

bluecaret:
@pkozlowski-opensource  Any chance this can be added to the next release? Seems like it's been fixed for a while now...

pkozlowski-opensource:
Merged. Thnx @burnedram for the PR and rebasing it, much appreciated!

PR #1902, opened 2017-10-18
fix(positioning): fix for issue 1876

mschoudry:
Fix for issue #1876

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -217,7 +217,7 @@ export class Positioning {
       availablePlacementArr: Array<string>) {
     let html = document.documentElement;
     // check for left-bottom
-    if ((window.innerHeight || html.clientHeight) - hostElemClientRect.left >= targetElemClientRect.width) {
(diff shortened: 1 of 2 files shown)


src/util/positioning.ts
```diff
@@ -217,7 +217,7 @@ export class Positioning {
       availablePlacementArr: Array<string>) {
     let html = document.documentElement;
     // check for left-bottom
-    if ((window.innerHeight || html.clientHeight) - hostElemClientRect.left >= targetElemClientRect.width) {
+    if ((window.innerWidth || html.clientWidth) - hostElemClientRect.left >= targetElemClientRect.width) {
```
pkozlowski-opensource: Hmm, I'm wondering why existing unit tests didn't catch it? Could you please try to add a test that fails when this fix is not in place?

pkozlowski-opensource:
@mschoudry thnx for the PR - I left 2 comments that I would like to be addressed before this one gets in. Thnx!

Issue #1904, opened 2017-10-19
Feature Request: Specify height of Progress bar


Juchar:
### Feature request:
Possibility to specify the height of the progress bar.
See [Bootstraps Possibility](https://getbootstrap.com/docs/4.0/components/progress/#height) of doing it.

### Link to minimally-working plunker that reproduces the issue:
Not applicable.

### Implementation suggestion
Add an attribute `height` to `<ngb-progressbar>` where one can specify the height as a String, e.g.:
`<ngb-progressbar height=".5rem">`

### Version of Angular, ng-bootstrap, and Bootstrap:
Angular:
4.2.4

ng-bootstrap:
1.0.0-beta.5

Bootstrap:
4.0.0-beta

Issue #905, opened 2016-10-18
Feature (Progressbar): Stacked Progressbar 


cnd5041:
Is there support for a stacked progress bar like in ui-bootstrap: https://angular-ui.github.io/bootstrap/#/progressbar?

pkozlowski-opensource:
No, we don't have support for stacked progressbars since Bootstrap doesn't allow them, see: https://github.com/twbs/bootstrap/issues/18835

We can add it as soon as Bootstrap 4 clarifies its stance.

DmitryEfimenko:
they have support for it now: http://getbootstrap.com/docs/4.0/components/progress/#multiple-bars

maxokorokov:
Don't think it would be the best idea to copy the angular-ui way of implementing this. Maybe come up with something like:

```html
<!-- single -->
<ngb-progressbar showValue="true" type="success" [value]="25"></ngb-progressbar>

<!-- stack -->
<ngb-progressbar-stack>
  <ngb-progressbar showValue="true" type="success" [value]="25"></ngb-progressbar>
  <ngb-progressbar showValue="true" type="warning" [value]="10"></ngb-progressbar>
</ngb-progressbar-stack>
```

Mind you it's just a quick idea.

chenyuzhcy:
Just submitted an initial implementation of the feature, let me know if you have any suggestions or better way to do this.

maxokorokov:
Ok, I spend some time trying several ideas and trying to make #1920 work (see comments there), but I don't think it's possible to make my previous suggestion work and reuse `ngb-progressbar`.

Alternative idea might be the following: introduce a new `ngb-bar` component and make `ngb-progressbar` use it internally to minimise duplication:

```html
<!-- single -->
<ngb-progressbar showValue="true" type="success" [value]="25" [height]="30px"></ngb-progressbar>

<!-- stack -->
<ngb-progressbar-stack [height]="30px">
  <ngb-bar showValue="true" type="success" [value]="25">...</ngb-bar>
  <ngb-bar showValue="true" type="warning" [value]="10">...</ngb-bar>
</ngb-progressbar-stack>
```

Opened short example PR with this one #1924. 
What do you think? 

If anybody can come up with simpler implementation, would be great

chenyuzhcy:
[...]

PR #1908, opened 2017-10-20
feat(progressbar): specify height

chenyuzhcy:
#1904


chenyuzhcy:
what's the best way to make sure to update that when we upgrade?

maxokorokov:
I guess the easiest would be for you to realign and update current PR once #1914 will land in master. Or just test locally with `beta.2`

maxokorokov:
Thanks for contributing, @chenyuzhcy!
If you're interested, there is another progress bar feature missing (#905) that need a little design first.

PR #1919, opened 2017-10-29
feat(tooltip): optionally get content from 'title' attribute

trumbitta:
Hi :)

I recently used tooltips in my app, and noticed we don't use `title` to get the content for the tooltip.  
That sounded like something I could contribute, so here's my pull-request.

Tooltips do what the HTML `title` attribute already does, so it's
semantically correct to "transform" the title attribute into a
JavaScript tooltip.

Also, Bootstrap does it: https://github.com/twbs/bootstrap/blob/v4-dev/js/src/tooltip.js#L426

### DIFF
--- a/demo/src/app/components/tooltip/demos/index.ts
+++ b/demo/src/app/components/tooltip/demos/index.ts
@@ -1,4 +1,5 @@
 import {NgbdTooltipBasic} from './basic/tooltip-basic';
+import {NgbdTooltipTitle} from './title/tooltip-title';
 import {NgbdTooltipContainer} from './container/tooltip-container';
 import {NgbdTooltipTplcontent} from './tplcontent/tooltip-tplcontent';
 import {NgbdTooltipTplwithcontext} from './tplwithcontext/tooltip-tplwithcontext';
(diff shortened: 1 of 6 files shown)


pkozlowski-opensource:
@trumbitta thnx for the PR but we are going to pass on it. The reason is that I can see complications with this PR with little added benefit:
* it is more verbose to write `<div title="blah" ngbTooltip></div>` as compared to `<div  ngbTooltip="blah"></div>`
* it would create a confusing API where one could _think_ that `title` is an `@Input` but in fact you can't do `<div [title]="exp" ngbTooltip>`
* impl requires low-level DOM manipulation 

Thnx for the suggestion but will skip this PR for the reasons outlined above. But hay, even if this PR is not merged we _love_ all the PRs so please keep them coming!

trumbitta:
@pkozlowski-opensource I get all the reasons listed above, and thanks for taking the time for addressing this. 

Do you have an "official" stance regarding accessibility and semantics?

I mean, I'll just have to repeat the info with `[ngbTooltip]="my tip" title="my tip"` and it's not that big of a deal. But the resulting experience is less than optimal for non-visually-impaired users because they'll see both the tooltip and the title popping up on the same element with the same content.

Do you have any pointers towards a win-win situation?

PR #1941, opened 2017-11-06

### LINKED ISSUE #1853: TypeAhead's dropdown is closed when clicking on the input

ymeine:
Related to #698, fixes #1853

A previous PR was open (#1895) but after discussing with @pkozlowski-opensource and @maxokorokov we went for as simpler solution consisting mostly of the bug fixes and a sample demo.

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -26,7 +26,7 @@ import {fromEvent} from 'rxjs/observable/fromEvent';
 import {positionElements, PlacementArray} from '../util/positioning';
 import {NgbTypeaheadWindow, ResultTemplateContext} from './typeahead-window';
 import {PopupService} from '../util/popup';
-import {toString, isDefined} from '../util/util';
+import {toString, isDefined, isElementOrAncestorOf} from '../util/util';
 import {NgbTypeaheadConfig} from './typeahead-config';
 
 enum Key {
@@ -69,7 +69,7 @@ let nextWindowId = 0;
   host: {
     '(blur)': 'handleBlur()',
     '[class.open]': 'isPopupOpen()',
-    '(document:click)': 'dismissPopup()',
+    '(document:click)': 'onDocumentClick($event)',
     '(keydown)': 'handleKeyDown($event)',
     'autocomplete': 'off',
     'autocapitalize': 'off',
@@ -214,6 +214,21 @@ export class NgbTypeahead implements ControlValueAccessor,
     this._renderer.setProperty(this._elementRef.nativeElement, 'disabled', isDisabled);
   }
 
+  onDocumentClick(event) {
+    if (this.isPopupOpen()) {
+      if (!this._isClickInWidget(event)) {
+        this.dismissPopup();
+      }
+    }
+  }
+
+  private _isClickInWidget(event) {
+    const target = event.target;
+
+    const input = this._viewContainerRef.element.nativeElement;
+    return isElementOrAncestorOf(input, target);
+  }
+
   /**
    * Dismisses typeahead popup window
    */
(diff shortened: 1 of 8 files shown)


src/typeahead/typeahead.ts
```diff
@@ -80,7 +80,8 @@ let nextWindowId = 0;
     '[attr.aria-owns]': 'isPopupOpen() ? popupId : null',
     '[attr.aria-expanded]': 'isPopupOpen()'
   },
-  providers: [NGB_TYPEAHEAD_VALUE_ACCESSOR]
+  providers: [NGB_TYPEAHEAD_VALUE_ACCESSOR],
+  exportAs: 'ngbTypeahead'
```
maxokorokov: @ymeine, maybe just move this to the top near the selector
@pkozlowski-opensource you had some concerns about exporting typeahead or I'm mistaken?

src/typeahead/typeahead.ts
```diff
@@ -213,6 +214,21 @@ export class NgbTypeahead implements ControlValueAccessor,
...
 
+  onDocumentClick(event) {
+    if (this.isPopupOpen()) {
+      if (!this._isClickInWidget(event)) {
+        this.dismissPopup();
+      }
+    }
+  }
+
+  _isClickInWidget(event) {
```
maxokorokov: should be private

src/util/util.ts
```diff
@@ -37,3 +37,15 @@ export function padNumber(value: number) {
 export function regExpEscape(text) {
   return text.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
 }
+
+/**
+ * @param possibleAncestor Element to verify as being the reference element or one of its ancestors
+ * @param reference The reference element to check against
+ */
+export function isElementOrAncestorOf(possibleAncestor, reference) {
```
pkozlowski-opensource: Couldn't we use https://developer.mozilla.org/en-US/docs/Web/API/Node/contains instead?

pkozlowski-opensource:
@ymeine could you please rebase and check my comment regarding the `isElementOrAncestorOf`?

Another thing: it seems like this PR tackles 2 separate issues:
* clicks inside typeahead popover
* opening popup on search.

We should _really_ try to have _very minimal_ PRs so those are easier to review / track / revert if needed. Would you mind splitting this PR into 2 separate PRs, each one tackling one issue?

src/util/util.ts
ymeine: Sure, I discovered a bit later that this was used in the rest of the code. However as discussed offline indeed we don't even need to use such a feature for specific case: I can just compare the event target with the input element on which the directive is applied. For now, there is no possibility of hierarchy.

ymeine:
@pkozlowski-opensource Comment checked and I am going to open two distinct PRs as requested, this makes sense. However, the "open on focus" depends on the "don't close on input click", so future possible single reverts is compromised.

pkozlowski-opensource:
Closing in favour of #1989 and #1990

PR #1372, opened 2017-03-12
feat(tooltip): add custom classes to tooltip

giampierobono:
Before submitting a pull request, please make sure you have at least performed the following:

 - [x] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) guide.
 - [x] built and tested the changes locally.
 - [x] added/updated any applicable tests.
 - [x] added/updated any applicable API documentation.
 - [x] added/updated any applicable demos.

PR #1951, opened 2017-11-09
feat(tooltip): Add 'class' option

### LINKED ISSUE #1349: Add custom class to tooltips - ngbTooltipClass
A great feature could making us able to set a custom class on tooltip for setting custom styles depending about this classe.

For example, in my case, I display a variable score in a `<i>` tag, and each tag have color depending of its score _(eg: 3/10 => .red class)_. 

It would be wonderfull if I could set a class by the way of interpolation : 

```
<ul>
    <template ngFor let-score [ngForOf]="scores | orderBy: ['label']">

        <template #tipScore>{{score.label}} : <strong>{{score.score}}</strong> /10</template>

        <li *ngIf="score.name != 'general'">
            <i [ngbTooltip]="tipScore" [ngbTooltipClass]="(score.note | colorScore)"></i>
        </li>
        
    </template>
</ul>
```

The class should set through a [ngbTooltipClass] _(in my case the classname would be returned by pipe colorScore)_.

**Edit :** 
There is a poor way to perform this, but which has its own limits and depends of the state of [container].

When the default behavior is running the container is the element where the [ngbTooltip] directive is attached. (in my case this is `<i>`), we could provide custom styles.

**Example :**

1) Attach a class to the element node which contains the tooltip :
`<i class="yellow" [ngbTooltip]="tipScore"></i>`

2) Then use this selector for cover the tooltip and setting some styles : 
```
[ng-reflect-ngb-tooltip].yellow + .tooltip {
    background-color: yellow;
}
```

But when the [container] is changes to _body_, I unknown a way to target the tooltip and apply a particular styles.

bastienmoulia:
Fix #1349

### DIFF
--- a/src/tooltip/tooltip.ts
+++ b/src/tooltip/tooltip.ts
@@ -27,7 +27,7 @@ let nextId = 0;
   selector: 'ngb-tooltip-window',
   changeDetection: ChangeDetectionStrategy.OnPush,
   host: {
-    '[class]': '"tooltip show bs-tooltip-" + placement.split("-")[0]+" bs-tooltip-" + placement',
+    '[class]': '"tooltip show bs-tooltip-" + placement.split("-")[0]+" bs-tooltip-" + placement + " " + class',
     'role': 'tooltip',
     '[id]': 'id'
   },
@@ -49,7 +49,7 @@ let nextId = 0;
     :host.bs-tooltip-left .arrow, :host.bs-tooltip-right .arrow {
       top: 50%;
     }
-    
+
     :host.bs-tooltip-left-top .arrow, :host.bs-tooltip-right-top .arrow {
       top: 0.7em;
     }
@@ -63,6 +63,7 @@ let nextId = 0;
 export class NgbTooltipWindow {
   @Input() placement: Placement = 'top';
   @Input() id: string;
+  @Input() class: string;
 
   constructor(private _element: ElementRef, private _renderer: Renderer2) {}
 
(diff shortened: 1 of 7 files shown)


maxokorokov:
Haven't checked your PR yet, but there is a similar open: #1372 (though haven't checked it either :)

src/tooltip/tooltip.ts
```diff
@@ -27,7 +27,7 @@ let nextId = 0;
   selector: 'ngb-tooltip-window',
   changeDetection: ChangeDetectionStrategy.OnPush,
   host: {
-    '[class]': '"tooltip show bs-tooltip-" + placement.split("-")[0]+" bs-tooltip-" + placement',
+    '[class]': '"tooltip show bs-tooltip-" + placement.split("-")[0]+" bs-tooltip-" + placement + " " + class',
```
pkozlowski-opensource: We should have a check that `class` is defined and not empty, otherwise we might end up with ` undefined` in the class list. Please a test for this case.

pkozlowski-opensource:
@bastienmoulia sorry for the delayed response. I've added a comment in the PR. It will need a rebase as well. Thnx!

WilliamRADFunk:
I could really use this functionality. Any estimate on when this change will get pulled in?

msosa:
@bastienmoulia are you going to make the review suggested by @pkozlowski-opensource ? if not I may just open up a new PR with the changes. This seems pretty useful

Sljux:
@msosa I think you should go ahead and create a PR, @bastienmoulia doesn't seem active. If not, I can do it

pkozlowski-opensource:
Going to close this one as it doesn't merge cleanly and we haven't heard from @bastienmoulia 

If anyone wants to open a fresh PR you can get inspiration from changes similar to ones done for popover in 483bd05

Issue #1970, opened 2017-11-17
Feature: accordion functions to collapse or expand and also collapse or expand all


Thorski:
Feature Request for ngbAccordion:

Add functions to collapse/hide or expand/show a pane
Add function to collapse/hide all panes
Add function returning if a pane is collapsed/hidden or expanded/shown.

I have a situation where it has been hard for me to keep track of which accordions are expanded/shown and which are closed/hidden.   All I have available is the toggle() function.   It would be very nice to have the above functions.

Thank you.

pkozlowski-opensource:
Makes sense. Would you like to send a PR for this?

Thorski:
Ok, I will try.

Thorski:
Is it possible to get my PR into a build or what do I need to do?

Thanks.

alignsoft:
This would be incredibly useful.

PR #1978, opened 2017-11-20
feat(accordion): add methods to expand/collapse/collapseAll panels

Thorski:
### feat(accordion): add methods to expand/collapse/collapseAll panels

Added
  expand, collapse, collapseAll, and isExpanded methods.

#1970

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -110,7 +110,7 @@ export interface NgbPanelChangeEvent {
             {{panel.title}}<ng-template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></ng-template>
           </a>
         </div>
-        <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'" 
+        <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'"
              class="card-body {{panel.isOpen ? 'show' : null}}" *ngIf="!destroyOnHide || panel.isOpen">
              <ng-template [ngTemplateOutlet]="panel.contentTpl.templateRef"></ng-template>
         </div>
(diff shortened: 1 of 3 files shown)


Thorski:
What needs to be done to get this into a build?   

Thanks.

christarczon:
Hopefully this is accepted. Can you add expandAll() as well?

Thorski:
@christarczon I don't know if I have missed a step or why this has not been accepted.   I wish someone would please advise, I would like to use this from an official release.   An expandAll could be added easily.

src/accordion/accordion.ts
```diff
@@ -175,6 +175,44 @@ export class NgbAccordion implements AfterContentChecked {
...
   }
 
+  /**
+   * Programmatically expand a panel with a given id if it is closed.
+   */
+  expand(panelId: string) {
+    const panel = this.panels.find(p => p.id === panelId);
+
+    if (panel && !panel.disabled && !panel.isOpen) {
+      this.toggle(panelId);
```
pkozlowski-opensource: IMO we should it differently - the logic of expanding should be in this method and `toggle` should call it based on the state. In other words - the `toggle` method should make use of the new `expand` / `collapse ` and not the other way around.

src/accordion/accordion.ts
```diff
@@ -175,6 +175,44 @@ export class NgbAccordion implements AfterContentChecked {
...
+    this.panels.forEach(panel => { panel.isOpen = false; });
+    this._updateActiveIds();
+  }
+
+  /**
+   * Returns if a panel is currently expanded.
+   */
+  isExpanded(panelId): boolean {
+    const panel = this.panels.find(p => p.id === panelId);
+    return panel.isOpen;
```
pkozlowski-opensource: This will crash if there is no panel with a given `panelId` if I'm not mistaken? Could you please add a test for this case and fix the code if needed?

src/accordion/accordion.ts
```diff
@@ -175,6 +175,44 @@ export class NgbAccordion implements AfterContentChecked {
...
+    if (panel && !panel.disabled && panel.isOpen) {
+      this.toggle(panelId);
+    }
+  }
+
+  /**
+   * Programmatically collapse all panels.
+   */
+  collapseAll() {
+    this.panels.forEach(panel => { panel.isOpen = false; });
```
pkozlowski-opensource: This should rather delegate to the `collapse` call - otherwise we are not taking the `disabled` flag into account. Please add a test for the case assuring that we are not collapsing an open, disabled panel.

pkozlowski-opensource:
@Thorski sorry for the delay reviewing this PR - I've finally got to it and left few comments. Generally speaking:
* the implementation is "light" on tests and might be missing some use-cases;
* I would approach implementation differently to make it simpler and make sure that we are not missing corner cases.

Additionally I would like to propose that we do the following renaming:
* `expand` => `open`
* `expandAll` => `openAll`
* `collapse` => `close`
* `collapseAll` => closeAll
* `isExpanded` => `isOpen`

This would align names with other directives in this library. WDYT?

src/accordion/accordion.ts
Thorski: I wasn't sure if the disabled flag should take effect here or not, because the 'closeOtherPanels' option I believe takes effect even when a panel is disabled (see toggle and _closeOthers).

Thorski: Also, I am not sure how to make a panel in the state of 'open and disabled'.   Is it possible?  If I set disabled on a panel, it closes.   How do I do this?

pkozlowski-opensource:
This PR doesn't merge cleanly, is very light on tests and has several review comments not addressed so we are going to work towards merging #2595 instead.

PR #1988, opened 2017-11-24
fix(buttons): mark the form control of a ngbRadioGroup as touched

### LINKED ISSUE #1987: ngbRadioGroup doesn't mark the form control as touched
### Bug description:

When a ngbRadioGroup (or, rather, one of its buttons) loses the focus, the underlying form control should be marked as touched. But it isn't. This is important to me because in case no option is selected, and the form control is marked as required, I want to display an error message only when the control has been touched. 

I'll submit a PR to fix this.

http://plnkr.co/edit/YNagzOqkgIBymo0amPVq?p=preview

Angular: 4.0.3 or 5.0.x

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta

jnizet:
fix #1987

### DIFF
--- a/src/buttons/radio.spec.ts
+++ b/src/buttons/radio.spec.ts
@@ -3,7 +3,7 @@ import {By} from '@angular/platform-browser';
 import {createGenericTestComponent} from '../test/common';
 
 import {Component} from '@angular/core';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
LGTM, thnx!

PR #1989, opened 2017-11-24
fix(typeahead): don't close dropdown on typeahead input click

ymeine:


src/typeahead/typeahead.ts
```diff
@@ -214,6 +214,12 @@ export class NgbTypeahead implements ControlValueAccessor,
     this._renderer.setProperty(this._elementRef.nativeElement, 'disabled', isDisabled);
   }
 
+  onDocumentClick(event) {
+    if (event.target !== this._viewContainerRef.element.nativeElement) {
```
pkozlowski-opensource: Shouldn't it be `this._elementRef. nativeElement`?

ymeine: Yes it should, my bad, fixing it now

pkozlowski-opensource:
LGTM, thnx!

PR #1990, opened 2017-11-24

ymeine:


pkozlowski-opensource:
@ymeine could you please rebase so only the second commit is part of this PR? Also, could you amend this commit message by changing `allowing` -> `allow`. Thnx!

demo/src/app/components/typeahead/demos/focus/typeahead-focus.ts
```diff
@@ -0,0 +1,38 @@
...
+  'South Carolina', 'South Dakota', 'Tennessee', 'Texas', 'Utah', 'Vermont', 'Virgin Islands', 'Virginia',
+  'Washington', 'West Virginia', 'Wisconsin', 'Wyoming'];
+
+@Component({
+  selector: 'ngbd-typeahead-focus',
+  templateUrl: './typeahead-focus.html',
+  styles: [`.form-control { width: 300px; }`]
+})
+export class NgbdTypeaheadFocus {
+  public model: any;
```
pkozlowski-opensource: Please remove `public` modifiers - it is public by default.

src/typeahead/typeahead.spec.ts
```diff
@@ -921,9 +1011,18 @@ class TestComponent {
 
   form = new FormGroup({control: new FormControl('', Validators.required)});
 
+  findOutput$: Observable<any[]>;
+
   @ViewChild(NgbTypeahead) typeahead: NgbTypeahead;
+  public focus$ = new Subject<string>();
```
pkozlowski-opensource: Please remove `public` modifiers.

src/typeahead/typeahead.spec.ts
```diff
@@ -194,6 +197,93 @@ describe('ngb-typeahead', () => {
...
+           expect(fixture.componentInstance.model).toBe(undefined);
+           expect(getNativeInput(compiled).value).toBe('');
+         };
+
+         const checkWindowIsOpen = () => { expect(getWindow(compiled)).not.toBeNull(); };
+
+         // focusing the input triggers a search and opens the dropdown
+         getNativeInput(compiled).focus();
+         // on IE the focus can be asynchronous, so we need to wait a tick before continuing
+         await new Promise(resolve => setTimeout(resolve, 0));
```
pkozlowski-opensource: Wouldn't `Promise.resolve` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) work here?

More importantly, do we need to trigger native `focus` event here? Wouldn't Angular testing trigger event be enough?

src/typeahead/typeahead.ts
```diff
@@ -302,6 +302,7 @@ export class NgbTypeahead implements ControlValueAccessor,
 
     if (!defaultPrevented) {
       this.writeValue(result);
+      this._userInput = result;
```
pkozlowski-opensource: This looks odd... `_userInput` should be a string while here we would be assigning an item which is potentially a complex object. Why do we need this assignment?

If you need to store `_userInput` you should probably store a stringified value inside the `this.writeValue` method.

src/typeahead/typeahead.spec.ts
```diff
@@ -194,6 +197,93 @@ describe('ngb-typeahead', () => {
...
+           fixture.detectChanges();
+
+           // select with click
+           getWindowLinks(fixture.debugElement)[0].triggerEventHandler('click', {});
+           fixture.detectChanges();
+           expectInputValue(compiled, 'one');
+
+           // open again but with focus
+           getNativeInput(compiled).blur();
+           getNativeInput(compiled).focus();
```
pkozlowski-opensource: Could you please add expectation here to prove that a window with results is open? I don't see `(focus)` handler on this `<input>` field so I'm not sure what would trigger opening?

src/typeahead/typeahead.ts
```diff
@@ -312,7 +313,7 @@ export class NgbTypeahead implements ControlValueAccessor,
   }
 
   private _showHint() {
-    if (this.showHint) {
+    if (this.showHint && this._userInput != null) {
```
pkozlowski-opensource: What is the exact condition under which `this._userInput` gets `null`? Maybe we should rather assume that user 's input is always a string, instead of checking this in several places?

src/typeahead/typeahead.spec.ts
ymeine: Sure, initially I tried with a timeout higher than 0, so I didn't realize that for this "next tick" use case using `Promise.resolve()` would work.

Regarding the native `focus` event it's indeed because we don't use a regular Angular event handler, it work with Observables on the user side.

src/typeahead/typeahead.spec.ts
ymeine: Yes, it was there initially but I removed it. However it might be better testing more than not enough. Adding it.

PR #2003, opened 2017-12-01
fix(datepicker): support `NgbDateAdapter` in `NgbInputDatepicker`

### LINKED ISSUE #2002: Issue with NgbDateAdapter when used with a Datepicker on an input
### Bug description:

Using NgbDateAdapter with anything other than the original NgbDateStruct will not work with a datepicker on an input. It looks like this was missed in my previous PR and when selecting a date from the popup it will show as blank on the input even though the model gets updated. I have something sort of working but having issues with dealing with any type of invalid input and it just causes it to crash. If i take out the second ternary here in datepicker-input on the function manualDateChange, it does end up fixing the crashing error after all my other changes. Not sure if that is the right solution though.

```typescript
 this._onChange(this._model ? this._model.toStruct() : (value === '' ? null : value));
```

 If it helps I can open up a PR with the potential fix

As this is not yet released I cannot make a plunker

ng-bootstrap: master branch

maxokorokov:
1. Support for the `NgbDateAdapter` in the datepicker with input
2. Default `NgbDateStructAdapter` is more forgiving now (`'2017-11'` transformed into `{year: 2017, month: 11, day: 1}`)so I changed one test case for this as well.

@msosa, could you take a look?

Fixes #2002

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -24,6 +24,7 @@ import {NgbDateParserFormatter} from './ngb-date-parser-formatter';
 
 import {positionElements, PlacementArray} from '../util/positioning';
 import {NgbDateStruct} from './ngb-date-struct';
+import {NgbDateAdapter} from './ngb-date-adapter';
 import {NgbCalendar} from './ngb-calendar';
 import {NgbDatepickerService} from './datepicker-service';
 
@@ -163,7 +164,8 @@ export class NgbInputDatepicker implements OnChanges,
   constructor(
       private _parserFormatter: NgbDateParserFormatter, private _elRef: ElementRef, private _vcRef: ViewContainerRef,
       private _renderer: Renderer2, private _cfr: ComponentFactoryResolver, ngZone: NgZone,
-      private _service: NgbDatepickerService, private _calendar: NgbCalendar) {
+      private _service: NgbDatepickerService, private _calendar: NgbCalendar,
+      private _ngbDateAdapter: NgbDateAdapter<any>) {
     this._zoneSubscription = ngZone.onStable.subscribe(() => {
       if (this._cRef) {
         positionElements(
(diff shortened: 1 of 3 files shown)


msosa:
LGTM, this is what I had but there seems to be an issue with the current `NgbDateNativeAdapter` in the demo when using the `manualDateChange`. Sort of crashes in the `formModel`, but maybe that's just a problem with the adapter itself? 

I added `typeof date.getFullYear === 'function'` in the ternary and the error doesn't show but change it in the input still seems a little weird. Not sure if there really anything that can be done about that

maxokorokov:
@msosa I think it's the adapter implementation issue. 

I think your example crashes when you set something like a string to the `model` value → your native adapter implementation will crash. It might happen in the demo if you add both `ngbDatepicker` and `ngbInputDatepicker` and share the same `model` between them (not your case by any chance?).

Actually the `NgbDateStructAdapter` also not working well, it's just not crashing. I'll open another PR to ignore invalid inputs for it.

msosa:
Ah yes I am using both `ngDatepicker` and `ngbInputDatepicker` together on the same model.

Awesome, thanks!

battistaar:
I found a problem with validation, the input field is always invalid using a custom `NgbDateAdapter`. I think the problem is in the `validate` method of `NgbInputDatepicker`.
I temporarely fixed it doing
```
validate(c: AbstractControl): {[key: string]: any} {   
   // convert value to NgbDateStruct using the adapter
   const value = this._ngbDateAdapter.fromModel(c.value);

    if (value === null || value === undefined) {
      return null;
    }

    if (!this._calendar.isValid(value)) {
      return {'ngbDate': {invalid: c.value}};
    }
```
I'm quite busy at the moment, I'll do more tests next week.

maxokorokov:
@battistaar yes, you're right, thanks. Forgot to add conversion here. It's obviously working without it for the default adapter...

Will push and update the demo to add datepicker with input

maxokorokov:
Opened a separate #2025 just with the demo update

liviu-c:
I'm trying to make the model as native Date, but the input format to be "dd-mm-yyyy" and not the standard iso.

when I pick a date from datepicker, the model changes back to ngbDateStruct... instead of Date.
Am I doing something wrong?

http://plnkr.co/edit/6xICvZtBCPNraavnhaz2?p=preview

msosa:
Yeah you you're not actually using the adapter, look at how the demo does it 

```
providers: [{provide: NgbDateAdapter, useClass: NgbDateNativeAdapter}]
})
```

liviu-c:
oh crap. I got the syntax wrong :) thx for the comment.
it's now working fine : http://plnkr.co/edit/7jU6TMamLQfqjZUNDyQN?p=preview

PR #2008, opened 2017-12-04
chore(): removed duplicate import (component, config) on alert module…

Rjoydip:


pkozlowski-opensource:
Those are _not_ duplicates! One set is of `import`s and another one - of `export`s. This change is not good and it makes the TravisCI to fail, see: https://travis-ci.org/ng-bootstrap/ng-bootstrap/builds/311196648?utm_source=github_status&utm_medium=notification

PR #2012, opened 2017-12-05
demo(typeahead): use HttpClient instead of deprecated Jsonp

maxokorokov:
* remove [deprecated `Jsonp` service](https://angular.io/api/http/Jsonp#deprecation-notes) from demo app
* use CORS instead of JSONP, as, apparently, wikipedia supports it when specifying `origin: *` → https://www.mediawiki.org/wiki/API:Cross-site_requests

### DIFF
--- a/demo/src/app/components/typeahead/demos/http/typeahead-http.ts
+++ b/demo/src/app/components/typeahead/demos/http/typeahead-http.ts
@@ -1,5 +1,5 @@
 import {Component, Injectable} from '@angular/core';
-import {Jsonp, URLSearchParams} from '@angular/http';
+import {HttpClient, HttpParams} from '@angular/common/http';
 import {Observable} from 'rxjs/Observable';
 import 'rxjs/add/observable/of';
 import 'rxjs/add/operator/catch';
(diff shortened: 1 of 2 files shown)


demo/src/app/components/typeahead/demos/http/typeahead-http.ts
```diff
@@ -10,25 +10,27 @@ import 'rxjs/add/operator/map';
...
-    params.set('search', term);
-    params.set('action', 'opensearch');
-    params.set('format', 'json');
-    params.set('callback', 'JSONP_CALLBACK');
-
-    return this._jsonp
-      .get(wikiUrl, {search: params})
-      .map(response => <string[]> response.json()[1]);
+    return this.http
+      .get(WIKI_URL, {params: PARAMS.set('search', term)})
```
pkozlowski-opensource: Looks kind of strange to mutate a shared object.. How about `new HttpParams(Object.assign(PARAMS, {'search': term})`?

pkozlowski-opensource:
Thnx for removing deprecated API usage! LGTM, generally speaking, just left one comment.

demo/src/app/components/typeahead/demos/http/typeahead-http.ts
maxokorokov: It's not mutating the existing one, see [implementation and doc](https://github.com/angular/angular/blob/6911a250ef82ef0cf41c42d34b75885a25c0301b/packages/common/http/src/params.ts#L164).

PR #1614, opened 2017-06-12
refactor(datepicker): simplify css handling of the navigation system

ExFlo:
Before submitting a pull request, please make sure you have at least performed the following:

 - [] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) guide.
 - [] built and tested the changes locally.
 - [] added/updated any applicable tests.
 - [] added/updated any applicable API documentation.
 - [] added/updated any applicable demos.

PR #2013, opened 2017-12-06
fix(datepicker): add class "show" to visible datepicker dropdown

wpanas:
Default behaviour of class "dropdown-menu" in bootstrap 4 is to make it invisble. To make it visible it is necessary to add class "show" to the element.  Forcing display: inline-block is dirty fix and cause problems with custom boostrap templates.

Class "d-inline-block" is deleted only when popup styles are applied, so it doesn't cause problems with inline element.

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -308,6 +308,7 @@ export class NgbInputDatepicker implements OnChanges,
   private _applyPopupStyling(nativeElement: any) {
     this._renderer.addClass(nativeElement, 'dropdown-menu');
     this._renderer.setStyle(nativeElement, 'padding', '0');
+    this._renderer.addClass(nativeElement, 'show');
   }
 
   private _subscribeForDatepickerOutputs(datepickerInstance: NgbDatepicker) {


src/datepicker/datepicker-input.ts
```diff
@@ -306,8 +306,10 @@ export class NgbInputDatepicker implements OnChanges,
   }
 
   private _applyPopupStyling(nativeElement: any) {
+    this._renderer.removeClass(nativeElement, 'd-inline-block');
```
pkozlowski-opensource: Hmmm, do we need to do this since the datepicker widget has `d-inline-block` on host?

wpanas: Yup. Cause some themes will not display this element. For example, they make it not visible by scaling it to 0 and changing opacity to 0, so display doesn't matter at all.

wpanas: We need to make it truly not visible, so it will have animation once made visible by class "show". It won't be a case with opacity 0, but not all themes use it.

ExFlo: Hi
There is no more the d-inline-block class on the datepicker element, I removed it in pull request 
#1614
This pull request will solve #2088.
BR

wpanas: Ok, so I will remove this line in next commit

mboughaba:
Hi,
Using `bootstrap 4.0.0`, `ngb 1.0.0-beta.9` and after applying this PR manually the `datepicker not showing issue` was fixed.

Cheers,

wpanas:
Ok. I removed the controversial line

wpanas:
How changes in typeahead components apply to datepicker popup?

pkozlowski-opensource:
> How changes in typeahead component apply to datepicker popup?

They don't "apply" but are _related_, as both components use dropdown HTML and styling to display popups

PR #2024, opened 2017-12-08
Patch license to be valid

bastienmoulia:
The old one was deprecated, see https://docs.npmjs.com/files/package.json#license


pkozlowski-opensource:
Thnx for the PR! Seems like sth went wrong with the rebase / merge? Could you make it so we've got only one commit (b191049) in this PR? Thnx!

bastienmoulia:
@pkozlowski-opensource I will try, but i'm not an expert with git

PR #2048, opened 2017-12-19
chore: use husky to handle git hooks

### LINKED ISSUE #1969: npm install in git worktrees fails when attempting to install git hook
### Bug description

Currently when doing a `npm install` the `misc/validate-commit.install.js` script is called and installs a git hook used to validate commit messages.

The problem occurs when the installation is done from a git worktree. Indeed in this case there is no `.git` folder but instead a `.git` file with some properties, including the `gitdir` one which indicates the path of the actual git directory associated to this worktree.

Currently the installation fails because indeed the target hook path does not exist, but writing to it is not possible since `.git` is not a folder.

The solution would be to implement logic to support worktrees or use [typicode/husky: Git hooks made easy](https://github.com/typicode/husky) which already supports them.

_Not applicable_

- `Angular`: _Not applicable_
- `ng-bootstrap`: `1.0.0-beta.5`
- `Bootstrap`: _Not applicable_

ymeine:
That simplifies things and also brings support for git worktrees.

fix #1969


pkozlowski-opensource:
LGTM. Travis build is failing for some strange reason, restarting it to see what is going on.

PR #2049, opened 2017-12-19
feat(modal): opened aria attribute aria-labelledby for accessibility

### LINKED ISSUE #1477: Modal - Missing aria-labelledby on modal container
To improve the modals accessibility, it could be useful to add the aria-labelledby attribute on the modal container (the element having the role=dialog set).

This attribute will help SR users in having a clear "description" about the modal content and should be linked to the modal header (normally the h tag).

According to W3C specifications(here: https://cdn.rawgit.com/w3c/aria-practices/master/aria-practices.html#dialog_modal) this attribute seems to be mandatory.

ymeine:
Opened attribute `aria-labelledby`.

close #1477

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -14,7 +14,6 @@ import {
 
 import {ModalDismissReasons} from './modal-dismiss-reasons';
 import {ngbFocusTrap} from '../util/focus-trap';
-import {isDefined} from '../util/util';
 
 @Component({
   selector: 'ngb-modal-window',
@@ -24,7 +23,7 @@ import {isDefined} from '../util/util';
     'tabindex': '-1',
     '(keyup.esc)': 'escKey($event)',
     '(click)': 'backdropClick($event)',
-    '[attr.aria-labelledby]': 'isDefined(ariaLabelledBy) ? ariaLabelledBy : null',
+    '[attr.aria-labelledby]': 'ariaLabelledBy',
   },
   template: `
     <div [class]="'modal-dialog' + (size ? ' modal-' + size : '') + (centered ? ' modal-dialog-centered' : '')" role="document">
@@ -37,13 +36,12 @@ export class NgbModalWindow implements OnInit,
   private _document: any;
   private _elWithFocus: Element;  // element that is focused prior to modal opening
 
+  @Input() ariaLabelledBy: string;
   @Input() backdrop: boolean | string = true;
   @Input() centered: string;
   @Input() keyboard = true;
   @Input() size: string;
   @Input() windowClass: string;
-
-  @Input() ariaLabelledBy: string;
 
   @Output('dismiss') dismissEvent = new EventEmitter();
 
@@ -51,8 +49,6 @@ export class NgbModalWindow implements OnInit,
     this._document = document;
     ngbFocusTrap(this._elRef.nativeElement, this.dismissEvent);
   }
-
-  isDefined() { return isDefined.apply(null, arguments); }
 
   backdropClick($event): void {
     if (this.backdrop === true && this._elRef.nativeElement === $event.target) {
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -14,6 +14,7 @@ import {
 
 import {ModalDismissReasons} from './modal-dismiss-reasons';
 import {ngbFocusTrap} from '../util/focus-trap';
+import {isDefined} from '../util/util';
 
 @Component({
   selector: 'ngb-modal-window',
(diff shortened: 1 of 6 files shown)


src/modal/modal.spec.ts
```diff
@@ -587,6 +587,45 @@ describe('ngb-modal', () => {
...
+      testAttributes({ariaLabel: label}, [['aria-label', label]]);
+    });
+
+    it('should support aria-labelledby', () => {
+      const id = 'aria-labelledby-id';
+
+      testAttributes({ariaLabelledBy: id}, [['aria-labelledby', id]]);
+    });
+
+    it('should make aria-label have priority over aria-labelledby', () => {
```
bastienmoulia: `aria-labelledby` should have priority over `aria-label`.

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

> When user agents compute the accessible name property of elements that have both an aria-labelledby attribute and an aria-label attribute, the user agents give precedence to aria-labelledby.

ymeine: That's actually what I implemented the first time, but I didnd't keep the logic for two reasons: 

- it's already done by the browser as you mentioned
- I prefer to keep the logic here focused on displaying rather than trying to guess what should be expected; let me explain below

If the user wants, for any reason, to have both `aria-label` and `aria-labelledby` present in the DOM, we shouldn't prevent him from doing so. Moreover, `aria-label` could be used as a fallback in case the id specified in `aria-labelledby` doesn't exist.

ymeine: Wait @bastienmoulia I just noticed I made a mistake and my branch is wrong, let me change it.

pkozlowski-opensource:
@ymeine thnx for PR. I can see that you are taking the same approach as https://angular-ui.github.io/bootstrap/#!#modal but I'm wondering if this is the right way to do things. Here are my concerns:
* it bloats the API (3 new options, not clear which one I should use: `ariaLabelledBy` and / or `ariaDescribedBy ` and / or `ariaLabel`)
* it requires users to specify `id` in 2 separate and distant places: in a template _and_ in the code that opens a modal (typically in TS code).

Given those concerns I would like to propose an alternative solution, see: https://github.com/ng-bootstrap/ng-bootstrap/issues/1477#issuecomment-353575476

I'm not saying that we shouldn't merge your PR but I would like to give us time to explore alternatives before we make this move. 

In fact we could probably combine 2 approaches by:
* adding only one option to modal opening options (ex. `ariaLabelledById`)
* expose this new option to a content template so users don't need to repeat it (and, potentially, we could generate it automatically if not provided).

WDYT?

src/modal/modal-window.ts
```diff
@@ -21,7 +22,8 @@ import {ModalDismissReasons} from './modal-dismiss-reasons';
     'role': 'dialog',
     'tabindex': '-1',
     '(keyup.esc)': 'escKey($event)',
-    '(click)': 'backdropClick($event)'
+    '(click)': 'backdropClick($event)',
+    '[attr.aria-labelledby]': 'isDefined(ariaLabelledBy) ? ariaLabelledBy : null',
```
maxokorokov: looks like an overkill for me, as attribute won't be there for `undefined` and `null` values anyway

src/modal/modal-window.ts
```diff
@@ -40,12 +42,16 @@ export class NgbModalWindow implements OnInit,
   @Input() size: string;
   @Input() windowClass: string;
 
+  @Input() ariaLabelledBy: string;
```
maxokorokov: Could you please put in in the alphabetical order?

https://github.com/ng-bootstrap/ng-bootstrap/wiki/Contributions%3A-Code-conventions#component-internal-organization

src/modal/modal.spec.ts
```diff
@@ -615,7 +615,26 @@ describe('ngb-modal', () => {
...
+
+      modalInstance.close('some result');
+      fixture.detectChanges();
+      expect(fixture.nativeElement).not.toHaveModal();
+    }
+
+    it('should support aria-labelledby', () => {
+      const id = 'aria-labelledby-id';
+
+      testAttributes({ariaLabelledBy: id}, [['aria-labelledby', id]]);
```
maxokorokov: Can't we simply write `expect(modalElement.getAttribute('ariaLabelledBy')).toBe('aria-labelled-by-id')`?
It's the only place `testAttributes` is used

src/modal/modal-window.ts
ymeine: true, that's something I used to be unaware of, I'll remove the check

src/modal/modal-window.ts
ymeine: I missed that, my bad, I'll change it

src/modal/modal.spec.ts
ymeine: Sure, it's a leftover from previous version, where it was used multiple times. I will remove the generic function.

PR #2050, opened 2017-12-19
feat: add keys utility to handle KeyboardEvent events

ymeine:
It helps both checking events consistently and with a safe abstract API and creating fake events for testing.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -222,7 +222,8 @@ gulp.task('enforce-format', function() {
 function doCheckFormat() {
   return gulp
       .src([
-        'gulpfile.js', 'karma-test-shim.js', 'misc/api-doc.js', 'misc/api-doc.spec.js', 'misc/demo-gen.js', PATHS.src
+        'gulpfile.js', 'karma-test-shim.js', 'misc/api-doc.js', 'misc/api-doc.spec.js', 'misc/demo-gen.js', PATHS.src,
+        '!src/util/keys.ts'
       ])
       .pipe(gulpFormat.checkFormat('file', clangFormat));
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -222,8 +222,7 @@ gulp.task('enforce-format', function() {
 function doCheckFormat() {
   return gulp
       .src([
-        'gulpfile.js', 'karma-test-shim.js', 'misc/api-doc.js', 'misc/api-doc.spec.js', 'misc/demo-gen.js', PATHS.src,
-        '!src/util/keys.ts'
+        'gulpfile.js', 'karma-test-shim.js', 'misc/api-doc.js', 'misc/api-doc.spec.js', 'misc/demo-gen.js', PATHS.src
       ])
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -225,7 +225,8 @@ gulp.task('enforce-format', function() {
 function doCheckFormat() {
   return gulp
       .src([
-        'gulpfile.js', 'karma-test-shim.js', 'misc/api-doc.js', 'misc/api-doc.spec.js', 'misc/demo-gen.js', PATHS.src
+        'gulpfile.js', 'karma-test-shim.js', 'misc/api-doc.js', 'misc/api-doc.spec.js', 'misc/demo-gen.js', PATHS.src,
+        '!src/util/keys.ts'
```
maxokorokov: I don't agree there should be exceptions for code formatting in the library source code at the file level

ymeine: I don't agree there should be code formatting automatic enforcement at all given its poor results.

I've let it until we have to integrate this PR, in case I have to change things, because honestly, working on the output of the formatter is a pain in the ass, to remain polite.

ymeine:
The additional advantage I see is also to avoid having to wonder: should I check `event.key`, `event.keyCode`, or `event.which`? Also knowing that the two latter are deprecated, but the first one works inconsistently between browsers (thanks IE).

So I see an advantage of having this utility also to perform the check directly.

Btw, a framework is something specific, which forces structure in addition to provide functionalities. Here, it's more like a library inside another library. Anyways, we already have some generic all purpose utilities in here, it's just another one.

I wanted to use a third party in order to avoid the NIH syndrome, but the project seems to be reluctant to including third-parties.

src/test/common.ts
```diff
@@ -59,3 +61,39 @@ export function isBrowser(browsers: Browser | Browser[], ua = window.navigator.u
...
   }
 }
+
+export interface FakeEvent extends Event {
+  which: number;
+  keyCode: number;
+  key: string;
+}
+
+export function createKeyboardEvent(input): Event {
```
pkozlowski-opensource: This seems useful but we should demonstrate this by actually using this helper in the existing test(s).

pkozlowski-opensource:
@ymeine is this PR still WIP? Generally speaking I don't want to merge any code that is not used. If you see value of those helpers it would be see them in action in the existing tests.

It would be really great to have small, focused PRs that improve existing tests rather than a bunch of helpers that we _might_ use in the future.

ymeine:
I will for sure extract the utility for testing, probably as a new PR, in which case we would close the current one. The new PR would then refactor the existing tests to use the utility, to make more sense.

I agree with your point and am now tending to do so (like for [feat(util): make common the enum Key by ymeine · Pull Request #2473 · ng-bootstrap/ng-bootstrap](https://github.com/ng-bootstrap/ng-bootstrap/pull/2473) which seemed to be a better alternative than this current PR).

pkozlowski-opensource:
#2507 was merged

PR #2051, opened 2017-12-19
 feat(dropdown): close on escape from anywhere 

### LINKED ISSUE #1741: dropdown: properly close on outside ESC
With the current implementation dropdowns are closed on ESC only if the trigger button is focused. It should close when ESC is pressed regardless of the element focused.

ymeine:
close #1741

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -1,5 +1,6 @@
 import {TestBed, ComponentFixture, inject} from '@angular/core/testing';
 import {createGenericTestComponent} from '../test/common';
+import {Key} from '../util/key';
 
 import {Component} from '@angular/core';
 import {By} from '@angular/platform-browser';
@@ -17,6 +18,15 @@ function getDropdownEl(tc) {
 
 function getMenuEl(tc) {
   return tc.querySelector(`[ngbDropdownMenu]`);
+}
+
+function createFakeEscapeKeyUpEvent(): Event {
+  const key = Key.Escape;
+  const event = document.createEvent('KeyboardEvent') as any;
+  let initEvent = (event.initKeyEvent || event.initKeyboardEvent).bind(event);
+  initEvent('keyup', true, true, window, 0, 0, 0, 0, 0, key);
+  Object.defineProperties(event, {which: {get: () => key}});
+  return event;
 }
 
 const jasmineMatchers = {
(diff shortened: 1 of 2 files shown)


devoto13:
Not sure listening for ESC on `document` is the right choice. Consider for example a dropdown inside a modal. What I would expect as a user is that first ESC click closes dropdown and second ESC click closes modal.

ymeine:
I don't know any other way to catch events from any element without listening to the `document` object.

The requirement is to close the dropdown on any <kbd>Escape</kbd> press.

What you are saying is that in addition to listening to the event and closing the dropdown, we should also prevent catching this event for let's say a modal to avoid closing it at the same time?

If so, that's a valid requirement, and we have to think about a generic solution working well for all widgets with an individual instance, but also for nested instances (dropdown inside modal, etc.).

src/dropdown/dropdown.ts
```diff
@@ -145,6 +146,13 @@ export class NgbDropdown implements OnInit {
       this._open = true;
       this._positionMenu();
       this.openChange.emit(true);
+      this._escapeListener = (event) => {
+        if (isEscape(event)) {
+          this.closeFromOutsideEsc();
+          this.changeDetector.detectChanges();
```
maxokorokov: hmm, why would you need to run the CD after closing?

src/dropdown/dropdown.ts
```diff
@@ -120,7 +121,7 @@ export class NgbDropdown implements OnInit {
    */
   @Output() openChange = new EventEmitter();
 
-  constructor(config: NgbDropdownConfig, ngZone: NgZone) {
+  constructor(config: NgbDropdownConfig, private ngZone: NgZone, private changeDetector: ChangeDetectorRef) {
```
maxokorokov: Should be `_ngZone` and `_changeDetector`

https://github.com/ng-bootstrap/ng-bootstrap/wiki/Contributions%3A-Code-conventions

ymeine: Damn, I thought I had the guidelines in mind, I'll read it again, more carefully. And I agree with that part.

src/dropdown/dropdown.ts
ymeine: Honestly I can't remember why, probably it wasn't working properly otherwise, to be tested again.

ymeine:
I can remove the `keys.ts` utility, it's subject to discussion and would slow down the acceptation of other features.

Regarding the event handler registered outside Angular: if we don't don't it, isn't the change detection triggered after every `keyup` then?

Now, about the Observable implementation, I'm just in love with it. Truly. That's totally how I want to write this kind of code.

src/dropdown/dropdown.ts
```diff
@@ -101,13 +110,11 @@ export class NgbDropdownToggle extends NgbDropdownAnchor {
...
-    '[class.show]': 'isOpen()',
-    '(keyup.esc)': 'closeFromOutsideEsc()',
-    '(document:click)': 'closeFromClick($event)'
-  }
+  host: {'[class.show]': 'isOpen()', '(document:click)': 'closeFromClick($event)'}
 })
-export class NgbDropdown implements OnInit {
+export class NgbDropdown implements OnInit,
+    OnDestroy {
+  private _closed$ = new Subject();
```
maxokorokov: hmm, why create a new one, if you already have `openChange`?

`takeUntil(this.openChange.pipe(filter(opened => opened === false)))` ?

src/dropdown/dropdown.ts
```diff
@@ -167,6 +176,13 @@ export class NgbDropdown implements OnInit {
       this._open = true;
       this._positionMenu();
       this.openChange.emit(true);
+      this._ngZone.runOutsideAngular(
+          () => fromEvent<KeyboardEvent>(this._document, 'keyup')
+                    .pipe(takeUntil(this._closed$), filter(e => e.which === Key.Escape))
+                    .subscribe(() => {
+                      this.closeFromOutsideEsc();
+                      this._changeDetector.detectChanges();
```
maxokorokov: so have you remembered why you need this one? :)

src/dropdown/dropdown.ts
```diff
@@ -9,11 +9,20 @@ import {
...
+  OnDestroy,
+  ChangeDetectorRef
 } from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+import {Subject, fromEvent} from 'rxjs';
+import {filter, takeUntil} from 'rxjs/operators';
 import {NgbDropdownConfig} from './dropdown-config';
 import {positionElements, PlacementArray, Placement} from '../util/positioning';
 
+enum Key {
```
maxokorokov: Do you mind rebasing to use common Keys, please?

src/dropdown/dropdown.ts
```diff
@@ -9,11 +9,20 @@ import {
   ContentChild,
   NgZone,
   Renderer2,
-  OnInit
+  OnInit,
+  OnDestroy,
+  ChangeDetectorRef
 } from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+import {Subject, fromEvent} from 'rxjs';
```
maxokorokov: Subject is not used anywhere

src/dropdown/dropdown.ts
```diff
@@ -101,13 +110,10 @@ export class NgbDropdownToggle extends NgbDropdownAnchor {
...
-  host: {
-    '[class.show]': 'isOpen()',
-    '(keyup.esc)': 'closeFromOutsideEsc()',
-    '(document:click)': 'closeFromClick($event)'
-  }
+  host: {'[class.show]': 'isOpen()', '(document:click)': 'closeFromClick($event)'}
 })
-export class NgbDropdown implements OnInit {
+export class NgbDropdown implements OnInit,
+    OnDestroy {
```
maxokorokov: Thanks for adding it 👍

src/dropdown/dropdown.spec.ts
```diff
@@ -361,44 +374,47 @@ describe('ngb-dropdown-toggle', () => {
...
-      <div ngbDropdown>
-          <button ngbDropdownToggle>Toggle dropdown</button>
-          <div ngbDropdownMenu></div>
-      </div>`;
-
-    const fixture = createTestComponent(html);
-    const compiled = fixture.nativeElement;
-    const buttonEl = compiled.querySelector('button');
+  describe('escape closing', () => {
+    const testEscapeClosing = ({autoClose, getElementForEventDispatch}) => {
```
maxokorokov: Personally, I would really much prefer reading 4 times duplicated code for many reasons:

- you have all you need in a single `it()` function, don't have to jump back and forth trying to understand what is passed where
- don't have to think when reading the tests (ex. conditions, remembering what were the arguments passed, reading `expectation.not.toBeShown()` vs `expect(compiled).not.toBeShown()`, etc.)
- all other tests (at least for this component) are written in this style

P.S. Would have nothing against this, if it was in the component code, not the test

@pkozlowski-opensource, should we change or leave like this?

PR #2052, opened 2017-12-19
chore: use @angular/cli for build and tests

### LINKED ISSUE #1474: Migrate demo server to use Angular CLI
Currently we use webpack to run our demo server and to build our documentation application. We should consider migrating over to usage of the CLI in order to offload those responsibilities to Angular's tooling.

ymeine:
fixes #1474


demo/src/polyfills.ts
```diff
@@ -2,7 +2,7 @@ import 'core-js/shim';
 
 require('zone.js/dist/zone');
```
maxokorokov: I think we should use `poyfills.ts` that comes with angular cli

package.json
```diff
@@ -50,27 +50,25 @@
...
     "@angular/http": "5.0.2",
     "@angular/platform-browser": "5.0.2",
     "@angular/platform-browser-dynamic": "5.0.2",
     "@angular/router": "5.0.2",
-    "@ngtools/webpack": "1.8.2",
-    "@types/jasmine": "2.5.41",
+    "@types/jasmine": "2.5.51",
+    "@types/jasminewd2": "~2.0.2",
     "@types/node": "^6.0.46",
     "angular2-template-loader": "^0.6.0",
```
maxokorokov: I think it should be also safe to remove all webpack-demo-related packages, like:
```
angular2-template-loader
css-loader
ts-loader
etc...
```

maxokorokov:
Bumped to cli 6, works perfectly fine!

Will now try to add e2e tests for ng-bootstrap and actually build ng-bootstrap itself with cli. 
If this works, I think it will be safe to merge

@ymeine, a short question: why all these changes to get the version from `package.json` in this PR ?

maxokorokov:
Hey, @ymeine, looks like you're accidentally removing commits I push to this PR branch when you rebasing your things on master. So I'll cherry-pick your initial changes into a new #2335 (to avoid losing everything) and add mine on top of them.

PR #2053, opened 2017-12-19
fix(timepicker): fixing tab navigation & handling keyboard arrows

### LINKED ISSUE #1836: Tab-based navigation in timepicker with enabled spinners
### Bug description:

Using a timepicker with spinners enabled, I expect that I can change from the hours field to the minutes field using tab on the keyboard. While this works without spinners, it doesn't work with spinners enabled. Here, pressing tab selects the first spinner control.

See demo page: https://ng-bootstrap.github.io/#/components/timepicker/examples

Example showing the requested behaviour: http://angular-ui.github.io/bootstrap/#!#%2Ftimepicker

Angular: 4.3.6

ng-bootstrap: 1.0.0-beta.2

Bootstrap: 4.0.0-beta

ymeine:
Fixing tab navigation between inputs and handling keyboard arrow up and down to change time.

close #1836

### DIFF
--- a/e2e-app/src/app/timepicker/navigation/timepicker-navigation.e2e-spec.ts
+++ b/e2e-app/src/app/timepicker/navigation/timepicker-navigation.e2e-spec.ts
@@ -1,0 +1,77 @@
+import {Key, ElementFinder} from 'protractor';
+
+import {openUrl, expectFocused, sendKey, getCaretPosition} from '../../tools.po';
+
+import {TimepickerNavigationPage} from './timepicker-navigation.po';
+
+describe('Timepicker', () => {
+  let page: TimepickerNavigationPage;
+
+  beforeAll(() => page = new TimepickerNavigationPage());
+  beforeEach(async() => await openUrl('timepicker/navigation'));
+
+  async function expectCaretPosition(field: ElementFinder, position: number) {
+    const {start, end} = await getCaretPosition(field);
+    expect(end).toBe(end, 'Caret should be at single position (no range selected)');
+    expect(start).toBe(position, `Caret is not at proper position for given field`);
+  }
+
+  describe('navigation', () => {
+    it(`should jump between inputs`, async() => {
+      await page.focusInputBefore();
+
+      const[hourField, minuteField, secondField] = page.getFields();
+
+      await sendKey(Key.TAB);
+      await expectFocused(hourField, 'Hour field should be focused');
+      await sendKey(Key.TAB);
+      await expectFocused(minuteField, 'Minute field should be focused');
+      await sendKey(Key.TAB);
+      await expectFocused(secondField, 'Second field should be focused');
+      await sendKey(Key.TAB);
+      await expectFocused(page.getInputAfter(), 'Input after should be focused');
+
+      await sendKey(Key.SHIFT, Key.TAB);
+      await expectFocused(secondField, 'Second field should be focused');
+      await sendKey(Key.SHIFT, Key.TAB);
+      await expectFocused(minuteField, 'Minute field should be focused');
+      await sendKey(Key.SHIFT, Key.TAB);
+      await expectFocused(hourField, 'Hour field should be focused');
+      await sendKey(Key.SHIFT, Key.TAB);
+      await expectFocused(page.getInputBefore(), 'Input before should be focused');
+    });
+  });
+
+  describe('arrow keys', () => {
+    it(`should keep caret at the end of the input`, async() => {
+      const testField = async(fieldElement: ElementFinder) => {
+        await fieldElement.click();
+
+        const endPosition = 2;
+        const expectCaretAtEnd = () => expectCaretPosition(fieldElement, endPosition);
+
+        await sendKey(Key.END);
+        await expectCaretAtEnd();
+
+        await sendKey(Key.ARROW_UP);
+        await expectCaretAtEnd();
+
+        await sendKey(Key.ARROW_DOWN);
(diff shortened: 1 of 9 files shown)


src/timepicker/timepicker.ts
```diff
@@ -81,51 +86,99 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
...
+            (click)="changeMinute(-minuteStep)"
+            [disabled]="disabled"
+            [class.disabled]="disabled"
+          >
             <span class="chevron bottom"></span>
             <span class="sr-only"  i18n="@@ngb.timepicker.decrement-minutes">Decrement minutes</span>
           </button>
         </div>
-        <div *ngIf="seconds" class="ngb-tp-spacer">:</div>
+        <div *ngIf="seconds" class="ngb-tp-spacer">:</div>=
```
maxokorokov: You have added the `=` symbol in the template

screen shot 2018-06-28 at 17 59 51

src/timepicker/timepicker.ts
```diff
@@ -1,11 +1,16 @@
 import {Component, forwardRef, Input, OnChanges, SimpleChanges} from '@angular/core';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
 
-import {isNumber, padNumber, toInteger} from '../util/util';
+import {isNumber, padNumber, toInteger, isDefined} from '../util/util';
 import {NgbTime} from './ngb-time';
 import {NgbTimepickerConfig} from './timepicker-config';
 import {NgbTimeAdapter} from './ngb-time-adapter';
 
+enum Key {
```
maxokorokov: Could you please rebase, to use common Keys?

maxokorokov:
Hey, @ymeine I just merged #2912 as a simpler implementation for the keyboard support (unfortunately with the same drawbacks as here, but it needs the whole component rewrite I suppose).

What's still missing though is `tabindex` on the arrows. Can I close this or you'll be updating it ?

maxokorokov:
@ymeine, you're right about this doing two more things:

- fixing the cursor that is jumping when pressing up/down arrows
- adding tabindex -1

Though you haven't left any comments about this and I missed the first point while testing your PR.

On the other hand I'd very much prefer adding `(keydown.ArrowUp)="changeHour(hourStep); $event.preventDefault()"` to adding 4 additional methods. 

WDYT about just adding `preventDefault()`s ?

maxokorokov:
@ymeine also this implementation breaks selection with `Shift + ArrowUp/Down`

ymeine:
I will update this PR and simplify the code

ymeine:
I pushed another version. It contains two commits:

- first is a refactoring, since there was a lot of duplicated code!
- second is the actual implementation, with e2e tests

I strongly advise to take the refactoring into account (and am open to suggestions to improve it maybe — naming, etc.) since we clearly have three parts (hour, minute, second) which have the same UI and logic.

ymeine:
So, I finally extracted the refactoring commit to put it in a new PR (see [chore(timepicker): refactoring, common code for hour/minute/second by ymeine · Pull Request #3128 · ng-bootstrap/ng-bootstrap](https://github.com/ng-bootstrap/ng-bootstrap/pull/3128)) since it might spoil the discussion thread here and slow down the adoption of the fix.

Now only the straightforward fix is available for this current PR. Hoping to make it through.

e2e-app/src/app/timepicker/navigation/timepicker-navigation.e2e-spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+describe('Timepicker', () => {
+  let page: Page;
+
+  beforeAll(() => page = new Page());
+  beforeEach(async() => await openUrl('timepicker/navigation'));
+
+  function focusInputBefore() { return page.getInputBefore().click(); }
+
+  function goNext() { return sendKey(Key.TAB); }
```
maxokorokov: I would argue that these keyboard helpers make it MORE difficult to read the test.

I would prefer to read the test:

```
focus input before
press TAB
expect hourField to be focused
press SHIFT+TAB
...
```

as opposed to using `goNext()` which I have no idea what it does.
What's the point of hiding of what you're actually doing and wrapping one function into another here?

P.S. talking about only `sendKey(...)` here

e2e-app/src/app/timepicker/navigation/timepicker-navigation.e2e-spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+import {Page} from './timepicker-navigation.po';
+
+describe('Timepicker', () => {
+  let page: Page;
+
+  beforeAll(() => page = new Page());
+  beforeEach(async() => await openUrl('timepicker/navigation'));
+
+  function focusInputBefore() { return page.getInputBefore().click(); }
```
maxokorokov: shouldn't this go to the `Page`?

e2e-app/src/app/timepicker/navigation/timepicker-navigation.po.ts
```diff
@@ -0,0 +1,11 @@
+import {$} from 'protractor';
+
+export type Field = 'hour' | 'minute' | 'second';
+
+export class Page {
```
maxokorokov: Could you please rename to `TimepickerPage`, please?
- all other pages have specific names
- having only `Page` makes navigation by class name difficult

ymeine:
Changes done!

Issue #1977, opened 2017-11-20
Alternative Date adapters


maxokorokov:
Now that we have an `NgbDateAdapter`, it might be nice to bundle a couple of default ones with the library:

- [x] `NgbDateNativeAdapter` - as seen in the datepicker adapter demo to use with native JS date
- [ ] `NgbDateISOAdapter` - to conform with ISO 8601 Dates
- [ ] `NgbDateMomentAdapter` - for use with moment js

msosa:
@maxokorokov do you want me to attempt in making this adapters? 

For the first one we take the one in the demo as you said, for the 2nd would you just take the substrings of the ISO 8601 dates and parse it that way? And for moment we would add the lib as a `devDepedency` and convert that object?

maxokorokov:
@msosa, please do, if you have some time for this!

Start with a PR per adapter + tests.

I would put them under `datepicker/adapters` for now:
- native one should be the same as in demos
- iso 8601 should be something similar to what we have in date-formatter (`iso string` ←→ `NgbDateStruct`)
- moment is tricker, as I don't think we would want to put momentjs in peer dependencies of ng-bootstrap, so will have to think about it...

Later, from `2.0.0` it will be possible to make native adapter a default one

msosa:
sounds good, i'll start off with those first 2 then

jeroenheijmans:
For what it's worth, I've just crafted a Moment.js adapter with a few basic smoke tests. The full deal is [in a public gist](https://gist.github.com/jeroenheijmans/af473dad8ace3c5c4734588fca7d8b84), here's the adapter bits:

```typescript
// Might need to polyfill depending on needed browser support...
const isInt = Number.isInteger;

@Injectable()
export class NgbMomentjsAdapter extends NgbDateAdapter<moment.Moment> {

  fromModel(date: moment.Moment): NgbDateStruct {
    if (!date || !moment.isMoment(date)) {
      return null;
    }

    return {
      year: date.year(),
      month: date.month() + 1,
      day: date.date(),
    };
  }

  toModel(date: NgbDateStruct): moment.Moment {
    if (!date || !isInt(date.day) || !isInt(date.day) || !isInt(date.day)) {
      return null;
    }

    return moment(`${date.year}-${date.month}-${date.day}`, 'YYYY-MM-DD');
  }
}
```

I think there's a `util` based version of `isInt` in the library itself that might be useful instead of `Number.isInteger`.

[...]

PR #2066, opened 2017-12-29
feat(datepicker): use the angular locale API

### LINKED ISSUE #2065: Datepicker i18n could now be automatic
### Bug description:

The day and month names displayed by the datepicker are currently hardcoded to English in a service (NgbDatepickerI18nDefault), and it's possible to provide a different implementation of that service to customize these names.

Since we now rely on Angular 5.x, and since Angular 5.x now comes with an i18n API, we could replace the default implementation of this service and get the day and month names from the API.

Advantages:
 - the names would be consistent with the ones used by the standard i18n pipes, and any other library using the i18n API
 - the datepicker would honor the configured locale automatically (provided the locale data have been provided with registerLocaleData, as it's necessary for the i18n of the pipes)
 - there is no need to provide a custom NgbDatepickerI18n implementation anymore to change the names
 - A single implementation can be used for all locales
 - It's still possible, by keeping our NgbDatepickerI18n service, and only changing the default implementation, to customize as it was done before, and we thus introduce no backward incompatibility in the API

Drawbacks:
 - I don't see any

I'm ready to work on a PR if it's OK with you.

jnizet:
the datepicker now honors the Angular LOCALE_ID and uses the registered
locale data to internationalize the months and days names. Unless you
really want custom names, providing a custom NgbDatepickerI18n service
isn't necessary anymore to internationalize the datepicker.

fix #2065

BREAKING CHANGE: if your application provides a LOCALE_ID other than
the default en-US, registers the locale data for this locale, and
doesn't use a custom NgbDatepickerI18n, then the days and months
of the datepicker won't be displayed in English anymore, but in the
langguage of the provided locale.

### DIFF
--- a/src/datepicker/datepicker-i18n.ts
+++ b/src/datepicker/datepicker-i18n.ts
@@ -1,14 +1,10 @@
-import {Injectable} from '@angular/core';
-
-const WEEKDAYS_SHORT = ['Mo', 'Tu', 'We', 'Th', 'Fr', 'Sa', 'Su'];
-const MONTHS_SHORT = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec'];
-const MONTHS_FULL = [
-  'January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November',
-  'December'
-];
+import {Inject, Injectable, LOCALE_ID} from '@angular/core';
+import {FormStyle, getLocaleDayNames, getLocaleMonthNames, TranslationWidth} from '@angular/common';
 
 /**
  * Type of the service supplying month and weekday names to to NgbDatepicker component.
+ * The default implementation of this service honors the Angular locale, and uses the registered locale data,
+ * as explained in the Angular i18n guide.
  * See the i18n demo for how to extend this class and define a custom provider for i18n.
  */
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
@maxokorokov I would love to land this one, but I would live to have your opinion first. Could you PTAL?

src/datepicker/datepicker-i18n.ts
```diff
@@ -34,9 +30,23 @@ export abstract class NgbDatepickerI18n {
 
 @Injectable()
 export class NgbDatepickerI18nDefault extends NgbDatepickerI18n {
-  getWeekdayShortName(weekday: number): string { return WEEKDAYS_SHORT[weekday - 1]; }
+  private weekdaysShort: Array<string>;
```
maxokorokov: If you'll be fixing `FormStyle.Format`, could you also prefix these with `_` as we usually do with private members

jnizet:
@maxokorokov sorry for the late update. I somehow missed your review, then was in holiday. I applied your suggested fixes and rebased on master.

nicoabie:
I would be nice to add a default implementation of NgbDateAdapter<Date> as well, it is very common case to, right? @pkozlowski-opensource

maxokorokov:
@nicoabie #1977

kukac7:
demo link how to use locale api with datepicker?
even though the language is registered, the datepicker is not translated.

PR #2078, opened 2018-01-12
demo: allow editing examples in StackBlitz

jnizet:
This adds an icon link to edit example in [StackBlitz](https://stackblitz.com/) (in addition to Plnkr). 

I find StackBlitz easier to use and more polished than Plnkr. It's becoming the tool of choice to share Angular examples, and angular material is already using it for the examples in its documentation.

I'd like your opinion on the following questions
 - is it a good idea?
 - should we keep both Plnkr and StackBlitz?

If you agree this is a good idea, there's till some work to be done though:

 - [x] check if it's allowed to use the StackBlitz logo (see question asked to StackBlitz in [this issue](https://github.com/stackblitz/core/issues/237)
 - [x] load the bootstrap CSS properly, using an external resource (see question asked to StackBlitz in [this issue](https://github.com/stackblitz/core/issues/237). I added a link directly in the body of the index.html file, but it causes issues in the datepicker popup examples. 
 - [x] see why the images in the carousel demo don't load in StackBlitz example (see question asked to StackBlitz in [this issue](https://github.com/stackblitz/core/issues/237)


pkozlowski-opensource:
@jnizet supporting StackBlitz makes sense IMO. If you can work toward completion of this PR I would love to land it. Thnx for looking into it!

jnizet:
Since I still haven't had an answer on how to link external resources, but had an answer showing that a CSS import is actually working fine, I used this technique. It requires bootstrap and its unwanted peer dependencies (jquery and popper.js) to be listed as dependencies, too. But I added a note warning about that in the generated StackBlitz demos.

jnizet:
@pkozlowski-opensource this can be reviewed now, if you have some time.

jnizet:
StackBlitz has improved its support for externally linked CSS files from the `head` of index.html now, and the StackBlitz examples don't need to depend on bootstrap, jquery and popper anymore. StackBlitz examples are now in a good shape, IMO.

I also rebased on master.

demo/src/app/components/carousel/demos/config/carousel-config.ts
```diff
@@ -7,6 +7,8 @@ import {NgbCarouselConfig} from '@ng-bootstrap/ng-bootstrap';
   providers: [NgbCarouselConfig] // add NgbCarouselConfig to the component providers
 })
 export class NgbdCarouselConfig {
+  images = [1, 2, 3].map(() => `https://picsum.photos/900/500?image=${Math.floor(Math.random() * 1000)}`);
```
maxokorokov: The problem with random is that sometimes the images are not there. I guess it's more of an exception, but this one broke for me: https://picsum.photos/900/500?image=207

screen shot 2018-03-09 at 09 42 24

I still like random images though...

demo/src/style/app.scss
```diff
@@ -178,12 +178,12 @@ div.component-demo {
...
     margin-bottom: 1rem;
 
-    .plunker {
+    .plunker, .stackblitz {
       opacity: 0.3;
       transition: opacity .5s;
     }
     &:hover {
-      .plunker {
+      .plunker, .stackblitz {
```
maxokorokov: I guess you should add `text-decoration: none` here, to remove the a:hover blue line glitch between two icons:

screen shot 2018-03-09 at 09 38 35

maxokorokov:
For the icons I can suggest a couple of options:

Your initial version (I would make the icon smaller):
screen shot 2018-03-09 at 09 30 12

Alternatives:
screen shot 2018-03-09 at 09 29 50

jnizet:
Regarding the icons, I like your first alternative, especially if we keep both plunkr and stackblitz, because the two icons would have a similar style. But I couldn't find this image anywhere, so feel free to amend my PR.

jnizet:
@maxokorokov thanks for the review. I made changes to the code to avoid the visual glitch, and to make sure only valid URLs are used for random images.

PR #2082, opened 2018-01-12
feat(dropdown): allow using any event to toggle dropdowns

### LINKED ISSUE #1115: Dropdown: Use other event than the click event
### Bug description:
Component: dropdown

It should be possible to specify the event that is used to toggle the menu. Currently it's hard coded to use the [click event](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/dropdown/dropdown.ts#L113). I think it should be possible to use the same directives for context menus which would require to listen to the the `contextmenu` event instead of the `click` event.

Unfortunately I'm not sure how to do this best, because it's not decided within the component but the host. If we can discuss what would be the best way to do it, I could create a PR.

Angular: 2.1.2
ng-bootstrap: 1.0.0-alpha.14
Bootstrap: 4.0.0-alpha.4

### LINKED ISSUE #1926: NgbDropdownToggle is not optional
Angular: 4.4.6
ng-bootstrap: 1.0.0-beta.5
Bootstrap: 4.0.0-beta.2

Documentation states that NgbDropdownToggle directive is optional, yet instantiating an NgbDropdown without an NgbDropdownToggle within it leads to poorly positioned drop-down.

Quick glance at the source code suggests positioning of menu element is skipped when NgbDropdown._toggle is undefined. See [dropdown.ts](https://github.com/ng-bootstrap/ng-bootstrap/blob/1.0.0-beta.5/src/dropdown/dropdown.ts), line 195.

pkozlowski-opensource:
Closes #1115
Fixes #1926

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -50,7 +50,28 @@ export class NgbDropdownMenu {
 }
 
 /**
- * Allows the dropdown to be toggled via click. This directive is optional.
+ * Marks an element to which dropdown menu will be anchored. This is a simple version
+ * of the NgbDropdownToggle directive. It plays the same role as NgbDropdownToggle but
+ * doesn't listen to click events to toggle dropdown menu thus enabling support for
+ * events other than click.
+ */
+@Directive({
+  selector: '[ngbDropdownAnchor]',
+  host: {'class': 'dropdown-toggle', 'aria-haspopup': 'true', '[attr.aria-expanded]': 'dropdown.isOpen()'}
+})
+export class NgbDropdownAnchor {
+  anchorEl;
+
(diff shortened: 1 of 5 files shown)


src/dropdown/dropdown.ts
```diff
@@ -49,6 +49,26 @@ export class NgbDropdownMenu {
   }
 }
 
+/**
+ * Marks an element to which dropdown menu will be anchored. This is a simple version
+ * of the NgbDropdownToggle directive. It plays the same role as NgbDropdownToggle but
+ * don't listen to click events to toggle dropdown menu thus enabling support for
```
maxokorokov: don't → doesn't

pkozlowski-opensource:
@maxokorokov thnx for the review, all the valid points! I believe I've fixed all the issues you've spotted - could you please have another look?

Issue #2111, opened 2018-01-23
Missing license file in npm release


readme42:
When installing this library via npm, the [license file](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/LICENSE) is missing. This makes automated license generation complicated. Could you add the file to upcoming npm releases?

pkozlowski-opensource:
@readme42 are you saying that we should distribute `LICENSE` file as part of our npm package? If so,  #2114 should take care of this

readme42:
@pkozlowski-opensource thats exactly what i mean. Thanks a lot!

PR #2113, opened 2018-01-24
docs(LICENSE): update copyright year to 2018

bastienmoulia:
Maybe related to #2111


pkozlowski-opensource:
Merged, thnx!

pkozlowski-opensource:
@bastienmoulia my understanding is that #2111 is about sth like #2114

PR #2142, opened 2018-02-05
fix(Modal): backdrop click events

### LINKED ISSUE #1950: Modals incorrectly dismissed by mouse release
### Bug description:
When you have a modal dialog that includes text which is expected to be copied with the mouse, users (inc. me) often somewhat imprecisely select the text and end up with the mouse outside of the modal dialog. They then release the mouse button used to select the text.

With ng-bootstrap modals this results in the unexpected behaviour of the whole modal gets dismissed.

I think the ideal solution would probably be to not listen for (click) in NgbModalWindow, and instead listen for (mousedown) events.
https://github.com/ng-bootstrap/ng-bootstrap/blob/0c03671c52500f13387deeaf31f815372d364c26/src/modal/modal-window.ts#L23

As a work-around I can disable the background click detection (option 'backdrop'), but it would be nice to still support that.

You can see the issue with [modal](https://ng-bootstrap.github.io/#/components/modal/examples)

Launch the first modal, then select all of the text ('One fine body…') and end with the mouse just outside the modal before you release the mouse button.

Angular: 4.4.6

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta.2

Browsers:
 Microsoft Edge 40.15063.674.0 (Win)
 Microsoft IE 11.674 (Win)
 Chrome Version 61.0.3163.100 (Win)
 Chrome  61.0.3163.100 (Mac)

 Firefox 56.0 Mac does not exhibit the problem
 Safari 11.0.1 Mac does not exhibit the problem

redshark1802:
Closes #1950

The ``click`` event was replaced with ``touchstart`` and ``mousedown``. This way the modal closes only when the backdrop was clicked directly and did not start anywhere else, as described in the bug report.

### DIFF
--- a/src/modal/modal-window.spec.ts
+++ b/src/modal/modal-window.spec.ts
@@ -107,6 +107,28 @@ describe('ngb-modal-dialog', () => {
 
       fixture.debugElement.triggerEventHandler('keyup.esc', {});
     });
+
+    it('should dismiss the modal on backdrop click on mousedown', (done) => {
+      fixture.detectChanges();
+
+      fixture.componentInstance.dismissEvent.subscribe(($event) => {
+        expect($event).toBe(ModalDismissReasons.BACKDROP_CLICK);
+        done();
+      });
+
+      fixture.nativeElement.dispatchEvent(new Event('mousedown'));
+    });
+
+    it('should dismiss the modal on backdrop click on touchstart', (done) => {
+      fixture.detectChanges();
+
+      fixture.componentInstance.dismissEvent.subscribe(($event) => {
+        expect($event).toBe(ModalDismissReasons.BACKDROP_CLICK);
+        done();
+      });
+
+      fixture.nativeElement.dispatchEvent(new Event('touchstart'));
+    });
   });
 
 });
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -20,7 +20,8 @@ import {ModalDismissReasons} from './modal-dismiss-reasons';
     'tabindex': '-1',
     'style': 'display: block;',
     '(keyup.esc)': 'escKey($event)',
-    '(click)': 'backdropClick($event)'
+    '(mousedown)': 'backdropClick($event)',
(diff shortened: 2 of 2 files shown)


redshark1802:
I just ran ``npm test`` instead of ``gulp test`` and didn't realise that the actual tests were not executed. Working on the tests now.

The browser emits ``mousedown``, ``click`` and ``mouseup`` events on a normal click, however the tests only emit ``click`` and the modal is therefore not closed. I'm going to do some research on this.

In the meantime, do you have an idea how this problem could be solved?

pkozlowski-opensource:
@redshark1802 are you sure that this change actually fixes the issue? I'm puzzled since clicks on "outside" of modal are actually captured here: https://github.com/ng-bootstrap/ng-bootstrap/blob/63559f02c2262581b9578b5f2cbd1d18709f9712/src/modal/modal-window.ts#L25

pkozlowski-opensource:
Closing for now as I haven't heard back from @redshark1802 - happy to re-open / continue discussion when I get more info on https://github.com/ng-bootstrap/ng-bootstrap/pull/2142#issuecomment-371773517

EpaXapate:
@pkozlowski-opensource I am also experiencing this issue, would it be possible to reopen this issue and perhaps provide a fix or some feedback on it?

mklein530:
+1 Bumping this issue. A fix for this is needed for our product.

vbakavets:
+1 We also need a fix for this issue

lucasklaassen:
@EpaXapate @mklein530 @vbakavets I've posted a PR which re-opens this one. I tested this fix myself and it works great.

enemysik:
@lucasklaassen It still happends in IE and Chrome browsers. It work fine only in Mozilla. You can see it at https://stackblitz.com/edit/angular-1qrkju?file=app/modal-config.ts.

DerZyklop:
finally could update my project and i need to say this: 
this fix is awesome!! :D 
🚀 🚀 🚀

PR #2180, opened 2018-02-20
fix(modal): support server rendering with lazy loading

### LINKED ISSUE #858: Server side rendering with universal
We should explore server-side rendering options with universal. IMO we should be in a pretty good shape to support it as we are not doing anything too crazy. There is direct DOM access in few places but it shouldn't be too hard to remove it.

Anyway, after we release a version with full AoT support (already in master) we should check what it means to support universal. Probably a good starting point would be to use the https://github.com/angular/universal-starter

---

Widgets confirmed to work with Angular 6+:
- [x] Accordion
- [x] Alert
- [x] Buttons
- [x] Carousel (fix in #2494)
- [x] Collapse
- [x] Datepicker*
- [x] Dropdown
- [x] Modal*
- [x] Pagination
- [x] Popover*
- [x] Progressbar
- [x] Rating
- [x] Tabs
- [x] Timepicker
- [x] Tooltip*
- [x] Typeahead*

### LINKED ISSUE #1968: modal service breaks universal server side rendering
**Bug description:**

The modal service seems to be causing a bug that prevents server side rendering from working properly. I realize this is still a feature in development, but I wanted to make a note of it, and it didn't seem like a bug report had been posted yet.

You can use the universal-starter repo to clone a test application:

[https://github.com/angular/universal-starter](url)

Compile and run on localhost using two npm scripts:

npm run build:ssr
npm run serve:ssr

When NgbModule.forRoot() is added to the root NgModule, everything is working fine. However, when NgbModal is injected into a component, the below error message occurs after compilation and attempting to visit localhost.

**Error message:**

`
Error: No component factory found for NgbModalBackdrop. Did you add it to @NgModule.entryComponents?
    at noComponentFactoryError (C:\Users\Alec\nodeprojects\testapp\dist\server.js:4059:34)
    at CodegenComponentFactoryResolver.resolveComponentFactory (C:\Users\Alec\nodeprojects\testapp\dist\server.js:4123:19)
    at new NgbModalStack (C:\Users\Alec\nodeprojects\testapp\dist\server.js:62589:59)
    at _createClass (C:\testapp\dist\server.js:10896:20)
    at _createProviderInstance$1 (C:\Users\Alec\nodeprojects\testapp\dist\server.js:10866:26)
    at resolveNgModuleDep (C:\Users\Alec\nodeprojects\testapp\dist\server.js:10851:17)
    at _createClass (C:\Users\Alec\nodeprojects\testapp\dist\server.js:10896:107)
    at _createProviderInstance$1 (C:\Users\Alec\nodeprojects\testapp\dist\server.js:10866:26)
    at resolveNgModuleDep (C:\Users\Alec\nodeprojects\testapp\dist\server.js:10851:17)
    at NgModuleRef_.get ``(C:\Users\Alec\nodeprojects\testapp\dist\server.js:12076:16)
`

Angular: 5.0.0

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta.2

Scoup:
Closes #1968
Fixes #858

This PR fix the problem ```No component factory found for NgbModalBackdrop. Did you add it to @NgModule.entryComponents? at noComponentFactoryError``` when using ng-bootstrap on server side.

We had 2 main issues:
- First, on lazy loading for some reason (I don't know how Angular Compiler really works) the service NgbModalStack is injected but the components ```NgbModalWindow``` and ```NgbModalBackdrop``` are not on the [factories list](https://github.com/angular/angular/blob/5.2.5/packages/core/src/linker/component_factory_resolver.ts#L57) yet. For that reason ```ComponentFactory<NgbModalBackdrop>``` and ```ComponentFactory<NgbModalWindow>``` will be initialized only when ```open``` is called.
- A lot of ```document``` access directly. I changed the code so we can inject the document to run on server side.

### DIFF
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -1,7 +1,9 @@
+import {DOCUMENT} from '@angular/common';
 import {
   ApplicationRef,
   Injectable,
   Injector,
+  Inject,
   ReflectiveInjector,
   ComponentFactory,
   ComponentFactoryResolver,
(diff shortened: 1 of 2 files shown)


src/modal/modal.ts
```diff
@@ -51,16 +51,13 @@ export interface NgbModalOptions {
...
+  constructor(private _modalStack: NgbModalStack) {}
 
   /**
    * Opens a new modal window with the specified content and using supplied options. Content can be provided
    * as a TemplateRef or a component type. If you pass a component type as content than instances of those
    * components can be injected with an instance of the NgbActiveModal class. You can use methods on the
    * NgbActiveModal class to close / dismiss modals from "inside" of a component.
    */
-  open(content: any, options: NgbModalOptions = {}): NgbModalRef {
-    return this._modalStack.open(this._moduleCFR, this._injector, content, options);
```
pkozlowski-opensource: This is not good, as it would change `ComponentFactoryResolver` instance and break lazy-loaded modules (I could not open a component from a lazy-loaded module).

src/modal/modal-stack.ts
```diff
@@ -20,75 +22,110 @@ import {NgbActiveModal, NgbModalRef} from './modal-ref';
...
 
     this._applyWindowOptions(windowCmptRef.instance, options);
-
     return ngbModalRef;
   }
 
+  // Cant build factories on constructor when using lazy loading
+  private _buildFactories() {
+    this._buildBackdropFactory();
+    this._buildWindowFactory();
```
pkozlowski-opensource: I don't think that caching factories here, in the code, adds any performance benefits. Remove those methods and replace with a direct call to `this._componentFactoryResolver.resolveComponentFactory(NgbModalBackdrop)`

src/modal/modal-stack.ts
```diff
@@ -20,75 +22,110 @@ import {NgbActiveModal, NgbModalRef} from './modal-ref';
...
-    const containerSelector = options.container || 'body';
-    const containerEl = document.querySelector(containerSelector);
+  open(content: any, options): NgbModalRef {
+    this._buildFactories();
+    const containerEl =
+        isDefined(options.container) ? this._document.querySelector(options.container) : this._document.body;
 
     if (!containerEl) {
-      throw new Error(`The specified modal container "${containerSelector}" was not found in the DOM.`);
+      throw new Error(`The specified modal container "${options.container}" was not found in the DOM.`);
```
pkozlowski-opensource: This will have a broken error message if a user doesn't specify the `container` option

src/modal/modal-stack.ts
```diff
@@ -20,75 +22,110 @@ import {NgbActiveModal, NgbModalRef} from './modal-ref';
...
       if (isDefined(options[optionName])) {
         windowInstance[optionName] = options[optionName];
       }
     });
   }
 
-  private _getContentRef(
-      moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any,
-      context: NgbActiveModal): ContentRef {
+  private _getContentRef(contentInjector: Injector, content: any, context: NgbActiveModal): ContentRef {
```
pkozlowski-opensource: The signature should stay as-is, we need to use a proper `ComponentFactoryResolver`

pkozlowski-opensource:
@Scoup thnx for the PR. I did the first pass on it and left the comments in the code, here is the gist:
* injecting `document` is a good change, thnx;
* we need to use proper `ComponentFactoryResolver`

Please resolve those issues and I will have another look.

src/modal/modal.ts
Scoup: Thanks for the tip!
I was looking for how ComponentFactoryResolver and now makes sense why the ComponentFactoryResolver was here!

src/modal/modal-stack.ts
Scoup: If a user doesn't specify the `container` option then will use `this._document.body` and not a query selector.
`this._document.body` will be always defined, right?

pkozlowski-opensource: > this._document.body will be always defined, right?

In theory not always, as it can potentially be `null`: https://developer.mozilla.org/en-US/docs/Web/API/Document/body

In practice it should be always defined, but it is better to be safe than sorry...

Scoup:
@pkozlowski-opensource I did the changes. Was very nice to learn how ComponentFactoryResolver works behind the scene. Thanks helping me.

pkozlowski-opensource:
Merged, thnx @Scoup 

One note - I've removed Fixes #858 from the commit as I believe that there is more work needed on other components dynamically inserting views (tooltip, popover, typeahead, datepicker) to fully support universal for all components.

PR #2193, opened 2018-02-26
chore(build): add 'rxjs' as a peer dependency

### LINKED ISSUE #2148: Rxjs is not specified as a dependency
### Bug description:

The library rxjs is used directly in the code but not specified as a dependency of any kind in package.json. 

Normally this is not an issue, as angular would not run without it anyway, but some analyzing tools (like bundle size tools) throw an error because of that. 

https://bundlephobia.com/result?p=@ng-bootstrap/ng-bootstrap

ng-bootstrap: current

maxokorokov:
Fixes #2148

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
   "dependencies": {
     "@angular/common": "6.0.0-rc.3",
     "@angular/core": "6.0.0-rc.3",
-    "@angular/forms": "6.0.0-rc.3"
+    "@angular/forms": "6.0.0-rc.3",
+    "rxjs": "6.0.0-rc.1"
   },
   "devDependencies": {
     "@angular/compiler": "6.0.0-rc.3",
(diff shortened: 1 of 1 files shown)


maxokorokov:
@pkozlowski-opensource, just wanted to double-check. Are you sure it's a breaking change? 
Angular 5.0.2 already has a peer dependency on rxjs ^5.5.0, I wouldn't imagine it breaking anybody.

maxokorokov:
I should just probably change it to ^5.5.0 also instead of ^5.5.2 in this PR

maxokorokov:
@pkozlowski-opensourceI guess I'll rebase when it will become `6.0.0`, because current one is `6.0.0-rc.x` and we have a script that adds `^` symbols for peer dependencies I think

maxokorokov:
Changed my mind, will forget later. Please merge

PR #2197, opened 2018-02-28
feat(typeahead): accessibility, options summary status message

### LINKED ISSUE #1945: Typeahead: suggested options not conveyed to screen readers
### Bug description:
Accessibility issue
As raised during an accessibility audit, when using an autocomplete feature, the list of suggested options should be conveyed to screen readers users.
The same issue is observed with the ng-bootstrap typeahead widget.
After typing a letter and options appear, this information is currently not conveyed to screen reader users, who can not know that options are available.
Same issue is occurring when no option is found.
The recommendation is to add a hidden sentence for screen readers like "2 options are available, please use up and down arrow keys to navigate"/"No results available" that will be read out after typing a letter.
An aria live region could be used to achieve this (role="status" and aria-live="polite").

Tested with NVDA/Firefox and JAWS/IE

https://ng-bootstrap.github.io/#/components/typeahead/examples

Angular: angular 4.4.6

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta.2

ymeine:
Closes #1945

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -12,6 +12,8 @@ import {NgbTypeaheadModule} from './typeahead.module';
 import {NgbTypeaheadConfig} from './typeahead-config';
 import {debounceTime, filter, map, merge} from 'rxjs/operators';
 
+import {ARIA_LIVE_DELAY} from '../util/accessibility/live';
+
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
 
@@ -78,7 +80,8 @@ describe('ngb-typeahead', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       declarations: [TestComponent, TestOnPushComponent, TestAsyncComponent],
-      imports: [NgbTypeaheadModule.forRoot(), FormsModule, ReactiveFormsModule]
+      imports: [NgbTypeaheadModule.forRoot(), FormsModule, ReactiveFormsModule],
+      providers: [{provide: ARIA_LIVE_DELAY, useValue: null}]
     });
   });
 
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -21,6 +21,7 @@ import {positionElements, PlacementArray} from '../util/positioning';
 import {NgbTypeaheadWindow, ResultTemplateContext} from './typeahead-window';
 import {PopupService} from '../util/popup';
 import {toString, isDefined} from '../util/util';
+import {Live} from '../util/accessibility/live';
 import {NgbTypeaheadConfig} from './typeahead-config';
 import {map, switchMap, tap} from 'rxjs/operators';
 
@@ -153,7 +154,7 @@ export class NgbTypeahead implements ControlValueAccessor,
   constructor(
       private _elementRef: ElementRef, private _viewContainerRef: ViewContainerRef, private _renderer: Renderer2,
       private _injector: Injector, componentFactoryResolver: ComponentFactoryResolver, config: NgbTypeaheadConfig,
-      ngZone: NgZone) {
+      ngZone: NgZone, private _live: Live) {
     this.container = config.container;
     this.editable = config.editable;
     this.focusFirst = config.focusFirst;
@@ -355,6 +356,8 @@ export class NgbTypeahead implements ControlValueAccessor,
 
         this._showHint();
       }
+      const count = results.length;
+      this._live.say(count === 0 ? 'No results available' : `${count} result${count === 1 ? '' : 's'} available`);
     });
   }
 
--- a/src/util/accessibility/live.ts
+++ b/src/util/accessibility/live.ts
@@ -1,0 +1,51 @@
+import {Injectable, Inject, InjectionToken, OnDestroy} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+
+
+
+// usefulness (and default value) of delay documented in Material's CDK
+// https://github.com/angular/material2/blob/6405da9b8e8532a7e5c854c920ee1815c275d734/src/cdk/a11y/live-announcer/live-announcer.ts#L50
+export type ARIA_LIVE_DELAY_TYPE = number | null;
+export const ARIA_LIVE_DELAY = new InjectionToken<ARIA_LIVE_DELAY_TYPE>('live announcer delay');
+export const DEFAULT_ARIA_LIVE_DELAY: ARIA_LIVE_DELAY_TYPE = 100;
+
+
+
+function createLiveElement(document): HTMLElement {
+  const element = document.createElement('div');
+
+  element.setAttribute('aria-live', 'polite');
+  element.setAttribute('aria-atomic', 'true');
+
+  element.classList.add('sr-only');
+
+  document.body.appendChild(element);
+
+  return element;
+}
+
+
+
+@Injectable()
+export class Live implements OnDestroy {
+  private _element: HTMLElement;
+
+  constructor(@Inject(DOCUMENT) document: any, @Inject(ARIA_LIVE_DELAY) private _delay: ARIA_LIVE_DELAY_TYPE) {
+    this._element = createLiveElement(document);
+  }
+
+  ngOnDestroy() { this._element.parentElement.removeChild(this._element); }
+
+  say(message: string): void {
+    const element = this._element;
+    const delay = this._delay;
+
+    element.textContent = '';
+    const setText = () => element.textContent = message;
+    if (delay === null) {
+      setText();
+    } else {
+      setTimeout(setText, delay);
+    }
+  }
+}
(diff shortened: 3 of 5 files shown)


src/util/accessibility/live.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@Injectable()
+export class Live implements OnDestroy {
+  private _element: HTMLElement;
+
+  constructor(@Inject(DOCUMENT) private _document: any) { this._element = this._createElement(); }
+
+  ngOnDestroy() { this._element.remove(); }
+
+  async say(message: string): Promise<void> {
```
pkozlowski-opensource: Could you please not use async-await for now? It transpiles to rather verbose code.

src/util/accessibility/live.ts
```diff
@@ -0,0 +1,36 @@
...
+    const document = this._document;
+
+    const element = document.createElement('div');
+
+    element.setAttribute('aria-live', 'polite');
+    element.setAttribute('aria-atomic', 'true');
+
+    element.classList.add('sr-only');
+
+    document.body.appendChild(element);
```
pkozlowski-opensource: Who is going to remove this element? We should have `onDestroy`, maybe?

pkozlowski-opensource: Oh, actually it is here, never mind

src/typeahead/typeahead.ts
```diff
@@ -357,9 +358,15 @@ export class NgbTypeahead implements ControlValueAccessor,
         // the change detection turn wouldn't be invoked automatically.
         this._windowRef.changeDetectorRef.detectChanges();
       }
+
+      this._live.say(this._srOptionsSummaryTemplate({count: results.length}));
```
pkozlowski-opensource: Why do you pass an object here instead of just a number? It results in the unnecessary object creation and you read only `count` from it anyway.

src/util/accessibility/live.ts
pkozlowski-opensource: More important - why do we return a `Promise` here? We don't care about any return value, do we? What would be the use-case for knowing when announcing starts?

src/util/accessibility/live.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  constructor(@Inject(DOCUMENT) private _document: any) { this._element = this._createElement(); }
+
+  ngOnDestroy() { this._element.remove(); }
+
+  async say(message: string): Promise<void> {
+    const element = this._element;
+
+    element.textContent = '';
+    await new Promise(resolve => setTimeout(resolve, 100));
```
pkozlowski-opensource: Why `100`? How was this number chosen? Why does it need to be async in the first place? As a very minimum we need documentation that explain this.

src/typeahead/typeahead.spec.ts
```diff
@@ -513,6 +513,7 @@ describe('ngb-typeahead', () => {
          changeInput(compiled, 'o');
          fixture.detectChanges();
          tick(250);
+         tick(100);  // for the aria-live announcer
```
maxokorokov: I think we should use a different announcer for tests or inject delay value 0 so it would become synchronous

src/typeahead/typeahead.ts
maxokorokov: or no need in method at all

src/util/accessibility/live.ts
```diff
@@ -0,0 +1,36 @@
+import {Injectable, Inject, OnDestroy} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+
+
+
+@Injectable()
+export class Live implements OnDestroy {
+  private _element: HTMLElement;
+
+  constructor(@Inject(DOCUMENT) private _document: any) { this._element = this._createElement(); }
```
maxokorokov: nitpick: no need to store document `privately`

src/typeahead/typeahead.spec.ts
ymeine: __ANYTHING__ to get rid of these weird fake things like `tick` would please me. I'll explore the injection of a different delay, seems the simplest and most flexible, that could even be useful for users.

src/util/accessibility/live.ts
ymeine: Why would I make it public? It's an internal thing, it's not part of the API. Otherwise this removes the whole purpose of having visibility specifiers in the typing system.

src/util/accessibility/live.ts
ymeine: This is all explained there: [material2/live-announcer.ts at 6405da9b8e8532a7e5c854c920ee1815c275d734 · angular/material2](https://github.com/angular/material2/blob/6405da9b8e8532a7e5c854c920ee1815c275d734/src/cdk/a11y/live-announcer/live-announcer.ts#L50).

It seems natural to wait a bit in case, knowing the unpredictability of screen readers. However, choosing a proper amount of delay is purely arbitrary.

src/util/accessibility/live.ts
ymeine: I think this is pity to deprive ourselves of one of the best __standard__ language features that was brought these last years: now asynchronous code remains explicit and controlled, but is aligned with synchronous code. It's what made me adopt async code widely, because otherwise dealing with Promises chaining is quite verbose.

This is a general comment, and of course here using a Promise is easy, since I even had to create one due to the lack of proper API for creating a delay.

However: 

- first, the code size shouldn't be an issue, especially once compressed
- then, we should configure the compilation so that it doesn't generate the helpers in each file, but instead relies on the TypeScript library to have this piece of code only once.
- last, we don't really care about the return value here. It's a `Promise` because async functions return Promises.

src/typeahead/typeahead.ts
ymeine: _I don't know why this appears as an outdated comment, but I'll reply anyways_

No need for that indeed. I started with a public API, so I made it generic:

- users would give a callback, which is now the private method we can see
- I would give them an object since this is more future-proof in case we want to give more information

I kept this design while making things private.

However I would not be concerned __AT ALL__ with performances. These are micro-optimizations, and I think you are well placed to know that actual performance is more complex than that. In this context, the feature call rate is __way too low__ to worry about extra nanoseconds spent creating an object, calling a method and reading from this object. My concerns would be more about maintainability. But here I can remove both the method and the object as long as it remains private, it's a matter of one line of code so it's perfectly fine.

ymeine:
Changes done.

src/util/accessibility/live.ts
```diff
@@ -0,0 +1,49 @@
...
+    element.textContent = '';
+    const setText = () => element.textContent = message;
+    if (delay === null) {
+      setText();
+    } else {
+      setTimeout(setText, delay);
+    }
+  }
+
+  private _createElement(document): HTMLElement {
```
maxokorokov: nitpick: this doesn't have to be private, can be a static outside function

src/util/accessibility/live.ts
```diff
@@ -0,0 +1,49 @@
+import {Injectable, Inject, InjectionToken, OnDestroy} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+
+
+
+// usefulness (and default value) of delay documented in Material's CDK
+// https://github.com/angular/material2/blob/6405da9b8e8532a7e5c854c920ee1815c275d734/src/cdk/a11y/live-announcer/live-announcer.ts#L50
+export type DELAY_TYPE = number | null;
```
maxokorokov: Could we name them more precisely, please? `{provide: DELAY, ...}` - not clear which delay.

Maybe something like `ARIA_LIVE_DELAY` ?

src/util/accessibility/live.ts
```diff
@@ -0,0 +1,49 @@
...
+
+@Injectable()
+export class Live implements OnDestroy {
+  private _element: HTMLElement;
+
+  constructor(@Inject(DOCUMENT) document: any, @Inject(DELAY) private _delay: DELAY_TYPE) {
+    this._element = this._createElement(document);
+  }
+
+  ngOnDestroy() { this._element.remove(); }
```
maxokorokov: Apparently `.remove()` is failing in IE 10-11 tests

src/typeahead/typeahead.ts
```diff
@@ -354,6 +355,8 @@ export class NgbTypeahead implements ControlValueAccessor,
 
         this._showHint();
       }
+      const count = results.length;
+      this._live.say(count === 0 ? 'No result available' : `${count} options available`);
```
maxokorokov: 0 → `No results available`
1 → `1 result available`
n → `n results available`

maxokorokov:
@ymeine, could you please fix a failing test and I'll merge:

```
Firefox 59.0.0 (Linux 0.0.0) LiveAnnouncer live announcer should remove the used element from the DOM on destroy FAILED
	TypeError: this._element.parentElement is null in node_modules/systemjs/dist/system.src.js line 1556 > eval (line 32)
	Live.prototype.ngOnDestroy@src/util/accessibility/live.ts:37:18 <- temp/util/accessibility/live.js:32:48
	@src/util/accessibility/live.spec.ts:33:20 <- temp/util/accessibility/live.spec.js:34:40
```

PR #2200, opened 2018-03-01
fix(accordion): refact active card class

kukac7:


pkozlowski-opensource:
@kukac7 what is the motivation for this PR?

kukac7:
@pkozlowski-opensource I want the entire `card` to be active, not a `card-header` because it can handle the active, open state. (eg. css style)

pkozlowski-opensource:
Hmm, seems like Bootstrap's accordion is _not_ using active class at all, so we should probably remove it altogether: https://getbootstrap.com/docs/4.0/components/collapse/#accordion-example

kukac7:
@pkozlowski-opensource on this basis, you can also remove the active class from the `card-header`. 🤷‍♂️ 

after that, how do I know which `card` is active so I can write a custom design?

pkozlowski-opensource:
@kukac7 I've just checked and Bootstrap 4 is not using `active` class any more in their markup for accordions: https://getbootstrap.com/docs/4.0/components/collapse/#accordion-example

Having this class set in our code is a mistake and we are going to remove it in a next release where we can do breaking changes (2.0.0). I've opened # to track this. Custom styling via CSS is the whole different topic so please open a separate issue with a feature request and your use-case description, so we can track this.

Given all the above I can't merge this PR as the `active` class shouldn't be here in the first place...

kukac7:
@pkozlowski-opensource never mind.
I wrote from myself and I will not use yours. 😉

PR #2213, opened 2018-03-10
test(alert): add more tests

Foxandxss:
The code coverage of alert wasn't 100%, now it should be.


Foxandxss:
Oh clang, forgot about you, I was happier, then I am misserable again.

PR #2214, opened 2018-03-12
fixed modal size

serekgrecki:

### DIFF
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -37,7 +37,7 @@ export interface NgbModalOptions {
   /**
    * Size of a new modal window.
    */
-  size?: 'sm' | 'lg';
+  size?: 'sm' | 'md' | 'lg';
 
   /**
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
Yeh, agree with @maxokorokov - this change would imply existence of the `modal-md` CSS class and I don't think one exists in Bootstrap's CSS, you can check here: https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.css

I'm going to close this PR based on the above and the fact that this PR misses tests. @serekgrecki if you want to introduce `md` as a default value it would also require:
* https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal-window.ts
* tests

Then again, it would make things inconsistent with Bootstrap where default window size doesn't require any CSS class.

Thnx for the PR but I'm going to close it for now based on the above. Happy to re-consider if / when you've got strong arguments to support this change.

PR #2217, opened 2018-03-14
feat(popover): only show a popover if the content is not falsy

### LINKED ISSUE #2188: Conditional popovers
In #737, ngbTooltip was amended to only show a tooltip if the content is not falsy. 

I'd like ngbPopover to behave the same way (currently, it shows an empty popover in this case).

### Usecase ###

I'd like to show a different (or no) popover depending on the state of the element being hovered over. I had this working for a tooltip:

      <div [ngbTooltip]="helpKey || invalid ? tooltip : null" placement="bottom"
          triggers="hover focusin:focusout">

but the same approach with a popover

      <div [ngbPopover]="helpKey || invalid ? tooltip : null" placement="bottom"
          triggers="hover focusin:focusout">

shows an empty popover if passed null. 

Angular: 5

ng-bootstrap: 1.0.0

Bootstrap: 4

medinarkhov:
ngbTooltip was amended to only show a tooltip if the content is not falsy. It would be nice if ngbPopover behave the same way
closes #2188

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -7,6 +7,7 @@ import {
   ChangeDetectionStrategy,
   OnInit,
   OnDestroy,
+  OnChanges,
   Injector,
   Renderer2,
   ComponentRef,
@@ -14,7 +15,8 @@ import {
   TemplateRef,
   ViewContainerRef,
   ComponentFactoryResolver,
-  NgZone
+  NgZone,
+  SimpleChanges
 } from '@angular/core';
 
 import {listenToTriggers} from '../util/triggers';
@@ -90,13 +92,13 @@ export class NgbPopoverWindow {
  * A lightweight, extensible directive for fancy popover creation.
  */
 @Directive({selector: '[ngbPopover]', exportAs: 'ngbPopover'})
-export class NgbPopover implements OnInit, OnDestroy {
+export class NgbPopover implements OnInit, OnDestroy, OnChanges {
   /**
-   * Content to be displayed as popover.
+   * Content to be displayed as popover. If title and content are empty, the popover won't open.
    */
   @Input() ngbPopover: string | TemplateRef<any>;
   /**
-   * Title of a popover.
+   * Title of a popover. If title and content are empty, the popover won't open.
    */
   @Input() popoverTitle: string;
   /**
@@ -133,6 +135,15 @@ export class NgbPopover implements OnInit, OnDestroy {
   private _windowRef: ComponentRef<NgbPopoverWindow>;
   private _unregisterListenersFn;
   private _zoneSubscription: any;
+  private _isDisabled(): boolean {
+    if (this.disablePopover) {
+      return true;
+    }
+    if (!this.ngbPopover && !this.popoverTitle) {
+      return true;
+    }
+    return false;
+  }
 
   constructor(
       private _elementRef: ElementRef, private _renderer: Renderer2, injector: Injector,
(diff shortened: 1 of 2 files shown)


medinarkhov:
@pkozlowski-opensource  please review

src/popover/popover.ts
```diff
@@ -150,13 +147,26 @@ export class NgbPopover implements OnInit, OnDestroy {
     });
   }
 
+  /**
+   * Content to be displayed as popover. If falsy, the popover won't open.
+   */
+  @Input()
+  set ngbPopover(value: string | TemplateRef<any>) {
+    this._ngbPopover = value;
+    if (!value && this._windowRef) {
```
pkozlowski-opensource: This check needs to be more strict - what if I want to open a popover with a single `0` value?

pkozlowski-opensource:
@medinarkhov I see that we've got similar logic in tooltips already but the situation with popovers is slightly more complicated, as a popover can have a title. What should we do if there is a non-falsy title but no content?

More generally, I would like to better understand your use-case. On my side I can see 2 reasons for having logic proposed in this PR:
* protect from developer errors (accidentally passing falsy value)
* have conditional popovers

I don't think we should be in the business of correcting for dev errors so it would be leaving use with the conditional popovers. If this is your use-case I would rather introduce a `disabled` (exact name to be defined).

Could you share more info on your particular use-case for which you need the change submitted in this PR?

medinarkhov:
@pkozlowski-opensource thank you for detailed comment.
Yes, I would like to use conditional popover the same way I can use conditional tooltip.
Example, http://plnkr.co/edit/qKEjdlS42FNuiKA3ZyWl?p=preview
In my case I would ignore `Title` when `Content == null`. 

@bedag-moo could you give your example of using conditional popover?

I like the idea with `disabled` input. I could try to create a new PR introducing `disabled` input. However, current PR is also makes sense because it is consistent with tooltip behavior.  So we could have both `conditional content` and `disabled`, what do you think?

src/popover/popover.ts
medinarkhov: Good point. There is no problem if the user passes value as string `[ngbPopover]="!shouldDisplayPopover ? '0' : null"`

medinarkhov:
@pkozlowski-opensource I read through Bootstrap specs https://getbootstrap.com/docs/4.0/components/popovers/
It says:
- Zero-length title and content values will never show a popover.

You were right, we need to check both content and title: hide only when both values are empty. I've updated my PR.

bedag-moo:
Thanks for the PR!

My usecase was a conditional popover without title, that contains a list of validation errors (which may become empty, in which case I want the popover to disappear).

Both proposed APIs (add a new disabled attribute, or check whether title and content are both falsy) work for me.

src/popover/popover.ts
```diff
@@ -150,14 +144,38 @@ export class NgbPopover implements OnInit, OnDestroy {
...
 
+  /**
+   * Content to be displayed as popover. If title and content are empty, the popover won't open.
+   */
+  @Input()
+  set ngbPopover(value: string | TemplateRef<any>) {
+    this._ngbPopover = value;
+    if (!this._ngbPopover && !this._popoverTitle && this._windowRef) {
+      this.close();
+    }
```
pkozlowski-opensource: @medinarkhov could you please extract this logic into a separate function? Even better, you could do this check in the `ngOnChanges` lifecycle hook - this way we wouldn't need a setter _and_ would remove code duplication.

pkozlowski-opensource:
@medinarkhov thnx for the PR, it looks good to me, generally speaking. I've just left one comment that I would like you to address before we merge this PR.

So please:
* move the closing logic to the `ngOnChanges` lifecycle hook
* squash all the commits into one so we've got only one commit in this PR
* your commits conform to the conventions established in this project (we follow the [conventional changelog](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md)

Also, this PR is a breaking change, technically speaking so marking it as such and scheduling for 2.0.0

PR #2218, opened 2018-03-14
docs: allow deprecation and new features in documentation

maxokorokov:
New features and deprecations should be reflected in documentation after `1.0.0`

Parsing additional `@deprecated` and `@since` annotations:

screen shot 2018-03-14 at 15 45 36

In documentation it will be reflected like this:

screen shot 2018-03-14 at 15 44 11

IDEs would also strikethrough deprecated methods and classes

If you deprecate the whole class:

screen shot 2018-03-14 at 15 44 44

Or a component/directive:

screen shot 2018-03-14 at 15 44 27

### DIFF
--- a/demo/src/app/components/shared/api-docs/api-docs-badge.component.ts
+++ b/demo/src/app/components/shared/api-docs/api-docs-badge.component.ts
@@ -11,13 +11,20 @@ const BADGES = {
 
 @Component({
   selector: 'ngbd-api-docs-badge',
-  template: `<h5><span class="badge" [ngClass]="badgeClass">{{text}}</span></h5>`,
+  template: `
+    <h5>
+      <span *ngIf="deprecated" class="badge badge-secondary" >Deprecated {{ deprecated.version }}</span> 
+      <span class="badge" [ngClass]="badgeClass">{{text}}</span>
+    </h5>
+  `,
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class NgbdApiDocsBadge {
 
   badgeClass;
   text;
+
+  @Input() deprecated: {version: string};
 
   @Input()
   set type(type: string) {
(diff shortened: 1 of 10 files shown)


demo/src/app/components/shared/api-docs/api-docs-badge.component.ts
```diff
@@ -11,14 +11,21 @@ const BADGES = {
 
 @Component({
   selector: 'ngbd-api-docs-badge',
-  template: `<h5><span class="badge" [ngClass]="badgeClass">{{text}}</span></h5>`,
+  template: `
+    <h5>
+      <span *ngIf="deprecated" class="badge badge-secondary" >Deprecated {{ deprecated.version }}</span>
```
pkozlowski-opensource: Shouldn't this be "Deprecated from" or "Deprecated in"?

nit: extra space before `<span>` tag close

demo/src/app/components/shared/api-docs/api-docs.component.html
```diff
@@ -58,9 +64,16 @@ <h4>Inputs</h4>
...
       <table class="table table-sm">
         <tbody>
-          <tr *ngFor="let output of apiDocs.outputs">
-            <td class="label-cell"><code>{{output.name}}</code></td>
-            <td class="content-cell">{{output.description}}</td>
+          <tr *ngFor="let output of apiDocs.outputs" [class.deprecated]="output.deprecated">
+            <td class="label-cell">
+              <code>{{output.name}}</code><br>
+              <span *ngIf="output.since" class="badge badge-info">since {{ output.since.version }}</span>
+              <span *ngIf="output.deprecated" class="badge badge-secondary">deprecated {{ output.deprecated.version }}</span>
```
pkozlowski-opensource: nit: maybe it makes sense to create a component for the 2 above lines as it seems that the same thing is repeated in several places?

pkozlowski-opensource:
LGTM. Left some nit-picking comments, feel free to fix / ignore and merge.

PR #2223, opened 2018-03-16
fix(modal): don't use inline styles

### LINKED ISSUE #2085: Modal does not work with strict Content Security Policy
### Bug description:
Opening a modal does not work if a Content Security Policy is enabled, unless style-src 'unsafe-inline' is also enabled.

It seems the modal uses the inline style 'display: block', which means that the CSP must allow inline styles for the modal to load correctly.

Could this be handled by moving the styles to a separate .css file, like in ui-bootstrap? (ui-bootstrap-csp.css)

http://plnkr.co/edit/pWadtAoosFePpru5nWIi?p=preview

Angular: 5.0.2

ng-bootstrap: 1.0.0-beta.9

Bootstrap: 4.0.0-beta,3

pkozlowski-opensource:
Fixes #2085

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -17,10 +17,9 @@ import {ModalDismissReasons} from './modal-dismiss-reasons';
 @Component({
   selector: 'ngb-modal-window',
   host: {
-    '[class]': '"modal fade show" + (windowClass ? " " + windowClass : "")',
+    '[class]': '"modal fade show d-block" + (windowClass ? " " + windowClass : "")',
     'role': 'dialog',
     'tabindex': '-1',
-    'style': 'display: block;',
     '(keyup.esc)': 'escKey($event)',
     '(click)': 'backdropClick($event)'
   },


pkozlowski-opensource:
Travis CI failure is a flake

src/modal/modal-backdrop.ts
```diff
@@ -1,5 +1,5 @@
 import {Component} from '@angular/core';
 
-@Component({selector: 'ngb-modal-backdrop', template: '', host: {'class': 'modal-backdrop fade show'}})
+@Component({selector: 'ngb-modal-backdrop', template: '', host: {'class': 'modal-backdrop show'}})
```
maxokorokov: Not sure about removing the `fade`class and leaving `show`. Take a look at BS styles:

screen shot 2018-03-16 at 16 02 41

Should we remove both?

pkozlowski-opensource: Yeh, probably... I was just afraid that people might be using `show` class for some reason (DOM manipulation). But in this case I should not be removing fade...

I will revert all this for now and only do inline style => `d-block` in this PR

PR #2229, opened 2018-03-17
feat(popover): introduce disabled flag

### LINKED ISSUE #2188: Conditional popovers

medinarkhov:
Fixes #2188

PR created as a continuation of PR #2217 

`[disabled]` would be a useful feature for popover\toltip, allowing user to create conditional popover\tooltip.
`<div [ngbPopover]="It can be conditionally hidden!" [disabled]="hidePopover"></div>`

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -118,7 +118,7 @@ export class NgbPopover implements OnInit, OnDestroy {
   /**
    * A flag indicating if a given popover is disabled and should not be displayed.
    */
-  @Input() disabled = false;
+  @Input() disablePopover = false;
   /**
    * Emits an event when the popover is shown
    */
@@ -159,7 +159,7 @@ export class NgbPopover implements OnInit, OnDestroy {
    * The context is an optional value to be injected into the popover template when it is created.
    */
   open(context?: any) {
-    if (!this._windowRef && !this.disabled) {
+    if (!this._windowRef && !this.disablePopover) {
       this._windowRef = this._popupService.open(this.ngbPopover, context);
       this._windowRef.instance.title = this.popoverTitle;
       this._windowRef.instance.id = this._ngbPopoverWindowId;
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -118,7 +118,7 @@ export class NgbPopover implements OnInit, OnDestroy {
   /**
    * A flag indicating if a given popover is disabled and should not be displayed.
    */
-  @Input() disablePopover = false;
+  @Input() disablePopover: boolean;
   /**
    * Emits an event when the popover is shown
    */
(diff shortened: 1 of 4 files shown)


medinarkhov:
@pkozlowski-opensource please review. If the change looks good then I will add another commit to update:
-  tooltip to have `[disabled]` flag (remove conditional content then?)
- update API
- update demo

src/popover/popover.ts
```diff
@@ -115,6 +115,10 @@ export class NgbPopover implements OnInit, OnDestroy {
    * Currently only supports "body".
    */
   @Input() container: string;
+  /**
+   * A flag indicating if a given popover is disabled and should not be displayed.
+   */
+  @Input() disabled = false;
```
maxokorokov: 1. Don't think it will be acceptable, too many collisions, ex:

```html
<button [disabled]="true" ngbPopover>...</button>
```

maybe `[showPopover]` or `[disablePopover]`?

```html
<div [ngbPopover]="text" [showPopover]="text !== ''" />
<div [ngbPopover]="text" [disablePopover]="text === ''" />
```

2. All other inputs initial values come from `NgbPopoverConfig`. I think we should put this one there as well. Not 100% sure though if there is a use case for massively initializing all popovers as disabled... So this is questionable

medinarkhov:
@maxokorokov thank you for prompt review. 

_Update:_
1. renamed `[disabled]` flag to `[disablePopover]`
2. updated `popover-config.ts`
3. won't touch Tooltip in this PR

API page looks good. Let me know if more changes required? Would you like demo example?

src/popover/popover.ts
```diff
@@ -115,6 +115,10 @@ export class NgbPopover implements OnInit, OnDestroy {
    * Currently only supports "body".
    */
   @Input() container: string;
+  /**
+   * A flag indicating if a given popover is disabled and should not be displayed.
+   */
+  @Input() disablePopover = false;
```
maxokorokov: You just forgot `this.disablePopover = config.disablePopover` in the constructor, and no need to initialize it here

medinarkhov: good catch! fixed)

pkozlowski-opensource:
LGTM, thnx!

PR #2230, opened 2018-03-17
feat(modal): Add option to drag the modal

### LINKED ISSUE #1378: How to make the modal can be dragged.

Angular:2.4

ng-bootstrap:

Bootstrap:

i want to know ...thanks

bastienmoulia:
Add an option `draggableSelector`.
By example, with the option `{ draggableSelector: '.modal-header' }` the header of the modal can be draggable.
Fix #1378

### DIFF
--- a/demo/src/app/components/modal/demos/draggable/modal-draggable.html
+++ b/demo/src/app/components/modal/demos/draggable/modal-draggable.html
@@ -1,0 +1,16 @@
+<ng-template #content let-c="close" let-d="dismiss">
+  <div class="modal-header">
+    <h4 class="modal-title">Drag me</h4>
+    <button type="button" class="close" aria-label="Close" (click)="d('Cross click')">
+      <span aria-hidden="true">&times;</span>
+    </button>
+  </div>
+  <div class="modal-body">
+    <p>One fine body&hellip;</p>
+  </div>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-outline-dark" (click)="c('Close click')">Close</button>
+  </div>
+</ng-template>
+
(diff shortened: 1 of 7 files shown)


maxokorokov:
A quick scan telling me that overall tests are missing.
Also maybe we should find a way to isolate and test the logic in `dragModal`, have a service for it or something?

pkozlowski-opensource:
> Also maybe we should find a way to isolate and test the logic in dragModal, have a service for it or something?

Agree with @maxokorokov - this should be something independent of the core modal functionality. Ideally I would see it as an extension living outside of `ng-bootstrap`, ideally in the user-land...

shyamal890:
This would be a great addition to `ng-bootstrap`

LastCodeDE:
We need this feature, too. We are using a lot of modal views but they are not draggable.

LastCodeDE:
We need this feature, too. We are using a lot of modal views but they are not draggable.

LastCodeDE:
We need this feature, too. We are using a lot of modal views but they are not draggable.

pkozlowski-opensource:
Thnx for the PR @bastienmoulia but we will pass on it for now. Bootstrap itself is not providing this functionality and we consider that it adds very little to the user's experience. On the other hand it would penalise other people using this library with additional code etc.

We could merge a PR that would introduce this functionality in a separate directive that would be tree-shakable, though.

deeg:
I am also looking for similar functionality, and might consider doing as requested @pkozlowski-opensource.

Just checking, is this something you would still consider accepting as a PR as long as it followed your state guidelines?

> We could merge a PR that would introduce this functionality in a separate directive that would be tree-shakable, though.

PR #2238, opened 2018-03-21
fix(typeahead) (document:click) does not work on iOS

linusbrolin:
This fixes closing the typeahead results on iOS when clicking outside the component.


src/typeahead/typeahead.ts
```diff
@@ -215,7 +216,11 @@ export class NgbTypeahead implements ControlValueAccessor,
   }
 
   onDocumentClick(event) {
-    if (event.target !== this._elementRef.nativeElement) {
+    if (event.target !== this._elementRef.nativeElement &&
+      (!this._windowRef ||
+       !this._windowRef.location ||
+       !this._windowRef.location.nativeElement ||
+       !this._elementClassNameContains(event.target, this._windowRef.location.nativeElement.className))) {
```
pkozlowski-opensource: I'm sorry, but it is not clear for me why we've got all this logic? Are you trying to check if the event is not coming from the results popup? If so, wouldn't check we use in other places (ex. https://github.com/ng-bootstrap/ng-bootstrap/blob/958b955fb3fdc96ff8292c472363adb326035bcf/src/dropdown/dropdown.ts#L31) work?

pkozlowski-opensource:
@linusbrolin thnx for the PR and sorry for the delayed review. I can see the bug you are trying to fix but this PR needs work before it can be considered for merging:
* it needs tests - we can't merge PRs without associated unit tests;
* see the in-line comment - I'm not sure why you are trying to check presence of CSS classes.
* it needs rebase / squashing (there should be just one commit in the PR and there should be no merge commits).

Please let us know when the above fixes are done so we can do another round of reviews.

pkozlowski-opensource:
I'm going to close this one as I haven't heard back with the review comments and we are trying to clean up PRs queue. Please open a new PR with tests and review comments addressed if you still care about this one.

src/typeahead/typeahead.ts
linusbrolin: @pkozlowski-opensource sorry, didn't see this until now. :/
Yes this is to check if the event is not coming from the results popup.
Unfortunately your suggested solution will not work on iOS if the results popup contains any other angular component.

For example:
I'm using a typeahead popup to show product results in a search box.
Each row in the typeahead popup is an angular component that shows a product image, product title, article number and price.

When tapping on any of the rows on iOS, it will simply close the popup nomatter what.
The code I provided is so far the only way I have found to make it behave like it should on iOS.

PR #2253, opened 2018-03-23
feat(tooltip): intoduce disabled flag

medinarkhov:
Introduce disable flag for `Tooltip` to be consistent with `Popover`

Example:
`<div [ngbTooltip]="Disabled!" [disableTooltip]="true"></div>`
`<div [ngbPopover]="Disabled!" [disablePopover]="true"></div>`


pkozlowski-opensource:
LGTM, thnx!

Issue #2177, opened 2018-02-19
Typeahead documentation adjustments for rxjs 5.5


andreas-aeschlimann:
The Typeahead documentation should be adapted to the changes coming in rxjs 5.5.

For example:
```
search = (text$: Observable<string>) =>
    text$
      .debounceTime(200)
      .distinctUntilChanged()
      .map(term => term.length < 2 ? []
        : states.filter(v => v.toLowerCase().startsWith(term.toLocaleLowerCase())).splice(0, 10));
}
```
This does not compile anymore, and could be changed to:
```
search = (text$: Observable<string>) =>
    text$.pipe(
      debounceTime(200),
      distinctUntilChanged(),
      map(term => term.length < 2 ? []
        : states.filter(v => v.toLowerCase().startsWith(term.toLocaleLowerCase())).splice(0, 10)
      )
    );
}
```

cristianmartinez:
This is a good one. Just in case we need more context about the change:
https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md

king612:
The new pipe operator pattern is not working for me.  Causes some weird typescript error and will not compile.  I am using:

@angular/cli:  1.6.1
rxjs:  5.5.5
typescript:  2.6.2
@ng-boostrap/ng-bootstrap:  1.0.0

Following the documentation at the link I do:

``import { debounceTime, distinctUntilChanged, map } from 'rxjs/operators';``

and the code in the previous post above.

Visual Studio Code reports no error, but ng build or serve with the --aot switch generates a strange error down in the bowels of typescript:  "TypeError: Cannot read property: 'text' of undefined"


Any help appreciated.  I'm trying to replace ng Material typeahead with ng-bootstrap typeahead.

king612:
Sorry, that is:

``@angular/cli:  1.6.1
@ng-bootstrap/ng-bootstrap:  1.0.0``

andreas-aeschlimann:
This is probably an issue you should talk about on rxjs or angular-cli.

If text is undefined, you probably need to make sure it has an initial value at component creation.

king612:
Well, the issue only occurs with ng-bootstrap.  I have lots of other modules in use in the app, and none of them are experiencing this problem, only ng-bootstrap typeahead.  And the only variable 'text' that is involved is the text$ observable taken directly from your sample code, so someone might want to look into it.

king612:
[...]

PR #2256, opened 2018-03-24
chore: use pipeable operators in library code

jnizet:
See https://github.com/ng-bootstrap/ng-bootstrap/issues/2177#issuecomment-375855692 for the rationale for this PR

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -19,15 +19,13 @@ import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
 import {Observable} from 'rxjs/Observable';
 import {BehaviorSubject} from 'rxjs/BehaviorSubject';
 import {Subscription} from 'rxjs/Subscription';
-import {letProto} from 'rxjs/operator/let';
-import {_do} from 'rxjs/operator/do';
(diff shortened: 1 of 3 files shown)


src/typeahead/typeahead.ts
```diff
@@ -182,19 +180,19 @@ export class NgbTypeahead implements ControlValueAccessor,
...
+    const inputValues$ = this._valueChanges.pipe(tap(value => {
       this._inputValueBackup = value;
       if (this.editable) {
         this._onChange(value);
       }
-    });
-    const results$ = letProto.call(inputValues$, this.ngbTypeahead);
-    const processedResults$ = _do.call(results$, () => {
+    }));
+    const results$ = inputValues$.pipe(this.ngbTypeahead);
```
maxokorokov: why would you keep these intermediate constants? (`results$`, `inputValues$`)
more readable?

jnizet: Thanks for your review!

To be honest, I just changed the usage of old operators to new ones, without changing anything else. But yes, the code is probably more readable this way than by chaining everything. In particular, the call to the ngbTypeahead input is more apparent this way.

Note that I added an additional tiny commit since your review, to make the code compatible with RxJS 6 (I tested the compatibility with a project of mine, and it works fine after this additional commit)

PR #2257, opened 2018-03-24
fix(pagination): emit page change only when collection size is set

### LINKED ISSUE #1841: Pagination. Default current page is 1 but documentation says 0
Because of default current page is 1 you can't ignore first loading. Furthermore you can't override it because any incorrect value is transformed to 1

```
if (page) {
 this.load(page); //always exist
}
```
Second question is why paginator calls `pageChange` callback with default page number after page was loaded?

http://plnkr.co/edit/VGeIFzcKzGeQV8MNGQus?p=preview

----
ng-bootstrap: v1.0.0-beta.4

maxokorokov:
See [my comment](https://github.com/ng-bootstrap/ng-bootstrap/issues/1841#issuecomment-375877649) in the related issue

Fixes #1841

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -562,6 +562,21 @@ describe('ngb-pagination', () => {
 
          expect(fixture.componentInstance.onPageChange).not.toHaveBeenCalled();
        }));
+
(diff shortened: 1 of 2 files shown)


jnizet:
LGTM

PR #2272, opened 2018-03-30
feat(autoclose): service & all popup based widgets

ymeine:
The autoclose feature brings a generic utility to handle the following use cases of popup widgets: 

- close popup on click inside
- close popup on click outside
- close popup on escape press

This PR also integrates it for each popup based widget.

There is 1 commit for the generic service, and 1 commit per modified widget.

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -19,6 +19,8 @@ import {
 import {AbstractControl, ControlValueAccessor, Validator, NG_VALUE_ACCESSOR, NG_VALIDATORS} from '@angular/forms';
 import {DOCUMENT} from '@angular/common';
 
+import {AutoCloseService, Subscriber} from '../util/autoclose.service';
+
 import {NgbDate} from './ngb-date';
 import {NgbDatepicker, NgbDatepickerNavigateEvent} from './datepicker';
 import {DayTemplateContext} from './datepicker-day-template-context';
@@ -70,6 +72,13 @@ export class NgbInputDatepicker<D = NgbDateStruct>
   private _disabled = false;
   private _model: NgbDate;
   private _zoneSubscription: any;
+  private _autoCloseSubscriber: Subscriber;
+
+  /**
+   * The element used to toggle the visibility of the datepicker.
+   * Used mainly to avoid handling automatically the datepicker visibility when clicking on this toggling element.
+   */
+  @Input('ngbDatepicker') togglingElement: HTMLElement | null;
 
   /**
    * Indicates whether the datepicker popup should be closed automatically after date selection / outside click or not.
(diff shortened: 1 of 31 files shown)


src/typeahead/typeahead.ts
```diff
@@ -267,17 +284,13 @@ export class NgbTypeahead implements ControlValueAccessor,
           }
           this._closePopup();
           break;
-        case Key.Escape:
```
maxokorokov: I prefer `case Key.Escape` to adding this to the service

src/datepicker/datepicker-input.ts
```diff
@@ -62,6 +64,13 @@ export class NgbInputDatepicker implements OnChanges,
   private _disabled = false;
   private _model: NgbDate;
   private _zoneSubscription: any;
+  private _autoCloseSubscriber: Subscriber;
+
+  /**
+   * The element used to toggle the visibility of the datepicker.
+   * Used mainly to avoid handling automatically the datepicker visibility when clicking on this toggling element.
+   */
+  @Input('ngbDatepicker') togglingElement: HTMLElement | null;
```
maxokorokov: This was a breaking change for the in #2440 
I should check why it doesn't look like a breaking change here...

src/dropdown/dropdown.ts
```diff
@@ -117,7 +113,7 @@ export class NgbDropdown implements OnInit {
    * When it is 'outside' dropdowns are automatically closed on outside clicks but not on menu clicks.
    * When it is 'inside' dropdowns are automatically on menu clicks but not on outside clicks.
    */
-  @Input() autoClose: boolean | 'outside' | 'inside';
+  @Input() autoClose: AutoCloseType;
```
maxokorokov: `AutoCloseType` will be displayed as-is in the documentation pages. I would prefer to leave `boolean | 'inside' | 'outside'` just for readability reasons

ymeine: Too bad :( Documentation tools should be improved, but that's out of scope. It's theoretically a potential double maintenance for us then (changing one requires changing the other since they are linked).

src/datepicker/datepicker-input.ts
ymeine: Well, if there is a toggling element but it is not passed here, the feature will not work properly when clicking on the toggling element. That's a breaking change.

src/typeahead/typeahead.ts
ymeine: This remains part of the debate: specific vs generic (or local vs service).

ymeine:
Initially I personally wanted to code the feature for each widget, one by one. I was told we should make a generic service for that. So I did it.

After gaining some experience doing that, i can tell some pros & cons of having a service, and things that I could already improve.

Pros: 

- one location for one common logic: we ensure consistency in behavior (and we also ensure DRY)
- greater control: centralizing proved to be useful in many cases... In this case, it is even required to handle the use case of a popup inside a popup (say, TypeAhead inside a Modal). Things can always be achieved otherwise, with smaller, distinct services/helpers for instance

Cons: 

- potential complexity (currently the case)
- indirection: when looking at the widget's code, the logic is not directly present
- unfortunately weak overlap between widgets: it looks like each widget specific use case adds a feature in the service, which won't be used by any other widget. This is not 100% the case of course, otherwise I would already have dropped the service myself)

Things I could changes: 

- simplify the service right away: that would maybe mean rewriting from scratch, so that's a bit time consuming (current complexity came from lack of design beforehand, and growing complexity while understanding the use cases one by one and just adding them, without taking a step back)
- keep the strict minimum in the service, put the rest in each widget

The real thing that requires a centralized processing is the use case of the popups in popups. We can't avoid the service completely. The stack needs to be known in order to avoid closing all the widgets at once.

pkozlowski-opensource:
After talking to @maxokorokov my understanding is that we are going to have separate PRs for popover / tooltip based on the same idea as in datepicker / dropdown. Closing this one then as my understanding is that we don't want to merge it in the current form.

PR #2275, opened 2018-03-31
feat(carousel): Allow to hide navigation items if only one slide

### LINKED ISSUE #2274: [Carousel] Hide indicators and navigation arrows if only one slide
### Feature description:
Allow the user to hide the navigation items when the carousel contains only one slide

I have a carousel that can contain 1 or more slides. It feels weird to display navigation items when navigation is impossible.

wilhelmhb:
Closes #2274

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -195,9 +195,9 @@ export class NgbCarousel implements AfterContentChecked,
     }
   }
 
-  showNavigation(): boolean { return !this.hideNavigation || (this.slides != null && this.slides.length > 1); }
+  showNavigation(): boolean { return !this.hideNavigation; }
 
-  showIndicator(): boolean { return !this.hideIndicator || (this.slides != null && this.slides.length > 1); }
+  showIndicator(): boolean { return !this.hideIndicator; }
 
   private _restartTimer() {
     this._stopTimer();
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -45,7 +45,7 @@ export class NgbSlide {
     '(keydown.arrowRight)': 'keyNext()'
   },
   template: `
-    <ol class="carousel-indicators">
+    <ol class="carousel-indicators" *ngIf="showIndicator()">
       <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId"
           (click)="cycleToSelected(slide.id, getSlideEventDirection(activeId, slide.id))"></li>
     </ol>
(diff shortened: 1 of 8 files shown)


wilhelmhb:
The tests seems to fail only due to Microsoft Edge and nbg-datepicker.
I can't test on the first one, and haven't touched the second.
@pkozlowski-opensource: do you have any hint as to where the issue could be ?

src/carousel/carousel.ts
```diff
@@ -183,6 +195,10 @@ export class NgbCarousel implements AfterContentChecked,
     }
   }
 
+  showNavigation(): boolean { return !this.hideNavigation || (this.slides != null && this.slides.length > 1); }
```
pkozlowski-opensource: @wilhelmhb could you please remove the ` this.slides.length > 1` check (and the corresponding tests(s) based on https://github.com/ng-bootstrap/ng-bootstrap/issues/2274#issuecomment-381173512 ?

src/carousel/carousel.ts
```diff
@@ -183,6 +195,10 @@ export class NgbCarousel implements AfterContentChecked,
...
+
+  showIndicator(): boolean { return !this.hideIndicator || (this.slides != null && this.slides.length > 1); }
```
pkozlowski-opensource: @wilhelmhb could you please remove the ` this.slides.length > 1` check (and the corresponding tests(s) based on https://github.com/ng-bootstrap/ng-bootstrap/issues/2274#issuecomment-381173512 ?

pkozlowski-opensource:
@wilhelmhb it looks good to me, generally speaking, could you just please:
* remove the explicit check for one slide based on https://github.com/ng-bootstrap/ng-bootstrap/issues/2274#issuecomment-381173512
* rebase

Thnx!

src/carousel/carousel.ts
wilhelmhb: if we have more than one slide, we probably still want to display the navigation item, right ?
Otherwise the user won't see that he/she can slide through the several slides.

src/carousel/carousel.ts
wilhelmhb: If I remove the `this.slides.length > 1` check, then the navigation items will be hidden, no matter how many slides you have.
If you have several slides, IMHO, it would never make sense to hide these navigations items. Hence the check.

wilhelmhb:
As navigation items are required when there are several slides, the explicit check on one slide is *needed*.
My branch has been rebased and all tests have passed.

pkozlowski-opensource:
> As navigation items are required when there are several slides

Why? Slides still change with time interval so I can imagine people wanting to have a carousel without navigation, even if there are multiple slides.

> the explicit check on one slide is needed.

I don't think I see why. Once again, I believe that an explicit flag to disable navigation (regardless of number of slides) would be enough (and less confusing at the same time).

wilhelmhb:
I get what you mean, I'll change it ASAP

PR #2286, opened 2018-04-08
demo: add anchor link to demo examples

jnizet:
The anchor links are similar to the ones used for the API doc sections.
They allow linking to a specific example from a stackoverflow comment or answer, for example.


pkozlowski-opensource:
Oh, great, meant to add it for a long time! 👍

PR #2294, opened 2018-04-10
chore: fix plunkers generation

pkozlowski-opensource:
There are 2 parts of this PR:
* making sure that the bundle doesn't contain RxJS
* correcting plunker config for RxJS 6.x compatibility


pkozlowski-opensource:
Here is a demo plunker with a new config and a bundle created with the new config: http://plnkr.co/edit/9frZfZUMiY7ocKYZH4na?p=preview

PR #2299, opened 2018-04-11
feat: clickoutside standalone service

benouat:
This is a preliminary work to prepare better handling of _clickoutside_ feature on any widget using a popup.

This PR provides a single feature: a service detecting click performed outside of an element.
For now this is an internal service, not meant to be used publicly.

The full picture here is to use it where widget would display something as a popup.
Hence its lifecycle should be the same one as popup. You create/destroy them together.

Usage is simple
+ provide the service in the widget module `providers: [NgbClickOutsideFactory]`
+ inject it in the widget constructor `private _clickoutsideFactory: NgbClickOutsideFactory`
+ call the create method specifying an element & a callback. It returns an instance of `NgbClickOutside`
```typescript
this._clickOutside = this._clickOutsideFactory.create(element, event => { /* User clicked outside of element */ };
```
+ later on, when not needed, properly destroy it `this._clickOutside.destroy()`

---
I could see possible evolution to this feature:
+ specify a list of elements instead of a single one.
+ register 2 callbacks: one executed on capture phase, one on bubbling phase (the actual implementation is using bubbling)

### DIFF
--- a/src/util/clickoutside.spec.ts
+++ b/src/util/clickoutside.spec.ts
@@ -1,0 +1,84 @@
+import {Component, ElementRef, OnDestroy, ViewChild} from '@angular/core';
+import {TestBed} from '@angular/core/testing';
+
+import {NgbClickOutside, NgbClickOutsideFactory} from './clickoutside';
+
+const getElement = (element: HTMLElement, selector: string): HTMLElement => {
+  return element.querySelector(selector) as HTMLElement;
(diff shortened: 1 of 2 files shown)


benouat:
Closing that one as it can be seen on #2440 we probably won't ever need such an external service for `clickOuside`.

Observables are just fine for that.

PR #2305, opened 2018-04-12
feat(datepicker): add native js date adapter

msosa:
Implements part 1 of #1977 and adds NgbDateNativeAdapter to the library

Let me know how this looks @maxokorokov and if there any changes you think are needed


msosa:
no problem! I'll take a look at ISO later

PR #2315, opened 2018-04-13
chore: update to Bootstrap 4.1

bastienmoulia:
Add bootstrap 4.1 dependency for 2.0 milestone.

Add a table with version compatibility in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ View all the directives in action at https://ng-bootstrap.github.io
 
 ## Dependencies
 * [Angular](https://angular.io) (tested with 5.0.2)
-* [Bootstrap 4](https://www.getbootstrap.com) (tested with 4.0.0)
+* [Bootstrap 4](https://www.getbootstrap.com) (tested with 4.1.0)
 
 ## Installation
 After installing the above dependencies, install `ng-bootstrap` via:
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
@bastienmoulia I was discussing this today with @maxokorokov and I think that we can leave 4.0.0 as a minimal _required_ Bootstrap CSS as all the releases so far work with both 4.0.0 _and_ 4.1.0.

WDYT?

bastienmoulia:
@pkozlowski-opensource It's true, just don't forget during the release to change the package.json `"bootstrap": "4.0.0"`in `"bootstrap": "^4.0.0"` and maybe the readme `tested with 4.1.0`.

pkozlowski-opensource:
> It's true, just don't forget during the release to change the package.json 

Well, I would much prefer that I don't have to remember about those things :-) 

In any case we can reasonably use / test only with one version of Bootstrap CSS so I would prefer that we are consistent with what we use and what we claim in our package.json.

I think that we should merge it when we are ready to require Bootstrap 4.1 from our users (and this is independent from the Angular version). Marking it as a breaking change.

pkozlowski-opensource:
As commented above today ng-bootstrap works with Bootstrap CSS `^4.0.0`. In practice it means that our library works perfectly with _any_ version of Bootstrap CSS. I see it as a good thing since bumping up _minimal_ Bootstrap CSS dependency is a breaking change.

For now we will stay with requirement for the minimal Bootstrap CSS version being `4.0.0`. We will update it only when we see there is a bug affecting ng-bootstrap that was fixed in any of 4.x versions.

Thnx for the PR but we are going to keep it as-is for now.

mendeza:
Thank you, it is good to understand the project stance clearly.

PR #2317, opened 2018-04-15
feat: add support for i18n of static text in component templates

### LINKED ISSUE #2314: Translation support
We need to make sure that people can translate strings in ng-bootstrap widget templates. Things to do:
* [ ] list widgets and their parts which might require tranlations
* [ ] select a set of supported translation solutions
* [ ] make the necessary changes to the existing templates and code

The current thinking is that we will support Angular translations so the majority of the process should boil down to adding `i18n` attributes in the templates of widgets needing translations

jnizet:
fixes #2314

Note: I didn't add any text for this feature, because I have no idea how to test this. It's not even possible to test that the i18n attributes are present in the rendered DOM, because Angular doesn't render them.

### DIFF
--- a/demo/src/app/getting-started/getting-started.component.html
+++ b/demo/src/app/getting-started/getting-started.component.html
@@ -68,6 +68,25 @@
   <br>
 
   <h3>
+    Internationalization
+  </h3>
+  <p>
+    Some components contain static English text or symbols that you might want to internationalize. Some of them appear
+    on the screen, like for example the placeholders used in the timepicker input fields. Others appear in aria attributes
+    used for accessibility.
+  </p>
+  <p>
+    Internationalizing the ng-bootstrap components is done the same way as for any of your components, using the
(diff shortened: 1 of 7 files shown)


jnizet:
I left out the messages for which I'm not sure if it's a good idea or not to mark them:
 -  `{{ i18n.getMonthFullName(month.number) }} {{ month.year }}` from datepicker-navigation.ts, useful only if we want to let users inverse the order. But that won't inverse the order of the select boxes anyway.
 - rating.ts: `&#9733;`, `&#9734;`: not sure if i18n is the best way of providing a custom symbol. Maybe the config service should be used instead.

Please tell me if you think those two should be translated (or if I missed some other static text to translate).

pkozlowski-opensource:
I'm going to merge this one before it gets stale. While this is not the complete solution to all our i18n problems, it does improve the situation considerably.

PR #2322, opened 2018-04-17
feat: focustrap standalone service

benouat:
PR introduces an internal service to setup a _focus-trap_ feature to be added to any element dynamically created by a widget.

Usage context: any datepicker popup, popover, modals...
Usage example:

+ provide the service in the widget module `providers: [NgbFocusTrapFactory]`
+ inject it in the widget constructor: `private _focusTrapFactory: NgbFocusTrapFactory`
+ call the create method specifying an element, it returns the newly created `NgbFocusTrap` instance

```typescript
this._focusTrap = this._focusTrapFactory.create(this.elementRef.nativeElement)
```

+ later on, when the dynamically created element is not needed, properly destroy the focustrap to release event listeners set on document: `this._focusTrap.destory()`

---

You can also decorate a node contained inside element with `NgbAutofocus` attribute, and specify a second boolean argument to the `create()` method:

```html
<div>
  <span></span>
  <input type="text" ngbAutofocus />
  <div></div>
</div>
```

```typescript
this._focusTrap = this._focusTrapFactory.create(this.elementRef.nativeElement, true)
```

this will automatically set the focus on the input when the focus gets initialised. On top it will remenber any previously focused element, and restore it when destroyed (feature needed to properly implement accessible modals)

### DIFF
--- a/src/util/focus-trap.spec.ts
+++ b/src/util/focus-trap.spec.ts
@@ -1,0 +1,123 @@
+import {AfterViewInit, Component, ElementRef, Inject, InjectionToken, OnDestroy, ViewChild} from '@angular/core';
+import {ComponentFixture, inject, TestBed} from '@angular/core/testing';
+
+import {NgbFocusTrap, NgbFocusTrapFactory} from './focus-trap';
+
+const getElement = (element: HTMLElement, selector: string): HTMLElement => {
+  return element.querySelector(selector) as HTMLElement;
+};
+
+const Autofocus = new InjectionToken('autofocus');
+
+describe('ngbFocusTrap', () => {
+  beforeEach(() => { TestBed.configureTestingModule({providers: [NgbFocusTrapFactory]}); });
+
+  it('should be instantiated manually', inject([NgbFocusTrapFactory], (focusTrapFactory: NgbFocusTrapFactory) => {
+       const element = document.createElement('div');
+       const focusTrap = focusTrapFactory.create(element);
+       expect(focusTrap).toBeDefined();
+       focusTrap.destroy();
+     }));
+
+  describe('navigation', () => {
+    let fixture: ComponentFixture<TestComponent>;
+    let instance;
+
+    beforeEach(() => {
+      TestBed.configureTestingModule({
+        declarations: [TestComponent, FocusTrapComponent],
+        providers: [NgbFocusTrapFactory, {provide: Autofocus, useValue: false}]
+      });
+      fixture = TestBed.createComponent(TestComponent);
+      instance = fixture.componentInstance;
+      fixture.detectChanges();
+    });
+
+    it('should intercept any outside focus when already instantiated', () => {
+      const initial = document.querySelector('#initial') as HTMLElement;
+      const link = getElement(fixture.nativeElement, 'a');
+      initial.focus();
+      expect(document.activeElement).toBe(link);
+    });
+
+    it('should create/remove the end of document anchor accordingly', () => {
+      expect(document.body.querySelector('.ngb-focustrap-eod')).toBeDefined();
+      instance.show = false;
+      fixture.detectChanges();
+      expect(document.body.querySelector('.ngb-focustrap-eod')).toBeNull();
+    });
+  });
+
+  it('should save/restore focused element when autofocus is set', () => {
+    TestBed.configureTestingModule({
+      declarations: [TestComponent, FocusTrapComponent],
+      providers: [NgbFocusTrapFactory, {provide: Autofocus, useValue: true}]
+    });
+    const fixture = TestBed.createComponent(TestComponent);
+    const instance = fixture.componentInstance;
+    instance.show = false;
+    fixture.detectChanges();
+
+    // put focus somewhere
+    const initial = document.querySelector('#initial') as HTMLElement;
+    initial.focus();
+
+    // let's create the focustrap with autofocus (via <focus-trapped>)
+    instance.show = true;
+    fixture.detectChanges();
+    const button = getElement(fixture.nativeElement, 'button');
+    expect(document.activeElement).toBe(button);
+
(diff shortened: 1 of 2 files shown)


src/util/focus-trap.ts
```diff
@@ -0,0 +1,140 @@
...
+      this._removeDocumentListener = () => {
+        this._document.removeEventListener('focusin', this._enforceFocus, true);
+        this._document.removeEventListener('keydown', this._detectDirection);
+      }
+    });
+  }
+
+  /** Detect if incoming focus event should be prevented or not */
+  private _enforceFocus(event) {
+    const {shiftKey, target} = event;
```
maxokorokov: `shiftKey` is not used

src/util/focus-trap.spec.ts
```diff
@@ -0,0 +1,139 @@
+import {AfterViewInit, Component, ElementRef, Inject, InjectionToken, OnDestroy, ViewChild} from '@angular/core';
+import {ComponentFixture, inject, TestBed} from '@angular/core/testing';
+
+import {NgbFocusTrap, NgbFocusTrapFactory} from './focus-trap';
+
+const getElement = (element: HTMLElement, selector: string): HTMLElement => {
+  return element.querySelector(selector) as HTMLElement;
+};
+
+function dispatchEvent(element, eventName: string, props: {[key: string]: any}) {
```
maxokorokov: `dispatchEvent` is not used

src/util/focus-trap.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Register the focustrap instance to initially put the focus on specific element
+   *
+   * Simply add the `ngbAutofocus` attribute on the given element such as `<button ngbAutofocus></button>`
+   */
+  set autofocus(value: boolean) {
+    if (value === true && this._previouslyFocused === null) {
+      this._previouslyFocused = document.activeElement as HTMLElement;
+      this._focusInitial();
```
maxokorokov: This would be called synchronously, I'm not sure it will work for dynamically instantiated components. 
I guess it would try to focus before the change detection is run, so should be checked

src/util/focus-trap.spec.ts
benouat: Removed !

src/util/focus-trap.ts
benouat: Removed !

src/util/focus-trap.ts
benouat: I am not sure it might be needed. Live tests seem to show that it is working in both datepicker and modal without that change.

Nonetheless, if we would ever need such a thing, I would rather prefer creating the focusTrap inside of an `ngAfterViewInit` to make sure that View is created and hence available to focus.

benouat:
@maxokorokov I pushed a new commit, and added a small change to make sure it would work properly in modal.

The issue with modal is simple. Modal are appended to the body, so when you hit <kbd>tab</kbd> on the last focusable element of the modal, you basically have reached the end of the document. So we never go inside the `focusin` in focustrap.

Fix was simple, when focustrap is created I create a focusable anchor and append it to the body.

src/util/focus-trap.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  /**
+   * Guess the next focusable element.
+   * Computation is based on specific CSS selector and [tab] navigation direction
+   */
+  get focusableElement(): HTMLElement {
```
maxokorokov: It is neither used publicly nor in tests. Why have a public getter for it instead of a private function?

src/util/focus-trap.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Create an instance of {@link NgbFocusTrap} and return it
+   * @param element HTMLElement to trap focus inside
+   * @param autofocus Whether the focustrap should automatically move focus into the trapped element upon
+   * initialization and return focus to the previous activeElement upon destruction.
+   */
+  create(element: HTMLElement, autofocus = false): NgbFocusTrap {
+    const trap = new NgbFocusTrap(element, this._document, this._ngZone);
+    trap.autofocus = autofocus;
```
maxokorokov: Same question for autofocus setter. Do you see anybody calling it after constructing the service?

I mean doing `service = factory.create(el)` and then doing `service.autofocus = true` later, as opposed to `service = factory.create(el, true)` right away ?

src/util/focus-trap.ts
benouat: I used to have tests in one of the previous version. I will make it private

src/util/focus-trap.ts
benouat: Yes, to me it was kind of related to one of the previous review comment.
I could see someone creating the focusTrap in `constructor`, and only activate autofocus in `ngAfterViewInit` to make sure the view element to focus is there.

maxokorokov:
Saucelabs failing as usual, will merge when green

benouat:
Awesome ! Thanks

ymeine:
You got a bit fast on merging this one - btw my PRs are forever pending - since it has some drawbacks.

First it doesn't handle the use case of coming from the outside of the viewport. Then it doesn't handle screen readers, where `aria-hidden` should be used. Finally, it re-implements a whole logic upon each <kbd>tab</kbd>/<kbd>shift+tab</kbd> keystroke while some other equivalent techniques don't. For people who tend to do remarks about micro-optimizations I find this one a bit borderline.

Also, since usually the Angular way is pushed, I thought that it would have been a good use case for a directive.

Not to mention some other use cases not handled such as elements hidden because of a parent which is not displayed.

So please don't overlook team work, I've worked on this task for quite a while now (doing researches to know what would be best for this product) so I have some knowledge about it, and also have an implementation already done before which handles those mentioned use cases.

PR #2328, opened 2018-04-19
fix(focustrap): more robust implementation based on focusout event

benouat:
### Context

Trying to implement focustrap on modal widget, I just discovered a not so great edge case side effect not handled by #2322.

On a big document with scrollbars, modals add a specific className on body to hide the overflow.

But when navigating backward with <kbd>tab</kbd> inside the modal, the first `focusin` event dispatched when outside of the modal might be located almost at the bottom of body (Simply because modals are actually appended to the body).

And even if focustrap would catch this `focusin` to immediately fire a `focus` elsewhere, the browser still manage to perform the scroll to place into the viewport the element that was about to be `focusin`, resulting in having the page to scroll behind the backdrop.

focustrap-scroll-fail

---

### Content

PR introduces a new way of managing focustrap based on `focusout` being backward compatible with previous implementation, and taking this edge case into consideration.

- [x] focus first focusable element when bottom boundary is reached by <kbd>tab</kbd> forward
- [x] focus last focusable element when top boundary is reached by <kbd>tab</kbd> backward
- [x] catch any outside focus (programmatic or tabIndex > 1) and put focus back inside the watched element

Not yet implemented (might not be needed, or not needed now, or could be done later if issues are opened):

- [ ] better querySelectorAll to strengthen the list of focusable elements
- [ ] manage `aria-hidden` attribute on elements outside of the watched element
- [ ] add dozens of Protractor tests

### DIFF
--- a/src/util/focus-trap.ts
+++ b/src/util/focus-trap.ts
@@ -14,12 +14,13 @@ enum DIRECTION {
 /**
  * Class that enforce the browser focus to be trapped inside a DOM element.
  *
- * The implementation is rather simple, the class add a `focusin` listener on the document with capture phase.
- * Any focus event will then be caught, and therefore the class will only allow the one for elements contained inside
- * it's own element.
+ * The implementation is rather simple, the class add a `focusout` listener on the element with capture phase.
+ * We then detect the navigation direction.
+ * Whenever we are about to exit the watched element from either top or bottom, we re-focus manually the right element
+ * accordinlgy.
  *
- * In case the element is not contained, the class will determine which new element has to be focused based on the `tab`
- * navigation direction.
+ * As a additional security, we also set a `focusin` listener on document, and immediately put focus back into the
+ * watched element.
  *
  * Should not be used directly. Use only via {@link NgbFocusTrapFactory}
  */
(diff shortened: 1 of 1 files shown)


src/util/focus-trap.ts
```diff
@@ -70,36 +80,42 @@ export class NgbFocusTrap {
...
+    }
+  }
+
+  /** Detects if we focus is reaching boundaries of the watched element, preparing to loop */
+  private _enforceFocus(event) {
+    const {target} = event;
+    const[first, last] = this._focusableBoundaryElements;
+    if ((target === first && this._direction === DIRECTION.BACKWARD) ||
+        (target === last && this._direction === DIRECTION.FORWARD)) {
+      this._focusElement(this._focusableElement as HTMLElement);
```
maxokorokov: You kind of know exaclty which node to focus here, so maybe no need to query document for the second time

Also `as HTMLElement` looks redundant as types match

benouat: I will split the condition in 2, and focus accordingly

benouat:
>do we need end-of-document node now with focusout?

We still need the end of document. Otherwise it is not working in FORWARD mode. When `focusout`is fired on the last focusable element of the document, even if you re-focus manually somewhere else, the browser is taking over, and focus end up in browser tabs.

> am I correct that the use-case that still won't work is when you initially focusing modal, but the next node is not in the viewport?

I don't understand your question.

benouat:
I am closing that one as I think it won't land anymore. Different direction has been taken with #2419

PR #2332, opened 2018-04-20
feat(datepicker): focustrap for popup datepickers

benouat:
Straight forward implementation of focustrap for any datepicker opened in a popup via `NgbDatepickerInput`

This PR does not have any tests, as writing UT for that would end up in duplicating NgbFocusTrap UT.

Nonetheless, as soon as Protractor is setup on the project, we then can write a whole bunch of new dedicated tests.


benouat:
Apparently, tests are failing on IE10

PR #2333, opened 2018-04-20
feat(modal): focustrap

benouat:
Same as #2332.

Very rough implementation of focustrap for modals.
With current master implementation of focustrap that landed via 6961cb3, the bug described and fixed in #2328 is present.

This PR does not have any tests, as writing UT for that would end up in duplicating NgbFocusTrap UT.

Nonetheless, as soon as Protractor is setup on the project, we then can write a whole bunch of new dedicated tests.


benouat:
Closing that one, because as soon as #2419 land we will be able to provide a much more simple version of it

Issue #1946, opened 2017-11-07
Datepicker: accessibility issues


hvinh:
### Bug description:
Accessibility issue
We noticed the following issues on the date picker widget:
- For the next/previous month buttons: there is no alternative text for screen readers and no tooltip on mouse over for the purpose of these buttons
- The days of the week are read in short format by screen readers: "Mon" instead of "Monday"
- Only the number is read by screen readers for the dates in the calendar: "1" instead of "Monday 1 September"
- After selecting a date, the selected date should be conveyed to screen reader users. This could be achieved by setting the focus back to the input field as raised in #1934 

Tested with NVDA/Firefox and JAWS/IE

### Link to minimally-working plunker that reproduces the issue:
Date picker in a pop-up
https://ng-bootstrap.github.io/#/components/datepicker/examples

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: angular 4.4.6

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta.2

maxokorokov:
Also see #1716, as not possible to focus navigation part of the datepicker currently

mendeza:
Driving the Datepicker via keyboard-only - an A11y requirement - is not possible with this component.

maxokorokov:
@mendeza it is possible to use datepicker with keyboard only, please see the [keymap service](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-keymap-service.ts) for available keys. Screen readers and appropriate aria attributes are not working yet though.

maxokorokov:
The remaining checklist to close is:
- [x] allow to focus items inside the datepicker (#1716)
- [ ] accessibility attributes for calendar (#2319)
- [ ] generic focus trap service (#2322)
- [ ] accessibility attributes for navigation (no PR yet)
- [ ] focus trap for datepicker input (no PR yet)

pkozlowski-opensource:
Seems like everything is done here 🍾

PR #2345, opened 2018-04-27
feat(datepicker): add aria attributes for navigation

maxokorokov:
Adding aria attributes for:
- previous/next month buttons
- months select box

Note: some screen readers ignore `aria-label` on option elements

Part of #1946 

cc @fbasso

### DIFF
--- a/src/datepicker/datepicker-navigation-select.spec.ts
+++ b/src/datepicker/datepicker-navigation-select.spec.ts
@@ -11,9 +11,8 @@ import {NgbDate} from './ngb-date';
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
 
-function getOptionValues(element: HTMLSelectElement): string[] {
-  return Array.from(element.options).map(x => (x as HTMLOptionElement).value);
-}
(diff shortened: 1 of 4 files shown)


maxokorokov:
@benouat, yep I tried putting both `title` and `aria-label` on the option didn't make any difference with any of the 3 screen readers, so I left `title` out

They read full month correctly with aria-label when you focus the select, but when you switch between options - only short names (what is actually displayed). Playing with `value` of the option didn't help either.

Haven't tried computing it on select and I doubt it will help when navigating between the options with the keyboard.

benouat:
@maxokorokov ok, then another idea would be to play the recent `liveAnnouncer` listening the `onChange` event

maxokorokov:
I think `onChange` is triggered only when you actually select an option. I was talking about the use case when you use `arrow up/down` to move between options here:
screen shot 2018-05-04 at 12 09 32

benouat:
Oh I see... I am afraid that this one will never ever be taken into account in different way that it is already today.

Myself I would use the arrow to navigate when the list is closed, I mean with just the field being focused. Thus computing the `select` aria-label on `onChange` should work

pkozlowski-opensource:
I've just merged 65c232d5c6a275ee1cef4db75b44e9fd00b348f0 so we would have to add i18n attributes so people can translate those new messages.

PR #2347, opened 2018-04-30
feat(timepicker): add custom time adapter support

### LINKED ISSUE #545: timepicker: allow two-way binding to something else
Users of timepicker (and, later, datepicker) will amost certainly want to bind the component to a string, a moment, a Date, or some other class that provides more functionality than a simple object with three fields.

I don't know if the component should provide this functionality (for example by passing a TimeFormat input which would be responsible for transforming the `{hour, minute, second}` object to something else and vice-versa), or if this should be done outside of the component. If the latter, a demo demonstrating how to do that would be helpful.

There might be two different needs BTW: creating a new string every time the model changes, and mutating the time part of a moment/Date.

jnizet:
This makes the timepicker consistent with the datepicker, by allowing
to use something other than an NgbTimeStruct as the model.

fix #545


pkozlowski-opensource:
Merged, thnx! Sorry for the delay

PR #2359, opened 2018-05-04
Fix: stop propagation from $event "closeFromClick"...

Carghaez:
Fix: stop propagation from $event "closeFromClick" of dropdown that cause execution of angular's digest lifecycle multiple times. 

More dropdown => more digest lifecycle.

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -190,14 +190,26 @@ export class NgbDropdown implements OnInit {
   }
 
   closeFromClick($event) {
-    if (this.autoClose && $event.button !== 2 && !this._isEventFromToggle($event)) {
+    if (!this.autoClose) {
+      return;
+    }
+    let doClose = false;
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@Carghaez we would need unit test(s) to prove / better understand your intended change.

pkozlowski-opensource:
> I have done another change to the behaviour of the document click listener. I recommend you to take a look

@Carghaez this is not how it works, I'm afraid... Your proposed change doesn't have tests _and_ breaks existing tests which indicates that it might introduce regressions. We can't accept PRs in such state and we are very clear about it in our contribution guidelines.

I'm going to close this PR based on the above, please open a new one if / when you've got time to work on a PR that follows our guidelines. Also relevant reading:  https://github.com/ng-bootstrap/ng-bootstrap/wiki/Contributions%3A-How-to-get-a-PR-merged-fast%3F

PR #2363, opened 2018-05-04
chore: use providedIn instead of forRoot()

bastienmoulia:
With the release of Angular 6, I made a PoC to replace the `.forRoot()` in the import by the new `providedIn` option in the `Injectable`.
https://angular.io/guide/providers#providing-a-service

It's work well, but it's a breaking change, so maybe for the 3.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,21 +42,20 @@ Once installed you need to import our main module:
 ```js
 import {NgbModule} from '@ng-bootstrap/ng-bootstrap';
 ```
-The only remaining part is to list the imported module in your application module. The exact method will be slightly
-different for the root (top-level) module for which you should end up with the code similar to (notice `NgbModule.forRoot()`):
+The only remaining part is to list the imported module in your application module:
 ```js
 import {NgbModule} from '@ng-bootstrap/ng-bootstrap';
 
 @NgModule({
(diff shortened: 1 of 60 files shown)


jnizet:
Couldn't we provide a smoother and faster upgrade path by
- keeping the forRoot() methods, but make them return the module without any provider (since providedIn: 'root' will take care of adding the necessary services to the root module)
- deprecating the forRoot() methods since they would actually not be useful anymore
- change the examples and the documentation to promote not using forRoot() anymore?

pkozlowski-opensource:
I agree with @jnizet suggestions - @bastienmoulia do you see this as being feasible? 

@bastienmoulia also please try to avoid mixing unrelated changes in one PR: 79c8ae8 should go into a separate PR. Could you please remove this commit here and open a new PR with this change? Thnx!

pkozlowski-opensource:
This should be rebased after #2483 gets merged

pkozlowski-opensource:
We've just merged #2483. I don't this that this PR is easy to rebase any more - IMO it would be much easier to just open a new when we are close to 4.0.0 release (we can make such breaking change only in a major release).

We are trying to cleanup PRs queue so I'm going to close this PR for now.

Issue #2194, opened 2018-02-27
typeahead: autocomplete="off" no longer prevents autocomplete in Chrome


TylerRick:
### Bug description:

Chrome shows its autocomplete suggestions when I start typing. These suggestions cover up the ngbTypeahead UI, making it virtually unusable when Chrome shows its suggestions:

![image](https://user-images.githubusercontent.com/14151/36749229-fe044b94-1bae-11e8-9de6-ab439a804a83.png)

It also makes it possible to input a value that's not in the list of allowed options defined by one's typeahead component, which is also not desirable. Worse still, if you select an option from Chrome's autocomplete, then it doesn't trigger the needed events and doesn't end up updating the bound model (model still says undefined)! 

#520 added `autocomplete="off"` to host, which was *supposed* to prevent this (and undoubtedly did until recently), but recent Chrome versions ignore this, much to the frustration of developers that depended on that behavior. See for example:

- https://stackoverflow.com/questions/47775041/disable-autofill-in-chrome-63
- https://bugs.chromium.org/p/chromium/issues/detail?id=587466
- https://stackoverflow.com/questions/15738259/disabling-chrome-autofill

### Link to minimally-working plunker that reproduces the issue:

I tried but could not reproduce in a [plunker](http://plnkr.co/edit/mMLQunlFOL9GLngLoDv5?p=preview). I'm not sure what the difference is in the plunker, or how to reproduce it there, but maybe someone else will know the answer.

Note that to reproduce you probably need to actually submit the form with some values so that it remembers them. I added `ngNoForm` in order to allow submitting the form, and tried this, but I can't get the autocomplete suggestions to show up in the plunker unless I remove the `[ngbTypeahead]`.

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 5.2.6

ng-bootstrap: 1.0.0

Bootstrap: 4

Chromium: 64.0.3282.140 (Official Build)

TylerRick:
Suggestion: Add a `@Input` for the `autocomplete` attribute so that users can override the default value of `off` if they need to (and if they think they know which value will be most effective at disabling autocomplete in the browser(s) that they know that their users will be using).

That or set it to a random value like `ignored-23195871650` to ensure it won't match any existing fields that it has autocomplete data for.

One of the SO posts suggested setting to `new-password` but that doesn't sound like a suitable option for a non-password field to me.

[...]

PR #2366, opened 2018-05-07
feat(typeahead): configurable autocomplete attribute

divdavem:
This PR makes the autocomplete attribute in the typeahead configurable. It allows to fix issue #2194.

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -67,7 +67,7 @@ let nextWindowId = 0;
     '[class.open]': 'isPopupOpen()',
     '(document:click)': 'onDocumentClick($event)',
     '(keydown)': 'handleKeyDown($event)',
-    'autocomplete': 'off',
+    '[autocomplete]': 'autocomplete',
     'autocapitalize': 'off',
     'autocorrect': 'off',
     'role': 'combobox',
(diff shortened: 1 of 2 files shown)


src/typeahead/typeahead.ts
```diff
@@ -89,6 +89,12 @@ export class NgbTypeahead implements ControlValueAccessor,
   private _windowRef: ComponentRef<NgbTypeaheadWindow>;
   private _zoneSubscription: any;
 
+  /**
+   * Value for the configurable autocomplete attribute.
+   * Defaults to 'off' to disable the native browser autocomplete, but this standard value does not seem
+   * to be always correctly taken into account (cf https://github.com/ng-bootstrap/ng-bootstrap/issues/2194).
```
pkozlowski-opensource: We are generating user documentation from those JS doc comments and I don't think that we want to have links to GH issues in the users docs.

Could you please split this message into 2 parts:
- relevant for end-users and keep it here;
- relevant for project developers and move it to the commit message?

src/typeahead/typeahead.ts
```diff
@@ -67,7 +67,7 @@ let nextWindowId = 0;
     '[class.open]': 'isPopupOpen()',
     '(document:click)': 'onDocumentClick($event)',
     '(keydown)': 'handleKeyDown($event)',
-    'autocomplete': 'off',
+    '[attr.autocomplete]': 'autocomplete',
```
pkozlowski-opensource: nit: there seems to be the `autcomplete` _property_ as well (https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement), so you could probably write it as `'[autocomplete]': 'autocomplete'`

pkozlowski-opensource:
@divdavem thnx, it LGTM, left 2 minor comments. Could you please have a look?

divdavem:
@pkozlowski-opensource Thank you for your review. I have updated my PR to take your comments into account.

PR #2369, opened 2018-05-07
fix(accordion): following bootstrap semantics

kukac7:


pkozlowski-opensource:
This should be part of #2368 - please merge code changes and tests into one PR.

Issue #2350, opened 2018-05-01
Logo design for bootstrap


harunbadilli:
I designed a logo for bootstrap. I think it is a more simple and stylish design with the new look. In this way, I designed a logo that is much more animated than a Monotype logo. With your consent, I will send you your logo in the color and size you want.

![1](https://user-images.githubusercontent.com/34319922/39473644-820ac276-4d58-11e8-85f8-593efaeed34e.png)
![2](https://user-images.githubusercontent.com/34319922/39473645-823c99cc-4d58-11e8-8b8a-dfc29ecfb907.png)
![3](https://user-images.githubusercontent.com/34319922/39473646-82767980-4d58-11e8-90a9-696182b80751.png)

mohammedzamakhan:
I think it should be the merge of bootstrap stack and angular logo

![bootstrap-stack](https://user-images.githubusercontent.com/2327532/39481401-fa7d2992-4d30-11e8-886d-c4a3ee88147f.png)

 and

![angular_solidblack](https://user-images.githubusercontent.com/2327532/39481454-22633b0e-4d31-11e8-8906-2b6eed566c7f.png)

velvok:
Agreed. It could be something like a bootstrap stack but in a shape of Angular shield instead of bootstrap's rhombus. Like a stack of shields. And the letter "A" on top of course. Would keep bootstrap color so that a final logo inherits 2 things from Angular (shape and letter) and 2 things from Bootstrap (stack and color)

benouat:
@velvok Something like that ?

![image](https://user-images.githubusercontent.com/1152740/39566683-b20c5254-4ebc-11e8-9c12-3f3a6e3981e6.png)

pkozlowski-opensource:
@benouat I love direction of your proposal! It looks good and have 2 nice properties:
* it is _clearly_ a combination of Bootstrap and Angular worlds
* the new logo is sufficiently different from the old one to warrant change yet similar enough so we are not loosing project's identity!

pkozlowski-opensource:
Thnx for the inspiration @harunbadilli and @mohammedzamakhan - we've decided to go with @benouat  proposal and website redesign!

PR #2380, opened 2018-05-15
chore(demo): new version of website

benouat:
Regarding discussion that started in #2350 I came up with a redesign of the homepage, which is as the previous versions, heavily inspired from Bootstrap itself.

Then with this new homepage, I got motivated to actually perform an entire re-design of the website.

This is what this PR is about.

Instead of forcing you to clone and setup up locally to have a look, I have pushed the aot build on https://benouat.github.io/ng-bootstrap-2018/

Feel free to comment. And please do try to play with responsive aspect of it. Don't hesitate to resize a lot the window, especially on component pages, as I tried to make it really dynamic.


pkozlowski-opensource:
Love it @benouat !

Just 2 nits:
* the 6 icons at the bottom of the page look kind of small as compared to other pieces on the page;
* not sure about the button to switch between examples / API on a component's page - maybe it is just about a color?

benouat:
> the 6 icons at the bottom of the page look kind of small as compared to other pieces on the page

We could try to make icons bigger when in desktop/wide screen mode, keeping them as it is here for tablet/mobile

> not sure about the button to switch between examples / API on a component's page - maybe it is just about a color?

For the button, I admit that the color here might be disturbing. First of all, I clearly wanted in the first place to get rid of the annoying tabbar I introduced myself one year ago. I did not like it back then, and it did not change, I can't stand it anymore. 😄 

So my idea was to stick with standard BS buttons. I tried several BS colors... none of them is better that each other. The one I picked sounded to be the less weird.

benouat:
Some tests here

image

image

I like it better 👍

pkozlowski-opensource:
I'm going to merge this since I believe that this is a wast improvement over the existing demo site. We can always iterate on improvements. Thnx @benouat !

PR #2384, opened 2018-05-17
feat(modal): add 'moduleCFR' option

codingnuclei:
Enable consumers to use NgModal service with lazy loaded module.

Using lazy loaded modules creates a separate ComponentFactoryResolver for that module.  Injecting NgModal service, into a shared services caused NgModal to use the root Component Factory Resolver. As such entryComponents used by lazy loaded modules have to be added to the root module. Increasing main.js bundle size.

This feature allows for passing in a ComponetFactoryResolver (i.e the one create for the lazy loaded module), so you can declare entryComponents on the modules that use them. Decrease main.js bundle size.

### DIFF
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -38,7 +38,8 @@ export class NgbModalStack {
     }
 
     const activeModal = new NgbActiveModal();
-    const contentRef = this._getContentRef(moduleCFR, options.injector || contentInjector, content, activeModal);
+    const contentRef =
+        this._getContentRef(options.moduleCFR || moduleCFR, options.injector || contentInjector, content, activeModal);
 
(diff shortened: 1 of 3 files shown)


pkozlowski-opensource:
This should not be needed - it should be enough to have `NgbModule` imported in your lazy-loaded module. Did you try it? If so, could you please provide a stackblitz demonstrating that this is not working?

codingnuclei:
What do you think?  [StackBlitz](https://stackblitz.com/edit/angular-rgr6pe)

Am I just missing something?

pkozlowski-opensource:
> What do you think? StackBlitz

So the problem is that you've got a "shared" service (`MyModalService`) which is provided in the root module therefore it can only "see" factories from root. What you should do is to create a _true_ shared module and import it in each individual module that needs modal functionality. Otherwise it is not clear who uses what.

PR #2392, opened 2018-05-22
fix(datepicker): prevent focustrap to restore focus on destroy

### LINKED ISSUE #2372: [this._focusTrap is null] Or [Cannot read property 'destroy' of null] When calling `close()`
### Bug description:
Sometimes, when calling `close()` error will be thrown.
Error: `this._focusTrap is null`
https://stackblitz.com/edit/angular-hobhrq

Steps:

- Make sure console is open.
- Click on input, this will open datepicker dialog. 
- Click now on text above(outside), this will trigger an event in code, and will close datepicker. (Sometimes need to click twice)
- Error will be thrown.

chrome: Cannot read property 'destroy' of null
FF: this._focusTrap is null

Angular: 6.0.0

ng-bootstrap: 2.0.0

Bootstrap: 4.1.1

benouat:
With this fix, focustrap on datepicker will no longer remember element focused prior initialization, and thus won't try to set it back on destroy.

fixes #2372

### DIFF
--- a/src/util/focus-trap.spec.ts
+++ b/src/util/focus-trap.spec.ts
@@ -48,7 +48,7 @@ describe('ngbFocusTrap', () => {
     });
   });
 
-  it('should save/restore focused element when autofocus is set', () => {
+  it('should focus element decorated with ngbAutofocus when created', () => {
     TestBed.configureTestingModule({
       declarations: [TestComponent, FocusTrapComponent],
       providers: [NgbFocusTrapFactory, {provide: Autofocus, useValue: true}]
@@ -71,7 +71,7 @@ describe('ngbFocusTrap', () => {
     // let's destroy the focustrap (removing <focus-trapped>)
     instance.show = false;
     fixture.detectChanges();
-    expect(document.activeElement).toBe(initial);
+    expect(document.activeElement).not.toBe(button);
   });
 });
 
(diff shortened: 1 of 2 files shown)


src/util/focus-trap.spec.ts
```diff
@@ -48,7 +48,7 @@ describe('ngbFocusTrap', () => {
     });
   });
 
-  it('should save/restore focused element when autofocus is set', () => {
+  it('should focus element decorated with ngbAutofocus when created', () => {
```
pkozlowski-opensource: Is this tests truly verifies what it claims? I can't see any element with `ngbAutofocus` anywhere....

pkozlowski-opensource:
@benouat I think that tests need more attention - left an in-line comment.

src/util/focus-trap.spec.ts
benouat: Here it is https://github.com/benouat/ng-bootstrap/blob/9c946f70d9fd382ea1386233c7fa85ad661ec01d/src/util/focus-trap.spec.ts#L84

pkozlowski-opensource: Oh, right, I see. But still, we should have an expectation verifying that this element is actually focused, no?

pkozlowski-opensource: Never mind, it was just GitHub diff view not showing the whole test :-/ LGTM

pkozlowski-opensource:
LGTM, going to merge in a bit.

PR #2393, opened 2018-05-22
chore: add missing documentation 'since' tags for 2.0.0 release

maxokorokov:


maxokorokov:
No idea what it was. I think I've seen it a couple of times randomly on IE

PR #2410, opened 2018-05-25
chore(demo): minor tweaks to latest UI revamp

benouat:
This PR would land a few fixes that previous PR did not addressed:

- The logo (both the stacked one, and the small header one) now has the 2 blue colours distinction image
image

- Both left menu and right end menu have same font size and spacings
image
image

- Component page now have a proper tab bar instead of isolated buttons
image


demo/src/style/app.scss
```diff
@@ -5,8 +5,9 @@ header.navbar {
 }
 
 .masthead {
-  padding-top: 5rem;
-  padding-bottom: 5rem;
+  a {
+    flex: 1;
```
maxokorokov: you should probably add `-ms` prefixes, as there is no autoprefixer currently

benouat: fixed by removing it entirely. It was messing with button sizes on mobile display under IE

benouat:
Fixes all done

PR #2416, opened 2018-05-28
feat(typeahead): Add a closeOnSelect property to typeahead

### LINKED ISSUE #2397: Is It possible to prevent popup from closing on select?
### Bug description:
I know there is an issue regarding a lot of items and that library wants to go way of google and amazon to show only 10 items.
In this case i just want to give user at least option to see that there are more options and when he presses it it will load full list or add 10 more items to list. Im adding specific option load 10 more to the list.
The problem is i cant get typeahead from not closing when user selects specific option. 
Is It possible to prevent popup from closing on select?

Angular: 6.02

ng-bootstrap: 2.0.0

Bootstrap:

samechikson:
Used to determine whether to close the dropdown after selection.

Resolves #2397.

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -1,4 +1,4 @@
-import {Component, Input, Output, EventEmitter, TemplateRef, OnInit} from '@angular/core';
+import {Component, Input, Output, EventEmitter, TemplateRef, OnInit, ElementRef} from '@angular/core';
 
 import {toString} from '../util/util';
 
@@ -79,9 +79,13 @@ export class NgbTypeaheadWindow implements OnInit {
 
   @Output('activeChange') activeChangeEvent = new EventEmitter();
 
+  constructor(private _elementRef: ElementRef) {}
+
   hasActive() { return this.activeIdx > -1 && this.activeIdx < this.results.length; }
 
   getActive() { return this.results[this.activeIdx]; }
+
+  isEventFrom($event) { return this._elementRef.nativeElement.contains($event.target); }
 
   markActive(activeIdx: number) {
     this.activeIdx = activeIdx;
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -113,6 +113,11 @@ export class NgbTypeahead implements ControlValueAccessor,
   @Input() focusFirst: boolean;
 
   /**
+   * A flag indication whether to close the popup after selecting an item
+   */
+  @Input() closeOnSelect: boolean;
+
+  /**
    * A function to convert a given value into string to display in the input field
    */
   @Input() inputFormatter: (value: any) => string;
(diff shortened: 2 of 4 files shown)


src/typeahead/typeahead.ts
```diff
@@ -112,6 +112,11 @@ export class NgbTypeahead implements ControlValueAccessor,
    */
   @Input() focusFirst: boolean;
 
+  /**
+   * A flag indication whether to close the popup after selecting an item
+   */
+  @Input() closeOnSelect: boolean;
```
maxokorokov: I think it should be named `autoClose: boolean` similar to dropdown and datepicker

src/typeahead/typeahead-window.ts
```diff
@@ -79,10 +79,14 @@ export class NgbTypeaheadWindow implements OnInit {
 
   @Output('activeChange') activeChangeEvent = new EventEmitter();
 
+  constructor(private _elementRef: ElementRef) {}
+
   hasActive() { return this.activeIdx > -1 && this.activeIdx < this.results.length; }
 
   getActive() { return this.results[this.activeIdx]; }
 
+  isEventFrom($event) { return this._elementRef.nativeElement.contains($event.target); }
```
maxokorokov: should be `event`, not `$event`

src/typeahead/typeahead.ts
```diff
@@ -373,4 +383,6 @@ export class NgbTypeahead implements ControlValueAccessor,
     }
     this._subscription = null;
   }
+
+  private _isEventFromMenu($event) { return this._windowRef ? !this._windowRef.instance.isEventFrom($event) : false; }
```
maxokorokov: what is `Menu`? maybe `_isEventFromWindow`?

pkozlowski-opensource:
@samechikson I wonder why the [NgbTypeaheadSelectItemEvent](https://ng-bootstrap.github.io/#/components/typeahead/api#NgbTypeaheadSelectItemEvent) is not working for you?

You can call `preventDefault()` on item selection and thus keep the popup open. Could you elaborate more on your use case? We are not taking API extensions lightly so I would like to make sure that we are giving people best API to solve their problems.

samechikson:
Sorry, haven't had time to look at this again. You're probably right @pkozlowski-opensource , `preventDefault()` would have probably worked fine. Just didn't think of it at the time. 

Our use case was simple: to keep the dropdown open when clicking on an option so that a user could select multiple options per search. I think the use case is valid, but sounds like it would be easily resolved with the solution above. 

Thanks for taking the time to review the PR!

src/typeahead/typeahead.ts
samechikson: Yes, realizing this option should have been named the same way it is in the dropdown component.

sajanipatel:
Using preventDefault() doesn't solve the issue. It would only clear the text inside the input field. Is there any better solution to avoid closing the popup after selection?

nickroberts:
So it looks like we still don't have a way to actually prevent the selection from happening, as `$event.preventDefault()` does not stop the dialog from closing.

Has anyone figured out a solution / workaround?

gustavobap:
found a workaround

fsomme2s:
Need this too,

PR #2421, opened 2018-05-29
refactor(datepicker): move 'outsideDays' generation to the service

maxokorokov:
The last in series of moving logic from components to the service and utility functions


src/datepicker/datepicker-month-view.ts
```diff
@@ -48,14 +48,14 @@ import {DayTemplateContext} from './datepicker-day-template-context';
       </div>
     </div>
     <ng-template ngFor let-week [ngForOf]="month.weeks">
-      <div *ngIf="!isCollapsed(week)" class="ngb-dp-week" role="row">
```
benouat: Why did you remove the `role="row"` here ?

maxokorokov: Unfortunate rebase, thanks for the catch!

PR #2436, opened 2018-06-05
feat(carousel): allow control mouse event

DheerajVislavath:
This commit resolves issue #1163 , which allows the user to enable or disable the mouse events. Please review the pull requests and let me know if everything is good to be merged into. Thanks

### DIFF
--- a/src/carousel/carousel.spec.ts
+++ b/src/carousel/carousel.spec.ts
@@ -41,6 +41,7 @@ describe('ngb-carousel', () => {
     expect(carousel.interval).toBe(defaultConfig.interval);
     expect(carousel.wrap).toBe(defaultConfig.wrap);
     expect(carousel.keyboard).toBe(defaultConfig.keyboard);
+    expect(carousel.mouse).toBe(defaultConfig.mouse);
   });
 
   it('should render slides and navigation indicators', fakeAsync(() => {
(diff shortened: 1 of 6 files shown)


DheerajVislavath:
Please review  code and merge the changes. This will be an important feature.

DheerajVislavath:
@benouat Thank you for your suggestions.
But , Carousel already has pause() and cycle() methods to set the interval values to 0 or give default interval.  Why not make use of those methods ?

benouat:
> Carousel already has pause() and cycle() methods to set the interval values to 0 or given default interval. Why not make use of those methods ?

You're right !! I did not see these ones.

Then it should be even easier. You just need to introduce the `pauseOnHover` as an input &amp; config

DheerajVislavath:
I implemented your suggested change. Please review and process to merge the code changes. 
Thank you.

src/carousel/carousel.spec.ts
```diff
@@ -321,9 +322,53 @@ describe('ngb-carousel', () => {
...
+       tick(6000);
+       fixture.detectChanges();
+       expectActiveSlides(fixture.nativeElement, [true, false]);
+       discardPeriodicTasks();
+     }));
+
   it('should pause / resume slide change with time passage on mouse enter / leave', fakeAsync(() => {
        const html = `
-      <ngb-carousel>
+      <ngb-carousel [pauseOnHover]='pauseOnHover'>
```
benouat: I don't see the point about adding that here. It basically duplicate with your added test.
Both `'should listen to mouse events based on pauseOnHover attribute'` and `'should pause / resume slide change with time passage on mouse enter / leave'` are the same.

src/carousel/carousel.ts
```diff
@@ -39,8 +39,8 @@ export class NgbSlide {
     'class': 'carousel slide',
     '[style.display]': '"block"',
     'tabIndex': '0',
-    '(mouseenter)': 'pause()',
-    '(mouseleave)': 'cycle()',
+    '(mouseenter)': 'mouseEventPause()',
```
benouat: I would rename it to `onMouseEnter()` as for now it won't be related to pause all the time.

src/carousel/carousel.ts
```diff
@@ -39,8 +39,8 @@ export class NgbSlide {
...
+    '(mouseleave)': 'mouseEventCycle()',
```
benouat: I would rename it to `onMouseLeave()`

src/carousel/carousel.ts
```diff
@@ -84,6 +84,11 @@ export class NgbCarousel implements AfterContentChecked,
    */
   @Input() keyboard: boolean;
 
+  /**
+   * A flag to enable or disable mouse events
```
benouat: `mouse events` is too generic in here. we only check for mouse enter/leave

I would rephrase it to `Flag to enable pause/resume on mouseover`.

pkozlowski-opensource:
I did changes requested by @benouat, squashed and merged as 8d54cac.

Issue #783, opened 2016-09-22
Datepicker Popup - Close when outside is clicked/touched


k-schneider:
### Bug description:

Datepicker popup does not close when an area outside of the popup/textbox is clicked/touched.  Perhaps an optional true/false setting could be introduced to allow for this functionality.

### Link to minimally-working plunker that reproduces the issue:

[Datepicker popup on examples page](https://ng-bootstrap.github.io/#/components/datepicker)

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 2.0.0 final

ng-bootstrap: 1.0.0-alpha.5

Bootstrap: 4.0.0-alpha.4

silentsod:
I also find this a desirable method of datepicker closing as having it hang open makes for an odd user experience.

davidalee:
Temporary workaround until this feature is added: http://stackoverflow.com/a/35713421/1176878

stephanoapiolaza:
I need that commit https://github.com/pkozlowski-opensource/core/commit/11fdf895a09565616208fd812dd65a7f6dbe813d#diff-89266438c9fe2066d20a7ca712e768bf in alpha 17 
what can i do?

metodribic:
What is the current status of this? As I see there is already commit, waiting to be included in release. Any reasons why is it not already included?

matt-iosh:
@pkozlowski-opensource @k-schneider  is there any update of when this will be committed into the master branch?

smartHasmukh:
@pkozlowski-opensource @k-schneider @matt-iosh 
 
Here is proper solutions of Datepicker hide on click outside

[Plnkr Demo](https://plnkr.co/edit/nRCpS4?p=preview)

k-schneider:
@smartHasmukh Thanks for the example - I'll keep that in my back pocket for now.  Still hoping there is something baked into the library itself that allows for this behavior!

Asharma86:
@pkozlowski-opensource could this be put on the upcoming [Alpha 30] (https://github.com/ng-bootstrap/ng-bootstrap/milestone/33) ?

superMDguy:
Currently, I'm using the [clickOutside](https://github.com/arkon/ng-click-outside) directive to close the datepicker.  It's a simple workaround.

Chris-DF:
@Asharma86 @pkozlowski-opensource 
[...]

PR #2440, opened 2018-06-07
feat(datepicker): close on outside click

maxokorokov:
I would prefer two options for the final API:
- introduce close on outside click as a default non-changeable behaviour
- change the API around the `autoClose` input. Current `autoClose` is for inside date selection and it looks impossible to combine the two behaviours (closing on click and closing on date selection) in one input. Have two separate ones, ex. `closeOnSelect` and `autoClose` 

Looks like both are breaking changes

See #783

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -32,8 +32,8 @@ import {NgbDateAdapter} from './adapters/ngb-date-adapter';
 import {NgbCalendar} from './ngb-calendar';
 import {NgbDatepickerService} from './datepicker-service';
 
-import {Subject, fromEvent} from 'rxjs';
-import {filter, takeUntil} from 'rxjs/operators';
+import {Subject, fromEvent, race} from 'rxjs';
+import {filter, take, takeUntil} from 'rxjs/operators';
 
 const NGB_DATEPICKER_VALUE_ACCESSOR = {
   provide: NG_VALUE_ACCESSOR,
@@ -64,6 +64,7 @@ const NGB_DATEPICKER_VALIDATOR = {
 })
 export class NgbInputDatepicker implements OnChanges,
     OnDestroy, ControlValueAccessor, Validator {
+  private _clickableElements = new Set<HTMLElement>();
   private _closed$ = new Subject();
   private _cRef: ComponentRef<NgbDatepicker> = null;
   private _disabled = false;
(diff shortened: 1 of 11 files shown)


benouat:
I would simply propose to not change anything in `2.1` nor `2.2` and just release a breaking change for `3.0` by introducing the `NgbDatepickerToggle` with `autoclose` working like this:

+ `inside`: close only on date selection
+ `outside`: close only when clicked outside (expect elements decorated with `NgbDatepickerToggle`
+ `true`: close both from inside and outside
+ `false`: don't close anything. One must call explicitly the `close` or `toggle` method

WDYT ?

maxokorokov:
I like the API proposed by @benouat!

The breaking change consists in adding the `[ngbDatepickerToggle]="d"` that references the datepicker instance and the fact that datepicker will now close by default on outside click

Ready for review.

P.S. Will update commit message with `BREAKING CHANGES` a bit later

maxokorokov:
@pkozlowski-opensource, could you PTAL?

We could make it less destructive by defaulting `[autoClose]` to `'inside'` (close only on date selection, behaviour as today). So people who didn't touch it would not be broken. 

But I think I would rather default behaviour to close on outside click and make people update the code.

---

Also as discussed alternative API options are:

```html
<!-- 1  (current) -->
<input ngbDatepicker #d="ngbDatepicker">
<button [ngbDatepickerToggle]="d" (click)="d.toggle()">

<!-- 2 -->
<input ngbDatepicker #d="ngbDatepicker" [toggle]="[b]">
<button #b (click)="d.toggle()">

<!-- 3 -->
<input ngbDatepicker #d="ngbDatepicker">
<button #b (click)="d.toggle([b])"> <!-- same for open here -->
```

src/datepicker/datepicker-input.ts
```diff
@@ -276,9 +280,30 @@ export class NgbInputDatepicker implements OnChanges,
       this._cRef.instance.focus();
 
       // closing on ESC
-      fromEvent<KeyboardEvent>(this._document, 'keyup')
-          .pipe(takeUntil(this._closed$), filter(e => e.which === Key.Escape))
-          .subscribe(() => this.close());
+      // clang-format off
```
pkozlowski-opensource: Please, no `clang-format off`...

benouat: @pkozlowski-opensource you don't want to maintain observables formatted with clang

pkozlowski-opensource: @benouat I don't want to maintain _any_ formatting. If we go down this path we should remove clang altogether. 

We either use a tool that works or we remove it. IMO being half-way is a worst place to be....

pkozlowski-opensource:
LGTM

PR #2448, opened 2018-06-11
fix(demo): carousel displays images in full width on large screens

benouat:
With the new demo layout, Carousel demo looks a bit broken on very large displays as images which are fetch have an hardcoded width of 900px

PR makes them fit the all space


benouat:
@maxokorokov  build would need to be restarted, Edge timed-out on Travis.

PR #2458, opened 2018-06-15
fix(live): use a single DOM element for Live Announcer

maxokorokov:
An attempt to lazily create and reuse `aria-live` element for the live announcer

- test execution order doesn't matter anymore
- multiple ng-bootstrap modules will use a single DOM element
- `ARIA_LIVE_DELAY_TYPE` → `any` for SSR

cc @ymeine


pkozlowski-opensource:
LGTM

PR #2460, opened 2018-06-15
fix(timepicker): width issues

### LINKED ISSUE #1578: TimePicker iOS placeholder is being cut off.
### Bug description:
In iPad Air 2 (10.3.1) - Safari and iPhone 6 (10.3.1) - Safari, the placeholder text is being cut off.
Might be solved by either making a new input to adjust the size of the input box or just setting the size to 3.
https://ng-bootstrap.github.io/app/components/timepicker/demos/basic/plnkr.html
Test this in the devices mentioned above.

Angular:
4.0.0
ng-bootstrap:
1.0.0-alpha.26
Bootstrap:
4.0.0 alpha-6

fbasso:
This is an attempt ti fix #1578.

The main idea is to avoid the sizing of the input with the size attribute. This one is calculated by the browser, and can create inconstancy of display between them.

The main ideas here are:

* let the input having a width 100% (done by the form-control class) and size its parent,
* follow the pattern : rem for widget, em for local (see https://css-tricks.com/rem-global-em-local/) :
  * set the font-size widget to 1rem, so that inner em 
  * size the input parent with em,

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -17,11 +17,20 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
 @Component({
   selector: 'ngb-timepicker',
   styles: [`
+
+    :host {
+      font-size: 1rem;
+    }
+
     .ngb-tp {
       display: -ms-flexbox;
       display: flex;
       -ms-flex-align: center;
       align-items: center;
+    }
+
+    .ngb-tp-input-container {
+      width: 4em;
     }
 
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
@benouat would you mind reviewing this one? Thnx!

benouat:
@pkozlowski-opensource code LGTM, but we definitely need to test it on live devices. Easiest way would be to build PR branch and deploy it somewhere on github pages.

@fbasso Could you have a look at my comment above, so we could test it asap ?

benouat:
LGTM. Can be merged

PR #2462, opened 2018-06-18
fix(datepicker): ExpressionChangedAfterItHasBeenCheckedError

### LINKED ISSUE #2408: DatePicker Error: ExpressionChangedAfterItHasBeenCheckedError
I am facing the following console error which appears when I am trying to go to next/prev months with a help of keyboard arrows.
ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'focused: true'. Current value: 'focused: false'.
at viewDebugError (core.js:9775)
at expressionChangedAfterItHasBeenCheckedError (core.js:9753)
at checkBindingNoChanges (core.js:9920)
at checkNoChangesNodeInline (core.js:13974)
at checkNoChangesNode (core.js:13942)
at debugCheckNoChangesNode (core.js:14771)
at debugCheckDirectivesFn (core.js:14673)
at Object.eval [as updateDirectives] (NgbDatepicker.html:3)

Bootstrap version 4, Angular 5. ng-bootstrap - 2
plunker - http://plnkr.co/edit/yzCMpNMb15oOQf5tJCn6

Do you have any assumptions why this is happening and whether it can be fixed?

divdavem:
Reuses data structures in the datepicker data model instead of building new objects each time the user navigates to a different month. As a consequence, DOM elements are no longer destroyed when changing month.

This fixes the following issue which occurred when BrowserAnimationsModule was used: when using the keyboard to navigate from one month to another, the focused element was destroyed during the change detection of the keyboard event, which synchronously triggered the focusout event in Chrome, leading to a change in the data model (and that change is still part of the data model only when BrowserAnimationsModule is used, because that module changes the behavior of Angular to not destroy views and their data model as quickly as when this module is not loaded), and this was causing the ExpressionChangedAfterItHasBeenCheckedError error.

Fixes #2408

### DIFF
--- a/src/datepicker/datepicker-tools.ts
+++ b/src/datepicker/datepicker-tools.ts
@@ -87,37 +87,61 @@ export function buildMonths(
     calendar: NgbCalendar, date: NgbDate, state: DatepickerViewModel, i18n: NgbDatepickerI18n,
     force: boolean): MonthViewModel[] {
   const {displayMonths, months} = state;
-  const newMonths = [];
-  for (let i = 0; i < displayMonths; i++) {
-    const newDate = calendar.getNext(date, 'm', i);
-    const index = months.findIndex(month => month.firstDate.equals(newDate));
+  // move old months to a temporary array
+  const monthsToReuse = months.splice(0, months.length);
 
-    if (force || index === -1) {
-      newMonths.push(buildMonth(calendar, newDate, state, i18n));
-    } else {
-      newMonths.push(months[index]);
+  // generate new first dates, nullify or reuse months
+  const firstDates = Array.from({length: displayMonths}, (_, i) => {
+    const firstDate = calendar.getNext(date, 'm', i);
+    months[i] = null;
+
+    if (!force) {
+      const reusedIndex = monthsToReuse.findIndex(month => month.firstDate.equals(firstDate));
+      // move reused month back to months
+      if (reusedIndex !== -1) {
+        months[i] = monthsToReuse.splice(reusedIndex, 1)[0];
+      }
     }
-  }
 
-  return newMonths;
+    return firstDate;
+  });
+
+  // rebuild nullified months
+  firstDates.forEach((firstDate, i) => {
+    if (months[i] === null) {
+      months[i] = buildMonth(calendar, firstDate, state, i18n, monthsToReuse.shift() || {} as MonthViewModel);
+    }
+  });
+
+  return months;
 }
 
 export function buildMonth(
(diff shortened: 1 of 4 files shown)


src/datepicker/datepicker-tools.ts
```diff
@@ -87,37 +87,68 @@ export function buildMonths(
     calendar: NgbCalendar, date: NgbDate, state: DatepickerViewModel, i18n: NgbDatepickerI18n,
     force: boolean): MonthViewModel[] {
   const {displayMonths, months} = state;
-  const newMonths = [];
+  const newFirstMonthStart = calendar.getNext(date, 'm', 0);
```
maxokorokov: What about the following implementation?
Not sure it's more/less optimised, but it is shorter and puts less strain on my brain :)
(all your tests pass as before)

```ts
  // move old months to a temporary array
  const monthsToReuse = months.splice(0, months.length);

  // generate new first dates, nullify or reuse months
  const firstDates = Array.from({length: displayMonths}, (_, i) => {
    const firstDate = calendar.getNext(date, 'm', i);
    months[i] = null;

    if (!force) {
      const reusedIndex = monthsToReuse.findIndex(month => month.firstDate.equals(firstDate));
      // move reused month back to months
      if (reusedIndex !== -1) {
        months[i] = monthsToReuse.splice(reusedIndex, 1)[0];
      }
    }

    return firstDate;
  });

  // rebuild nullified months
  firstDates.forEach((newDate, i) => {
    if (months[i] === null) {
      months[i] = buildMonth(calendar, newDate, state, i18n, monthsToReuse.shift() || {} as MonthViewModel);
    }
  });
```

divdavem:
@maxokorokov Thank you for your review.
To solve the issue of the 1000-line exception message, maybe I can write a [https://jasmine.github.io/tutorials/custom_matcher](custom matcher) instead of my `expectSameMonthsDataStructure` function that calls `expect` many times.
I agree your `buildMonths ` function is more readable than mine (and probably a bit less efficient with the `findIndex` that is repeated for each month, but the impact is probably negligible, as the number of months should never be very high, probably less than 5 in most cases). I will replace mine by yours.

divdavem:
@maxokorokov I have updated the code to take your remarks into account. Here is the diff: https://github.com/divdavem/ng-bootstrap/commit/97124b8b71468f2bebac2f15408db7e2947b7ca0

divdavem:
@maxokorokov Thank you!

PR #2483, opened 2018-06-28
chore: deprecate '.forRoot()' in all modules

maxokorokov:
Alternative to #2363
Contains only deprecations and documentation updates
We should deprecate it in 2.2.0 and remove in 3.0.0

We should also keep #2363 to only actually remove `.forRoot()` and  update BREAKING CHANGES.
I think I even would have opened a separate PR to update our own code (tests + demo)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,31 +47,28 @@ npm install --save @ng-bootstrap/ng-bootstrap
 Once installed you need to import our main module:
 ```js
 import {NgbModule} from '@ng-bootstrap/ng-bootstrap';
-```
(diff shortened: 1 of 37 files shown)


maxokorokov:
Tests are failing in IE10
Looks like it's blocked by https://github.com/angular/angular/issues/24339

maxokorokov:
Added a `breaking change`, as we need `6.1` for tests to pass

maxokorokov:
Done. And it's not a breaking change actually... my bad... another PR with `6.1` was

pkozlowski-opensource:
Merged, thnx!

PR #2485, opened 2018-06-29
demo(timepicker): don't use ngModel and FormControl simultaneously

maxokorokov:
Fixes the deprecation in Angular 6:

```
    It looks like you're using ngModel on the same form field as formControl. 
    Support for using the ngModel input property and ngModelChange event with 
    reactive form directives has been deprecated in Angular v6 and will be removed 
    in Angular v7.
    
    For more information on this, see our API docs here:
    https://angular.io/api/forms/FormControlDirective#use-with-ngmodel
```


pkozlowski-opensource:
LGTM! Will merge as soon as TravisCI is green.

PR #2492, opened 2018-07-02
demo(datepicker): add sections in the datepicker overview

maxokorokov:
Sections like for examples and API:

screen shot 2018-07-02 at 15 43 20


benouat:
I have an idea here for the sections navigation sidebar (meanwhile we still need to refactor it all, but it could be the transitional period) :

What about we simply just set a convention that whatever the documentation page, headers are always `<h2>` with inside `<a>` (which is already the case), and then on `NavigationSuccess` in the sidebar component we simply `<pageContent>.querySelectorAll('h2')` and just display them.

It would be much more easier, no data model involved, no inputs to be cascaded.

@pkozlowski-opensource @maxokorokov WDYT ?

pkozlowski-opensource:
@benouat `jQuery` FTW :-)

benouat:
> @benouat jQuery FTW :-)

true...

But I would say _**just using the platform**_ instead of _**jQuery**_ because after all this is what it is.

We have been biased through the years by saying DOM manipulation === jQuery. DOM manipulation is a standard primitive provided by the platform. We should kill this _jQuery_ wording once and for all 😄

maxokorokov:
Will merge this one and @benouat will have a way to simplify demo structure in a separate PR

PR #2497, opened 2018-07-05
chore: extract and display generic type info in the API

maxokorokov:
Allow extracting things like:

```typescript
class NgbDatepicker<T = NgbDateStruct> {
}

interface NgbDateAdapter<T> {
}
```

To be displayed like this:

screen shot 2018-07-05 at 12 22 09

screen shot 2018-07-05 at 12 22 23


pkozlowski-opensource:
LGTM

PR #2502, opened 2018-07-07
chore(demo): fix social buttons alignment

maxokorokov:
Before (current live site):

screen shot 2018-07-07 at 14 16 54

After:

screen shot 2018-07-07 at 14 16 28

cc @benouat

### DIFF
--- a/demo/src/app/app.component.html
+++ b/demo/src/app/app.component.html
@@ -30,13 +30,11 @@
     </ul>
   </div>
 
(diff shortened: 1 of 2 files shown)


benouat:
> not sure why you've added this to styles before:

I don't have any clue at all....

maxokorokov:
Just in case checked with IE 11, seems ok too

PR #2506, opened 2018-07-09
chore: fix umd bundle name for plunker

maxokorokov:
After the migration to CLI, bundle name has changed to from `bundles/ng-bootstrap.js` to `bundles/ng-bootstrap.umd.js`


pkozlowski-opensource:
LGTM

PR #2507, opened 2018-07-09
chore(test): add util to create fake keyboard events

ymeine:
closes #2050

### DIFF
--- a/src/test/common.ts
+++ b/src/test/common.ts
@@ -1,4 +1,7 @@
 import {TestBed, ComponentFixture} from '@angular/core/testing';
+import {Key} from '../util/key';
+
+
 
 export function createGenericTestComponent<T>(html: string, type: {new (...args: any[]): T}): ComponentFixture<T> {
   TestBed.overrideComponent(type, {set: {template: html}});
@@ -58,4 +61,13 @@ export function isBrowser(browsers: Browser | Browser[], ua = window.navigator.u
   } else {
     return browsersStr.indexOf(browser) > -1;
   }
+}
+
+export function createKeyEvent(key: Key, options: {type: 'keyup' | 'keydown'} = {type: 'keyup'}) {
+  const event = document.createEvent('KeyboardEvent') as any;
+  let initEvent = (event.initKeyEvent || event.initKeyboardEvent).bind(event);
+  initEvent(options.type, true, true, window, 0, 0, 0, 0, 0, key);
+  Object.defineProperties(event, {which: {get: () => key}});
+
+  return event;
 }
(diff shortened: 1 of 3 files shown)


ymeine:
The thing is I don't know what's planned, so in this case I usually prefer using an object to have more flexibility.

However: 

- that's true that most of the time the key only might be provided
- it's testing code, so no problem to change the API later on

In the current state, we could indeed have the key as first parameter, and maybe an object as second (which for now would accept only `type`). What do you think?

benouat:
I am fine with your proposal, `key` as the first argument and second as options

src/test/common.ts
```diff
@@ -59,3 +63,21 @@ export function isBrowser(browsers: Browser | Browser[], ua = window.navigator.u
     return browsersStr.indexOf(browser) > -1;
   }
 }
+
+export function createKeyEvent(key: Key, options?: {type?: 'keyup' | 'keydown'}) {
```
pkozlowski-opensource: Couldn't we write it as `createKeyEvent(key: Key, options?: {type?: 'keyup' | 'keydown'} = {})` so you could drop subsequent lines:

```
if (!isDefined(options)) {
  options = {};
}
```

pkozlowski-opensource: Or better yet: `reateKeyEvent(key: Key, options?: {type?: 'keyup' | 'keydown'} = {type: 'keyup'})`

so we can drop more:

```
let {type} = options;
if (!isDefined(type)) {
  type = 'keyup';
}
```

ymeine: I'm used to write this kind of code - despite the fact it's indeed more verbose - because it tended to be more robust in the general case (for specific cases, we can always adapt).

First, the potential advantage of using `isDefined` instead of the default parameter assignment is to be able to handle not only `undefined` values but also `null` ones.

Then, why not `createKeyEvent(key: Key, options?: {type?: 'keyup' | 'keydown'} = {type: 'keyup'})`, it's because if an empty `options` object is given, we won't have any value for `options.type` anymore.

In both cases we could say that a user shouldn't pass a `null` value or an empty object for `options`, and this would be a contract. That's why I said the code I wrote is more robust, not that it is mandatory in our case (especially that we are dealing with internal, testing code).

pkozlowski-opensource: >  it's because if an empty options object is given, we won't have any value for options.type anymore.

This can be easily fixed with:

```
createKeyEvent(key: Key, options?: {type: 'keyup' | 'keydown'} = {type: 'keyup'})
```

This is shorter,  more explicit and readable, IMO so I would prefer that we use it instead.

pkozlowski-opensource: > In both cases we could say that a user shouldn't pass a null value or an empty object for options, and this would be a contract. 

Agreed. But this is precisely why we should express it in a type - we can explicitly say that you can either omit a params object  or - if you decide to provide one - have it conform to exact specification.

Issue #641, opened 2016-08-31
Modal: content shifts to the right


pkozlowski-opensource:
Same as https://github.com/angular-ui/bootstrap/issues/6125

For the ref it is still a pb in the BS itself...

Martin-Wegner:
Workaround:

`<body style="overflow-y: visible !important;">`

555ea:
@pkozlowski-opensource Any progress on it ? 

It should be fixed by adding padding to the body and the modal, as in bootstrap. Take a look at your examples page https://ng-bootstrap.github.io/#/components/modal/examples and bootstrap modal demo https://getbootstrap.com/docs/4.0/components/modal/

PR #2508, opened 2018-07-09
fix(modal): adjust modal background to avoid shifting

ymeine:
Shifting potentially occurs when a scrollbar present on the container disappears once the modal opens.

The solution is taken from Bootstrap itself.

fixex #641 

Remaining:

- tests

### DIFF
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -12,6 +12,7 @@ import {
 
 import {ContentRef} from '../util/popup';
 import {isDefined, isString} from '../util/util';
+import {Scrollbar} from '../util/scrollbar';
 
 import {NgbModalBackdrop} from './modal-backdrop';
 import {NgbModalWindow} from './modal-window';
@@ -19,19 +20,19 @@ import {NgbActiveModal, NgbModalRef} from './modal-ref';
 
 @Injectable()
 export class NgbModalStack {
-  private _document: any;
   private _windowAttributes = ['ariaLabelledBy', 'backdrop', 'centered', 'keyboard', 'size', 'windowClass'];
   private _backdropAttributes = ['backdropClass'];
 
   constructor(
       private _applicationRef: ApplicationRef, private _injector: Injector,
-      private _componentFactoryResolver: ComponentFactoryResolver, @Inject(DOCUMENT) document) {
-    this._document = document;
-  }
+      private _componentFactoryResolver: ComponentFactoryResolver, @Inject(DOCUMENT) private _document,
+      private _scrollbar: Scrollbar) {}
 
   open(moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any, options): NgbModalRef {
     const containerEl =
         isDefined(options.container) ? this._document.querySelector(options.container) : this._document.body;
+
+    const revertPaddingForScrollbar = this._scrollbar.compensateScrollbar();
 
     if (!containerEl) {
       throw new Error(`The specified modal container "${options.container || 'body'}" was not found in the DOM.`);
@@ -45,6 +46,7 @@ export class NgbModalStack {
     let windowCmptRef: ComponentRef<NgbModalWindow> = this._attachWindowComponent(containerEl, contentRef);
     let ngbModalRef: NgbModalRef = new NgbModalRef(windowCmptRef, contentRef, backdropCmptRef, options.beforeDismiss);
 
+    ngbModalRef.result.then(revertPaddingForScrollbar, revertPaddingForScrollbar);
     activeModal.close = (result: any) => { ngbModalRef.close(result); };
     activeModal.dismiss = (reason: any) => { ngbModalRef.dismiss(reason); };
 
--- a/src/util/scrollbar.ts
+++ b/src/util/scrollbar.ts
@@ -1,0 +1,47 @@
+import {Injectable, Inject} from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
+
+import {isDefined} from './util';
+
+
+
+@Injectable()
+export class Scrollbar {
+  constructor(@Inject(DOCUMENT) private _document) {}
+
+  compensateScrollbar() {
+    if (!this.isScrollbarPresent()) {
+      return () => {};
+    }
+    const scrollbarWidth = this.getScrollbarWidth();
+    return this.adjustElementForScrollbar(scrollbarWidth);
+  }
+
+  adjustElementForScrollbar(scrollbarWidth: number) {
+    const element = this._document.body;
+    const userSetPadding = element.style.paddingRight;
+    const paddingAmount = parseFloat(window.getComputedStyle(element)['padding-right']);
+    element.style['padding-right'] = `${paddingAmount + scrollbarWidth}px`;
+    return () => element.style['padding-right'] = userSetPadding;
+  }
+
+  isScrollbarPresent(): boolean {
+    const element = this._document.body;
+    const rect = element.getBoundingClientRect();
+    return rect.left + rect.right < window.innerWidth;
+  }
+
+  getScrollbarWidth(): number {
+    const document = this._document;
+    const element = document.body;
+
+    const measurer = document.createElement('div');
+    measurer.className = 'modal-scrollbar-measure';
+
+    element.appendChild(measurer);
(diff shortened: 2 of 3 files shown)


src/util/scrollbar.ts
```diff
@@ -0,0 +1,46 @@
+import {Injectable, Inject} from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
```
maxokorokov: `@angular/common`, this one should be deprecated

src/util/scrollbar.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import {isDefined} from './util';
```
maxokorokov: unused

src/util/scrollbar.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  compensate() {
+    if (!this.isPresent()) {
+      return () => {};
+    }
+    const width = this.getWidth();
+    return this.adjustBody(width);
+  }
+
+  adjustBody(width: number) {
```
maxokorokov: `adjustBody`, `isPresent`, `getWidth` → `private _adjustBody`, `private _isPresent`, `private _getWidth`

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
...
+      return () => {};
+    }
+    const width = this._getWidth();
+    return this._adjustBody(width);
+  }
+
+  private _adjustBody(width: number) {
+    const {body} = this._document;
+    const userSetPadding = body.style.paddingRight;
+    const paddingAmount = parseFloat(window.getComputedStyle(body)['padding-right']);
```
maxokorokov: Hmm, I was just fixing SSR for the carousel and realised that `window` might break it too...

maxokorokov: Regardless, I don't think the fix for this should be a part of this PR

src/modal/modal-stack.ts
```diff
@@ -45,6 +46,7 @@ export class NgbModalStack {
     let windowCmptRef: ComponentRef<NgbModalWindow> = this._attachWindowComponent(containerEl, contentRef);
     let ngbModalRef: NgbModalRef = new NgbModalRef(windowCmptRef, contentRef, backdropCmptRef, options.beforeDismiss);
 
+    ngbModalRef.result.then(revertPaddingForScrollBar, revertPaddingForScrollBar);
```
pkozlowski-opensource: Maybe write as `ngbModalRef.result.finally(revertPaddingForScrollBar)` ?

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
+import {Injectable, Inject} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+
+
+
+@Injectable()
+export class ScrollBar {
+  constructor(@Inject(DOCUMENT) private _document) {}
+
+  compensate() {
```
pkozlowski-opensource: Could you please add explicit return type so it is easier to read? We've got a `return` statement in this function but it is not immediately clear what is returned.

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
+import {Injectable, Inject} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+
+
+
+@Injectable()
+export class ScrollBar {
```
pkozlowski-opensource: Could you add a bit of JS doc to this class with a brief description on how it works, what it can be used for etc.?

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
...
+  compensate() {
+    if (!this._isPresent()) {
+      return () => {};
+    }
+    const width = this._getWidth();
+    return this._adjustBody(width);
+  }
+
+  private _adjustBody(width: number) {
+    const {body} = this._document;
```
pkozlowski-opensource: Wouldn't explicit `const body = this._document.body;` be more readable here?

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
...
+    body.style['padding-right'] = `${paddingAmount + width}px`;
+    return () => body.style['padding-right'] = userSetPadding;
+  }
+
+  private _isPresent(): boolean {
+    const rect = this._document.body.getBoundingClientRect();
+    return rect.left + rect.right < window.innerWidth;
+  }
+
+  private _getWidth(): number {
```
pkozlowski-opensource: This could be called only once, if I'm not mistaken? (I mean, we don't have to call it in each and every `compensate()` call.

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
...
+
+
+
+@Injectable()
+export class ScrollBar {
+  constructor(@Inject(DOCUMENT) private _document) {}
+
+  compensate() {
+    if (!this._isPresent()) {
+      return () => {};
```
pkozlowski-opensource: nit: this is always the same, you could extract it as `const noop = () => {};`

pkozlowski-opensource:
@ymeine thnx for this PR, it fixes a long-lasting bug! 

I've added few comments in the code - I would mostly like to see more docs (or at least types!).

src/util/scrollbar.ts
maxokorokov: Same on line 33, if you want to do something with it. But it's just a question of style to me...

src/util/scrollbar.ts
```diff
@@ -0,0 +1,44 @@
...
+  private _isPresent(): boolean {
+    const rect = this._document.body.getBoundingClientRect();
+    return rect.left + rect.right < window.innerWidth;
+  }
+
+  private _getWidth(): number {
+    const document = this._document;
+    const {body} = document;
+
+    const measurer = document.createElement('div');
```
maxokorokov: By the way just realised that this is a lot more confusing to me. Unclear if it's a global `document` (wrong) or a local one.

(nitpicking, feel free to ignore)

src/modal/modal-stack.ts
pkozlowski-opensource: Forget it, it seems like browser support is still lacking... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

src/util/scrollbar.ts
ymeine: Yes, since I finally made it as a separate utility it makes sense

src/util/scrollbar.ts
ymeine: Sure I will do

src/util/scrollbar.ts
ymeine: Well, destructuring is standard feature, used elsewhere in this project, and avoids repetition. Then I could say it makes more sense when destructuring pure data objects (like parameters). I don't know 😛

src/util/scrollbar.ts
ymeine: Yeah... again that's a matter of style. And here the context is not far to check. I made like that to avoid repetition, but also to write in a more familiar style - which is exactly what is making you confused ^^

src/util/scrollbar.ts
ymeine: I don't know... could there be any situation where the width of the scrollbar would change (with custom styling then) throughout the application lifetime? (and would we support that?)

If not, I can cache the result.

src/util/scrollbar.ts
ymeine: Yes, I'll add it along with documentation about it

src/util/scrollbar.ts
pkozlowski-opensource: Let's leave it as-is for now.

src/util/scrollbar.ts
pkozlowski-opensource: If you ask me I would:
* drop the `document` const (and use this._document everywhere)
* have `const documentBody = this._document.createElement('div');`

maxokorokov: I see the word 'document' 4 times in the snippet above, so I won't agree with the repetition argument :)

vs
```typescript
const body = this._document.body; // or const {body} = this._document;
const measurer = this._document.createElement('div')
```

I can read your version fine too though

PR #2511, opened 2018-07-10
fix(accordion): the markup should follow the boostrap one

fbasso:
Following the bootstrap example: https://getbootstrap.com/docs/4.1/components/collapse/#accordion-example

The card-body is inside the collapse element. It can be useful for animation.

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -123,8 +123,10 @@ export interface NgbPanelChangeEvent {
           </h5>
         </div>
         <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'"
-             class="card-body collapse" [class.show]="panel.isOpen" *ngIf="!destroyOnHide || panel.isOpen">
-             <ng-template [ngTemplateOutlet]="panel.contentTpl?.templateRef"></ng-template>
+             class="collapse" [class.show]="panel.isOpen" *ngIf="!destroyOnHide || panel.isOpen">
+          <div class="card-body">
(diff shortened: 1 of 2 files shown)


src/accordion/accordion.ts
```diff
@@ -123,8 +123,10 @@ export interface NgbPanelChangeEvent {
           </h5>
         </div>
         <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'"
-             class="card-body collapse" [class.show]="panel.isOpen" *ngIf="!destroyOnHide || panel.isOpen">
-             <ng-template [ngTemplateOutlet]="panel.contentTpl?.templateRef"></ng-template>
+             class="collapse" [class.show]="panel.isOpen" *ngIf="!destroyOnHide || panel.isOpen">
+          <div class="card-body">
```
benouat: I think we are also missing `card` on here

From BS [demo site](https://getbootstrap.com/docs/4.1/components/collapse/)

```html
<div class="collapse" id="collapseExample">
  <div class="card card-body">
    Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid. Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident.
  </div>
</div>
```

benouat: Nevermind, I overlooked it. `card` and `card-body` on the same element is for just a single collpase.

pkozlowski-opensource:
@maxokorokov I would consider this a fix. Merged.

PR #2520, opened 2018-07-12
Dynamic generation of widget demos/api pages (Lazy loading postponed)

benouat:
This PR is a WIP.
I just used the alert widget to investigate a new way of managing demos/api pages. Main goal is to assess if it is better than what we have currently.

If answer is yes I will update the PR to migrate all demos at once. One PR per widget would be too much I think.

---

All routes configuration for a given widget are now defined at widget module level.
First benefit of it is that lazy loading comes for 🆓 🎊 🍾 !

I also created a few new versions of existing components (that will replace old ones as soon as we will merge that PR)
+ `ComponentWrapper` (this one will also handle dynamic table of content)
+ `ExampleBox`

And also introduced 2 new generic ones that display both examples and api pages
+ `NgbdExamplesPage`
+ `NgbdApiPage`


demo/src/app/components/shared/api-page/api.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+
+@Component({
+  template: `
+    <ngbd-api-docs *ngFor="let component of components" [directive]="component"></ngbd-api-docs>
+    <ngbd-api-docs-class *ngFor="let klass of classes" [type]="klass"></ngbd-api-docs-class>
+    <ngbd-api-docs-config *ngFor="let config of configs" [type]="config"></ngbd-api-docs-config>
+  `
+})
+export class NgbdApiPage {
```
maxokorokov: Maybe add `OnPush` to all components?

benouat:
Things I have done: 

+ Updated regarding @maxokorokov comments
+ Run a pass of Prettier on all files
+ Activated OnPush as much as possible

---

> extract and ship in api-docs.ts only classes that are actually should be documented

I also did that. A class with no top jsdoc is not taken into account (considered as private). What we should do in a separate PR is to manage files that might have top jsdoc but decorated @internal (useful in IDE, but do not want to see it on demo site)

--- 

Lazy loading rocks 🚀 (the weird second line below is a common module between all demos)

```bash
yarn demo
````
image

```bash
yarn build demo --prod
```
image

demo/src/app/app.routing.ts
```diff
@@ -1,135 +1,94 @@
...
+import { DefaultComponent } from './default';
+import { GettingStarted } from './getting-started';
 
-const componentRoutes = [{
+const DEFAULT_API_PATH = {
+  path: '',
+  pathMatch: 'full',
+  redirectTo: 'examples'
+};
+const DEFAULT_API_PATH_OVERVIEW = {
```
maxokorokov: I don't think you need this and the one above anymore

demo/src/app/app.component.ts
```diff
@@ -12,21 +15,29 @@ export class AppComponent implements OnInit {
...
+    router.events
+      .pipe(filter(event => event instanceof NavigationEnd))
+      .subscribe(event => {
+        const { fragment } = router.parseUrl(router.url);
         if (fragment) {
-          const element = document.querySelector(`#${fragment}`);
-          if (element) {
-            element.scrollIntoView();
-          }
+          setTimeout(() => {
```
maxokorokov: this is going to be fixed separately with router changes from `6.1` later, right ?

demo/src/app/shared/component-wrapper/component.component.ts
```diff
@@ -0,0 +1,88 @@
+import { Component, NgZone } from '@angular/core';
+import { ActivatedRoute, NavigationEnd, Router } from '@angular/router';
+import { filter } from 'rxjs/operators';
+
+import { NgbdApiPage } from '../../components/shared/api-page/api.component';
+import { NgbdExamplesPage } from '../../components/shared/examples-page/examples.component';
+
+
+@Component({
+  selector: 'component-wrapper',
```
maxokorokov: why is current file called `component.component.ts`, but the selector is `component-wrapper`?

demo/src/app/app.component.ts
benouat: Yes

benouat:
@maxokorokov thanks for the review. I have updated your minor comments.

pkozlowski-opensource:
@benouat this needs rebase, could you PTAL?

benouat:
The PR has been rebased, I added a few commits. It can be squashed and merge.

pkozlowski-opensource:
LGTM, thnx! Going to squash and merge it.

PR #2525, opened 2018-07-16
chore: add application for e2e tests

maxokorokov:
This is meant to be the PR to set up tooling for writing e2e tests, not for tests themselves!

Adding:
- new application `e2e-app` with test cases
- new application `e2e` with actual tests
- an example with two e2e tests (to be changed later)
- CI configuration with headless Chrome

Folder structure (`e2e` is inside `e2e-app`):
```
  /e2e                      ← `e2e-app` root
      /e2e                  ← `e2e` root
          /src              ← specs
          protractor.conf.js
          tsconfig.e2e.json
      /src                  ← test app source code
      tsconfig.json
```
Tests can be accessed via the routing configuration:
```
  /datepicker/test1
  /datepicker/test2
  /modal/test1
  /modal/test2
  ...
```

Any ideas/suggestions are very welcomed!


maxokorokov:
Seems to work fine in CI:

screen shot 2018-07-17 at 11 28 46

maxokorokov:
@benouat @fbasso code review :)

package.json
```diff
@@ -13,6 +13,7 @@
     "build": "yarn ngb:build && yarn demo:build",
     "test": "yarn check-format && yarn ngb:lint && yarn ngb:test",
     "tdd": "yarn ngb:tdd",
+    "e2e": "yarn ngb:e2e",
```
fbasso: For debugging purpose, it would be fine to have a task in order to launch the e2e app only, without running the tests. i.e. "e2e-app"

fbasso: In the same idea of developping/debugging tests, a good idea would be to be able launch the test without the compilation. This way we can have a console with the incremental compilation, and another one where we will launch 'yarn e2e:dev' (without compilation) when needed.

fbasso:
Regarding the folders architecture, and as the purpose at the end is to have two e2e suites (one for the e2e-app, one for the demo) I would prefer something like this : 

- demo
- e2e-app
- e2e
  - demo
  - e2e-app

to have all the tests gathered in one folder, and demo and e2e-app (at the root) being the corresponding applications.

package.json
maxokorokov: Thanks, I've added:
```
"e2e-app:serve": "ng serve e2e-app" → to serve e2e app
"ngb:e2e-noserve": "ng e2e -c noserve" → to only run protractor tests
```

maxokorokov:
Thanks for the review!
Amended:
- renamed `e2e` application to `e2e-app` for more clarity (but still kept structure as it was before)
- added `--prod` flag when running tests
- added separated targets for serving e2e app and running protractor for development
- updated developer docs for `yarn e2e`

PR #2527, opened 2018-07-17
chore: use headless chrome for CI

maxokorokov:
An experiment to use headless chrome instead of Firefox for CI

### DIFF
--- a/src/karma.conf.js
+++ b/src/karma.conf.js
@@ -1,7 +1,7 @@
 // Configuration used for local testing and Travis CI
 
 const reporters = process.env.TRAVIS ? ['dots'] : ['progress'];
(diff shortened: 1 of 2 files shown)


maxokorokov:
It seems to work and at least twice as fast (~40sec for Chrome vs ~1:40 for Firefox) and it will be better for e2e tests

src/karma.conf.js
```diff
@@ -19,6 +19,12 @@ module.exports = function (config) {
       reports: ['html', 'json', 'lcovonly'],
       fixWebpackSourcePaths: true
     },
+    customLaunchers: {
+      ChromeHeadlessNoSandbox: {
```
maxokorokov: See https://docs.travis-ci.com/user/chrome#Sandboxing for explanation

PR #2537, opened 2018-07-19
demo(datepicker): update demos to use 'NgbDate'

maxokorokov:
Demos for #2536


pkozlowski-opensource:
@maxokorokov nice, those demos are significantly simpler now!  Could you please rebase on top of the latest master so we've got green CI build and merge it?

PR #2542, opened 2018-07-23
refactor(alert): add custom class to alert component

ulrichdohou:


maxokorokov:
Hey, @ulrichdohou!

Thanks for the PR, however could you please provide a use case for this?

I think you can achieve exactly the same by doing:

```html
<ngb-alert class="myclass">
```

```css
ngb-alert.myclass > div { ... }
```

ulrichdohou:
Hi @maxokorokov 

I was trying to solve the issue related here on stackoverflow https://stackoverflow.com/questions/51450796/apply-classes-to-sub-elements-of-components/51451075?noredirect=1#comment89915498_51451075

<ngb-alert class="myclass"> this way would add the class to the ngb-alert tag and thats not what I was looking for
But this way ngb-alert.myclass > div { ... } should work. But I think adding that option would be fine.

pkozlowski-opensource:
@ulrichdohou thnx for the PR but I believe that we should do #2544 instead.

pkozlowski-opensource:
@ulrichdohou thnx for the PR but we are going to focus on  #2544 and #2552 instead.

PR #2548, opened 2018-07-26
chore: bump minimal required versions of Angular and Bootstrap

maxokorokov:
BREAKING CHANGE: for ng-bootstrap 3.0.0 minimal required versions are Angular 6.1.0 and Bootstrap 4.1.0


maxokorokov:
required for #2483

PR #2551, opened 2018-07-26
fix(demo): fix closing of popover & tooltip using popup context

ymeine:
Change the design of this demo to make it more consistent and fix the issue at the same time.

### DIFF
--- a/demo/src/app/components/popover/demos/tplwithcontext/popover-tplwithcontext.html
+++ b/demo/src/app/components/popover/demos/tplwithcontext/popover-tplwithcontext.html
@@ -3,15 +3,24 @@
 </p>
 
 <ng-template #popContent let-greeting="greeting">{{greeting}}, <b>{{name}}</b>!</ng-template>
-<p>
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
> and fix the issue at the same time.

Could you please amend the commit message (or at least PR description) to describe what issue is fixed? It is not immediately obvious to me...

pkozlowski-opensource:
OK, so after having another look at those demos / code I think that we are demonstrating 2 things here:
* passing context to a template
* swapping context in a template

But the truth is that we can't really swap context as it requires closing / re-opening of a tooltip / popover. We can get away with this today but it just won't work if we add delay to tooltips.

So I think that we shouldn't pretend that we can swap context and instead I would simply open 3 different tooltips / popovers (one for each greeting). This way we could remove all the code and just have a template.

How does it sound?

ymeine:
You're right, swapping is fake and won't work with the new feature as you said. I agree for keeping the demo to show passing context, but not swapping it.

Will change the commit message / PR description as you asked and will especially reflect the changes implied by your suggestion.

pkozlowski-opensource:
Thnx @ymeine !

PR #2552, opened 2018-07-27
refactor(alert): remove intermediate 'div' element

### LINKED ISSUE #2544: alert: remove superfluous <div> tag
As of today `<ngb-alert>` doesn't set Bootstrap classes on a host but [adds a `<div>` inside](https://github.com/ng-bootstrap/ng-bootstrap/blob/e0c005082f11f811ecc6e723e27ec029eb29770a/src/alert/alert.ts#L18). This is an issue for animations and prevents people from properly styling the alert's host (see https://stackoverflow.com/questions/51450796/apply-classes-to-sub-elements-of-components/51451075 and #2542 as an example). 

We should remove [this `<div>`](https://github.com/ng-bootstrap/ng-bootstrap/blob/e0c005082f11f811ecc6e723e27ec029eb29770a/src/alert/alert.ts#L18) and set classes (and `role="alert"` directly on the host).

maxokorokov:
Fixes #2544

### DIFF
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -4,6 +4,11 @@ import {
   Output,
   EventEmitter,
   ChangeDetectionStrategy,
+  Renderer2,
+  ElementRef,
+  OnChanges,
+  OnInit,
+  SimpleChanges
 } from '@angular/core';
 
 import {NgbAlertConfig} from './alert-config';
@@ -14,17 +19,22 @@ import {NgbAlertConfig} from './alert-config';
 @Component({
   selector: 'ngb-alert',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {'role': 'alert', 'class': 'alert', '[class.alert-dismissible]': 'dismissible'},
   template: `
-    <div [class]="'alert alert-' + type + (dismissible ? ' alert-dismissible' : '')" role="alert">
-      <button *ngIf="dismissible" type="button" class="close" aria-label="Close" i18n-aria-label="@@ngb.alert.close"
-        (click)="closeHandler()">
-        <span aria-hidden="true">&times;</span>
-      </button>
-      <ng-content></ng-content>
-    </div>
-    `
+    <button *ngIf="dismissible" type="button" class="close" aria-label="Close" i18n-aria-label="@@ngb.alert.close"
+      (click)="closeHandler()">
+      <span aria-hidden="true">&times;</span>
+    </button>
+    <ng-content></ng-content>
+    `,
+  styles: [`
+    :host {
+      display: block;
+    }
+  `]
 })
-export class NgbAlert {
+export class NgbAlert implements OnInit,
+    OnChanges {
   /**
    * A flag indicating if a given alert can be dismissed (closed) by a user. If this flag is set, a close button (in a
    * form of an ×) will be displayed.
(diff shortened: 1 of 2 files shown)


src/alert/alert.ts
```diff
@@ -38,12 +43,22 @@ export class NgbAlert {
...
     this.dismissible = config.dismissible;
     this.type = config.type;
   }
 
   closeHandler() { this.close.emit(null); }
+
+  ngOnChanges(changes: SimpleChanges) {
+    const {type} = changes;
+    if (type && !type.firstChange) {
+      this._renderer.removeClass(this._element.nativeElement, `alert-${type.previousValue}`);
```
pkozlowski-opensource: Why do we do this through low-level DOM manipulation? Couldn't we move the existing bindings to the host?
Or is binding overriding classes set by a user on a host?

maxokorokov: Double-checked, and it overrides user-set classes. So I see three options:

1. Leave as it is done currently with `Renderer2`
2. hardcode all supported classes and remove the demo with `alert-customclass`:

```
  host: {
    'role': 'alert',
    'class': 'alert d-block',
    '[class.alert-dismissible]': 'dismissible',
    '[class.alert-success]': `type === 'success'`,
    '[class.alert-warning]': `type === 'warning'`,
    '[class.alert-info]': `type === 'info'`,
    '[class.alert-danger]': `type === 'danger'`,
    '[class.alert-primary]': `type === 'primary'`,
    '[class.alert-secondary]': `type === 'secondary'`,
    '[class.alert-light]': `type === 'light'`,
    '[class.alert-dark]': `type === 'dark'`
  },
```
3. do nothing and close this :)

src/alert/alert.ts
```diff
@@ -38,12 +43,22 @@ export class NgbAlert {
...
-  constructor(config: NgbAlertConfig) {
+  constructor(config: NgbAlertConfig, private _renderer: Renderer2, private _element: ElementRef) {
     this.dismissible = config.dismissible;
     this.type = config.type;
   }
 
   closeHandler() { this.close.emit(null); }
+
+  ngOnChanges(changes: SimpleChanges) {
+    const {type} = changes;
```
pkozlowski-opensource: While it works I find the `type` name confusing / not very descriptive  here. A much better name would be `typeChange`. I would prefer to be explicit here.

src/alert/alert.ts
```diff
@@ -38,12 +43,22 @@ export class NgbAlert {
...
+    if (type && !type.firstChange) {
+      this._renderer.removeClass(this._element.nativeElement, `alert-${type.previousValue}`);
+      this._renderer.addClass(this._element.nativeElement, `alert-${type.currentValue}`);
+    }
+  }
+
+  ngOnInit() { this._renderer.addClass(this._element.nativeElement, `alert-${this.type}`); }
```
pkozlowski-opensource: Couldn't we move this logic to `ngOnChanges` and set the class on `firstChange`? This way the whole logic would be centralised in one place

pkozlowski-opensource:
@maxokorokov I left few nits but generally speaking I think that we should merge it. It is not an ideal solution but this is the best we've got given current Angular...

src/alert/alert.ts
maxokorokov: Will change

maxokorokov:
@pkozlowski-opensource , fixed comments and:
- added a `BREAKING CHANGE` comment about the markup change
- replaced `d-block` with `:host { display: block; }` as it doesn't set `!important` flag

PR #2555, opened 2018-07-27
fix(accordion): collapsed class not correct

### LINKED ISSUE #2553: NgbAccordion collapsed class is added in open tab instead of collapsed one
### Bug description:

Collapsed class in added in case of open tab instead of closed tab.

See the differences in class added in case of bootstrap 
https://getbootstrap.com/docs/4.1/components/collapse/
and 
ng-bootstrap
https://ng-bootstrap.github.io/#/components/accordion/examples

Last version of ng-bootstrap

I will submit a PR to correct this.

ExFlo:
fixes #2553

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -33,7 +33,11 @@ function expectOpenPanels(nativeEl: HTMLElement, openPanelsDef: boolean[]) {
   expect(panels.length).toBe(openPanelsDef.length);
 
   const panelsTitles = getPanelsTitle(nativeEl);
-  const result = panelsTitles.map((titleEl: HTMLButtonElement) => titleEl.getAttribute('aria-expanded') === 'true');
+  const result = panelsTitles.map((titleEl: HTMLButtonElement) => {
+    const isAriaExpanded = titleEl.getAttribute('aria-expanded') === 'true';
+    const isCSSCollapsed = titleEl.classList.contains('collapsed');
+    return isAriaExpanded === !isCSSCollapsed ? isAriaExpanded : fail('inconsistent state');
+  });
 
   const panelContents = getPanelsContent(nativeEl);
   panelContents.forEach(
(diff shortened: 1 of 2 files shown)


src/accordion/accordion.spec.ts
```diff
@@ -322,6 +322,21 @@ describe('ngb-accordion', () => {
     });
   });
 
+  it('should have the appropriate CSS collapsed classes', () => {
```
pkozlowski-opensource: Couldn't we add assert to https://github.com/ExFlo/ng-boostrap/blob/e1064feca33b001b9cad323edb4f5ffbb0c0373a/src/accordion/accordion.spec.ts#L37 instead of writing whole new test?

The advantage is that it would be verified with all the existing tests.

pkozlowski-opensource: @ExFlo I was more thinking of sth like:

```
panelsTitles.map((titleEl: HTMLButtonElement) => {
   const isAriaExpanded = titleEl.getAttribute('aria-expanded') === 'true';
   const isCSSCollapsed = titleEl.classList.contains('collapsed');
   return isAriaExpanded === !isCSSCollapsed ? isAriaExpanded : fail('inconsistent state');
});
```

not well thought-through, just a rough idea...

ExFlo: So we just test that collapsed is correctly removed, but not that it is in place the first time right?

pkozlowski-opensource: During the first time rendering we should have:
* `isAriaExpanded=false`
* `isCSSCollapsed=true`

Given this the `isAriaExpanded === !isCSSCollapsed` condition should be `true` and the return value `false`.

If the class wouldn't be here initially we would get:
* `isAriaExpanded=false`
* `isCSSCollapsed=false`

and the `isAriaExpanded === !isCSSCollapsed` test would fail breaking the test. So I believe that the case of initial display would be covered as well.

But hey, this is just a rough idea, feel free to adapt / discard.

ExFlo: right I missed the use case of the function. Thx

PR #2558, opened 2018-07-30
feat(datepicker): close popup on outside click

maxokorokov:
* rollback of a previously introduced breaking change
* new implementation with `requestAnimationFrame` to ignore initial click


pkozlowski-opensource:
LGTM!

PR #2565, opened 2018-08-01
fix(typeahead): add hover class instead of active class on mouseover

kevin-andrews:
Current behavior for typeahead does not align with CSS standards. mouseenter effectively adds the "active" class to elements when it should add the "hover" class.

From https://www.w3.org/TR/selectors-3/#pseudo-classes:

- The :hover pseudo-class applies while the user designates an element with a pointing device, but does not necessarily activate it. For example, a visual user agent could apply this pseudo-class when the cursor (mouse pointer) hovers over a box generated by the element. User agents that do not support interactive media do not have to support this pseudo-class. Some conforming user agents that support interactive media may not be able to support this pseudo-class (e.g., a pen device that does not detect hovering).

- The :active pseudo-class applies while an element is being activated by the user. For example, between the times the user presses the mouse button and releases it. On systems with more than one mouse button, :active applies only to the primary or primary activation button (typically the "left" mouse button), and any aliases thereof.

### DIFF
--- a/src/test/typeahead/common.ts
+++ b/src/test/typeahead/common.ts
@@ -19,10 +19,10 @@ export function expectResults(nativeEl: HTMLElement, resultsDef: string[]): void
     let classIndicator = resultDef.charAt(0);
 
     if (classIndicator === '+') {
-      expect(pages[i]).toHaveCssClass('active');
+      expect(pages[i]).toHaveCssClass('hover');
       expect(normalizeText(pages[i].textContent)).toEqual(resultDef.substr(1));
     } else {
-      expect(pages[i]).not.toHaveCssClass('active');
+      expect(pages[i]).not.toHaveCssClass('hover');
       expect(normalizeText(pages[i].textContent)).toEqual(resultDef);
     }
   }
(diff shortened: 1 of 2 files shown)


kevin-andrews:
I see. I think the correct PR would be to simply not add any class then. I assumed ng-bootstrap was trying to use an explicit .active class to replace the functionality of pseudo-classes for some reason. Is there a purpose to adding the .active class here that couldn't be achieved using the :hover pseudo-class? At the very least, the class is confusingly named considering "active" is an established CSS concept.

pkozlowski-opensource:
> I see. I think the correct PR would be to simply not add any class then.

We can't do this as the `active` class is need to add visual effect.

> I assumed ng-bootstrap was trying to use an explicit .active class to replace the functionality of pseudo-classes for some reason. 

This is not the case.

> Is there a purpose to adding the .active class here that couldn't be achieved using the :hover pseudo-class? At the very least, the class is confusingly named considering "active" is an established CSS concept.

We do follow Bootstrap CSS so if you believe that this CSS class should be renamed or styling achieved in the different way, please open an issue in https://github.com/twbs/bootstrap/issues

Going to close this PR based on the above.

Issue #2498, opened 2018-07-05
Docs: new lines are ignored in API descriptions


maxokorokov:
Issue: even though JSDoc we have contains new line breaks:

```typescript
/**
 * Hello
 *
 * world
 */
class MyComponent {
}
```

and they're also there in exracted `api-docs.ts`:
```ts
"MyComponent": {
    "description": "Hello\n\nworld"
}
```

they're NOT displayed on the main site, so it would be like `Hello world`.
It becomes a problem for long descriptions when everything start to stick together.

So two things to do:

- [ ] add `whitespace: pre-wrap;` for description paragraphs in the API components
- [ ] go through all documentation descriptions and see how they look afterwards :)

maxokorokov:
Another idea: render them as markdown directly, so this would look better? ex. `NgbDatepicker` rendered as code

![screen shot 2018-07-05 at 17 48 43](https://user-images.githubusercontent.com/8074436/42333717-c0dab208-807b-11e8-96de-e3e358ebd802.png)

PR #2568, opened 2018-08-02
Doc/ngb highlight

jnizet:
I noticed that the `NgbHighlight` directive is exported from the ng-bootstrap module, but not documented. 
This PR thus adds documentation for this directive and shows how to use it in the *custom template* demo of `NgbTypeahead`.

A second commit also makes a tiny refactoring of the template of the directive.

If the directive was not documented on purpose, because it's intended to leave it private in a future release, then I think it would be better to document it, but to deprecate it and thus make it clear that it shouldn't be used. But I'm not sure why making it private would be desirable. Anyway, in that case, the second commit can be cherry-picked.

----

### DIFF
--- a/src/typeahead/highlight.ts
+++ b/src/typeahead/highlight.ts
@@ -1,6 +1,10 @@
 import {Component, Input, OnChanges, ChangeDetectionStrategy, SimpleChanges} from '@angular/core';
 import {regExpEscape, toString} from '../util/util';
 
+/**
+ * Directive that can be used inside a custom result template in order to highlight the term inside the text of the
+ * result
+ */
 @Component({
   selector: 'ngb-highlight',
   changeDetection: ChangeDetectionStrategy.OnPush,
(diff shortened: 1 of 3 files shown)


src/typeahead/highlight.ts
```diff
@@ -1,11 +1,15 @@
 import {Component, Input, OnChanges, ChangeDetectionStrategy, SimpleChanges} from '@angular/core';
 import {regExpEscape, toString} from '../util/util';
 
+/**
+ * Directive that can be used inside a custom result template in order to highlight the term inside the text of the
```
maxokorokov: Technically it's a component, so it makes it look a bit confusing:

screen shot 2018-08-03 at 10 15 04

pkozlowski-opensource: Changed to Component and merged.

jnizet: FYI, @pkozlowski-opensource and @maxokorokov, I just reused the documentation conventions from other components. I looked at progressbar, carousel, datepicker, rating: they're all components, and their documentation describes them as a directive.

maxokorokov: Good to know, I'll do a pass through all documentation comments when fixing #2498

Issue #1052, opened 2016-11-16
ngbTooltip: add new option "delay"


Coleim:
Hello,

In bootstrap 4, it seems that we have the possibility to add a delay on the tooltip display.
`<div data-toggle="tooltip" data-placement="top" title="Tooltip on top" data-delay='{"show":"5000", "hide":"3000"}'></div>`

But I think this feature is not available with ngbTooltip.
Is it possible to add it ?

#### Version of Angular, ng-bootstrap, and Bootstrap:
Angular: 2.1.1
ng-bootstrap: 1.0.0-alpha.3
Bootstrap: 4.0.0-alpha.5

Thanks

monkatar:
Based on the docs "tooltipPopupDelay" was available for use at some point but has since been tagged as deprecated without a replacement. It doesn't seem to work correctly anyway, so I would advise against using it.
Any updates on the state of the feature request?

KissBalazs:
I think some of us are still interested in this feature - any updates or known solutions for this feature maybe?

One possible workaround I did is that you could do this with mouse events + html bind manually, but with dynamically generated divs it is quite ugly to manage.

dagangalarneau:
Here's my workaround, using pure CSS:

.tooltip > div {
  animation-name: delayedFadeIn;
  animation-duration: 1s;           /* Adjust this duration */
}

@keyframes delayedFadeIn {
  0% {opacity: 0;}
  75% {opacity: 0;}              /* Set this to 99% for no fade-in. */
  100% {opacity: 1;}
}

KissBalazs:
@dagangalarneau really nice solution, thank you very much!

PR #2569, opened 2018-08-03
feat(tooltip): add openDelay and closeDelay

divdavem:
This PR is an implementation of the feature requested in #1052.
It adds the `triggerOpenDelay` and `triggerCloseDelay` properties on the `tooltip` and `popover` components.

Notes:
- **The API documentation will have to be updated for the new properties to include the `@since` keyword with the right version number.**

### DIFF
--- a/demo/src/app/components/tooltip/demos/delay/tooltip-delay.html
+++ b/demo/src/app/components/tooltip/demos/delay/tooltip-delay.html
@@ -1,0 +1,15 @@
+<p>
+  When using non-manual triggers, you can control the delay to open and close the tooltip through the <code>openDelay</code> and
+  <code>closeDelay</code> properties. Note that the <code>autoClose</code> feature does not use the close delay, it closes the tooltip immediately.
+</p>
+
+<button type="button" class="btn btn-outline-secondary mr-2"
+  ngbTooltip="You see, I show up after 300ms and disappear after 500ms!"
+  [openDelay]="300" [closeDelay]="500">
+  Hover 300ms here
+</button>
+<button type="button" class="btn btn-outline-secondary mr-2"
+  ngbTooltip="You see, I show up after 1s and disappear after 2s!"
+  [openDelay]="1000" [closeDelay]="2000">
+  Hover 1s here
+</button>
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -109,6 +109,14 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
    */
   @Input() popoverClass: string;
   /**
+   * Delay (in ms) before opening the popover after a non-manual opening trigger.
+   */
+  @Input() openDelay: number;
+  /**
+   * Delay (in ms) before closing the popover after a non-manual closing trigger.
+   */
+  @Input() closeDelay: number;
+  /**
    * Emits an event when the popover is shown
    */
   @Output() shown = new EventEmitter();
@@ -142,6 +150,8 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
     this.container = config.container;
     this.disablePopover = config.disablePopover;
     this.popoverClass = config.popoverClass;
+    this.openDelay = config.openDelay;
+    this.closeDelay = config.closeDelay;
     this._popupService = new PopupService<NgbPopoverWindow>(
         NgbPopoverWindow, injector, viewContainerRef, _renderer, componentFactoryResolver);
 
@@ -213,8 +223,8 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
 
   ngOnInit() {
     this._unregisterListenersFn = listenToTriggers(
-        this._renderer, this._elementRef.nativeElement, this.triggers, this.open.bind(this), this.close.bind(this),
-        this.toggle.bind(this));
+        this._renderer, this._elementRef.nativeElement, this.triggers, this.isOpen.bind(this), this.open.bind(this),
+        this.close.bind(this), +this.openDelay, +this.closeDelay);
   }
 
   ngOnChanges(changes: SimpleChanges) {
--- a/src/tooltip/tooltip.ts
+++ b/src/tooltip/tooltip.ts
@@ -89,6 +89,14 @@ export class NgbTooltip implements OnInit, OnDestroy {
    */
   @Input() tooltipClass: string;
   /**
+   * Delay (in ms) before opening the tooltip after a non-manual opening trigger.
+   */
+  @Input() openDelay: number;
+  /**
+   * Delay (in ms) before closing the tooltip after a non-manual closing trigger.
+   */
+  @Input() closeDelay: number;
+  /**
    * Emits an event when the tooltip is shown
    */
   @Output() shown = new EventEmitter();
@@ -114,6 +122,8 @@ export class NgbTooltip implements OnInit, OnDestroy {
     this.container = config.container;
     this.disableTooltip = config.disableTooltip;
     this.tooltipClass = config.tooltipClass;
+    this.openDelay = config.openDelay;
+    this.closeDelay = config.closeDelay;
     this._popupService = new PopupService<NgbTooltipWindow>(
         NgbTooltipWindow, injector, viewContainerRef, _renderer, componentFactoryResolver);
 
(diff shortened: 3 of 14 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "ng-packagr": "^3.0.1",
     "prismjs": "^1.12.2",
     "protractor": "~5.3.1",
-    "rxjs": "6.0.0",
+    "rxjs": "6.2.2",
```
divdavem: Note that updating the version of rxjs to 6.2.2 is needed for the fake async test with observables to pass.
The following commit in rxjs fixes the problem:
https://github.com/ReactiveX/rxjs/commit/01eefff2132f29dfe51476782c3e7f2db48f88f2

src/tooltip/tooltip.ts
```diff
@@ -107,6 +107,14 @@ export class NgbTooltip implements OnInit, OnDestroy {
    * @since 1.1.0
    */
   @Input() disableTooltip: boolean;
+  /**
+   * Delay (in ms) before opening the tooltip after a non-manual opening trigger.
+   */
+  @Input() triggerOpenDelay = 0;
```
maxokorokov: - these inputs should come from `NgbTooltipConfig`
- related tests are in `ngb-tooltip-config.ts`

src/popover/popover.ts
```diff
@@ -156,6 +156,14 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
    * @since 2.2.0
    */
   @Input() popoverClass: string;
+  /**
+   * Delay (in ms) before opening the popover after a non-manual opening trigger.
+   */
+  @Input() triggerOpenDelay = 0;
```
maxokorokov: - these inputs should come from `NgbPopoverConfig`
- related tests are in `ngb-popover-config.ts`

package.json
maxokorokov: I guess the ideal solution is for the fix to be released in `rxjs 6.0.x`... I see three options in order of preference:
- use a workaround if any
- `src/pacakge.json` → `rxjs: ^6.0.0`, `package.json` → `rxjs: 6.2.2`. And avoid using features from `6.1` and `6.2` →  `endWith` operator and `isObservable` methods https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#620-2018-05-22
- `src/pacakge.json` → `rxjs: ^6.2.2`, `package.json` → `rxjs: 6.2.2`

demo/src/app/components/tooltip/demos/delay/tooltip-delay.html
```diff
@@ -0,0 +1,15 @@
+<p>
+  When using non-manual triggers, you can control the delay to open and close the tooltip through the <code>triggerOpenDelay</code> and
+  <code>triggerCloseDelay</code> properties.
+</p>
+
+<button type="button" class="btn btn-outline-secondary mr-2"
+  ngbTooltip="You see, I show up after 300ms and disappear after 500ms!"
+  [triggerOpenDelay]="300" [triggerCloseDelay]="500">
+  Hover 200ms here
```
maxokorokov: `300ms`

divdavem:
@maxokorokov Thank you for your review!
I have renamed the properties to remove the `trigger` prefix.
The default values for the new properties are now taken from the injected config.
I have also fixed the inconsistency in the tooltip-delay example.

benouat:
Of course you both guys @maxokorokov @divdavem, would have translated my twisted mind 🤣 `ngrx` -> `rxjs`

divdavem:
@benouat I have rebased the PR.

divdavem:
I have now updated my PR to better cooperate with `autoClose` (or any other open/close done outside of the automatic triggers of the popover/tooltip).

divdavem:
@benouat I have added a commit with an ugly workaround for tests to pass with rxjs < 6.2.2

pkozlowski-opensource:
@benouat @divdavem whatever we do we can _not_ bump up RxJS dependency version in the published npm package as it would be:
* a breaking change;
* place more restriction on our users (Angular works with RxJS 6.x)

I would be also very much against testing against one RxJS version and publishing with dependency on another - it would be way too easy to introduce dependency on 6.2.x without noticing it.

One idea would be to omit problematic tests in IE, but this has risks as well.

In any case please hold off any further steps on this PR until the RxJS version issue is properly addressed.

divdavem:
@pkozlowski-opensource Thank you for your comment.

> whatever we do we can not bump up RxJS dependency version in the published npm package as it would be:
> 
> -   a breaking change;
> -   place more restriction on our users (Angular works with RxJS 6.x)
> 

OK! So this removes option 3 suggested by @maxokorokov 

> I would be also very much against testing against one RxJS version and publishing with dependency on another - it would be way too easy to introduce dependency on 6.2.x without noticing it.

I understand. This removes option 2 suggested by @maxokorokov.
Note that, according to @maxokorokov, there are not that many new features: `endWith` operator and `isObservable` methods https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#620-2018-05-22

> One idea would be to omit problematic tests in IE, but this has risks as well.

Note that the same problem occurs in all browsers, not only in IE.

So, I can think about different solutions:

1) Maybe checking with people from rxjs if they can release a 6.0.1 version that only contains the required fix.

2) Using the workaround I did in the second commit of this PR. The workaround is ugly but it does not harm, and it is used only in tests, it is not part of the released code. Even if we forget to remove it after updating to rxjs >= 6.2.2, it is not a big deal, it will probably not break anything.

3) Using a rewritten version of the `delay` operator instead of the one from `rxjs`. This would allow to avoid relying on the code that uses the saved copy of `Date.now` and prevents proper testing with `fakeAsync`.

4) Using real `async` tests instead of `fakeAsync` ones. Even if we use shorter delays than the ones used in the current tests, it would for sure be much slower than the current `fakeAsync` tests, but it would not require any workaround. Maybe it would also not be as accurate as `fakeAsync` tests (there might me uncontrolled delays due to the execution time that we do not see in `fakeAsync` tests, and waiting 1ms less than the `openDelay` to check that it is not opened yet may not work as reliably, I have not tested).

5) Disabling the problematic tests. As you said, this has risks, especially as the tests would have to be disabled in all browsers, not only IE.

pkozlowski-opensource:
@divdavem thnx for the detailed analysis and proposals - seriously good stuff in there. Let's go over it together tomorrow and decide.  I will talk to Ben tonight to see if they can release 6.0.x patch.

benouat:
Next major version, ie `4.x` will probably bump `rxjs@6.3`. We should then be able to move on with this PR.

divdavem:
I have rebased and removed the ugly workaround.
In fact, the workaround was not so ugly, but it is now useless so I removed it (as we now use `rxjs@6.3`).

src/util/triggers.ts
```diff
@@ -37,24 +40,74 @@ export function parseTriggers(triggers: string, aliases = DEFAULT_ALIASES): Trig
...
+                pending = null;
+                return event.open !== isOpenedFn();
+              }
+              return false;
+            }),
+            map(event => event.open));
+  };
+}
+
+export function listenToTriggers(
```
maxokorokov: Just as an alternative, while I was trying to read the code with a fresh look.

I think I'd prefer it to be like this:

```ts
  const parsedTriggers = parseTriggers(triggers);

  if (parsedTriggers.length === 1 && parsedTriggers[0].isManual()) {
    return () => {};
  }

  const subscription = observeTriggers(renderer, nativeElement, parsedTriggers, isOpenedFn)
    .pipe(triggerDelay(openDelay, closeDelay, isOpenedFn))
    .subscribe(open => (open ? openFn() : closeFn()));

  return () => subscription.unsubscribe();
```

Reasoning being:
- I see the whole flow undisturbed without jumping: parse → create observable → pipe through the delay → subscribe → call open/close
- `observeTriggers` **only** sets up an observable
- one function less → no need in `subscribeTriggers`

divdavem:
@maxokorokov Thank you for reviewing again my PR. I have updated it to take into account your feedback.
Regarding the demo of the popover, it may feel buggy because people are used to have the tooltip closed when clicking outside (autoclose), but it was disabled (because otherwise there is no visible delay). I have changed the demo to use mouseenter and mouseleave event instead, which is closer to real use cases and similar to the tooltip demo.

maxokorokov:
@divdavem as mentioned, I just split it into two commits for tooltip and popover. No history seems to be lost :)

divdavem:
@maxokorokov Thank you for integrating my PR!

PR #2571, opened 2018-08-03
feat(tooltip): autoclose

ymeine:


src/tooltip/tooltip-config.ts
```diff
@@ -8,6 +8,7 @@ import {PlacementArray} from '../util/positioning';
  */
 @Injectable({providedIn: 'root'})
 export class NgbTooltipConfig {
+  autoClose: boolean | 'inside' | 'outside' = true;
```
maxokorokov: No big deal, but could you please add test for this?

https://github.com/ymeine/ng-bootstrap/blob/2ac9798db9c7d76f3af4d891ed28cdf36cc1faf1/src/tooltip/tooltip.spec.ts#L730-L738

(probably forgot to do it for popover too, if so, preferably, in a separate PR)

PR #2578, opened 2018-08-07
feat(pagination): add class to be able customize the look

### LINKED ISSUE #899: Custom pagination boundary and directional link text

ExFlo:
Closes  #899

(Lint was not passing without changing the radix for the rating...)
I changed view encapsulation to none, because if we add some style in the pagination feature then the encapsulation does not allow to use the class to custom the inner content of the li (and this is what we want here).

### DIFF
--- a/src/pagination/pagination.ts
+++ b/src/pagination/pagination.ts
@@ -11,7 +11,7 @@ import {NgbPaginationConfig} from './pagination-config';
   host: {'role': 'navigation'},
   template: `
     <ul [class]="'pagination' + (size ? ' pagination-' + size : '')">
-      <li *ngIf="boundaryLinks" class="page-item"
+      <li *ngIf="boundaryLinks" class="page-item ngb-pagination-first"
         [class.disabled]="!hasPrevious() || disabled">
         <a aria-label="First" i18n-aria-label="@@ngb.pagination.first-aria" class="page-link" href
           (click)="!!selectPage(1)" [attr.tabindex]="(hasPrevious() ? null : '-1')">
@@ -19,7 +19,7 @@ import {NgbPaginationConfig} from './pagination-config';
         </a>
       </li>
 
-      <li *ngIf="directionLinks" class="page-item"
+      <li *ngIf="directionLinks" class="page-item ngb-pagination-previous"
         [class.disabled]="!hasPrevious() || disabled">
         <a aria-label="Previous" i18n-aria-label="@@ngb.pagination.previous-aria" class="page-link" href
           (click)="!!selectPage(page-1)" [attr.tabindex]="(hasPrevious() ? null : '-1')">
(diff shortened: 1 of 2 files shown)


src/rating/rating.ts
```diff
@@ -210,7 +210,7 @@ export class NgbRating implements ControlValueAccessor,
       return 100;
     }
     if (diff < 1 && diff > 0) {
-      return Number.parseInt((diff * 100).toFixed(2));
+      return Number.parseInt((diff * 100).toFixed(2), 10);
```
maxokorokov: Saw your comment, but I absolutely don't see the link to this PR and this file didn't have any issues previously. Could you please remove it?

src/pagination/pagination.ts
```diff
@@ -11,44 +20,45 @@ import {NgbPaginationConfig} from './pagination-config';
...
+      <li *ngIf="boundaryLinks" class="page-item ngb-pagination-last" [class.disabled]="!hasNext() || disabled">
         <a aria-label="Last" i18n-aria-label="@@ngb.pagination.last-aria" class="page-link" href
           (click)="!!selectPage(pageCount)" [attr.tabindex]="(hasNext() ? null : '-1')">
           <span aria-hidden="true" i18n="@@ngb.pagination.last">&raquo;&raquo;</span>
         </a>
       </li>
     </ul>
-  `
+  `,
+  encapsulation: ViewEncapsulation.None
```
maxokorokov: I don't think we need to add it now, this component has no styles, so it won't have any effect if I understand correctly. Using classes you're adding doesn't need this.

src/pagination/pagination.ts
```diff
@@ -11,44 +20,45 @@ import {NgbPaginationConfig} from './pagination-config';
...
+      <li *ngIf="directionLinks" class="page-item ngb-pagination-previous"
         [class.disabled]="!hasPrevious() || disabled">
         <a aria-label="Previous" i18n-aria-label="@@ngb.pagination.previous-aria" class="page-link" href
           (click)="!!selectPage(page-1)" [attr.tabindex]="(hasPrevious() ? null : '-1')">
           <span aria-hidden="true" i18n="@@ngb.pagination.previous">&laquo;</span>
         </a>
       </li>
       <li *ngFor="let pageNumber of pages" class="page-item" [class.active]="pageNumber === page"
-        [class.disabled]="isEllipsis(pageNumber) || disabled">
+        [class.disabled]="isEllipsis(pageNumber) || disabled" [class.ngb-pagination-dots]="isEllipsis(pageNumber)">
```
maxokorokov: - should be `ngb-pagination-ellipsis` as it's the name of the symbol and in all APIs we use `ellipsis`
- there is only test case for when this class is present, but never checked that is is not present (should be same as for `.disabled`)

maxokorokov:
@ExFlo, could you please reformat, as build is failing

maxokorokov:
@ExFlo, I guess this one was blocked because of adding non-bootstrap classes to the bootstrap markup. The idea is to avoid doing it for standard bootstrap components. For our custom ones we can do whatever we want.

By the way the template-way of customising landed in #2971. So I'll close this one.

I think we might/should revisit it once we ship our own custom theme and make CSS classes the part of the API officially in #2947.

PR #2579, opened 2018-08-07
chore: fix linting configuration issues

maxokorokov:
Two commits here:
- `semicolon` tslint rule wasn't working for a looong time → turned on an fixed issues
- `member-ordering` rule also had syntax issues


pkozlowski-opensource:
LGTM, please merge when Travis gets green.
BTW, can prettier (or any other tool) do this stuff for us when re-formatting?

maxokorokov:
tslnt can fix some of the rules automatically, but we should make sure its configuration is not in conflict with prettier/clang-format (these two just code format though)

PR #2587, opened 2018-08-09
refactor(datepicker): use 'formatDate' instead of 'DatePipe'

maxokorokov:


pkozlowski-opensource:
LGTM, didn't know that `formatDate` was public!

PR #2599, opened 2018-08-13
fix(carousel): make it work inside the 'OnPush' components

### LINKED ISSUE #2597: ChangeDetectionStrategy.OnPush on ngbCarousel not working
when I implement ChangeDetectionStrategy.OnPush ngbcarousel doesn't work with atuomatic intervals

maxokorokov:
Also the carousel itself is now 'OnPush'

Fixes #2597


pkozlowski-opensource:
LGTM, thnx!

Issue #1632, opened 2017-06-23
Allow component routing for NgbTabset


tilowestermann:
### Bug description:

I'd like to load different components using the tab set. These components should be accessed by their URL. 
Currently, I see two ways of achieving this:
1. Provide a link with a `routerLink` in the `ngbTabTitle` template
2. Use the `tabChange` event and tell the router to navigate.

[Option 1](https://stackoverflow.com/a/41592058/184245) has the drawback that routing only happens if the title is clicked (not the tab)

[Option 2](https://stackoverflow.com/a/44717581/184245) doesn't allow for copying the link when hovering it

While both options may achieve the desired result, I think there should be a better way (e.g., without adding an empty `ngbTabContent` template).

### Link to minimally-working plunker that reproduces the issue:

See links provided for Option 1 and 2.

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.x

ng-bootstrap: 1.0.0-alpha.26

Bootstrap: v4.0.0-alpha.6

chuckconway:
I second this issue.

suchinb:
I am still facing the issue, I am using routerLink in ngb-tab tag, But when I click it does not navigate , but If I use the simple buttons it navigates, 

<ngb-tab title="Simple" [routerLink]="['./nestedChild1']">

chuckconway:
I'm not sure if this will help, but I create a post on StackOverflow how I got tab navigation working. 

https://stackoverflow.com/questions/42446434/how-can-i-access-the-select-method-from-ngbtabset-in-the-component-in-angular2/44940187#44940187

masiorama:
Is it going to be handled in the next future? I'm trying to achieve the same result and none of the two options proposed by the opener are perfect. 
Thanks in advance.

divdavem:
For info, I have opened PR #2603 to implement this feature request.
What do you think about it?

dries-2002:
I've run into the same problem, and instead of trying to get the ngb-tabset to behave like a navbar, I refactored it into a navbar. This effectively solves the routing problem.

The key is to create a wrapper component, and to define the tabs as children in the routing. Don't forget to add the `<router-outlet>` tag in the wrapper component.

kashpatel:
For anyone who come looking for solution on how to make routing works with ng-bootstrap tabs. I have made an example app to demonstrate that. I used secondary router outlet for to render tab component.

See the demo here:
[...]

Issue #1584, opened 2017-05-26
tabset destroyOnHide=false inits all tabs immediately


achimha:
### Bug description:

When a `ngb-tabset` is created with `[destroyOnHide]=false`, the content of all tabs is immediately created. With the default value of `true`, content is only created when switching to the respective tab. From the naming and documentation I would expect the only difference in behavior to be that already created tab content is not destroyed on tab change. I would not expect it to immediately construct all tab content.

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.1.3

ng-bootstrap: v1.0.0-alpha.26

Bootstrap: 4.0.0-beta-6

pkozlowski-opensource:
The current behaviour is by design. We might consider having lazy-initiated tabs in the future so marking this as a feature request.

achimha:
A simple workaround:

```
tabstatus = {};
```
and
```
<ngb-tabset [destroyOnHide]="false" (tabChange)="tabstatus[$event.nextId]=true">
  ...
  <ngb-tab id="tab2" title="Second Tab">
    <ng-template ngbTabContent>
      <ng-container *ngIf="tabstatus['tab2'] === true">
      ...
      </ng-container>
    </ng-template>
  </ngb-tab>
  ...
</ngb-tabset>
``` 
I reported it as a bug because the behavior conflicts with the name `destroyOnHide` as it also makes the tabset construct all tabs at load time while with the default value, it would not do that.

PhilFlash:
Same problem,
@achimha, thanks for the tip

rakesh2293:
Hi achimha ,
I facing some issue in Ngb-tabs so please help me 
1. How we **Hiding**  the tabs shift or disable the tab's 
2. After clicking on **t.select('tabid1')** button then only tabs can shift to the next tabs?

Bat-Orshikh:
pkozlowski-opensource @maxokorokov

Hello, I am having same problem here,

So, is there any news?

yohny:
with `Tabset` being deprecated, the same problem is valid for its recommended alternative `Nav`

JasonPierce:
@achimha's wonderful workaround, updated for `ngbNav` (now that `ngb-tabset`  is deprecated)

JS/TS
```js
tabStatus = {};
```

HTML
```html
<ul ngbNav #nav="ngbNav" class="nav-tabs" [destroyOnHide]="false" (navChange)="tabStatus[$event.nextId] = true">
[...]

PR #2603, opened 2018-08-13
feat(tabset): bind tabs to routes

divdavem:
This PR is a suggestion to solve the issue raised in #1632. It allows to easily bind tabs to routes by taking into account the `routerLink` directives on `<ngb-tab>` elements.

For this to work, the new `NgbTabsetRoutingModule` module has to be included in the application and the `routerControlled` attribute has to be defined on the `<ngb-tabset>` element.

Also, the corresponding routes have to be defined in the configuration of the router.
The content of the tabs can be defined either in the corresponding `<ng-template ngbTabContent>` elements, or as the `component` property defined in the router configuration (if the  `<router-outlet>` element is included in the page).

Here is an example:
```html
<ngb-tabset routerControlled>
  <ngb-tab title="First" routerLink="firstTab">
    <ng-template ngbTabContent>
      First content
    </ng-template>
  </ngb-tab>
  <ngb-tab title="Second" routerLink="secondTab">
    <ng-template ngbTabContent>
      Second content
    </ng-template>
  </ngb-tab>
  <ngb-tab routerLink="thirdTab">
    <ng-template ngbTabTitle>Third</ng-template>
    <ng-template ngbTabContent>
      Third content
    </ng-template>
  </ngb-tab>
</ngb-tabset>
```

Note that `NgbTabsetRoutingModule` is a separate module in a different entry point (`"@ng-bootstrap/ng-bootstrap/tabset/routing"`) because it depends on `@angular/router` (and the main entry point does not have this dependency).

Note that this PR also adds an `href` input on `<ngb-tab>` elements, which can be used when `routerControlled` is not set on `<ngb-tabset>` in order to set the `href` used on the tab title anchor.

---

### DIFF
--- a/demo/src/app/components/tabset/demos/router/tabset-router.html
+++ b/demo/src/app/components/tabset/demos/router/tabset-router.html
@@ -1,0 +1,31 @@
+<p>In order to easily synchronize the <code>&lt;ngb-tabset&gt;</code> with the Angular router, some directives are provided in the
+  <code>NgbTabsetRoutingModule</code> module, which is not in the main entry point of ng-bootstrap (to avoid adding a general dependency
+  on the Angular router).</p>
+<p>The <code>NgbTabsetRoutingModule</code> module has to be imported from <code>'@ng-bootstrap/ng-bootstrap/tabset/routing'</code>.</p>
+<p>All child <code>&lt;ngb-tab&gt;</code> elements must be linked with a route using the standard
+  <a href="https://angular.io/api/router/RouterLink#description">routerLink</a> directive.</p>
+<p>The corresponding routes have to be defined in the configuration of the Angular router.</p>
+<p>The <code>&lt;ngb-tabset&gt;</code> should only be used on those configured routes. If no child <code>&lt;ngb-tab&gt;</code> element matches,
+  the <code>&lt;ngb-tabset&gt;</code> will not change the active tab (and if there is no active tab, the default behavior of the normal
+  <code>&lt;ngb-tabset&gt;</code> component is to select the first tab).</p>
+<p>The content of the tabs can be defined either in the corresponding <code>&lt;ng-template ngbTabContent&gt;</code> elements, or as the
+component property defined in the router configuration (if the <code>&lt;router-outlet&gt;</code> element is included in the page).</p>
+
+<ngb-tabset routerControlled>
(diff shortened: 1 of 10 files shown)


divdavem:
@benouat Thank you for your comment.

Currently, it is not easy to both have the convenient markup offered by &lt;ngb-tabset&gt; and the binding to the router. That's why feature request #1632 was opened. I have opened this PR to suggest a solution. If you have a better idea on how to do it, feel free to share it.

As you are talking about the demo website, the routable tabs there ("Overview", "Example" and "Api") are currently done without the &lt;ngb-tabset&gt; directive at all, so the full bootstrap tab markup is written manually. I find the markup I suggest in this PR much simpler. I have implemented this as an independent module (which is not fake, by the way!) so that there is no negative impact for people who prefer not using it.

benouat:
@divdavem don't get me wrong here, I started a discussion, but I am mostly fine with the changes behind this PR. To me 'fake' just meant that you have to deal with another 'embedded' module as compared to a 'real' one that would come from npm.

So to me, ideally it should be shipped as a standalone module and we still don't know how to do that. How to maintain? how to publish simply several modules? (I am talking about several repos vs. mono-repo, independent lifecycle vs. synchronised version number)

Then it's true that current demo site does not involve the widget now, it used to be the case, and was not that complex to setup. I think people opened #1632 because they simply did not know how to do it. If we document this how to properly with code snippets to copy/paste, then I think we don't need this PR.

Maybe I am wrong, maybe I am totally biased due to my experience, especially with demo website.

divdavem:
@benouat Thank you for your answer.

I don't think people opened #1632 only because they did not know how to do it, but also because the `<ngb-tabset>` component itself makes it more complex to do it than when using the raw markup from bootstrap, because it wraps `<a>` elements, so it is no longer possible to put `routerLink` directives on them. So, with `<ngb-tabset>`, it is not possible anymore to link tabs with the router in a declarative way, some javascript/typescript code is required too. With this PR, I am trying to restore the ability to do it only with markup, so I think it makes sense to share the code and avoid having people re-write that JS/TS code each time they need to bind tabs to the Angular router.

asvishnyakov:
This feature is not about markup, You need it if you use server side rendering and want good SEO

maxokorokov:
Closing this in favour of #1584, where you can use navs with router links directly

Issue #2601, opened 2018-08-13
Some stackblitz demos are broken


maxkorz:
- [ ] Datepicker - [Popup](https://ng-bootstrap.github.io/#/components/datepicker/examples#popup)
https://stackblitz.com/angular/gaqnmyyjxdm icon is missing
- [ ] Datepicker - [Alternative Calendars](https://ng-bootstrap.github.io/#/components/datepicker/examples#calendars)
https://stackblitz.com/angular/nlogpelxlme 
`Import error, can't find file: app/islamic-calendars.html`
- [ ] Popover - [Automatic closing with keyboard and mouse](https://ng-bootstrap.github.io/#/components/popover/examples#autoclose)
https://stackblitz.com/angular/kdqrkadrdga `Error in /turbo_modules/@angular/compiler@6.1.0/bundles/compiler.umd.js (1087:21)
Unexpected value 'undefined' declared by the module 'AppModule'
Evaluating main.ts
Booting application`
- [ ] Popover - [Popover with custom class](https://ng-bootstrap.github.io/#/components/popover/examples#custom-class)
https://stackblitz.com/angular/mrjavlmbrvp `Error in app/app.module.ts (8:27) ',' expected.`
- [ ] Tooltip - [Automatic closing with keyboard and mouse](https://ng-bootstrap.github.io/#/components/tooltip/examples#autoclose)
https://stackblitz.com/angular/rdlnbxqlenb
`Error in /turbo_modules/@angular/compiler@6.1.0/bundles/compiler.umd.js (1087:21)
Unexpected value 'undefined' declared by the module 'AppModule'
Evaluating main.ts
Booting application`

I did not test plunker demos

maxkorz:
Updated

maxokorokov:
@benouat, I think something is wrong with tabset stackblitz url generation after you changed page generation, could you PTAL?

stackblitzes are generated with

```
https://ng-bootstrap.github.io/app/components/tabs/demos/pills/stackblitz.html
```

instead of 

```
https://ng-bootstrap.github.io/app/components/tabset/demos/pills/stackblitz.html
```

benouat:
@maxkorz @maxokorokov thanks for spotting that one out.

@maxokorokov It probably went through my filter before merging, if you remember, I changed some demos names before we realised that plunkr/stackblitz generation was _hardcoded_. I will open a PR for that asap

PR #2604, opened 2018-08-13
demo(fix): use correct demo file and class names

maxokorokov:
Part of #2601

cc @ymeine


pkozlowski-opensource:
LGTM, thnx

pkozlowski-opensource:
I'm going to merge it and re-generate demo site so we've got it working.

PR #2609, opened 2018-08-14
demo(fix): missing icon

kukac7:
add @FortAwesome/Font-Awesome icons
#2601


pkozlowski-opensource:
> It might be an overkill to have the whole fa just for one icon.

Yeh, I was thinking the same... I'm not sure how big fa is but it would probably slow down demo site load...

maxokorokov:
~12Kb gzipped, I don't think it will slow the site load in any way, but still think it's an overkill.

There is an Angular API to include only one icon as an alternative: https://fontawesome.com/how-to-use/on-the-web/using-with/angular, but I'm not a fan of this either :)

kukac7:
yes, I know it's a bit overkill.
I just wanted a simple method, and many people used it when it comes to icons. (including us)

yes, I know there is angular api, but for me it is not simpler and overkill than font.

maxokorokov:
On a side note any `yarn.lock` changes shouldn't be a part of this PR

kukac7:
sry, I fucked up. 😭

pkozlowski-opensource:
@kukac7 thnx for the PR but we are going to merge #2624 instead as it doesn't require us to add / load new dependency.

Appreciate your contribution and time, though!

PR #2614, opened 2018-08-15
feat(buttons): uncheckable radio buttons

### LINKED ISSUE #2598: RadioButton uncheckable
I did not find any issue about that, so I thought I open one:

I would like to have the feature of uncheckable radio buttons like in [ui-bootstrap](https://angular-ui.github.io/bootstrap/#!#buttons)

Is the feature planned, dissmissed or did just nobody thought about it ;) ?

Zwergenpunk:
Radio buttons can now be unchecked, if the radioGroup has the uncheckable flag set to true.

closes #2598

### DIFF
--- a/src/buttons/radio.ts
+++ b/src/buttons/radio.ts
@@ -36,12 +36,17 @@ export class NgbRadioGroup implements ControlValueAccessor {
    */
   @Input() name = `ngb-radio-${nextId++}`;
 
+  /**
+   * A flag indicating if the radio buttons can be unchecked
+   */
+  @Input() uncheckable = false;
+
   onChange = (_: any) => {};
   onTouched = () => {};
 
-  onRadioChange(radio: NgbRadio) {
-    this.writeValue(radio.value);
-    this.onChange(radio.value);
+  onRadioChange(value: any) {
+    this.writeValue(value);
+    this.onChange(value);
   }
 
   onRadioValueUpdate() { this._updateRadiosValue(); }
@@ -80,6 +85,7 @@ export class NgbRadioGroup implements ControlValueAccessor {
     '[disabled]': 'disabled',
     '[name]': 'nameAttr',
     '(change)': 'onChange()',
+    '(click)': 'onClick()',
     '(focus)': 'focused = true',
     '(blur)': 'focused = false'
   }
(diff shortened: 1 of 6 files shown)


src/buttons/radio.ts
```diff
@@ -133,7 +139,13 @@ export class NgbRadio implements OnDestroy {
 
   ngOnDestroy() { this._group.unregister(this); }
 
-  onChange() { this._group.onRadioChange(this); }
+  onChange() { this._group.onRadioChange(this.value); }
+
+  onClick() {
```
pkozlowski-opensource: So, I must say that I'm super-worried about the sequence of `click` / `change` events here - in this code you are assuming that `click` will fire before `change` but this not the case in all browsers, AFAIK.

Also, the current code would trigger model mutation twice and I don't think this is right.

Generally speaking we should make sure that only one of `click` / `change` is processed.

Zwergenpunk: thanks for the feedback.

yeah I had some bad feelings as well, but didn't want to remove the onchange, but it seems like we have to remove the onchange to make it work. 

I didn't know that browser differ so much in that regard.

pkozlowski-opensource: > but it seems like we have to remove the onchange to make it work.

We can't do this as it would kill keyboard support, if I'm not mistaken

Zwergenpunk:
I tried to remove the onchange method and it is still working with arrow left / arrow right. 

Deselecting with space is not working in every browser, but I'm not sure if that is completely needed.

I tried to add some tests with keyboard events, but I have very little experience with tests, and that seemed to be not so simple =/

Zwergenpunk:
I found a really strange behaviour in the examples page (also currently live): you cannot access the radio buttons of the normal radio buttons example with tabbing without clicking on it first. But you can access the other radio button.

That happened in chrome and edge

benouat:
@Zwergenpunk Your PR need a rebase, be aware that we recently changed the way demos are defined in #2520. You will also have to change a few things after your rebase. Thanks!

Zwergenpunk:
@benouat  thanks for the hint. I saw the merge conflicts but totally forgot about it =/

Zwergenpunk:
Is there anything left todo?

Zwergenpunk:
will this ever be merged? Actually I need that feature =/

If I need to change something I will do it, but I need to know what.

adam-marshall:
What is the status of this please?

Zwergenpunk:
if I knew whats the problem I would fix it, but it seems nobody cares about it... =(

Zwergenpunk:
Since this will never be merged i just close it

PR #2624, opened 2018-08-17
demo(datepicker): ship calendar icon to plunkers and stackblitzes

maxokorokov:
Alternative to #2609 without font awesome icons.
We have only one icon, but stackblitz doesn't allow to ship images, so shipping it in css.

Looks almost the same as before: 

screen shot 2018-08-17 at 14 04 20

Part of #2601


misc/plunk-gen.js
```diff
@@ -84,6 +88,7 @@ function generatePlnkrContent(componentName, demoName) {
     <input type="hidden" name="description" value="Example usage of the ${componentName} widget from https://ng-bootstrap.github.io">
 ${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName)])}
     <input type="hidden" name="files[index.html]" value="${he.encode(generateIndexHtml())}">
+    <input type="hidden" name="files[css/demos.css]" value="${he.encode(generateDemosCSS())}">
```
pkozlowski-opensource: Is there a reason for having this file name different in plunker and stackblitz?

pkozlowski-opensource:
@maxokorokov LGTM, just left one nit comment / question. Feel free to ignore / change and merge.

misc/plunk-gen.js
maxokorokov: The only reason is that stackblitz tries to follow cli structure. Will align plunker with it actually, thanks!

PR #2627, opened 2018-08-20
fix: provide overrideable services with at the root level

### LINKED ISSUE #2618: NgbDateNativeAdapter provided in app.module sems not be considered
### Bug description:
When providing the NgbDateNativeAdapter in app.module.ts with 
`typescript
providers: [
    { provide: NgbDateAdapter, useClass: NgbDateNativeAdapter }
  ]
`

my model value is bound to a NgbDateStruct-Value representing the selected date. 
I expected it to be a "native" date, as transformed by NgbDateNativeAdapter.

http://plnkr.co/edit/lkfKKxk63EjuoVWJrpiT?p=preview

Angular: 6.1.3

ng-bootstrap: 3.0.0

Bootstrap: 4.1.3

maxokorokov:
Fixes a regression introduced in 3.0.0, where you had to duplicate overrides for the services in question at the module level in case of lazy-loading

Note: can only use `useFactory` as a named function that needs to be exported for ngc, `useClass` with `forwardRef` doesn't work with `providedIn`

Fixes #2618


pkozlowski-opensource:
LGTM!

Issue #2543, opened 2018-07-23
Guidelines: WCAG 2.0 (Level AA) not followed for ngb date-picker


kedar9444:
### Bug description: The auto generated select fields for month and year violates the **WCAG 2.0 (Level AA) guidelines**.

Error code : Check 91: select element missing an associated label.
Repair: Add a label element that surrounds the control's label. Set the for attribute on the label element to the same value as the id attribute of the control. And/or add a title attribute to the input element. And/or create a label element that contains the input element.

since the html is auto-generated we can not add the label for it.

### Version of Angular, ng-bootstrap, and Bootstrap:

**Angular: "^5.1.3"**

**ng-bootstrap:  "1.1.0"**

**Bootstrap: "4.0.0"**

kedar9444:
website used for checking : https://achecker.ca/

you can use the html from this link to reproduce the error : 
https://ng-bootstrap.github.io/#/components/datepicker/overview

maxokorokov:
An option would be to add a title something like:

```html
<select title="months"></select>
<select title="years"></select>
```

But these strings should be added to i18n in this case

kedar9444:
@maxokorokov :  **But these strings should be added to i18n in this case**?
 i am not clear about this step can you please elaborate it further.

kedar9444:
This is what my code looks like:

![image](https://user-images.githubusercontent.com/20212215/43083498-4ae6f100-8eb4-11e8-8a53-74500b0d9810.png)

Add this is how it appears : 
![image](https://user-images.githubusercontent.com/20212215/43083531-627a6eb4-8eb4-11e8-9b05-7a2601f47ce9.png)

and here is dynamically generated DOM elements for date-picker, which are clearly missing labels for select boxes used for month and year selection : 

![image](https://user-images.githubusercontent.com/20212215/43083709-ca726bb6-8eb4-11e8-910c-aed62d2a6119.png)

which violates the WCAG 2.0 guild lines  which states : **Check 91: select element missing an associated label**

maxokorokov:
@kedar9444, sure, I see the problem, it was clear from the description. Adding the `title` attribute should fix it.

> But these strings should be added to i18n in this case

Just wanted to say that if we put a hardcoded string as a select box title, we should not forget to translate it.

kedar9444:
@maxokorokov : yes, got it.
do you guys have any plan to add/implement these WCAG guidelines for ng-bootstrap?

[...]

PR #2636, opened 2018-08-23
feature(date-picker): added aria-labels to year and month select elem…

RonnBlack:
…ents


pkozlowski-opensource:
@RonnBlack based on the discussion in #2543 it doesn't seem like aria labels alone are enough...
Did you tests with screen readers and found different results in your testing?

rblack-a3digital:
Thanks for pointing out the discussion. I hadn't seen this.
I was only basing my changes on recommendations from WAVE (Web Accessibility Evaluation Tool) and I didn't actually test with a screen reader. I'll update the code to add the title property and update the pull request.

RonnBlack:
I added titles to the select elements and tested in Chrome, Firefox, and Edge. I saw now display issues on widows 10 or Mac OS High Sierra (10.13.6). You mentioned that there would be visual impact... do you know of something specific? 

I also tested with Voice Over on Mac and Narrator on Windows and the drop downs seemed to behave as I would expect. It looks like the pull request shows the additional changes I made. If that isn't the case let me know.

benouat:
> You mentioned that there would be visual impact... do you know of something specific?

I think he was mentioning the fact that browsers add a native native _yellow_ tooltip on elements with `title` attribute when you hover them and wait for some milliseconds.

RonnBlack:
Yes the hover behavior is there but the next and previous month buttons also have that behavior. 
In my testing with screen readers it seemed to work fine with just the aria-label even before I added the title. The only problem I noticed is that the Microsoft Narrator didn't announce the label on the Month select but it did properly announce it on the Year. Once title was added it was fine with both. Mac Voice over didn't have a problem either way but IMHO the way it announced what you were on was more confusing (perhaps there is some control over this)

Issue #2635, opened 2018-08-22
Disabled class is not added to radio button label when used with *ngIf


Fredx87:
### Bug description:
I'm using radio buttons with reactive forms. Normally when the control is disabled, the "disabled" class is added to the label, but when on the label there is also a *ngIf, the disabled class is not applied anymore. The same problem is not present with checkbox buttons.

### Link to minimally-working plunker that reproduces the issue:
https://stackblitz.com/edit/angular-ycgkcs

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 6.1

ng-bootstrap: 2.2 / 3.0

Bootstrap: 4.0

pkozlowski-opensource:
Thnx for in-depth investigation done by @benouat we can confirm that this is a bug on our end. Will fix it for the next release. Thnx for reporting @Fredx87 !

PR #2639, opened 2018-08-24
fix(radio): initialiase disabled according to parent in constructor

benouat:
Proper fix with test for #2635


pkozlowski-opensource:
LGTM, thnx!

PR #2640, opened 2018-08-24
feat(modal): stacked modals

maxkorz:

P.S. if you think the project is not ready to have stacked modals in the demo, at least consider merging changes related to the `modal-open` class.

### DIFF
--- a/src/modal/modal-backdrop.ts
+++ b/src/modal/modal-backdrop.ts
@@ -3,7 +3,8 @@ import {Component, Input} from '@angular/core';
 @Component({
   selector: 'ngb-modal-backdrop',
   template: '',
-  host: {'[class]': '"modal-backdrop fade show" + (backdropClass ? " " + backdropClass : "")'}
+  host:
+      {'[class]': '"modal-backdrop fade show" + (backdropClass ? " " + backdropClass : "")', 'style': 'z-index: 1050'}
 })
 export class NgbModalBackdrop {
   @Input() backdropClass: string;
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -6,7 +6,8 @@ import {
   Inject,
   ComponentFactoryResolver,
   ComponentRef,
-  TemplateRef
+  TemplateRef,
+  RendererFactory2
 } from '@angular/core';
 
 import {ContentRef} from '../util/popup';
@@ -25,13 +26,19 @@ export class NgbModalStack {
 
   constructor(
       private _applicationRef: ApplicationRef, private _injector: Injector, @Inject(DOCUMENT) private _document,
-      private _scrollBar: ScrollBar) {}
+      private _scrollBar: ScrollBar, private _rendererFactory: RendererFactory2) {}
 
   open(moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any, options): NgbModalRef {
     const containerEl =
         isDefined(options.container) ? this._document.querySelector(options.container) : this._document.body;
+    const renderer = this._rendererFactory.createRenderer(null, null);
 
     const revertPaddingForScrollBar = this._scrollBar.compensate();
+    const removeBodyClass = () => {
+      if (!this._modalRefs.length) {
+        renderer.removeClass(this._document.body, 'modal-open');
+      }
+    };
 
     if (!containerEl) {
       throw new Error(`The specified modal container "${options.container || 'body'}" was not found in the DOM.`);
@@ -47,10 +54,14 @@ export class NgbModalStack {
 
     this._registerModalRef(ngbModalRef);
     ngbModalRef.result.then(revertPaddingForScrollBar, revertPaddingForScrollBar);
+    ngbModalRef.result.then(removeBodyClass, removeBodyClass);
     activeModal.close = (result: any) => { ngbModalRef.close(result); };
     activeModal.dismiss = (reason: any) => { ngbModalRef.dismiss(reason); };
 
     this._applyWindowOptions(windowCmptRef.instance, options);
+    if (this._modalRefs.length === 1) {
+      renderer.addClass(this._document.body, 'modal-open');
+    }
 
     if (backdropCmptRef && backdropCmptRef.instance) {
       this._applyBackdropOptions(backdropCmptRef.instance, options);
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -9,7 +9,9 @@ import {
   TemplateRef,
   RendererFactory2
 } from '@angular/core';
+import {Subject} from 'rxjs';
 
+import {ngbFocusTrap} from '../util/focus-trap';
 import {ContentRef} from '../util/popup';
 import {isDefined, isString} from '../util/util';
 import {ScrollBar} from '../util/scrollbar';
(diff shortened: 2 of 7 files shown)


src/modal/modal-backdrop.ts
```diff
@@ -3,7 +3,8 @@ import {Component, Input} from '@angular/core';
 @Component({
   selector: 'ngb-modal-backdrop',
   template: '',
-  host: {'[class]': '"modal-backdrop fade show" + (backdropClass ? " " + backdropClass : "")'}
+  host:
+      {'[class]': '"modal-backdrop fade show" + (backdropClass ? " " + backdropClass : "")', 'style': 'z-index: 1050'}
```
pkozlowski-opensource: I'm surprised that we don't need to adjust `z-index` on windows themselves...

src/modal/modal-stack.ts
```diff
@@ -47,10 +54,12 @@ export class NgbModalStack {
 
     this._registerModalRef(ngbModalRef);
     ngbModalRef.result.then(revertPaddingForScrollBar, revertPaddingForScrollBar);
+    ngbModalRef.result.then(removeBodyClass, removeBodyClass);
     activeModal.close = (result: any) => { ngbModalRef.close(result); };
     activeModal.dismiss = (reason: any) => { ngbModalRef.dismiss(reason); };
 
     this._applyWindowOptions(windowCmptRef.instance, options);
+    renderer.addClass(this._document.body, 'modal-open');
```
pkozlowski-opensource: Could we check if there are no open modals open already and skip DOM manipulation if there is already an open modal?

pkozlowski-opensource:
@maxkorz thnx for the PR - I did a first pass on it and it looks really good! Just left minor nit comment. 

I would like to spend more time testing it but if all goes well we will land it in 3.2.0. Thnx again!

src/modal/modal-stack.ts
maxkorz: @pkozlowski-opensource why do you think this is necessary? How is it possible to open a modal without using NgbModalStack.open method?

maxkorz: Done. `addClass` will be called only when `_modalRefs` has 1 modal.

src/modal/modal-stack.ts
```diff
@@ -25,13 +26,19 @@ export class NgbModalStack {
 
   constructor(
       private _applicationRef: ApplicationRef, private _injector: Injector, @Inject(DOCUMENT) private _document,
-      private _scrollBar: ScrollBar) {}
+      private _scrollBar: ScrollBar, private _rendererFactory: RendererFactory2) {}
 
   open(moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any, options): NgbModalRef {
     const containerEl =
         isDefined(options.container) ? this._document.querySelector(options.container) : this._document.body;
+    const renderer = this._rendererFactory.createRenderer(null, null);
```
benouat: Is there any specific reason you are using `RendererFactory2` here to create a new renderer instance instead of injecting directly `private _renderer: Renderer2` in constructor ?

src/modal/modal-backdrop.ts
benouat: @pkozlowski-opensource it's ok to not adjust manually `z-index`.
According to [W3C](
http://www.w3.org/TR/CSS2/visuren.html#propdef-z-index):

> Each box belongs to one stacking context. Each positioned box in a given stacking context has an integer stack level, which is its position on the z-axis relative other stack levels within the same stacking context. Boxes with greater stack levels are always formatted in front of boxes with lower stack levels. Boxes may have negative stack levels. **Boxes with the same stack level in a stacking context are stacked back-to-front according to document tree order.**

So it should be fine for us.

benouat:
The overall code LGTM! Nonetheless I have a few comments:

- **Keyboard**: it seems broken (or simply missing) 😄. I was just playing with the new demo, open both modals, hit <kbd>Esc</kbd> -> it is closing the first one!

- **Backdrops**: Each new modal created has its own `NgbModalBackdrop`. Do we really want that behaviour ? Can't we have only one backdrop, and move it dynamically in the DOM to be in between opened modals ? Maybe it sounds kind of unnecessary optimisation ?

maxkorz:
@benouat Thanks for the review!
- **Keyboard:** I'll look into it. Sound like a bug to me. Nice catch!
- **Backdrops:** With only one backdrop the appearance of stacked modals is different:
  - 2 modals with two backdrops: https://i.imgur.com/4iv4HDu.png
  - 2 modals with one backdrop: https://i.imgur.com/LLZ27uA.png
  - 3 modals with three backgdrops: https://i.imgur.com/OGPAtX8.png
  - 3 modals with one backdrop: https://i.imgur.com/XURmXrb.png

  I personally prefer the darker(stacked) backdrop for multiple modals, so I really don't know what's the best solution here. Some users might not like the stacked backdrop. Should it be configurable?

maxkorz:
@benouat Keyboard issue has been fixed.

src/modal/modal-stack.ts
```diff
@@ -150,4 +163,17 @@ export class NgbModalStack {
...
     ngbModalRef.result.then(unregisterModalRef, unregisterModalRef);
   }
+
+  private _registerWindowCmpt(ngbWindowCmpt: ComponentRef<NgbModalWindow>) {
+    this._windowCmpts.push(ngbWindowCmpt);
+    this._activeWindowCmptHasChanged.next();
+
+    ngbWindowCmpt.onDestroy(() => {
+      const index = this._windowCmpts.indexOf(ngbWindowCmpt);
+      if (index > -1) {
```
maxkorz: I copied this check from `_registerModalRef`. Is it really necessary to check the existence of an element before removing it from an array?

benouat: With `splice()`yes it is needed, because `splice( negative-index-here, ...)` will remove from the end, e.g. : `[1, 2 , 3].splice(-1, 1)` -> `[1,  2]`

As a sidenote, I think we could only fire `this._activeWindowCmptHasChanged.next();` if we remove a componentRef. If nothing is removed it is not needed.

src/modal/modal.spec.ts
```diff
@@ -283,6 +285,51 @@ describe('ngb-modal', () => {
...
+             modalRef2.close('bar result');
+             fixture.detectChanges();
+             fixture.whenStable().then(() => {
+               expect(fixture.nativeElement).not.toHaveModal();
+               expect(document.body).not.toHaveCssClass('modal-open');
+             });
+           });
+         }));
+
+      it('should dismiss modals on ESC in correct order', () => {
```
maxkorz: <kbd>ESC</kbd> didn't work because `ngbFocusTrap` was preventing focus to go out of the first modal window.

I fixed it by moving `ngbFocusTrap` to the `NgbModalStack` service, but I'm not sure how to test it. The test seems right, but it still passes, no matter where I call `ngbFocusTrap`.

src/modal/modal-stack.ts
maxkorz: > As a sidenote, I think we could only fire this._activeWindowCmptHasChanged.next(); if we remove a componentRef. If nothing is removed it is not needed.

I subscribe to `_activeWindowCmptHasChanged` in constructor in order to call `ngbFocusTrap` on the top/activate/latest (you get the idea) `WindowCmpt`. So it should fire on `push`.

src/modal/modal.spec.ts
benouat: Your changes look good to me. No matter where we call the `ngbFocusTrap`, test does not need to be changed I think.

benouat:
@maxkorz for the backdrop I don't think we need to change anything right now. Let's keep it as it, with one new backdrop per modal. We can always refactor and change it later.

src/modal/modal-stack.ts
benouat: I was just talking about moving the call inside the `if `:

```typescript
ngbWindowCmpt.onDestroy(() => {
    const index = this._windowCmpts.indexOf(ngbWindowCmpt);
    if (index > -1) {
        this._windowCmpts.splice(index, 1);
        this._activeWindowCmptHasChanged.next();
    }
    // this._activeWindowCmptHasChanged.next();  move it inside the upper if
});

maxkorz: oh, sure, I misunderstood you.

PR #2646, opened 2018-08-27
feat(typeahead): clear input on blur and dismiss (non editable)

### LINKED ISSUE #1276: clear input (on blur) when editable false

ymeine:
Clear input on blur and popup dismiss when not editable, and when content doesn't match current model anymore.

close #1276

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -149,6 +149,7 @@ export class NgbTypeahead implements ControlValueAccessor,
   activeDescendant: string;
   popupId = `ngb-typeahead-${nextWindowId++}`;
 
+  private _model: any;
   private _onTouched = () => {};
   private _onChange = (_: any) => {};
 
@@ -183,13 +184,13 @@ export class NgbTypeahead implements ControlValueAccessor,
     const inputValues$ = this._valueChanges.pipe(tap(value => {
       this._inputValueBackup = value;
       if (this.editable) {
-        this._onChange(value);
+        this._callOnChange(value);
       }
     }));
     const results$ = inputValues$.pipe(this.ngbTypeahead);
     const processedResults$ = results$.pipe(tap(() => {
       if (!this.editable) {
-        this._onChange(undefined);
+        this._callOnChange(undefined);
       }
     }));
     const userInput$ = this._resubscribeTypeahead.pipe(switchMap(() => processedResults$));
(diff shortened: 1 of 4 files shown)


src/typeahead/typeahead.ts
```diff
@@ -370,4 +377,13 @@ export class NgbTypeahead implements ControlValueAccessor,
...
     this._subscription = null;
   }
+
+  private _callOnChange(value: any) {
+    this._model = value;
+    this._onChange(value);
+  }
+
+  private _inputDifferentFromModel(): boolean {
+    return this._formatItemForInput(this._model) !== this._elementRef.nativeElement.value;
```
pkozlowski-opensource: I don't think this is right since:
* several model values can lead to the same formatted string;
* we are reading from the DOM

We should probably listen to model changes instead to know if it was changed or not.

jnizet:
FYI, I do that on one of my applications by applying a separate directive on the typeaheads:

```
/**
 * Directive which automatically clears the input on blur when an ngbTypeahead is applied to the input
 * if its editable input is false and the value in the model is falsy (indicating that the entered value is not
 * a valid value). It also applies an is-warning CSS class in the same conditions.
 */
@Directive({
  selector: '[ngbTypeahead]'
})
export class NonEditableTypeaheadDirective {

  constructor(private ngControl: NgControl,
              private elementRef: ElementRef<HTMLInputElement>,
              private typeahead: NgbTypeahead) { }

  @HostListener('blur')
  onBlur() {
    if (!this.ngControl.value && !this.typeahead.editable) {
      this.elementRef.nativeElement.value = '';
    }
  }

  @HostBinding('class.is-warning')
  get isWarning() {
    return !this.typeahead.editable  && this.elementRef.nativeElement.value && !this.ngControl.value;
  }
}
```

ymeine:
I'm going to close this one due to the long inactivity, but also because of those facts: 

- I am myself quite confused with the code, and even if it's true that two different model items can lead to a same formatted input value, it's also true that the feature is just related to keeping a relevant input value, no matter what the real committed model item is. There's this round trip "I type, it searches and formats to text to build an options list, user selects and it formats to string - potentially differently - to write into the input", and it gets confusing
- the feature is controversial in my opinion: I better not mess with data entered by the user. Processing, giving feedback, yes, but things like removing automatically, especially on something as easy as a blur which can be accidental, I would say it's bad user experience
- apparently, it can be done on the user side if they could assess that it is actually wanted, which is the better choice for controversial features

PR #2649, opened 2018-08-27
fix(demo): NgbdExamplesPage must not be OnPush

benouat:
`NgbdExamplesPage` has been defined to be `ChangeDetectionStrategy.OnPush` on last demos website refactor (#2520).

This breaks some demos. It must *not* be OnPush


pkozlowski-opensource:
>  It must not be OnPush

This is interesting - doesn't it mean that some of our components can't be used inside of `OnPush` components?

> This breaks some demos.

Could you share a list of demos that were broken?  Do we know why?

PR #2661, opened 2018-08-29
fix(popover): don't show header if title is empty

### LINKED ISSUE #2653: popover shows title area even if no title is specified
### Bug description:

If a project sets `{preserveWhitespaces: true}` and opens a popover without a title, the popover shows a title area without text rather than no title area.

http://next.plnkr.co/edit/8rFzdu1oGmpv5HpW

Angular: 6.1.3

ng-bootstrap: 3.1.0

Bootstrap: 4.1.3

### Analysis

Bootstrap attempts to suppress empty titles with 
```
.popover-header:empty {
  display: none
}
```

but a popover is not empty if it contains whitespace ...

### Suggested fix

Use ngIf to suppress an empty title - or set `preserveWhitespaces: false` in your component decorator.

bedag-moo:
by removing the element from the DOM

Closes #2653


src/popover/popover.ts
```diff
@@ -43,7 +43,7 @@ let nextId = 0;
   },
   template: `
     <div class="arrow"></div>
-    <h3 class="popover-header">
+    <h3 class="popover-header" *ngIf="title">
```
pkozlowski-opensource: Could you please change it to `title != null` (which will catch both `null` and `undefined`) to be explicit and avoid all discussion around passing `0, ""` and other falsy values?

bedag-moo: Good point; done.

PR #2668, opened 2018-08-30
fix(datepicker): no longer trigger onChange when the value is equivalent

divdavem:

### DIFF
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -149,6 +149,7 @@ export class NgbDatepicker implements OnDestroy,
     OnChanges, OnInit, ControlValueAccessor {
   model: DatepickerViewModel;
 
+  private _controlValue: NgbDate;
   private _subscription: Subscription;
   private _selectSubscription: Subscription;
   /**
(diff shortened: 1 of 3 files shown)


src/datepicker/datepicker.ts
```diff
@@ -149,6 +149,7 @@ export class NgbDatepicker implements OnDestroy,
     OnChanges, OnInit, ControlValueAccessor {
   model: DatepickerViewModel;
 
+  private _controlValue: NgbDate;
```
maxokorokov: Do you mind if we move this to the datepicker service?

Not sure it it makes a huge difference, but the idea was to keep any state away from component.

I would have added an option:

```ts
this._service.select(..., {fromControl: true});
```

in this case the service's `select` would to something like:

```ts
if (isChangedDate(this._state.selectedDate, selectedDate)) {
        this._nextState({selectedDate, controlValue: options.fromControl ? selectedDate : undefined});
      }
```

P.S. names `formControl` and `controlValue` are just examples

divdavem: Thank you for your comment! It is a good idea! I will update my PR.

divdavem: Well, in fact, as discussed, I think it is better as it is now (as we need to update `controlValue` again when calling `onChange`).

divdavem:
@maxokorokov Thank you for your review!

src/datepicker/datepicker.ts
maxokorokov: Yep, as discussed, simpler to leave it as is

PR #2699, opened 2018-09-10
fix(popover): unregister event listener functions only if present

### LINKED ISSUE #2688: Popover: this._unregisterListenersFn is not a function
### Bug description:

The same issue that happened with tooltip described in issue #2199 is also present in popover due to the same code being used in the ngOnInit and ngOnDestroy

maxokorokov:
Same case as for the tooltip in #2205 

Fixes #2688


pkozlowski-opensource:
LGTM

PR #2716, opened 2018-09-14
feat(datepicker): allow to pass user data to the day template context

maxokorokov:
Many custom-day-display customisation cases can be solved by allowing to pass arbitrary user data to the day template context:

```html
<ng-template #t let-date="date" let-data="data">
   {{ data }}    <!-- will display 'custom data' -->
<ng-template>

<ngb-datepicker [dayTemplate]="t" [dayTemplateData]="dayTemplateData"></ngb-datepicker>
```
```ts
dayTemplateData(date: NgbDate, currentMonth: {month, year}): any {
  return 'custom data';
}
```

The big advantage is that this data is computed once when month view is constructed.

### DIFF
--- a/src/datepicker/datepicker-tools.ts
+++ b/src/datepicker/datepicker-tools.ts
@@ -119,7 +119,7 @@ export function buildMonths(
 export function buildMonth(
     calendar: NgbCalendar, date: NgbDate, state: DatepickerViewModel, i18n: NgbDatepickerI18n,
     month: MonthViewModel = {} as MonthViewModel): MonthViewModel {
-  const {minDate, maxDate, firstDayOfWeek, markDisabled, outsideDays} = state;
+  const {dayTemplateData, minDate, maxDate, firstDayOfWeek, markDisabled, outsideDays} = state;
 
   month.firstDate = null;
   month.lastDate = null;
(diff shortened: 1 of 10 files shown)


src/datepicker/datepicker.spec.ts
```diff
@@ -320,14 +322,6 @@ describe('ngb-datepicker', () => {
     expectMaxDate(10000, 1);
   });
 
-  it('should support disabling dates via callback', () => {
```
maxokorokov: This was just a duplicated test

src/datepicker/datepicker-tools.ts
```diff
@@ -170,9 +174,13 @@ export function buildMonth(
         dayObject = days[day] = {} as DayViewModel;
       }
       dayObject.date = newDate;
-      dayObject.context = Object.assign(
-          dayObject.context || {},
-          {date: newDate, currentMonth: month.number, disabled, focused: false, selected: false});
+      dayObject.context = Object.assign(dayObject.context || {}, {
+        date: newDate,
```
benouat: Could we take the opportunity here to also introduce `$implicit` usage ?

```ts
$implicit: newDate,
date: newDate, // we can keep that one for backward compat
```

It simply allows people to use the short `let` notation such as:

```html
<ng-template #t let-date>
```

maxokorokov: Thanks for the suggestion and the review. 
I'll just open another PR with this feature to keep things separate.

PR #2729, opened 2018-09-20
feat(datepicker): add $implicit property for day template context

maxokorokov:
Allows using $implicit shortcut for the day template context (previous `let-d='date'` also works):

```html
<ng-template #t let-d> {{ d.day }}</ng-template>
<ngb-datepicker [dayTemplate]="t"></ngb-datepicker>
```

cc @benouat

### DIFF
--- a/demo/src/app/components/datepicker/overview/datepicker-overview.component.html
+++ b/demo/src/app/components/datepicker/overview/datepicker-overview.component.html
@@ -209,6 +209,13 @@
   </p>
 
   <ngbd-code lang="html" [code]="snippets.dayTemplate"></ngbd-code>
+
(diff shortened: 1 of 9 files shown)


demo/src/app/components/datepicker/overview/datepicker-overview.component.html
```diff
@@ -209,6 +209,12 @@ <h4>Input date parsing and formatting</h4>
   </p>
 
   <ngbd-code lang="html" [code]="snippets.dayTemplate"></ngbd-code>
+
+  <p>
```
benouat: what about using an `ngb-alert` here ? type `warning` or `info`

PR #2733, opened 2018-09-20
fix(datepicker): mark year 0 as invalid in Gregorian calendar

### LINKED ISSUE #2721: Year 0000 considered as valid in Datepicker
**Bug description**:
In the popup Datepicker a date string in which the year is 0000 is marked as valid. The value of the ngModel is null, but the status is "VALID" which is confusing. In my opinion these dates should be marked as "INVALID".
E.g 0000-04-09

**Link to plunker**: http://plnkr.co/edit/4dTDk1?p=preview

**Version of Angular, ng-bootstrap, and Bootstrap**:

Angular: 6.1.3

ng-bootstrap: 3.2.0

Bootstrap: 4.0.0-alpha.6

maxokorokov:
Fixes #2721


pkozlowski-opensource:
LGTM

Issue #2728, opened 2018-09-20
Improve modal focus handling with NgbAutoFocus directive and first element focus


maxokorokov:
- Allow users to set the initial focus inside the modal with `NgbAutoFocus` directive. 

```html
<!-- will be focused after the modal opens -->
<input type="text" ngbAutoFocus />
``` 

- If there is no `ngbAutoFocus` in the modal, focus first focusable element by default. Allow user to cancel this behaviour and focus nothing by doing 

```ts
modal.open(content, {focusFirst: false})
```

- add a demo for focus handling in a separate PR

---
Not sure if this is a breaking change and not sure if we need `ngbAutoFocus` as a directive or a simple attribute would do.

benouat:
I'll work on that one.

peterblazejewicz:
hi folks, just wondering why the custom (and new) directive instead of using native `autofocus`?
(I know that part, it does not work in first place, dynamic content behaviour, etc). But in the guts of the pending PR plain selector is being using:
```js
const elementToFocus = this._elRef.nativeElement.querySelector(`[ngbAutofocus]`) as HTMLElement;
```
https://github.com/ng-bootstrap/ng-bootstrap/pull/2737/files#diff-8fb1cd1cc4cf9a5207c30f1474b6c8b9R67

and I think this does not differ much from:

```js
const elementToFocus = this._elRef.nativeElement.querySelector(`[autofocus]`) as HTMLElement;
```

Thanks!

benouat:
@peterblazejewicz we decided not to use the native `[autofocus]` attribute mostly for 2 reasons:

1. As you said, it does not work in first place, dynamic content behaviour, etc...
2. Using it would be confusing for people not aware of point 1. They would think it is natively working, which is not true.

The feature is provided by ng-bootstrap, hence the `ngb` prefix, no one can be fooled.
As a sidenote, material is using the same technique with a plain attribute, and theirs is name `cdkFocusInitial`

kedar9444:
@maxokorokov  : modal.open(content, {focusFirst: false}) is this feature removed from V4, because i am not able to see `focusFirst` property in NgbModalOptions. if yes how to achieve a situation where i don't want to set any default focus but should be able to close the modal with `ESC` key.

janwidmer:
[...]

PR #2737, opened 2018-09-21
feat(modal): [ngbAutofocus] option

### LINKED ISSUE #2718: bug(modal): the first press of shift + tab does not focus an element in the modal, focus is not trapped
https://stackblitz.com/angular/ygpjvdbxbre

1. Launch demo modal
2. Press <kbd>Tab</kbd> - first element in the modal container gets the focus - focus is trapped ✔
3. Keep pressing <kbd>Tab</kbd> - focus is looped ✔
4. Press <kbd>Shift</kbd> + <kbd>Tab</kbd> - focus is going to a previous element, focus is still trapped  ✔

<hr>

1. Launch demo modal
2. Press <kbd>Shift</kbd> + <kbd>Tab</kbd> - element outside of the modal container gets the focus - focus is not trapped ❌

P.S. it works fine in [Datepicker in a popup](https://stackblitz.com/angular/nvpvpagxmoxo). 
1. Click on the calendar icon
2. Press <kbd>Shift</kbd> + <kbd>Tab</kbd> - focus is trapped and looped ✔

benouat:
As mentioned in #2728 this PR introduced 2 new features:

+ ability to use a new `ngbAutofocus` attribute in your modal templates 
  ```html
  <!-- will be focused after the modal opens -->
  <input type="text" ngbAutofocus />
  ```

+ introduction of a new `focusFirst` Modal Config option (will basically works as `ngbAutofocus` but finding automatically the first focusable element)
  ```typescript
  /**
   * Assign focus to the first focusable element when opened (true by default)
   */
  focusFirst?: boolean;
  ```

---

Any newly opened modal will now do in that order:

- If `ngbAutofocus` is detected, it will be used to focus the decorated element.
- Else, it will look at `focusFirst` value (default to be `true`), find the first focusable element, and focus it.
- At this point, if nothing has been focus (not focusable element found, or `focusFirst` set to `false`) it will simply focus itself (the modal window).

Another PR will be opened to demonstrate the usage of these new features in the Modal demos

Closes #2718

### DIFF
--- a/misc/stackblitz-gen.js
+++ b/misc/stackblitz-gen.js
@@ -10,7 +10,8 @@ const versions = getVersions();
 
 const ENTRY_CMPTS = {
   'modal-component': ['NgbdModalContent'],
-  'modal-stacked': ['NgbdModal1Content', 'NgbdModal2Content']
+  'modal-stacked': ['NgbdModal1Content', 'NgbdModal2Content'],
+  'modal-focus': ['NgbdModalConfirm', 'NgbdModalConfirmAutofocus']
 };
 
 function generateDemosCSS() {
@@ -29,7 +30,7 @@ function generateStackblitzContent(componentName, demoName) {
 <body>
   <form id="mainForm" method="post" action="${stackblitzUrl}">
     <input type="hidden" name="description" value="Example usage of the ${componentName} widget from https://ng-bootstrap.github.io">
-${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName)])}  
+${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName)])}
 
     <input type="hidden" name="files[.angular-cli.json]" value="${he.encode(getStackblitzTemplate('.angular-cli.json'))}">
     <input type="hidden" name="files[index.html]" value="${he.encode(generateIndexHtml())}">
@@ -41,7 +42,7 @@ ${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName
     <input type="hidden" name="files[app/app.component.html]" value="${he.encode(generateAppComponentHtmlContent(componentName, demoName))}">
     <input type="hidden" name="files[app/${fileName}.ts]" value="${he.encode(codeContent)}">
     <input type="hidden" name="files[app/${fileName}.html]" value="${he.encode(markupContent)}">
-    
+
     <input type="hidden" name="dependencies" value="${he.encode(JSON.stringify(generateDependencies()))}">
   </form>
   <script>document.getElementById("mainForm").submit();</script>
@@ -60,12 +61,13 @@ function generateIndexHtml() {
   <head>
     <title>ng-bootstrap demo</title>
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/${versions.bootstrap}/css/bootstrap.min.css" />
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.15.0/themes/prism.css" />
   </head>
 
   <body>
     <my-app>loading...</my-app>
   </body>
-  
+
 </html>`;
 }
 
@@ -74,7 +76,7 @@ function generateAppComponentHtmlContent(componentName, demoName) {
 
   return `
 <div class="container-fluid">
-    
+
   <hr>
 
   <p>
@@ -110,10 +112,10 @@ import { AppComponent } from './app.component';
 import { ${demoImports} } from '${demoImport}';
 
 @NgModule({
-  imports: [BrowserModule, FormsModule, ReactiveFormsModule, HttpClientModule, NgbModule.forRoot()], 
+  imports: [BrowserModule, FormsModule, ReactiveFormsModule, HttpClientModule, NgbModule],
   declarations: [AppComponent, ${demoImports}]${entryCmptClasses ? `,\n  entryComponents: [${entryCmptClasses}],` : ','}
   bootstrap: [AppComponent]
-}) 
+})
 export class AppModule {}
 `;
 }
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -1,16 +1,17 @@
 import {DOCUMENT} from '@angular/common';
 import {
-  Component,
-  Output,
-  EventEmitter,
-  Input,
-  Inject,
-  ElementRef,
-  OnInit,
   AfterViewInit,
-  OnDestroy
+  Component,
+  ElementRef,
+  EventEmitter,
+  Inject,
+  Input,
+  OnDestroy,
+  OnInit,
+  Output
 } from '@angular/core';
 
+import {getFocusableBoundaryElements} from '../util/focus-trap';
 import {ModalDismissReasons} from './modal-dismiss-reasons';
 
 @Component({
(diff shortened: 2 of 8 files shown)


benouat:
After discussing with @pkozlowski-opensource I will update PR to drop the config option `focusFirst` and will hardcode `focusFirst: true` behaviour in the code.

So new rules are like so:

- If `ngbAutofocus` is detected, it will be used to focus the decorated element.
- Otherwise first tabbable element will get focused.
- If no element matches these 2 previous rules, fallback is to give focus to modal window.

src/modal/modal-config.ts
```diff
@@ -62,6 +62,12 @@ export interface NgbModalOptions {
    * @since 1.1.0
    */
   backdropClass?: string;
+
+  /**
+   * Assign focus to the first focusable element when opened (true by default)
+   * @since 3.3.0
```
pkozlowski-opensource: Please don't add the version since we don't know when it is going to be merged. I will add those when cutting a release.

src/modal/modal-config.ts
```diff
@@ -62,6 +62,12 @@ export interface NgbModalOptions {
...
+   */
+  focusFirst?: boolean;
```
pkozlowski-opensource: My understanding was that we don't want to introduce this option for now. Did anything change?

src/modal/modal-config.ts
```diff
@@ -75,4 +81,5 @@ export interface NgbModalOptions {
 export class NgbModalConfig implements NgbModalOptions {
   backdrop: boolean | 'static' = true;
   keyboard = true;
+  focusFirst = true;
```
pkozlowski-opensource: My understanding was that we don't want to introduce this option for now. Did anything change?

src/modal/modal-stack.ts
```diff
@@ -21,7 +21,8 @@ import {NgbModalWindow} from './modal-window';
 
 @Injectable({providedIn: 'root'})
 export class NgbModalStack {
-  private _windowAttributes = ['ariaLabelledBy', 'backdrop', 'centered', 'keyboard', 'size', 'windowClass'];
+  private _windowAttributes =
+      ['ariaLabelledBy', 'backdrop', 'centered', 'keyboard', 'size', 'windowClass', 'focusFirst'];
```
pkozlowski-opensource: My understanding was that we don't want to introduce this option for now. Did anything change?

src/modal/modal-config.ts
benouat: Ok, I will remove it !

src/modal/modal-config.ts
benouat: Did you see any new code push to this PR branch between my comment where I said I will remove it and now ? 😜

It will be removed...

benouat:
@pkozlowski-opensource I made changes according to your review and also added a new demo for both explanation and illustration of the feature.

src/modal/modal-window.ts
```diff
@@ -62,7 +63,17 @@ export class NgbModalWindow implements OnInit,
 
   ngAfterViewInit() {
     if (!this._elRef.nativeElement.contains(document.activeElement)) {
-      this._elRef.nativeElement['focus'].apply(this._elRef.nativeElement, []);
+      let elementToFocus = this._elRef.nativeElement.querySelector(`[ngbAutofocus]`) as HTMLElement;
+      if (!elementToFocus) {
```
mboughaba: suggestion: Can you please merge this two level if statement, it looks maybe confusing (at least form me)

src/modal/modal-window.ts
```diff
@@ -62,7 +63,17 @@ export class NgbModalWindow implements OnInit,
...
+        const firstFocusableEl = getFocusableBoundaryElements(this._elRef.nativeElement)[0];
+        if (firstFocusableEl) {
+          elementToFocus = firstFocusableEl;
+        }
+      }
+      if (!elementToFocus) {
+        elementToFocus = this._elRef.nativeElement;
+      }
+      elementToFocus.focus();
```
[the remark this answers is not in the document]
benouat: Changes have been applied

benouat:
I am having a look at failing stackblitz

misc/stackblitz-gen.js
```diff
@@ -57,15 +58,16 @@ function generateIndexHtml() {
   return `<!DOCTYPE html>
 <html>
 
-  <head>
+  <head>xx
```
pkozlowski-opensource: xx?

misc/stackblitz-gen.js
```diff
@@ -57,15 +58,16 @@ function generateIndexHtml() {
...
     <title>ng-bootstrap demo</title>
     <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/${versions.bootstrap}/css/bootstrap.min.css" />
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.15.0/themes/prism.css" />
```
pkozlowski-opensource: Why?

misc/stackblitz-gen.js
```diff
@@ -110,10 +112,10 @@ import { AppComponent } from './app.component';
 import { ${demoImports} } from '${demoImport}';
 
 @NgModule({
-  imports: [BrowserModule, FormsModule, ReactiveFormsModule, HttpClientModule, NgbModule.forRoot()], 
+  imports: [BrowserModule, FormsModule, ReactiveFormsModule, HttpClientModule, NgbModule.forRoot()],
```
pkozlowski-opensource: There are ws changes in this file and others - why?

BTW, I know it is not related to your change, but why do we have `NgbModule.forRoot()` here?

src/modal/modal-window.ts
```diff
@@ -76,8 +81,7 @@ export class NgbModalWindow implements OnInit,
     } else {
       elementToFocus = body;
     }
-    elementToFocus['focus'].apply(elementToFocus, []);
-
+    elementToFocus.focus();
```
pkozlowski-opensource: Do you know why we've called it through `apply` before and don't do this here any more?

src/modal/modal.module.ts
```diff
@@ -1,8 +1,8 @@
-import {NgModule, ModuleWithProviders} from '@angular/core';
+import {ModuleWithProviders, NgModule} from '@angular/core';
 
+import {NgbModal} from './modal';
 import {NgbModalBackdrop} from './modal-backdrop';
 import {NgbModalWindow} from './modal-window';
-import {NgbModal} from './modal';
```
pkozlowski-opensource: Why there are changes in this file?

misc/stackblitz-gen.js
benouat: We never updated it i think... I will remove them

misc/stackblitz-gen.js
benouat: what the o_O ?? I have no clue why there are here, will remove them of course

misc/stackblitz-gen.js
benouat: Because of a snippet of code I have in the demo

image

src/modal/modal-window.ts
benouat: I have no clue at all why _you_ wrote all of these with `apply()`. The only thing I know is that it seems to be not needed anymore. It works fine without it. So it makes the code easier to read

pkozlowski-opensource:
LGTM

PR #2740, opened 2018-09-21
feat(modal): allow checking if there are open modals

### LINKED ISSUE #1600: Add possibility to check if modalRef is shown
### Bug description:
Pure bootstrap has event listener to check when modal is shown. NgbModal - no :(
But I think, this is a good idea (for example in some bad UI cases i need to hide other modals, and i need to check whether current modal is shown or hidden)

Angular: 4.0.1

ng-bootstrap: 1.0.0-alpha.26

Bootstrap:

pkozlowski-opensource:
Closes #1600

A tiny PR for a good start of the weekend!

### DIFF
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -31,4 +31,9 @@ export class NgbModal {
    * @since 3.1.0
    */
   dismissAll(reason?: any) { this._modalStack.dismissAll(reason); }
+
+  /**
+   * Indicates if there are currently any open modal windows in the application.
+   */
+  hasOpenModals(): boolean { return this._modalStack.hasOpenModals(); }
 }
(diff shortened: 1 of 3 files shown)


src/modal/modal.ts
```diff
@@ -31,4 +31,9 @@ export class NgbModal {
    * @since 3.1.0
    */
   dismissAll(reason?: any) { this._modalStack.dismissAll(reason); }
+
+  /**
+   * Indicates if there are currently any open modal windows in the application.
+   */
+  hasOpenWindows(): boolean { return this._modalStack.hasOpenWindows(); }
```
benouat: Some nitpicking here, but I would have named it `hasOpenModals()`
Even your commit message is using open modals and not windows

pkozlowski-opensource: Ok, thnx, will change it to `hasOpenModals()`

Issue #2626, opened 2018-08-17
Popover opens over button with floating html content


ImNightwing:
### Bug description:
Popover opens over button with left-top with floating dynamic html
### Link to minimally-working plunker that reproduces the issue:
https://angular-a5funs.stackblitz.io

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 6.1.3

ng-bootstrap: 3.0.0

Bootstrap: 4.1.3

m-przybylski:
Working on that issue.

m-przybylski:
I did some investigations and this is what I found:
Popover apply position in 2 places. The first one is just after element has been added to DOM and the second time is when component is stable (CD has nothing to do any more). But still this is not enough because browser is some edge cases is not able to apply all the styles to the element, so placement is wrongly applied.

![screen shot 2018-10-05 at 12 29 30](https://user-images.githubusercontent.com/1764395/46530544-616f4100-c89a-11e8-8ff0-89490721ddd1.png)

IMO to avoid that is to wait till next frame and than calculate proper position of the element

fbasso:
#2972 has fixed the issue. In the stackblitz example, there is a width issue (see screenshot above), but here it's a matter of html/bootstrap markup inside the popover content.

This issue can be closed.

![image](https://user-images.githubusercontent.com/1152930/52415241-c061f280-2ae6-11e9-9676-49933e10f795.png)

maxokorokov:
Closed in #2972

fbasso:
To be complete about this issue, the popover width has a  'constraint' because of its container. Adding container="body" solve the width problem:

![image](https://user-images.githubusercontent.com/1152930/52416036-f43e1780-2ae8-11e9-884e-fadea27795b9.png)

PR #2751, opened 2018-09-29
fix(popover): align popover correctly

m-przybylski:
closes: #2626

### DIFF
--- a/src/popover/popover.spec.ts
+++ b/src/popover/popover.spec.ts
@@ -9,7 +9,7 @@ import {
   Injectable,
   OnDestroy,
   TemplateRef,
-  ViewContainerRef
+  ViewContainerRef,
 } from '@angular/core';
 
 import {Key} from '../util/key';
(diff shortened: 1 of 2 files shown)


pkozlowski-opensource:
I've got 2 issues with this PR:
* I don't see any analysis of the problem - what is the root cause? Why adding RAF fixes it?
* this approach makes all tests async which wasn't the case before.

I'm very uncomfortable merging this one before we clearly understand what is the root cause and if adding RAF is the only solution.

m-przybylski:
@maxokorokov sorry, no can do. It requires to be formatted in that. Commit will not pass validation. The issue is that tests are now asyc, so additional decorator method is applied to all tests.

maxokorokov:
@m-przybylski was talking about reformatted trailing commas, removed/added new lines and reformatted module imports/providers. Not test cases themselves.

I also don't see how just marking tests as `fakeAsync` without anything else makes it work.

Issue #2660, opened 2018-08-29
Failing commit hook prevents all commits to ng-bootstrap


bedag-moo:
### Bug description:
```
$ git commit
husky > npm run -s commitmsg (node v8.11.3)

C:\dev\git-repositories\ng-bootstrap\misc\validate-commit.js:93
  var incorrectLogFile = commitMsgFile.replace('COMMIT_EDITMSG', 'logs/incorrect-commit-msgs');
                                       ^

TypeError: Cannot read property 'replace' of undefined
    at Object.<anonymous> (C:\dev\git-repositories\ng-bootstrap\misc\validate-commit.js:93:40)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:191:16)
    at bootstrap_node.js:612:3

husky > commit-msg hook failed (add --no-verify to bypass)
```

This doesn't appear to be related to the commit message I used, because the validation scripts fails before it even finds the message.

I am now going to disable this useless script so I can get on with my PR.

mendeza:
Did you try `--no-verify` ?

bedag-moo:
No -- neutering the script seemed a more permanent solution. Besides, never having needed `--no-verify` before, it was not immediately clear where the option needed to be specified (git, or to the validation script directly?), and whether and how it could be specified when committing through an IDE.

pkozlowski-opensource:
Can't reproduce it locally I'm afraid and none of the contributors bumps into it. As such you are in the best position to reproduce, investigate and fix. If you care about this item please investigate and send a PR. We can't take action on our end as we can't reproduce...

divdavem:
Actually, I had the same issue and I did exactly the same thing: removing the hooks.
This issue seems to be reproducible on Windows only.

divdavem:
@m-przybylski Thank you for taking the time to investigate and open a PR. 👍 
It solves the issue on my Windows machine too!

kukac7:
this is a problem for me.

```
> git add .
> git commit -m "sry validate commit"
husky > npm run -s commitmsg (node v10.11.0)

/Users/XXX/ng-bootstrap/misc/validate-commit.js:93
[...]

PR #2753, opened 2018-09-29
fix(hook): git commit msg hook fail

m-przybylski:
closes: #2660

### DIFF
--- a/misc/validate-commit.js
+++ b/misc/validate-commit.js
@@ -89,7 +89,7 @@ exports.validateMessage = validateMessage;
 
 // hacky start if not run by jasmine :-D
 if (process.argv.join('').indexOf('jasmine-node') === -1) {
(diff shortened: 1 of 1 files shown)


pkozlowski-opensource:
Couple of thoughts: 
- I can see that there is a "fix" that works for people but I don't see analysis of the root cause - how can we be sure that it fixes things all and for good?
- more generally - do we need a hook locally? 

Personally my vote would go to removing this hook from local altogether and execute the check on CI only.

I don't mind merging this one as a temporary measure but I would like to move out of local git hooks. Opened #2761 for this.

kukac7:
it looks good in me at local.

pkozlowski-opensource:
Going to merge it as it seems to help some people, but we are going to remove local hooks via #2761 in the long run.

Issue #2743, opened 2018-09-24
Track "Open demo in Stackblitz" button clicks


benouat:
Changing the code of the Open is Github button for #2742 I realised we have analytics setup here

```typescript
trackSourceClick() {
  this._analytics.trackEvent('Source File View', this.apiDocs.className);
}`
```

We should also add this kind of tracking when people click on Stackblitz buttons.

PR #2756, opened 2018-10-01
 chore(demo): add stackblitz ga event tracker

kukac7:
#2743


pkozlowski-opensource:
LGTM (apart from multiple commits, of course :-))

PR #2762, opened 2018-10-05
fix(accordion): don't submit HTML form when toggling panels

### LINKED ISSUE #2760: Toggle method of NgbAccordion is executed on pressing Enter key in case of using accordion inside form
### Bug description:
In case of using NgbAccordion inside a form, its panels can be opened/closed by pressing the Enter key.
The reason of bug: 
https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/accordion/accordion.ts#L119
Button's **type** attribute isn't specified. The default is 'submit'.

pkozlowski-opensource:
Fixes #2760

I've also checked all other widgets and all seems fine - only accordion was missing `type="button"`

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -116,7 +116,8 @@ export interface NgbPanelChangeEvent {
       <div class="card">
         <div role="tab" id="{{panel.id}}-header" [class]="'card-header ' + (panel.type ? 'bg-'+panel.type: type ? 'bg-'+type : '')">
           <h5 class="mb-0">
-            <button class="btn btn-link" (click)="!!toggle(panel.id)" [disabled]="panel.disabled" [class.collapsed]="!panel.isOpen"
+            <button type="button" class="btn btn-link"
+              (click)="toggle(panel.id)" [disabled]="panel.disabled" [class.collapsed]="!panel.isOpen"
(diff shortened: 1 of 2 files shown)


src/accordion/accordion.ts
```diff
@@ -116,7 +116,8 @@ export interface NgbPanelChangeEvent {
       <div class="card">
         <div role="tab" id="{{panel.id}}-header" [class]="'card-header ' + (panel.type ? 'bg-'+panel.type: type ? 'bg-'+type : '')">
           <h5 class="mb-0">
-            <button class="btn btn-link" (click)="!!toggle(panel.id)" [disabled]="panel.disabled" [class.collapsed]="!panel.isOpen"
+            <button type="button" class="btn btn-link"
+              (click)="!!toggle(panel.id)" [disabled]="panel.disabled" [class.collapsed]="!panel.isOpen"
```
benouat: You could remove the `!!` here as it is not needed anymore. It used to be needed when the `<button>` was an `<a>`.

Returning `false` to prevent default only work on `<a>` tag (and it is not standard, and should not be used anymore BTW).

On a button it is effect-less. Main issue here was definitely only related to the missing type.

pkozlowski-opensource: Good point @benouat - thnx, removed it.

pkozlowski-opensource:
@maxokorokov thnx, it is merged now.

PR #2763, opened 2018-10-05
fix: remove void type expressions in templates for TS 3.1

maxokorokov:


pkozlowski-opensource:
LGTM. Let's merge when TravisCI gets green.

PR #2769, opened 2018-10-08
chore: use TS for static file generation instead of Gulp

maxokorokov:
- migrate JS scripts we have to TS
- remove not relevant anymore `gen-doc.js`

New scripts to replace old gulp tasks are:
- `copy-static-files.ts` → copy `LICENSE` and `README.md` for npm
- `generate-docs.ts` → extract docs from ts sources
- `generate-stackblitzes.ts` → generate stackblitzes from demo sources

Stackblitz generation is not in `types: strict` mode yet, as there would be huge changes


misc/generate-stackblitzes.ts
```diff
@@ -37,7 +42,7 @@ ${generateTags(['Angular', 'Bootstrap', 'ng-bootstrap', capitalize(componentName
     <input type="hidden" name="files[main.ts]" value="${he.encode(getStackblitzTemplate('main.ts'))}">
     <input type="hidden" name="files[polyfills.ts]" value="${he.encode(getStackblitzTemplate('polyfills.ts'))}">
     <input type="hidden" name="files[styles.css]" value="${he.encode(generateDemosCSS())}">
-    <input type="hidden" name="files[app/app.module.ts]" value="${he.encode(generateAppModuleTsContent(componentName, demoName, `${basePath}.ts`))}">
+    <input type="hidden" name="files[app/app.module.ts]" value="${he.encode(generateAppModuleTsContent(componentName, demoName, basePath + '.ts'))}">
```
maxokorokov: this is because of `clang-format`

maxokorokov:
@benouat and you don't need to specify which files to watch

PR #2776, opened 2018-10-10
fix(carousel): change sildes on prev/next/select API calls with OnPush

### LINKED ISSUE #2766: Bug in Carousel for external Previous/Next buttons
### Bug description:

Steps to reproduce:

1. Click "Next" button - observe nothing happens.
2. Move your mouse into the image - observe the Carousel advance on mouseenter.

Other Notes:

* If you click the Next button twice, then mouseenter the image, it will advance 2 images (corresponding exactly to number of clicks). It's correctly keeping track of active slide, but not advancing until mouse enter.
* Same behavior (in reverse) for Previous button.

StackBlitz sample that reproduces issue:  https://stackblitz.com/edit/ng-bootstrap-carousel-bug

Angular: 6.1.0

ng-bootstrap: 3.3.0

Bootstrap:

maxokorokov:
Regression introduced after marking carousel OnPush due to lack of tests. Now calling `detectChanges()` on interval and all public API actions.

Fixes #2766

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -147,10 +147,7 @@ export class NgbCarousel implements AfterContentChecked,
             .pipe(
                 map(() => this.interval), filter(interval => interval > 0 && this.slides.length > 0),
                 switchMap(interval => timer(interval).pipe(takeUntil(this._stop$))))
-            .subscribe(() => this._ngZone.run(() => {
-              this.next();
-              this._cd.detectChanges();
-            }));
+            .subscribe(() => this._ngZone.run(() => this.next()));
 
         this._start$.next();
       });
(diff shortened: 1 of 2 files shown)


src/carousel/carousel.ts
```diff
@@ -202,6 +199,9 @@ export class NgbCarousel implements AfterContentChecked,
       this._start$.next();
       this.activeId = selectedSlide.id;
     }
+
+    // we get here after the interval fires or any external API call like next(), prev() or select()
+    this._cd.detectChanges();
```
pkozlowski-opensource: Shouldn't this be `markForCheck`? What would happen in a scenario where click on prev / next is calling prev / next on a carousel _and_ is adding a new slide (ex. from `ngFor`). Would the content of a carousel be checked if we just call detect changes on the carousel itself?

pkozlowski-opensource: > markForCheck would only refresh it on next CD, isn't it correct ?

I would say: in the current CD (the one that would be run at the end of JS loop, before any rendering is done). So I believe that this is the behaviour we actually want.

We should add a test for the above scenario to be sure.

maxokorokov:
@pkozlowski-opensource, sure shouldn't run CD synchronously right away, will change to `markForCheck`

Actually you're right, you're example won't work with `detectChanges`, didn't get it at first. Added a test for it

pkozlowski-opensource:
LGTM. Feel free to merge when CI gets green.

Issue #1875, opened 2017-09-28
Tooltip arrows misaligned if ViewEncapsulation.None is set globally.


bfbares:
### Bug description:

Tooltip arrows misaligned if ViewEncapsulation.None is set globally.

### Link to minimally-working plunker that reproduces the issue:

[http://plnkr.co/edit/cAtUk1W8m9emOlZHBkwZ?p=preview](http://plnkr.co/edit/cAtUk1W8m9emOlZHBkwZ?p=preview)

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.0.3

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta

mschoudry:
Hi,
Component level styles are used in ngbootstrap by some components(tooltip, popover, datepicker, timepicker), therefore ngbootsrap is dependent on default ViewEncapsulation i.e. Emulated, to work properly. If ViewEncapsulation is changed then these components will not work properly.

encapsulation can be applied at each component level of ngbootstrap to avoid such issue, so that it should not have any impact from the parent ViewEncapsulation.

@pkozlowski-opensource @maxokorokov  comments please?

Regards,

Adondriel:
In the plunkr, I messed around with the class. 

```CSS
.tooltip.bs-tooltip-auto[x-placement^=top] .arrow::before, .tooltip.bs-tooltip-top .arrow::before {
    /* margin-left: -3px; */
    content: "";
    border-width: 5px 5px 0;
    border-top-color: #000;
}
```
Fixes it, but this also creates an issue with the border-radius on the actual tooltip itself, so you also have to fix that, using this CSS:

```CSS
.tooltip-inner {
    max-width: 200px;
    padding: 3px 8px;
    color: #fff;
    text-align: center;
    background-color: #000;
    /* border-radius: .25rem; */
```

PR #2784, opened 2018-10-11
feat(widgets): turn on viewEncapsulation.None

### LINKED ISSUE #2564: CSS Variables: allow styling of ng-Bootstrap by their host angular components css file
### Feature Request description:
Styling ng-Bootstrap widgets can be a challenge. 
Especially with /deep/ ::shadow being deprecated. 
By using CSS variables in the bootstrap styling, consumers of the widgets (i.e. Other angular components, with ViewEncapsulation enabled) could then specify override values for these CSS variables. 

Just think, in the local CSS for your component, you could specify a CSS variable value for the background color of the ngb-accordion, and it would set it 'accordingly' ;0

See Excerpt below:
https://developers.google.com/web/updates/2016/02/css-variables-why-should-you-care?hl=en
"This technique is especially useful for theming Web Components that use Shadow DOM, as custom properties can traverse shadow boundaries. A Web Component author can create an initial design using fallback values, and expose theming “hooks” in the form of custom properties."

fbasso:
The purpose of this PR is to allow a better widgets customization with css.

- Fix #2564 and #1875
- It removes the /deep/ selector used in the datepicker, which is deprecated

### DIFF
--- a/src/datepicker/datepicker-navigation-select.scss
+++ b/src/datepicker/datepicker-navigation-select.scss
@@ -1,4 +1,4 @@
-:host>select {
+ngb-datepicker-navigation-select > .custom-select {
   flex: 1 1 auto;
   padding: 0 0.5rem;
   font-size: 0.875rem;
--- a/src/popover/popover.scss
+++ b/src/popover/popover.scss
@@ -1,4 +1,4 @@
-:host {
+ngb-popover-window {
   &.bs-popover-top .arrow,
   &.bs-popover-bottom .arrow {
     left: 50%;
--- a/src/tooltip/tooltip.scss
+++ b/src/tooltip/tooltip.scss
@@ -1,4 +1,4 @@
-:host {
+ngb-tooltip-window {
   &.bs-tooltip-top .arrow,
   &.bs-tooltip-bottom .arrow {
     left: calc(50% - 0.4rem);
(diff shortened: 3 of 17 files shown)


src/datepicker/datepicker-navigation-select.scss
```diff
@@ -1,4 +1,4 @@
-:host>select {
+ngb-datepicker-navigation-select > select {
```
maxokorokov: this should be more specific without encapsulation, ex:
```css
ngb-datepicker-navigation-select > .custom-select
```

src/tooltip/tooltip.scss
```diff
@@ -1,32 +1,32 @@
-:host {
-  &.bs-tooltip-top .arrow,
-  &.bs-tooltip-bottom .arrow {
+.bs-tooltip {
```
maxokorokov: I guess it should be specific to `ngb-popover-window`, not to any `.bs-tooltip`

src/popover/popover.scss
```diff
@@ -1,35 +1,35 @@
-:host {
-  &.bs-popover-top .arrow,
-  &.bs-popover-bottom .arrow {
+.bs-popover {
```
maxokorokov: I guess it should be specific to `ngb-popover-window`, not to any `.bs-popover`

fbasso:
Thanks for the review ! The changes have been done.

Issue #2779, opened 2018-10-10
TinyMCE loses focus on modal dialog


igor-shubin:
Add official TinyMCE component on a modal window.

**Step to reproduce:**
 - To run the example application on https://tinymce-ngbmodal-focus-issue-demo-5b9nrp.stackblitz.io ;
 - To click the button "OPEN MODAL";
 - To close warning message regarding free trial;
 - To click on the first input to make it being in focus;
 - To type e.g. "Test" inside TineMCE editor window;
 - To click the button "bold".

**Expected result:**
 - "Test" style changed to bold;
 - Focus is returned to the editor window.

**Actual result:**
 - "Test" style changed to bold;
 - Focus is returned to the first input.

As I can see the problem is with focus-trap component. 
Could you please help me to resolve this issue?

**Library versions:**
 angular: 6.0.3
 tinymce: 4.8.3
 @tinymce/tinymce-angular: 2.3.0

pkozlowski-opensource:
@benouat could you PTAL if / when you've got free cycles?

benouat:
I'll have a look asap.
First investigation leads to weird result, for example I can reproduce it on Chrome, but not Safari...
Safari is working as expected.

`focustrap` is the culprit here. I'll have to dive into that.

benouat:
I can actually also reproduce it in Safari, but only when the first input contains something.

benouat:
@maxokorokov in this case I don't see any room for improvement here... The iframe is messing things around.
@igor-shubin do you know if it could be initialised without an iframe ? like using `content-editable` instead ?

igor-shubin:
@maxokorokov thank you for simplifying repro steps. 

@benouat I could not find how to avoid using iframe. However, if I get rid of iframe, I will face with the styling issues. So I do not think that it can be the way of fixing the focus issue there. 
In addition, I have found an example how to use focus-trap with iframes. See here: http://davidtheclark.github.io/focus-trap/demo/ Maybe it will help.

PR #2798, opened 2018-10-16
fix(focustrap): modal won't refocus last focused element on click

benouat:
Working on finding a solution for #2779, I realised that `ngbFocustrap()` implementation was refocusing any last focused element whenever clicking inside the element being trapped.

This behaviour might be great/needed for component such as `<input ngbDatePicker />` (ie clicking on any header week day will refocus the selected date), but on modal where plenty of elements _could be_ defined as focusable with `tabindex="-1"` it does not make sense as it will most transparently refocus itself...

The idea behind this PR is to deactivate this part of this code:
https://github.com/ng-bootstrap/ng-bootstrap/blob/a76d7fe4981b93667a9936d40a0a40a804a7ab72/src/util/focus-trap.ts#L50-L53

And only activate it in the datepicker for now.

---

With that fix, using TinyMCE editor (as suggested in #2779) would work with any hack/patch/implementation needed.

### DIFF
--- a/src/util/focus-trap.ts
+++ b/src/util/focus-trap.ts
@@ -24,8 +24,10 @@ export function getFocusableBoundaryElements(element: HTMLElement): HTMLElement[
  * @param element The element around which focus will be trapped inside
  * @param stopFocusTrap$ The observable stream. When completed the focus trap will clean up listeners
  * and free internal resources
+ * @param refocusOnClick Put the focus back to the last focused element whenever a click occurs on element (default to
+ * false)
  */
-export const ngbFocusTrap = (element: HTMLElement, stopFocusTrap$: Observable<any>) => {
+export const ngbFocusTrap = (element: HTMLElement, stopFocusTrap$: Observable<any>, refocusOnClick = false) => {
   // last focused element
   const lastFocusedElement$ =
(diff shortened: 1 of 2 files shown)


maxokorokov:
Added missing focustrap tests for this in #2802 and #2800. Will relaunch CI after both are merged

src/util/focus-trap.ts
```diff
@@ -25,7 +25,7 @@ export function getFocusableBoundaryElements(element: HTMLElement): HTMLElement[
  * @param stopFocusTrap$ The observable stream. When completed the focus trap will clean up listeners
  * and free internal resources
  */
-export const ngbFocusTrap = (element: HTMLElement, stopFocusTrap$: Observable<any>) => {
+export const ngbFocusTrap = (element: HTMLElement, stopFocusTrap$: Observable<any>, refocusOnClick = false) => {
```
maxokorokov: could you please add `refocusOnClick` to jsdoc, just because two other arguments are there?

benouat: done

benouat:
@maxokorokov Implementing a tinymce-like usecase would be kind of big, and very specific, hence not worth it.

PR #2806, opened 2018-10-17
datepicker: improve Hebrew demo

### LINKED ISSUE #2682: Hebrew Calendar
Following up on the request from Twitter, here are some notes:

1. It's very uncommon to have a Hebrew calendar without Gregorian calendar dates displayed alongside. In fact, in a quick search, I couldn't find any example doing so. (see references below). I suggest to add the Gregorian dates as well, otherwise this calendar will not be very useful.

2. Day names are better written as words, as it is much more common:
ראשון (Sunday)
שני
שלישי
רביעי
חמישי
שישי
שבת (Saturday)

If you decided to use the letter notation due to short space, I'd remove the apostrophe. 

3. Some dates have the quotes misaligned - basically, everything between 11 (י"א) and 19 (י"ט). When you have two letters, quotes are always in the middle.

Some examples I found online:

https://www.kipa.co.il/לוח-שנה/
https://www.uzit.co.il/newcalendar/CalendarWithTimes.aspx
https://calendar.walla.co.il
http://xn--5dbhgjl2b4b.com

Thank you!

maxokorokov:
(do not squash, several commits)

- changed weekday names to a longer format
- added Gregorian day/month to display
- fixed Hebrew numerals for 11-19

screen shot 2018-10-17 at 16 16 13

Fixes #2682

### DIFF
--- a/demo/src/app/components/datepicker/demos/hebrew/datepicker-hebrew.html
+++ b/demo/src/app/components/datepicker/demos/hebrew/datepicker-hebrew.html
@@ -3,8 +3,16 @@
   In Israel it is used for religious purposes and frequently as an official calendar for civil purposes.
 </p>
 
-<ngb-datepicker class="rtl" #dp [(ngModel)]="model" [displayMonths]="2" [firstDayOfWeek]="7">
+<ngb-datepicker class="hebrew rtl" #dp [(ngModel)]="model" [displayMonths]="1" [firstDayOfWeek]="7" [dayTemplate]="dt"
+[dayTemplateData]="dayTemplateData">
 </ngb-datepicker>
+
+<ng-template #dt let-date let-data="data" let-selected="selected" let-currentMonth="currentMonth">
+  <div class="hebrew-day" [class.outside]="date.month !== currentMonth" [class.selected]="selected">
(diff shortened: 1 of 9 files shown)


demo/src/app/components/datepicker/demos/hebrew/datepicker-hebrew.html
```diff
@@ -3,9 +3,17 @@
   In Israel it is used for religious purposes and frequently as an official calendar for civil purposes.
 </p>
 
-<ngb-datepicker class="rtl" #dp [(ngModel)]="model" [displayMonths]="2" [firstDayOfWeek]="7">
+<ngb-datepicker class="hebrew rtl" #dp [(ngModel)]="model" [displayMonths]="1" [firstDayOfWeek]="7" [dayTemplate]="dt"
+[dayTemplateData]="dayTemplateData">
 </ngb-datepicker>
 
+<ng-template #dt let-date="date" let-data="data" let-selected="selected" let-currentMonth="currentMonth">
```
benouat: This being a demo, shouldn't we use the short version `let-date` instead of `let-date="date"` ?

maxokorokov: Yep, thanks. Created this PR before $implicit was introduced. Will fix

PR #2811, opened 2018-10-18
test: better e2e test performance and navigation

fbasso:
The main idea is to avoid an transitional click for each 'it'

### DIFF
--- a/e2e-app/src/app/navigation.component.ts
+++ b/e2e-app/src/app/navigation.component.ts
@@ -1,24 +1,38 @@
 import {Component} from '@angular/core';
+import {Router} from '@angular/router';
 
 import {routes} from './app.routing';
+import {NavigationEnd} from '@angular/router';
 
 @Component({
   selector: 'app-navigation',
   template: `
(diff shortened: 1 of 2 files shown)


e2e-app/src/app/navigation.component.ts
```diff
@@ -1,24 +1,46 @@
...
+        </li>
+      </ul>
     </div>
   </div>
-  <a role="button" class="btn btn-outline-primary ml-3" id="navigate-home" href="#/">Reset</a>
   `
 })
 
 export class NavigationComponent {
   urls: string[] = [];
```
maxokorokov: I don't think you need `urls` at all now

e2e-app/src/app/navigation.component.ts
```diff
@@ -1,24 +1,46 @@
...
+  groupedUrls: any = {};
```
maxokorokov: To take it a step further, can't we just expose `routes` to the template and simply iterate over them without creating `groupedUrls` ?

fbasso:
Thanks for the review ! Changes done.

PR #2817, opened 2018-10-19
feat(animation): bootstrap animations

### LINKED ISSUE #295: animations: initial push for animations

fbasso:
The purpose of this PR is to fix #295.

This is the starting point to talk about animations in ngBootstrap.  This is done in the Bootstrap way, i.e. adding and removing Bootstrap css classes.

This branch is still in development and discussions are open. All widgets have been done, nevertherless, the accordion needs a refactor (mainly because of the activeIds management), that's why their tests have been excluded at the moment.

### Pending tasks

- [ ] Take a decision to know if we can manage the runningTransition inside ngbTransition directly instead of on each widget (for example on the alert). 

It could save to manage a flag inside the widget, but it's maybe not possible as at one time, there will be necessary to implement the events as specified by Bootstrap (i.e. `close` and `closed`for the alert).

We can add the transition option `ignore` or `stop` as described by @maxokorokov, but in this case we will need a `beforeTransition` callback.

Waiting for this decision before changing anything in the code.

- [ ] By the way, is it the right time to implement the Bootstrap events (i.e. close and closed in the alert). Or is it better to implement them in another PR (to not complexify more with PR) ?

- [ ] I was not able to able to reproduce the modal issue (described by @maxokorokov): need a recheck to see if it happens again after the last changes

- [ ] Refactor of the NgbRunTransition function: re-implement the `zip` part, remove the `$done`, ....

### DIFF
--- a/src/collapse/collapse.ts
+++ b/src/collapse/collapse.ts
@@ -1,4 +1,7 @@
-import {Directive, Input} from '@angular/core';
+import {Directive, Input, Renderer2, ElementRef, Output, EventEmitter} from '@angular/core';
+import {NgbRunTransition} from '../util/transition/ngbTransition';
+import {NgbCollapsingTransition} from '../util/transition/ngbCollapseTransition';
+import {NgbConfig} from '../ngb-config';
 
 /**
  * A directive to provide a simple way of hiding and showing elements on the page.
@@ -13,4 +16,23 @@ export class NgbCollapse {
    * If `true`, will collapse the element or show it otherwise.
    */
   @Input('ngbCollapse') collapsed = false;
+
+  /**
+   * A flag to enable/disable the animation
+   */
+  @Input() animation = false;
+
+  @Output() ngbCollapseChange = new EventEmitter<boolean>();
+
+  constructor(_renderer: Renderer2, private _element: ElementRef, ngbConfig: NgbConfig) {
+    this.animation = ngbConfig.animation;
+  }
+
+  toggle(open: boolean = this.collapsed) {
+    this.collapsed = !open;
+    this.ngbCollapseChange.next(this.collapsed);
+    NgbRunTransition(
+        this._element.nativeElement, NgbCollapsingTransition,
+        {animation: this.animation, data: {direction: open ? 'show' : 'hide'}});
+  }
 }
--- a/src/util/transition/ngbCollapseTransition.ts
+++ b/src/util/transition/ngbCollapseTransition.ts
@@ -1,0 +1,61 @@
+import {NgbTransitionOptions, NgbTransitionDefReturn} from './ngbTransition';
+import {reflow} from '../util';
+
+const collapseClass = 'collapse';
+const collapsingClass = 'collapsing';
+const showClass = 'show';
+const dataMaxHeight = 'data-max-height';
+
+export const NgbCollapsingTransition =
+    (panelElement: HTMLElement, options: NgbTransitionOptions = {}): NgbTransitionDefReturn => {
+      const direction = options.data.direction;
+      const classList = panelElement.classList;
+      if (options.animation) {
+        let maxHeight = panelElement.getAttribute(dataMaxHeight);
+        if (!maxHeight) {
+          // No transition is running
+          const hasShown = classList.contains(showClass);
+          if (!hasShown) {
+            classList.add(showClass);
+          }
+          panelElement.style.height = '';
+          maxHeight = panelElement.getBoundingClientRect().height + 'px';
+          panelElement.setAttribute(dataMaxHeight, maxHeight);
+          if (!hasShown) {
+            classList.remove(showClass);
+          }
+
+          // Fix the height before starting the animation
+          panelElement.style.height = direction !== 'show' ? maxHeight : '0px';
+
+          // Remove the collapse classes to let the nbgAnimationEngine works by itself
+          classList.remove(collapseClass);
+          classList.remove(collapsingClass);
+          classList.remove(showClass);
+
+          reflow(panelElement);
+
+          // Start the animation
+          classList.add(collapsingClass);
+        }
+
+        // Start or revert the transition
+        panelElement.style.height = direction === 'show' ? maxHeight : '0px';
+      }
+      return {
+        elements: panelElement,
+        end: () => {
+          // Togle classes
+          classList.remove(collapsingClass);
+          classList.add(collapseClass);
+          if (direction === 'show') {
+            classList.add(showClass);
+          } else {
+            classList.remove(showClass);
+          }
+
+          panelElement.removeAttribute(dataMaxHeight);
+          panelElement.style.height = '';
+        }
+      };
+    };
--- a/src/util/transition/ngbFadingTransition.ts
+++ b/src/util/transition/ngbFadingTransition.ts
@@ -1,0 +1,58 @@
+import {NgbTransitionOptions, NgbTransitionDefReturn} from './ngbTransition';
+import {reflow} from '../util';
+
+export const NgbNavFadingTransition = (node: HTMLElement, options: NgbTransitionOptions): NgbTransitionDefReturn => {
+  const animation = options.animation;
+  let end;
+  const mustBeHidden = options.data.direction !== 'show';
+  const classList = node.classList;
+  if (mustBeHidden) {
+    classList.remove('show');
+    end = () => { classList.remove('active'); };
+  } else {
+    classList.add('active');
+    if (animation) {
+      reflow(node);
+    }
+    classList.add('show');
+  }
+  return {elements: animation ? node : null, end};
+};
+
+export const NgbAlertFadingTransition =
+    (node: HTMLElement, options: NgbTransitionOptions = {}): NgbTransitionDefReturn => {
+      node.classList.remove('show');
+      return {elements: node};
+    };
+
+const _showClass = 'show';
+export const NgbModalFadingTransition =
+    (node: HTMLElement, options: NgbTransitionOptions = {}): NgbTransitionDefReturn => {
+      const classList = node.classList;
+      if (classList.contains(_showClass)) {
+        classList.remove('show');
+      } else {
+        reflow(node);
+        classList.add('show');
+      }
+      return ({ elements: [node, node.querySelector('div.modal-dialog')] } as NgbTransitionDefReturn);
+    };
+
+export const NgbPopupFadingTransition = (node: HTMLElement, options: NgbTransitionOptions = {}) => {
+  let addClass;
+  const showElement = options.data.showElement;
+  const classList = node.classList;
+  if (showElement != null) {
+    addClass = showElement === true;
+  } else {
+    addClass = !classList.contains(_showClass);
+  }
+
+  if (addClass) {
+    classList.add('show');
+  } else {
+    classList.remove('show');
+  }
+
+  return {elements: node};
+};
--- a/src/util/transition/ngbTransition.ts
+++ b/src/util/transition/ngbTransition.ts
@@ -1,0 +1,79 @@
+import {Observable, AsyncSubject, fromEvent, Subject, zip, of} from 'rxjs';
+import {filter, takeUntil, skip} from 'rxjs/operators';
+
+export interface NgbTransitionDefReturn {
+  elements?: HTMLElement | Array<HTMLElement>| null;
+  end?();
+}
+
+export interface NgbTransitionOptions {
+  animation?: boolean;
+  data?: any;
+  callback?(data?: any);
+}
+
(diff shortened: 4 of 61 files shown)


fbasso:
Back to the animations. The last push enhances the following points:

- If animations are turned off, everything is now synchronous: promises have been replaced by observers and setTimeout by ngZone (to start animations only when all html nodes ready.
- enableAnimation is now false by default. The tests code is now close to master (the changes concerns the new global configuration object).
- A new global injector has been created, to quickly switch the animation. The purpose is to be able to propose animations in a beta easily actionable.
- On the demo site, he animations are turned on.
- The PR is green for the first time :)

razzildinho:
@fbasso if you need any help on this I'd be happy to take parts of it on.

asvishnyakov:
Any progress on this?

kcexn:
@fbasso I was wondering where this PR is at? Is there anything that you need help looking at?

joerter:
Is there any extra documentation that needs to be added to help this merge? I'm willing to pitch in wherever.

asegarra:
Any update on why this hasn't been merged?

joerter:
@fbasso, @maxokorokov, or @benouat is there any help you need to get this PR merged?

src/util/transition/ngbCollapseTransition.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    // set the height before the animation
+    panelElement.style.height = heightInit;
+    panelElement.setAttribute(dataHeightFrom, heightFrom);
+    panelElement.setAttribute(dataHeightTo, heightTo);
+
+    // Remove the collapse classes to let the nbgAnimationEngine works by itself
+    classList.remove(collapseClass);
+    classList.remove(collapsingClass);
+    classList.remove(showClass);
```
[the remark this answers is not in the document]
fbasso: Unfortunately, several parameters doesn't work on ie : https://caniuse.com/#search=classList

> Does not support multiple parameters for the add() & remove() methods

src/util/transition/ngbCollapseTransition.ts
```diff
@@ -0,0 +1,105 @@
...
+      panelElement.style.height = '';
+
+      return callbackParameters;
+    }
+  };
+};
+
+
+export const showTransition = {
+  classname: 'show'
```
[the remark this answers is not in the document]
fbasso: It's dead code here, after many refactors :). I'll remove it

joerter:
I've tested this branch with my project. I'm using Modals and NgbCollapse. Everything seems to be working perfectly, thank you!

asvishnyakov:
Is there any chance it will be merged to master?

ArnauAregall:
@fbasso checked latest master commits, and the following should resolve the failing test of your branch: https://github.com/ng-bootstrap/ng-bootstrap/commit/b4dd160668dc9ca3ad1e8adf6be0d9a15d80444f 

This is promising, awesome work!

src/util/util.ts
```diff
@@ -73,3 +73,29 @@ export function closest(element: HTMLElement, selector): HTMLElement {
...
+}
+
+
+/**
+ *
+ * @param container
+ * @param element
+ * @returns true if element is a child of parent
+ */
+export function contains(container, element) {
```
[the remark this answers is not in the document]
ArnauAregall: May I help with this? Is the only remaining change requested so awesome work of @fbasso can be merged?

KevinBassaDevelopment:
Is there an estimated date when this will get into master?
Animations for these components would be greatly appreciated

src/collapse/collapse.ts
```diff
@@ -9,8 +14,31 @@ import {Directive, Input} from '@angular/core';
...
+    collapsingTransition.run(element, {enableAnimation: this._firstChange ? false : this.enableAnimation});
+    this._firstChange = false;
+  }
+
+  /**
+   * A flag to enable/disable the animation
+   */
+  @Input() enableAnimation = false;
+
+  constructor(_renderer: Renderer2, private _element: ElementRef, ngbConfig: NgbConfig) {
```
maxokorokov: version with `ngOnChanges` that does the same:

```ts
ngOnChanges({collapsed}: SimpleChanges) {
    if (!collapsed.firstChange) {
      collapsingTransition.run(this._element.nativeElement, {enableAnimation: this.enableAnimation});
    }
  }
```

src/util/transition/ngbTransition.ts
```diff
@@ -0,0 +1,81 @@
+import {Observable, AsyncSubject, fromEvent, Subject, zip} from 'rxjs';
+import {filter, takeUntil, skip} from 'rxjs/operators';
+
+export interface NgbTransitionDefReturn {
+  elements?: HTMLElement | Array<HTMLElement>| null;
```
maxokorokov: code would be much simpler if we just use `elements: HTMLElement[]`

src/util/transition/ngbFadingTransition.ts
```diff
@@ -0,0 +1,55 @@
...
+  return {elements: enableAnimation ? node : null, end};
+};
+
+export const NgbAlertFadingTransition =
+    (node: HTMLElement, options: NgbTransitionOptions = {}): NgbTransitionDefReturn => {
+      node.classList.remove('show');
+      return {elements: node};
+    };
+
+const _showRegExp = /\bshow\b/;
```
maxokorokov: can't we use `classList.contains()` instead of the regexp ?

src/util/transition/ngbFadingTransition.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  let addClass;
+  if (options.showElement != null) {
+    addClass = options.showElement === true;
+  } else {
+    addClass = !_showRegExp.test(node.className);
+  }
+
+  const classList = node.classList;
+  if (addClass) {
```
maxokorokov: Can't we use `classList.toggle()` instead?

fbasso:
The last push brings:

- ngbTransition now manages the case where an animation is already running on the given element. I think this is particularly important as its role is to:
  - Register the transition
  - Complete the previous transition observer without emitting a value, so that the previous transitionend event and the subscribers are released
  - Start and register (in a map) the new transition function with a new observer.

It enables the transition to be reverted several times, any subscriber in widget will be called once the last transition is finished.

- The collapse transition has been simplified (revert still possible)
- The collapse demo has been changed in order to show the two-way data binding.
- The accordion transition is reversible
- Prevents alert to be close during the transition process, so that close event is emitted only once.

fbasso:
This last push implements the nav and tab behavior as described by @maxokorokov (changing nav during an animation)

fbasso:
New changes with:

- The toast actions are prevented during the transition (as the alerts)
- The carousel transitions can now be reverted properly. This is allowed only to return to the previous slide (any cycle to others slides than the previous one is prevented)

benouat:
> 6. You're handling stopping current element transition to complete and reverse it in the `NgbTransition` → `runningTransition` map (ex. for accordion, nav).
> 
> Why not use the same technique to handle `alert` and avoid having flags `_transtionRunning` in the component?

👍 Same remark here!

fbasso:
> 
> 
> > 1. You're handling stopping current element transition to complete and reverse it in the `NgbTransition` → `runningTransition` map (ex. for accordion, nav).
> > 
> > Why not use the same technique to handle `alert` and avoid having flags `_transtionRunning` in the component?
> 
> 👍 Same remark here!

Yes, that's was my first though afer reading this comment. But after a deeper thinking, I wonder if we can get rid of it. At one time, we will need to manage the `hide` and `hidden` events, and the first one will be emitted before the transition, and should not re-emitted if a transition is already running.

A solution would be to pass also a callback to run before the transition, but we need to decide if it's a better way or not. Or find another solution....

fbasso:
> 1. Accordion CD issue. It is not updating the state correctly, unless you trigger CD manually after open/collapse (ex. clicking elsewhere):

This is now fixed. By the way, the collapse all didn't work well in the same sample, and has been fixed as well.

fbasso:
>     2. Modal animation should not be reversed. You can check this by double-clicking backdrop:

Done, at least when the modal is dismissed. It seems acceptable to close the modal during its transition in.

>     3. There is no backdrop fadeout/in animation

Done.

fbasso:
>     4. Modal sometimes gives this error, can't reproduce consistently unfortunately:

I was not able to reproduce it. To be checked if it still exist with the last modal changes

fbasso:
https://getbootstrap.com/docs/4.4/getting-started/accessibility/#reduced-motion is now managed. By the, it also covers the case where the theme override `transition-property` to `none`

fbasso:
The last changes include:

- enableAnimation -> animation
- Animations enabled by default
- Disable animations for e2e and tdd

Devqon:
Does this also enable animations on the nav component when using angular's routing? Or should that be handled manually?

fbasso:
Good point, @Devqon, thanks !

I checked on the demo site, and indeed, the navs driven by the router don't work with transitions. Basically, it's because they don't manage their contents directly, their roles are just to change the url. Then the router-outlet works by itself, but there is no animation, as it's not part of the navs.

We need to think deeper about it.

maxokorokov:
@Devqon, **current thinking** is `<div [ngbNavOutlet]></div>` will be animated, but `<router-outlet></router-outlet>` will not be. 

There is no way of telling the router to 'wait' until we do the fade out css transition, even if we try doing something like `<router-outlet [ngbNavOutlet]></router-outlet>` and integrating it with navs.

So you would have two options, if you want to animate:
- use Angular Animations for the router → https://angular.io/guide/route-animations
- use `ngbNav` without router outlet / router config

We also do not want to add `@angular/router` dependency to the library at the moment.

I understand it might be a pretty common use case though, so as @fbasso said, we have to think more about this one. Any ideas are welcome...

Devqon:
I understand that it would be difficult because it should not be the goal to control the `<router-outlet/>` component. I would already be happy if there is a default animation that is used for the `[ngbNavOutlet]` that would be reusable for other components so I could just do something like this:
```typescript
@Component({
  selector: 'my-tabs',
  template: `
    <!-- ... -->
    <div [@ngbNavOutletAnimation]="outlet.activatedRoute.url">
      <router-outlet></router-outlet>
    </div>
  `,
  animations: [ngbNavOutletAnimation]
```

fbasso:
I'm not sure that even with an `ngbNavOutletAnimation`, things would be simpler. This is due to the behavior of the router-outler itself. The old dom component is destroyed as far as the new one is ready (as far as I know).

So it means that the new component provided by the router can be displayed before the end of the fade out, giving a bad visual effect.

Another idea would be to 'delay' the url change after the fade out, but the fade delay would be lost for asynchronous operations (i.e. fetching data from the server). 

So still not sure what we can do for this use-case.

My guess is that the best (for now) would be to use the navs the usual way: 
- panels with component, as defined in the router,
- The tab click change the url and run the panel change (and by the way run the router for asynchronous stuff)
- The navs are initialized with the ActiveRoute information, to manage page reload or url change

Not ideal too, but I have no other idea :(

maxokorokov:
Hey, @fbasso!

I have big comment → I don't understand why did you go into trouble and added `[animation]="false"` to ALL the tests. We could do this in `.configureTestingModule()` by providing something like:

```ts
providers: [
  { provide: NgbConfig, useClass: NgbTestConfig }
]
```

Or providing something like `NgbAlertModule.forTests()` with overridden `NgbConfig`.

But, I think even better option to introduce enviroments like we do for normal apps:
- [ ] let's rename `tokens.ts` to `ngb-config.ts`
- [ ] create `src/environment.ts` and `src/enviroment.test.ts`
- [ ] override `fileReplacements` for the `test` target in `angular.json` like so:

```json
"fileReplacements": [
  {
    "src": "src/environment.ts",
    "replaceWith": "src/environment.test.ts"
  }
],
```

```ts
// environment.ts
export const environment = {
  animation: true,
};

// ngb-config.ts
import {environment} from './environment';

export class NgbConfig {
  animation = environment.animation;
}


Cheers,
Max

cc @benouat 

Will review the rest after this change :)

fbasso:
Hi @maxokorokov, changes done. 

Much cleaner code for the unit tests, indeed !

fbasso:
For information, animations are now actively developed in a dedicated branch https://github.com/ng-bootstrap/ng-bootstrap/tree/animations

maxokorokov:
Closing, final implementations are in #3747

PR #2820, opened 2018-10-19
feat(popover): add ability to toggle popover based on boolean value

kwall2004:
I have a requirement to toggle a popover based on a boolean input value.  Without this change, I have to write this code for each instance:

```
@Component({
  selector: 'ngbd-popover-openif',
  templateUrl: './popover-openif.html'
})
export class NgbdPopoverOpenif {
  @ViewChild('popover') popover: NgbPopover;

  @Input()
  set openIf(value: boolean) {
    if (value) {
      this.popover.open();
    } else {
      this.popover.close();
    }
  }
}
```

### DIFF
--- a/demo/src/app/components/popover/demos/openif/popover-openif.html
+++ b/demo/src/app/components/popover/demos/openif/popover-openif.html
@@ -1,0 +1,22 @@
+<p>You can set the popover to open or close based on a boolean value.</p>
+
+<form>
+  <div class="form-group">
+    <div class="form-check">
+      <input class="form-check-input" type="radio" name="openIf" id="openTrue" [value]="true" [(ngModel)]="openIf">
+      <label class="form-check-label" for="openTrue">
+        open
+      </label>
+    </div>
+    <div class="form-check">
+      <input class="form-check-input" type="radio" name="openIf" id="openFalse" [value]="false" [(ngModel)]="openIf">
+      <label class="form-check-label" for="openFalse">
+        close
+      </label>
+    </div>
+  </div>
+  <div class="form-group">
+    <input type="text" class="form-control" ngbPopover="What a great tip!" popoverTitle="Pop title" [autoClose]="false"
+      triggers="manual" [openIf]="openIf">
+  </div>
+</form>
--- a/demo/src/app/components/popover/demos/openif/popover-openif.ts
+++ b/demo/src/app/components/popover/demos/openif/popover-openif.ts
@@ -1,0 +1,11 @@
+import {Component} from '@angular/core';
+
+
+
+@Component({
+  selector: 'ngbd-popover-openif',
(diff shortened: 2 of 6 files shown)


benouat:
I don't really see why you would have to create a Component and use a ViewChild query to do that... You were **really close** with your idea :)

I simply used composition here, by enriching the existing popover directive with an extra one, adding some superpower to it.

Here is a working example:

```typescript
@Directive({
  selector: '[ngbPopover][openIf]'
})
export class PopoverBoolean implements OnChanges, OnDestroy {
  destroyed$ = new Subject();
  
  @Input() openIf: boolean;
  @Output() openIfChange = new EventEmitter();

  constructor(private popover: NgbPopover) {
    this.popover.autoClose = false;
    this.popover.shown.pipe(takeUntil(this.destroyed$)).subscribe(() => this.openIfChange.next(true));
    this.popover.hidden.pipe(takeUntil(this.destroyed$)).subscribe(() => this.openIfChange.next(false));
  }

  ngOnChanges(changes: SimpleChanges) {
    if (changes.openIf) {
      const { currentValue } = changes.openIf;
      if (currentValue) {
        this.popover.open();
      } else {
        this.popover.close();
      }
    }
  }

  ngOnDestroy() {
    this.destroyed$.next();
  }
}
```

This directive could be used like that:

```html
<ng-template #popContent>Hello, <b>{{name}}</b>!</ng-template>
<ng-template #popTitle>Fancy <b>content!!</b></ng-template>


<button 
  type="button" class="btn btn-outline-secondary"
  [ngbPopover]="popContent" [popoverTitle]="popTitle"
  [(openIf)]="open"
>
  I've got markup and bindings in my popover!
</button>
<br />
<br />
<div class="form-check">
  <input class="form-check-input" type="checkbox" id="defaultCheck1"
    [(ngModel)]="open" [ngModelOptions]="{ standalone: true }">
  <label class="form-check-label" for="defaultCheck1">Toggle Popover</label>
</div>
```

Feel free to experiment more, [Stackblitz is here](https://stackblitz.com/edit/angular-4tsm9t?file=app%2Fpopover-tplcontent.html)

---

With that in mind, I don't think we would need to merge your PR.
Adding such a new feature inside ng-bootstrap itself would need a complete refactor of `triggers` functionality IMHO.

maxokorokov:
Thanks for the idea and for the effort to open a PR, @kwall2004!

However as @benouat mentioned we won't merge this PR as it is now. We have `triggers` to open/close popovers, `show/hide` events and `.open()`, `.close()` and `.trigger()` methods. Adding an additional `@Input() openIf` seems like too much at the moment. Also it would work only in one-way leading to states when you bind something that is `true` when the popup is actually will be closed.

I'd encourage you to open an issue to propose this API change to discuss design first.

Otherwise you have a workaround suggested by @benouat.

PR #2823, opened 2018-10-23
feat(dropdown): enable container="body"

### LINKED ISSUE #1012: Dropdown append to body

fbasso:
fix #1012

### DIFF
--- a/e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.e2e-spec.ts
+++ b/e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.e2e-spec.ts
@@ -1,115 +1,119 @@
 import {Key, ElementFinder} from 'protractor';
 import {sendKey, openUrl, rightClick} from '../../tools.po';
 import {DropdownAutoClosePage} from './dropdown-autoclose.po';
+import {DropdownPage} from '../dropdown.po';
 
-describe('Dropdown Autoclose', () => {
-  let page: DropdownAutoClosePage;
+const containers = [null, 'body'];
+containers.forEach((container) => {
+  describe('Dropdown Autoclose with container = ' + container, () => {
+    let page: DropdownAutoClosePage;
 
-  const expectDropdownToBeVisible = async(dropdown: ElementFinder, message: string) => {
-    expect(await page.isOpened(dropdown)).toBeTruthy(message);
-    expect(await page.getOpenStatus().getText()).toBe('open', message);
-  };
+    const expectDropdownToBeVisible = async(dropdown: ElementFinder, message: string) => {
+      expect(await page.isOpened(dropdown)).toBeTruthy(message);
+      expect(await page.getOpenStatus().getText()).toBe('open', message);
+    };
 
-  const expectDropdownToBeHidden = async(dropdown: ElementFinder, message: string) => {
-    expect(await page.isOpened(dropdown)).toBeFalsy(message);
-    expect(await page.getOpenStatus().getText()).toBe('closed', message);
-  };
+    const expectDropdownToBeHidden = async(dropdown: ElementFinder, message: string) => {
+      expect(await page.isOpened(dropdown)).toBeFalsy(message);
+      expect(await page.getOpenStatus().getText()).toBe('closed', message);
+    };
 
-  const openDropdown = async(dropdown: ElementFinder, message: string) => {
-    await page.open(dropdown);
-    await expectDropdownToBeVisible(dropdown, message);
-  };
+    const openDropdown = async(dropdown: ElementFinder, message: string) => {
+      await page.open(dropdown);
+      await expectDropdownToBeVisible(dropdown, message);
+    };
 
-  beforeAll(() => page = new DropdownAutoClosePage());
+    beforeAll(() => page = new DropdownAutoClosePage());
 
-  beforeEach(async() => await openUrl('dropdown/autoclose'));
+    beforeEach(async() => await openUrl('dropdown/autoclose'));
 
-  it(`should not close when right clicking`, async() => {
-    await page.selectAutoClose('true');
-    const dropdown = page.getDropdown('#dropdown');
+    it(`should not close when right clicking`, async() => {
+      await page.selectAutoClose('true');
+      const dropdown = page.getDropdown('#dropdown');
 
-    await openDropdown(dropdown, `Opening dropdown for right clicks`);
-    await rightClick(page.getFirstItem(dropdown));
-    await expectDropdownToBeVisible(dropdown, `Dropdown should NOT be closed on right click inside`);
+      await openDropdown(dropdown, `Opening dropdown for right clicks`);
+      await rightClick(page.getFirstItem(dropdown));
+      await expectDropdownToBeVisible(dropdown, `Dropdown should NOT be closed on right click inside`);
 
-    await page.rightClickOutside();
-    await expectDropdownToBeVisible(dropdown, `Dropdown should NOT be closed on right click outside`);
-  });
+      await page.rightClickOutside();
+      await expectDropdownToBeVisible(dropdown, `Dropdown should NOT be closed on right click outside`);
+    });
 
-  it(`should work when autoClose === true`, async() => {
-    await page.selectAutoClose('true');
-    const dropdown = page.getDropdown('#dropdown');
+    it(`should work when autoClose === true`, async() => {
+      await page.selectAutoClose('true');
+      const dropdown = page.getDropdown('#dropdown');
 
-    // escape
-    await openDropdown(dropdown, `Opening dropdown for escape`);
-    await sendKey(Key.ESCAPE);
-    await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on ESC`);
+      // escape
+      await openDropdown(dropdown, `Opening dropdown for escape`);
+      await sendKey(Key.ESCAPE);
+      await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on ESC`);
 
-    // outside click
-    await openDropdown(dropdown, `Opening dropdown for outside click`);
-    await page.clickOutside();
-    await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on outside click`);
+      // outside click
+      await openDropdown(dropdown, `Opening dropdown for outside click`);
+      await page.clickOutside();
+      await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on outside click`);
 
-    // inside click
-    await openDropdown(dropdown, `Opening dropdown for inside click`);
-    await page.getFirstItem(dropdown).click();
-    await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on inside click`);
-  });
+      // inside click
+      await openDropdown(dropdown, `Opening dropdown for inside click`);
+      await page.getFirstItem(dropdown).click();
+      await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on inside click`);
+    });
 
-  it(`should work when autoClose === false`, async() => {
-    await page.selectAutoClose('false');
-    const dropdown = page.getDropdown('#dropdown');
+    it(`should work when autoClose === false`, async() => {
+      await page.selectAutoClose('false');
+      const dropdown = page.getDropdown('#dropdown');
 
-    // escape
-    await openDropdown(dropdown, `Opening dropdown for escape`);
-    await sendKey(Key.ESCAPE);
-    await expectDropdownToBeVisible(dropdown, `Dropdown should NOT be closed on ESC`);
+      // escape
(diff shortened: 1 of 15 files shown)


src/dropdown/dropdown.ts
```diff
@@ -36,24 +36,38 @@ export class NgbDropdownMenu {
...
+    const nativeElement = this._elementRef.nativeElement;
+    if (container === 'body') {
+      document.body.appendChild(nativeElement);
+    }
+  }
+
+  removeFromBody() {
+    const nativeElement = this._elementRef.nativeElement;
+    if (nativeElement.parentNode === document.body) {
+      document.body.removeChild(nativeElement);
```
pkozlowski-opensource: I don't think this is correct - we should rather return it back to the original parent. Think of this situation:
* we start with `[container]="exp"` where `exp` evaluates to `"body"`;
* we open and close dropdown;
* we flip `exp` to evaluate to `null`
* we try to open a dropdown - it is detached from the DOM now!

fbasso: Thanks for your comment. I'm going to have a look at this use case

fbasso:
Changes done. 
I also fixed an issue regarding the positioning. In Bootstrap, the menu size is dependent to its parent (with dropdown ou dropup). To get the same behavior when it's appended to the body, the menu is first appended to a container with the same classes, and this container is appended to the body.

Everything seems fine.  The tests checked that the positions are the same, with or without container="body"

src/dropdown/dropdown.ts
```diff
@@ -30,29 +31,69 @@ export class NgbDropdownMenu {
...
+  position(triggerEl, placement, container) {
+    this.applyPlacement(
+        positionElements(
+            triggerEl, this._bodyContainer || this._elementRef.nativeElement, placement, container === 'body'));
+  }
+
+  applyContainer(container: null | 'body' = null) {
+    const nativeElement = this._elementRef.nativeElement;
+    const renderer = this._renderer;
+    const body = document.body;
```
benouat: Don't access `document` like that. See my comment below.

src/dropdown/dropdown.ts
```diff
@@ -30,29 +31,69 @@ export class NgbDropdownMenu {
...
   isOpen = false;
 
+  private _dropdownElement;
+  private _bodyContainer: HTMLElement;
+
   constructor(
       @Inject(forwardRef(() => NgbDropdown)) public dropdown, private _elementRef: ElementRef<HTMLElement>,
-      private _renderer: Renderer2) {}
+      private _renderer: Renderer2) {
+    this._dropdownElement = this._elementRef.nativeElement.parentNode;
```
benouat: I don't really like to rely on DOM here accessing `this._elementRef.nativeElement.parentNode`. (I know you did not write this code in the first place, it was already here, you just moved it...)

Why can't we use already `dropdown` for that? It's already here in the constructor, and we could add a public get on it to return its `nativeElement`

src/dropdown/dropdown.ts
```diff
@@ -30,29 +31,69 @@ export class NgbDropdownMenu {
...
+    const dropdownElement = this._dropdownElement;
+
     // remove the current placement classes
-    this._renderer.removeClass(this._elementRef.nativeElement.parentNode, 'dropup');
-    this._renderer.removeClass(this._elementRef.nativeElement.parentNode, 'dropdown');
+    renderer.removeClass(dropdownElement, 'dropup');
+    renderer.removeClass(dropdownElement, 'dropdown');
     this.placement = _placement;
+
     /**
```
benouat: Could you please take the opportunity of this PR to correct this comment by using a standard `/*` instead of a documentation one `/**` ?

fbasso:
Thanks @benouat for your comments. 

To fix them, I refactored the directives to have the dropdown driving the dropdownmenu position (before the dropdown menu position itself related to its parentNode), which is more logical.

jgrenat:
Hey! Is there any work left to do on this that we can help with?

src/dropdown/dropdown.ts
```diff
@@ -143,22 +135,27 @@ export class NgbDropdown implements OnInit, OnDestroy {
...
+    this.container = config.container;
     this.autoClose = config.autoClose;
     this._zoneSubscription = _ngZone.onStable.subscribe(() => { this._positionMenu(); });
   }
 
   ngOnInit() {
-    if (this._menu) {
-      this._menu.applyPlacement(Array.isArray(this.placement) ? (this.placement[0]) : this.placement as Placement);
-    }
+    this._setupMenuAttributes();
```
maxokorokov: Maybe better to do this on open only? what's the point of moving DOM around on widget initialization? If you have a table with a 100 of dropdowns it will do lots of useless operations

src/dropdown/dropdown.ts
```diff
@@ -242,7 +252,63 @@ export class NgbDropdown implements OnInit, OnDestroy {
...
+    }
+  }
+
+  private _position(triggerEl, placement, container) {
+    this._applyPlacement(
+        positionElements(
+            triggerEl, this._bodyContainer || this._menuElement.nativeElement, placement, container === 'body'));
+  }
+
+  private _applyContainer(container: null | 'body' = null) {
```
maxokorokov: I would split this method at least in two:
- one that does the setup
- another that does the cleanup

Currently cleanup is confusing (you have to pass `container = null` to do it?)

src/dropdown/dropdown.ts
```diff
@@ -242,7 +252,63 @@ export class NgbDropdown implements OnInit, OnDestroy {
...
+
+    /*
+     * apply the new placement
+     * in case of top use up-arrow or down-arrow otherwise
+     */
+    const dropdownClass = _placement.search('^top') !== -1 ? 'dropup' : 'dropdown';
+    renderer.addClass(dropdownElement, dropdownClass);
+
+    const bodyContainer = this._bodyContainer;
+    if (bodyContainer) {
```
maxokorokov: So in case of a valid `bodyContainer` you'll add placement and dropdown classes on BOTH `bodyContainer` and `dropdownElement`? don't get why

benouat:
My above approval is just for my previous initial Review, not @maxokorokov's one

src/dropdown/dropdown.ts
fbasso: It's because of the arrow (and, by extension, on the theme customization that can be applied). This is what is get if I just set it in the body container only :

image

src/dropdown/dropdown.ts
fbasso: I need to perform at least the applyPlacement, as it was done before, otherwise there are unit tests failing (i.e. ngb-dropdown-toggle Custom config should initialize inputs with provided config FAILED).

 I don't know if it worth to do it, but I'll add a condition to avoid the this._applyContainer(this.container); if the dropdown is not open.

fbasso:
Thanks for your review @maxokorokov  ! 

I changed all changes that can be applied directly, without refactor (ie without the popup service usage). Nevertheless, I noticed some placement issues with container=body and autoclose=true, so there is still things to investigate.

Regarding the popup service usage, it's a good idea as the code for the positioning is shared among the widgets. Just a remark: it should be used only with container=body, a wrapper is added around the dropdown. While it's fine when it's appended to the body (a container is required with a specific bootstrap class), it breaks the markup when appended to the widget (ie don't follow the bootstrap markup).

Talking about refactor, I still wonder if, finally, it's the right way to fix the issue. I had a closer look on how Bootstrap manage this use case. There is no markup to specify (nor through the api) to have menu appended to the body. There are some important points:

* The positioning is done with popper.js (https://github.com/FezVrasta/popper.js/),
* Bootstrap has some positionning options (https://getbootstrap.com/docs/4.1/components/dropdowns/#options) to adapt how the dropdown is positioned.
* By default, we have flip=true, boundary=scrollParent.

A quickly tested did a test page (with Bootstrap/js only). With the default options values, the dropdown simply 'flip' to the top if there is no space at the bottom of the scrolled element. So another way to fix the issue could be to do the same.

fbasso:
The last push add an observer on the first 'scrolled' parent, to keep the positioning sync with the element. This part needs a full review !

fbasso:
> Could you please remove all scrolling-related stuff from here and rebase ?
> Let's just do dropdown append to body here.
> 
> Close-on-scrolling when attached to body is not dropdown-related if I understand correctly and a feature that could be added either before or later?

@maxokorokov Done ! I also rewrite some little parts as it didn't work due to some recent changes on master.

src/dropdown/dropdown.ts
```diff
@@ -310,7 +342,67 @@ export class NgbDropdown implements OnInit,
 
   private _positionMenu() {
     if (this.isOpen() && this._menu) {
```
maxokorokov: maybe this can also be moved inside the `_applyPlacement()` ?

src/dropdown/dropdown.ts
```diff
@@ -145,8 +147,10 @@ export class NgbDropdown implements OnInit,
     OnDestroy {
   private _closed$ = new Subject<void>();
   private _zoneSubscription: Subscription;
+  private _bodyContainer: HTMLElement;
 
   @ContentChild(NgbDropdownMenu) private _menu: NgbDropdownMenu;
+  @ContentChild(NgbDropdownMenu, {read: ElementRef}) private _menuElement: ElementRef;
 
   @ContentChild(NgbDropdownMenu, {read: ElementRef}) private _menuElementRef: ElementRef<HTMLElement>;
```
maxokorokov: this is unused (maybe after refactoring after adding `_menuElement` ?)

e2e-app/src/app/dropdown/dropdown.po.ts
```diff
@@ -0,0 +1,28 @@
+import {$, $$, ElementFinder, by} from 'protractor';
+
+export class DropdownPage {
+  getDropdown(dropDownSelector = '') { return $(`${dropDownSelector}[ngbDropdown]`); }
+
+  getDropdownMenu(dropDownMenuSelector = 'div') { return $(`${dropDownMenuSelector}[ngbdropdownmenu]`); }
+
+  getDropdownMenuParent(dropdownMenu: ElementFinder) { return dropdownMenu.element(by.xpath('..')); }
+
+  getFirstItem(dropDownMenuSelector = 'div') { return $(dropDownMenuSelector).$$(`.dropdown-item`).first(); }
```
maxokorokov: unused method

e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.po.ts
```diff
@@ -26,4 +26,8 @@ export class DropdownAutoClosePage {
     const classNames = await dropdown.getAttribute('class');
     return classNames.includes('show');
   }
+
+  async appendToWidget() { await $('#container-null').click(); }
```
maxokorokov: unused method

e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.po.ts
```diff
@@ -26,4 +26,8 @@ export class DropdownAutoClosePage {
...
+
+  async appendToBody() { await $('#container-body').click(); }
```
maxokorokov: unused method

e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.e2e-spec.ts
```diff
@@ -1,115 +1,120 @@
...
-    // inside click
-    await page.getFirstItem(dropdown).click();
-    await expectDropdownToBeHidden(dropdown, `Dropdown should be closed on inside click`);
+import {DropdownPage} from '../dropdown.po';
+
+const containers = [null, 'body'];
+containers.forEach((container) => {
+  describe('Dropdown Autoclose with container = ' + container, () => {
+    let page: DropdownAutoClosePage;
+    let dropdownPage: DropdownPage;
```
maxokorokov: unused

e2e-app/src/app/dropdown/position/dropdown-position.po.ts
```diff
@@ -0,0 +1,15 @@
+import {$} from 'protractor';
+
+export class DropdownPositionPage {
```
maxokorokov: please check unused methods

src/dropdown/dropdown.ts
```diff
@@ -245,6 +272,11 @@ export class NgbDropdown implements OnInit,
   }
 
   ngOnDestroy() {
+    const menu = this._menu;
+    if (menu) {
```
maxokorokov: I don't think this check is necessary

crissty:
Great!! This property is very necessary. Thanks!
When are you plan to publish a new release with this PR?

PR #2830, opened 2018-10-26
feat(timepicker): use Angular locale data for AM and PM

### LINKED ISSUE #2822: Request: TimePicker should use Angular's i18n data to get meridian names
### Feature Request:

Angular's i18n data [contains the strings for AM and PM](https://github.com/angular/angular/blob/6737e91974c2f5d3bcbed68235cd9ff1730773b3/packages/common/src/i18n/locale_data_api.ts#L152-L165). The TimePicker currently [hard-codes the strings to "AM" and "PM"](https://github.com/ng-bootstrap/ng-bootstrap/blob/dac43f4b271fcf3724f3aec4d2dc1822f0bf3f8e/src/timepicker/timepicker.ts#L85-L86) (though it does supply an i18n attribute). 

The DatePicker uses Angular's locale data for day names and month names via the `NgbDatepickerI18n` service. The TimePicker should also use Angular's locale data via its own i18n service.

I'm more than happy to submit a PR if you agree with this idea. 😄 

Angular: 6+
ng-bootstrap: 3.3.1
Bootstrap: N/A

reduckted:
Fixes #2822.

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -13,6 +13,7 @@ import {isInteger, isNumber, padNumber, toInteger} from '../util/util';
 import {NgbTime} from './ngb-time';
 import {NgbTimepickerConfig} from './timepicker-config';
 import {NgbTimeAdapter} from './ngb-time-adapter';
+import {NgbTimepickerI18n} from './timepicker-i18n';
 
 const NGB_TIMEPICKER_VALUE_ACCESSOR = {
   provide: NG_VALUE_ACCESSOR,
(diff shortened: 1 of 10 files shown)


src/timepicker/timepicker.ts
```diff
@@ -82,8 +83,8 @@ const NGB_TIMEPICKER_VALUE_ACCESSOR = {
           <button type="button" class="btn btn-outline-primary" [class.btn-sm]="isSmallSize" [class.btn-lg]="isLargeSize"
             [disabled]="disabled" [class.disabled]="disabled"
                   (click)="toggleMeridian()">
-            <ng-container *ngIf="model?.hour >= 12; else am" i18n="@@ngb.timepicker.PM">PM</ng-container>
-            <ng-template #am i18n="@@ngb.timepicker.AM">AM</ng-template>
+            <ng-container *ngIf="model?.hour >= 12; else am">{{ i18n.getAfternoonPeriod() }}</ng-container>
```
benouat: I would have just replaced the `"PM"` with `{{i18n.getAfternoonPeriod()}}`.

Which lead us to one question: shouldn't we keep that extra level of translation via Angular i18n to let user use it, if they would need to ?

like

```html
<ng-container
  *ngIf="model?.hour >= 12; else am"
  i18n="@@ngb.timepicker.PM">{{ i18n.getAfternoonPeriod() }}</ng-container>
```

Default content would come from Angular Locale Data, If they want to override it with a custom translation, they can.

I know that in your example, you mention that they could still use a custom implementation and provide it, but translating something outside of a template is now quite cumbersome.

reduckted: OK. I just followed what the `Datepicker` component does and used only the Angular locale data, but if you think having the ability to translate via i18n is worth it, then I'll add those translation IDs back in.

I guess leaving the translation IDs there would also mean this isn't a breaking change.

I'll make the changes and push them up tomorrow. 😄

reduckted: Done! 👍

maxokorokov:
Sorry, will have to move this to `5.0.0` as technically might be a breaking change for those using non-default locale in their applications

benouat:
Hey @reduckted would you mind rebasing this PR in order to land it in `5.0.0` as soon as `5.0.0-rc.0` is built ?

reduckted:
Hi @benouat, I've just rebased onto master. Thanks for taking a look :)

Issue #2829, opened 2018-10-25
Feature request: Rating


pamidivenkat:
### Link to minimally-working StackBlitz that reproduces the issue:

You can fork a StackBlitz from one of our [demos](https://ng-bootstrap.github.io/#/components) and use it as a starting point.

Please note that we can _not_ act on bug reports without a _minimal_ reproduction scenario in a StackBlitz. Here is why:
https://github.com/ng-bootstrap/ng-bootstrap#you-think-youve-found-a-bug

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 6

ng-bootstrap: latest

Bootstrap: latest

baruchvlz:
Why? It's good UX to show where the mouse is hovering over so the user knows how many stars they're applying. Google, Yelp, Booking, they all do it with their own components.

pkozlowski-opensource:
We could expose the info to a custom rating template so users would know if `fill` value comes from hovering or from selection (essentially we could expose 2 fill values). At the same time agree with @baruchvlz - I'm not sure this is the pattern we want to "promote".

Marking as a feature request but don't get too much hope - it is very low priority for us so probably won't get fixed anytime soon (unless someone sends a good PR, of course).

pamidivenkat:
> Why? It's good UX to show where the mouse is hovering over so the user knows how many stars they're applying. Google, Yelp, Booking, they all do it with their own components.

I also like the feature, but my client didn't want it.

baruchvlz:
@pkozlowski-opensource Why not have another `Input` in the component called `mode` that defaults to `hover` and checks for the value before updating the state? This way it can also be used with the default rating template and not necessarily depend on a custom template.

```typescript
@Input() mode: 'hover' | 'selection' = 'hover';

// ...
enter(value: number): void {
  if (this.mode !== 'hover') {
    return;
  }

  if (!this.readonly && !this.disabled) {
    this._updateState(value);
  }

  this.hover.emit(value);
}
```

benouat:
Looking at this issue, I think it should be fixed in a totally different way, but way that would be a breaking change (so probably available in 5+)

Main idea here would be to not introduce a new `@Input()`, but more rely on the platform.

Let me explain:

We already have today an `@Output() hover`
https://github.com/ng-bootstrap/ng-bootstrap/blob/e0fe9f566061d6a346aab1742f4162ff23e10b70/src/rating/rating.ts#L104-L108

[...]

PR #2834, opened 2018-10-28
feat(rating): add flag to only do enter state update if mode is hover

baruchvlz:

Issue: #2829


src/rating/rating.ts
```diff
@@ -95,6 +95,11 @@ export class NgbRating implements ControlValueAccessor,
    */
   @Input() resettable: boolean;
 
+  /**
+   * A flag to know if rating should show when mouse enters or only on selection
+   */
+  @Input() mode: 'hover' | 'selection';
```
benouat: It might be a stupid suggestion.
Why not naming the input `updateOn` with 2 different possible values: `hover` and `click`
Default one being `click` so that we are backward compatible.

src/rating/rating.ts
```diff
@@ -130,12 +135,16 @@ export class NgbRating implements ControlValueAccessor,
...
   enter(value: number): void {
+    if (this.mode === 'selection') {
+      return;
+    }
+
     if (!this.readonly && !this.disabled) {
       this._updateState(value);
     }
+
     this.hover.emit(value);
```
benouat: We still want to emit this output whenever we hover a rating. Your test should only prevent the state update.

src/rating/rating.ts
baruchvlz: That's not a bad suggestion, but default should be `hover` to keep the current behaviour intact.

benouat: My bad, you're totally right !! 👍

baruchvlz:
@benouat I messed up my rebase with master, so those 2 last commits didn't really apply the changes. In any case, I am now getting a e2e error with the DatePicker. I'll see what I can do and if any of my changes are affecting this, though I don't see how that's possible.

PR #2843, opened 2018-11-05
fix(typehead): #2719 & #2786 & #2816 

### LINKED ISSUE #2719: NgbTypeahead required validation with reactive form works incorrectly
### Bug description:

In markup I have this code 

```html
<div class="first-group-margin form-group">
      <label for="recipient">{{ 'To' | translate }}:</label>
      <input id="recipient" type="text" class="tatoo-form-control" name="recipient" [ngbTypeahead]="search"
        [resultFormatter]="autocompleteFormatter" [inputFormatter]="autocompleteFormatter" [class.is-invalid]="recipient.invalid && recipient.touched"
        formControlName="recipient" [editable]="false" placeholder="{{ 'Placeholder.UserRecipient' | translate }}"
        autocomplete="false" (selectItem)="changeEvent($event)">
      <div class="alert-absolute alert alert-danger mt-2" *ngIf="recipient.touched && recipient.invalid">
        <span *ngIf="recipient.errors['required']">{{'Validation.BlankRecipient' | translate}}</span>
      </div>
</div>
```

and when I select value from dropdown the input field is 'blinking red' because change event occures earlier then typeahead select item event I think. So my form control become invalid for a while

Angular: 6.0.2

ng-bootstrap: 2.0.0

Bootstrap: 4.1.1

### LINKED ISSUE #2786: (typeagead, bug): Model value updated incorrectly when input in { updateOn: "blur" } mode
### Bug description:
There are 2 issues when `input`'s `FormControl` is set to `updateOn: "blur"` mode with Typeagead:
```
public formControl = new FormControl(null, { updateOn: "blur" });
```

1. When typing some text (`de`) and then clicking an item (`Delaware`) in the dropdown, it immediately sends `de` to a model. Since the focus is still meant to be on the `input`, this is not what's expected.

2. After that, if you **immediately** give the focus to some other element, the model would be left with `de` which is incorrect since `Delaware` is actually displayed in the `input`. However, if after clicking you give the focus to the `input` itself first and then move it to another element, the model gets correctly updated to `Delaware`.

[StackBlitz](https://stackblitz.com/edit/ngb-typeahead-blut-not-updatoing).

Angular: 6.1.10
ng-bootstrap: 3.3.0
Bootstrap: 4.1.3

### LINKED ISSUE #2816: it is not possible to reset input value for typeahead component
### Bug description:

### Steps to reproduce:
- introduce text "aaa"
- click on button "Clear"
- the input is cleared
- introduce text "ala"
- click on button "Clear"
- the input is not cleared

Angular: 6.1.0

ng-bootstrap: 3.3.0

Bootstrap: 4.0

wannadream:
Fix #2816

If input value has been emptied by changing model value, writing backup value back to input value should be skipped.

Fix #2786 

If updateOn is set to 'blur', update input value back to FormControl to coerce the synchronization.

Fix #2719 

If related target of blur() is from drop down list selected item, do not proceed to execute _onTouched().


benouat:
Thank you very much for your willingness to make things better, but we usually prefer to have small atomic PRs.

I find it to be hard to follow to do proper review.
Could you try to restart a fresh rebase from master, and create only one commit per issue fixed? Thanks!

wannadream:
Okay, will do.

PR #2844, opened 2018-11-05
chore(demos): dynamic generation of stackblitzes, reading names from sources

benouat:
Stackblitzes generation used to be relying on a lot of naming conventions.
Via this PR, generation for all widgets demos is using a set of regular expressions to read/extract/discover all the needed things:

- extract the list of demos (looking up for `const DEMO_DIRECTIVES = [...]`).
- get the actual filesystem path from the EcmaScript import of each demo directive.
- get the demo selector and template filesystem path from the @Component decorator
- read entryComponents classes from the @NgModule decorator

Creating a new demo is now **much more simple** as I does not rely on any hardcoded naming convention (apart form DEMO_DIRECTIVES, that already exists in each widget module)

+ demo folder name can be anything
+ component name / filename could be anything
+ template filename could be anything
+ demo selector could be anything

### DIFF
--- a/misc/generate-stackblitzes.ts
+++ b/misc/generate-stackblitzes.ts
@@ -1,9 +1,18 @@
 // tslint:disable:max-line-length
-import * as glob from 'glob';
 import * as fs from 'fs-extra';
+import * as glob from 'glob';
 import * as he from 'he';
+import {basename, dirname, join, normalize} from 'path';
+
+/**
+ * Generates StackBlitzes for all demos of all components and puts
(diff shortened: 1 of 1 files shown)


benouat:
ah ah nice ! at least some stuff to work on :) and improve.
I will have a look

benouat:
I just tested the 3 first ones, and they are all working

benouat:
Ohhh @maxokorokov you were right.. all the _others_ demos (the one outside of Examples tab) are broken !

PR #2845, opened 2018-11-05
feat(Accordion): add PanelControl template to accordion

### LINKED ISSUE #717: Accordion Feature Request - Title Content that does not cause expand/collapse

wclaibor:
Closes #717 
Adds a template outside of the accordion title button that allows controls to be added to the accordion header.


tplk:
It's quite an important feature for my current project, is there an estimation which release will include this PR?

maxokorokov:
Closing because of #2981, thanks for pushing for this change !

PR #2850, opened 2018-11-07
fix(typeahead): input was not always updated when model changes

### LINKED ISSUE #2816: it is not possible to reset input value for typeahead component

fbasso:
fix #2816

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -462,6 +462,31 @@ describe('ngb-typeahead', () => {
          tick(250);
          expectWindowResults(compiled, ['+one', 'one more']);
        }));
+    it('should stick model on click outside', async(() => {
+         const fixture = createTestComponent('<input [(ngModel)]="model" [ngbTypeahead]="find" />');
+         const compiled = fixture.nativeElement;
+         const el = fixture.nativeElement;
+
+         const comp = fixture.componentInstance;
+         expectInputValue(el, '');
+
+         fixture.whenStable()
+             .then(() => {
+               changeInput(compiled, 'o');
+               fixture.detectChanges();
+
+               expectInputValue(el, 'o');
+               comp.model = 'a';
+               fixture.detectChanges();
+               return fixture.whenStable();
+             })
+             .then(() => {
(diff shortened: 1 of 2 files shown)


benouat:
Could you please change your commit message and PR title to be fix(typeahead) ?

src/typeahead/typeahead.spec.ts
```diff
@@ -462,6 +462,31 @@ describe('ngb-typeahead', () => {
          tick(250);
          expectWindowResults(compiled, ['+one', 'one more']);
        }));
+    it('should stick model on click outside', async(() => {
```
maxokorokov: Sorry, I don't understand the message

src/typeahead/typeahead.ts
```diff
@@ -271,6 +270,7 @@ export class NgbTypeahead implements ControlValueAccessor,
           event.preventDefault();
           this._resubscribeTypeahead.next(null);
           this.dismissPopup();
+          this._writeInputValue(this._inputValueBackup);
```
maxokorokov: I don't get the fix, why do it only on ESC and not on document click for instance?
Also I think `dismissPopup` can be called manually from code, will it work?

src/typeahead/typeahead.ts
```diff
@@ -224,7 +224,6 @@ export class NgbTypeahead implements ControlValueAccessor,
   dismissPopup() {
     if (this.isPopupOpen()) {
       this._closePopup();
-      this._writeInputValue(this._inputValueBackup);
```
maxokorokov: As this is `showHint`-related, maybe just wrap this in:

```ts
if (this.showHint) {
  this._writeInputValue(this._inputValueBackup);
}
```

benouat:
After discussion with @maxokorokov and @fbasso I will work today on a different fix for that underlying issue #2816.

I will close that PR as soon as I open the new one.

PR #2855, opened 2018-11-08
fix(typehead): do not exec _onTouched if related target is from ddl.

### LINKED ISSUE #2719: NgbTypeahead required validation with reactive form works incorrectly

wannadream:
Fix #2719

If related target of blur() is from drop down list selected item, do not proceed to execute _onTouched().

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -701,6 +701,25 @@ describe('ngb-typeahead', () => {
            expect(fixture.componentInstance.model).toBeUndefined();
          });
        }));
+
+    it('should validate value after selecting an item from drop down list', async(() => {
+         const html = `
+          <form>
+            <input type="text" [(ngModel)]="model" name="control" required [ngbTypeahead]="find" />
+          </form>`;
+         const fixture = createTestComponent(html);
+         fixture.whenStable().then(() => {
+           fixture.detectChanges();
+           const compiled = fixture.nativeElement;
+
+           changeInput(compiled, 'o');
+           fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


maxokorokov:
Hey, @wannadream, thanks for the PR and for adding the test!

I've done some tests and the underlying issue was actually fixed by #2869 (the PR fixed several issues). However I was looking into the test you added to check if we could keep it for this use case.

Unfortunately, the unit test you added passes both before and after the fix.
The problem is that the `blur` is not triggered by your unit test neither by doing `.click()`, nor by simulating with `.triggerEventHandler()`.

This is a use case that is not easily testable by unit tests, so we'll go with an e2e test for this with #2872

PR #2857, opened 2018-11-08
fix(typehead): write value back to FormControl if updateOn is 'blur'.

### LINKED ISSUE #2786: (typeagead, bug): Model value updated incorrectly when input in { updateOn: "blur" } mode

wannadream:
Fix #2786

If updateOn is set to 'blur', update input value back to FormControl to coerce the synchronization.

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -13,9 +13,10 @@ import {
   Output,
   Renderer2,
   TemplateRef,
-  ViewContainerRef
+  ViewContainerRef,
+  Type
 } from '@angular/core';
-import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
+import {ControlValueAccessor, NG_VALUE_ACCESSOR, NgControl, FormControl} from '@angular/forms';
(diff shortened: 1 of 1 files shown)


maxokorokov:
Thanks for the contribution, however, the issue in question was also fixed by #2872

PR #2869, opened 2018-11-15
fix(typeahead): prevent dropdown to be reopen after an item click

### LINKED ISSUE #2854: Typeahead "Open on focus" instantly re-opens popup menu after clicking
### Bug description:

The gist is that you can't click on entries in the popup menu anymore if you follow the "Open on focus" example of typeahead and work with remote data. It's not directly reproducible on ng-bootstrap's website but I've posted a minimally-working StackBlitz below which is a combination of the "Wikipedia search" and "Open on focus" examples.

The core issue is probably that you can't distinguish between an actual focus by the user and an auto-focus by the breaking change.

Please see the full bug description here: https://github.com/ng-bootstrap/ng-bootstrap/issues/2792#issuecomment-436356694

@lucasklaassen was able to confirm the bug as well.

https://stackblitz.com/edit/angular-f1e9jn

Angular: 7.0.0 or higher

ng-bootstrap: 3.4.0 (but probably 3.3.0/3.3.1 or higher)

Bootstrap: 4.0.0 or higher

fbasso:
fix #2854

### DIFF
--- a/e2e-app/src/app/typeahead/focus/typeahead-focus.component.ts
+++ b/e2e-app/src/app/typeahead/focus/typeahead-focus.component.ts
@@ -1,7 +1,7 @@
 import {Component, ViewChild} from '@angular/core';
 import {NgbTypeahead} from '@ng-bootstrap/ng-bootstrap';
 import {Observable, Subject, merge} from 'rxjs';
-import {distinctUntilChanged, filter, map} from 'rxjs/operators';
+import {distinctUntilChanged, filter, map, delay} from 'rxjs/operators';
 
 const states = [
   'Alabama',
@@ -81,6 +81,7 @@ export class TypeaheadFocusComponent {
     return merge(debouncedText$, inputFocus$, clicksWithClosedPopup$)
         .pipe(
             map(term => (term === '' ? states : states.filter(v => v.toLowerCase().indexOf(term.toLowerCase()) > -1))
-                            .slice(0, 10)));
+                            .slice(0, 10)),
+            delay(0));
   }
 }
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -310,7 +310,6 @@ export class NgbTypeahead implements ControlValueAccessor,
   private _selectResultClosePopup(result: any) {
     this._selectResult(result);
     this._closePopup();
-    this._elementRef.nativeElement.focus();
   }
 
   private _showHint() {
--- a/e2e-app/src/app/typeahead/focus/typeahead-focus.e2e-spec.ts
+++ b/e2e-app/src/app/typeahead/focus/typeahead-focus.e2e-spec.ts
@@ -58,7 +58,7 @@ describe('Typeahead', () => {
 
        await page.getDropdownItems().get(0).click();
        expectTypeaheadValue('Colorado');
-       // expectTypeaheadFocused();
+       expectTypeaheadFocused();
 
        await page.getInputBefore().click();
        await sendKey(Key.TAB);
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -20,7 +20,7 @@ export interface ResultTemplateContext {
 @Component({
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
-  host: {'class': 'dropdown-menu show', 'role': 'listbox', '[id]': 'id'},
+  host: {'(mousedown)': '$event.preventDefault()', 'class': 'dropdown-menu show', 'role': 'listbox', '[id]': 'id'},
   template: `
     <ng-template #rt let-result="result" let-term="term" let-formatter="formatter">
       <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>


fbasso:
The last push is a complete refactor of the solution proposed in #2805. The mousedown is listen on the dropdown, to prevent the event default action. It simply prevents the input to loose the focus.

We should definitively consider this PR, as it's a better way to manage the typeahead and it comes with some advantages:

- The focus is not lost on item click, so no need to give it back (fix #2805)
- If open on focus is set, the dropdown is not reopen after an item click, neither a search perform again (fix #2869)
- I haven't been checked yet, but it possibly fixes #2786, as the blur on the input is never triggered on an item click
- The same for #2719 (to be checked too)

src/typeahead/typeahead.ts
```diff
@@ -284,13 +285,20 @@ export class NgbTypeahead implements ControlValueAccessor,
       this._windowRef.instance.selectEvent.subscribe((result: any) => this._selectResultClosePopup(result));
       this._windowRef.instance.activeChangeEvent.subscribe((activeId: string) => this.activeDescendant = activeId);
 
+      const nativeElement = this._windowRef.location.nativeElement;
+      this._mousedownRemoveListener =
+          this._renderer.listen(nativeElement, 'mousedown', (event) => { event.preventDefault(); });
+
       if (this.container === 'body') {
-        window.document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
+        window.document.querySelector(this.container).appendChild(nativeElement);
```
maxokorokov: you should `@Inject(DOCUMENT) private _document: any`, not use `window.document`

src/typeahead/typeahead.ts
```diff
@@ -284,13 +285,20 @@ export class NgbTypeahead implements ControlValueAccessor,
       this._windowRef.instance.selectEvent.subscribe((result: any) => this._selectResultClosePopup(result));
       this._windowRef.instance.activeChangeEvent.subscribe((activeId: string) => this.activeDescendant = activeId);
 
+      const nativeElement = this._windowRef.location.nativeElement;
```
maxokorokov: Can you please rename `nativeElement` → `windowElement` and add a 1-2 line comment on why have this 'mousedown'. I don't think anybody will remember this in a year :)

e2e-app/src/app/typeahead/focus/typeahead-focus.component.ts
```diff
@@ -1,86 +1,43 @@
...
+  constructor(private _service: SearchService) {}
+
   search = (text$: Observable<string>) => {
-    const debouncedText$ = text$.pipe(distinctUntilChanged());
+
+    const debouncedText$ = text$.pipe(debounceTime(200), distinctUntilChanged());
     const clicksWithClosedPopup$ = this.click$.pipe(filter(() => !this.instance.isPopupOpen()));
     const inputFocus$ = this.focus$;
 
     return merge(debouncedText$, inputFocus$, clicksWithClosedPopup$)
```
maxokorokov: I think the problem is that current code is synchronous. To reproduce we should just add the `delay(0)` operator, maybe ?

```ts
return merge(...)
    .pipe(
        map(...),
        delay(0)
    );
```

And remove all the changes with HTTP to keep use cases as simple as possible

src/typeahead/typeahead.ts
```diff
@@ -80,6 +80,7 @@ export class NgbTypeahead implements ControlValueAccessor,
   private _valueChanges: Observable<string>;
   private _resubscribeTypeahead: BehaviorSubject<any>;
   private _windowRef: ComponentRef<NgbTypeaheadWindow>;
+  private _mousedownRemoveListener: Function;
```
maxokorokov: maybe `_removeMousedownListener` ?

src/typeahead/typeahead.ts
maxokorokov: Oh, see it was like that before, but is worth fixing anyway!

fbasso:
@maxokorokov Thanks for your comments ! I changed the code accordingly.

src/typeahead/typeahead.ts
```diff
@@ -25,6 +26,7 @@ import {Key} from '../util/key';
 import {Live} from '../util/accessibility/live';
 import {NgbTypeaheadConfig} from './typeahead-config';
 import {map, switchMap, tap} from 'rxjs/operators';
+import {DOCUMENT} from '@angular/platform-browser';
```
maxokorokov: this is deprecated, should be:
```ts
import {DOCUMENT} from '@angular/common';
```

src/typeahead/typeahead.ts
```diff
@@ -284,13 +288,20 @@ export class NgbTypeahead implements ControlValueAccessor,
       this._windowRef.instance.selectEvent.subscribe((result: any) => this._selectResultClosePopup(result));
       this._windowRef.instance.activeChangeEvent.subscribe((activeId: string) => this.activeDescendant = activeId);
 
+      const windowElement = this._windowRef.location.nativeElement;
+      this._removeMousedownListener =
+          this._renderer.listen(windowElement, 'mousedown', (event) => { event.preventDefault(); });
```
maxokorokov: @benouat has just suggested to remove all this renderers and listeners and move this all to the `typeahead-window.ts`... it would be just a one-liner, no idea how we missed it before :)

```ts
@Component({
  selector: 'ngb-typeahead-window',
  host: {
    '(mousedown)': '$event.preventDefault()',
    /* ... */
  },
```

Issue #2880, opened 2018-11-20
Tooltip isn't shown correctly when placement is left and has a longer text


irega:
### Bug description:

Tooltip isn't shown correctly when placement is left and has a longer text, directive calculates a position left negative (we think should be greater or equals to zero)

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-ashlnx

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 7.0.1

ng-bootstrap: 4.0.0

Bootstrap: 4.0.0

pkozlowski-opensource:
@irega this works as expected in my books. You've asked the tooltip to be positioned `left` so it happily complied. If you want to prevent it going off the screen you can use the more positioning options, ex.: `[placement]="['left', 'top-left', 'auto']"`. This would mean - try on the left, if there is not enough space try `top-left` and if this doesn't fit try all other placements. Obviously you can specify an arbitrary number of placements to try: https://stackblitz.com/edit/angular-iyqree?file=src/app/app.component.html

Going to close this one as it works as designed / expected.

irega:
@pkozlowski-opensource sorry we don't know we can set an array with multiple options as placement. Even so in our production app we have a wrapper container with a scroll, so we had to move all tooltip elements to left and use "top-left" as placement. 

Thanks so much!

PR #2881, opened 2018-11-20
fixed position left negative for tooltip #2880

irega:
Fix for https://github.com/ng-bootstrap/ng-bootstrap/issues/2880 issue.

This fix is which we are using in our application in production. We send you PR to remove our duplicate code.


pkozlowski-opensource:
@irega thnx for the PR but I don't think this is a correct fix, see https://github.com/ng-bootstrap/ng-bootstrap/issues/2880#issuecomment-440997405

PR #2882, opened 2018-11-21
chore(test): move e2e specs to their corresponding e2e-app folders

fbasso:
The purpose of this PR is to discuss about the e2e specs location. 

At the moment, the e2e-app and the e2e-specs are in two separated structures. It seems to be more intuitive to have them merged, the same way the unit tests are located in their corresponding widget folders.


e2e-app/src/app/datepicker/autoclose/datepicker-autoclose.e2e-spec.ts
```diff
@@ -1,6 +1,6 @@
 import {Key} from 'protractor';
 import {DatepickerAutoClosePage} from './datepicker-autoclose.po';
-import {openUrl, sendKey} from '../tools';
+import {openUrl, sendKey} from '../../tools.po';
```
maxokorokov: didn't get why you renamed `tools.ts` → `tools.po.ts` as its just helper functions and not a page object. ex `openUrl`?

PR #2888, opened 2018-11-28
fix(focustrap): discarding tabindex='-1' when finding element

### LINKED ISSUE #2884: Modal: Elements with tabindex="-1" get focused
When opening a modal, the documentation says the first focusable element is focused (if there is no ngbAutofocus marker). But when I assign `tabindex="-1"` to the first focusable element, it is still focused.

Link to StackBlitz reproduction: [https://stackblitz.com/edit/angular-aoqoet](https://stackblitz.com/edit/angular-aoqoet)
The expection here is, that instead of the close button, the date field gets focused. But if you check with `document.activeElement`, the close button is still focused

Angular: 6.1.9

ng-bootstrap: 3.3.0, but is not yet fixed in master branch

Bootstrap: 4.1.3

I inspected the code and found the place, where the bug occurs:
src/util/focus-trap.ts
Line: 6-9

Today it is
`const FOCUSABLE_ELEMENTS_SELECTOR = [
  'a[href]', 'button:not([disabled])', 'input:not([disabled]):not([type="hidden"])', 'select:not([disabled])',
  'textarea:not([disabled])', '[contenteditable]', '[tabindex]:not([tabindex="-1"])'
].join(', ');`

should be like this (note the `:not([tabindex="-1"])`) after each selector

`var FOCUSABLE_ELEMENTS_SELECTOR = [
    'a[href]:not([tabindex="-1"])', 'button:not([disabled]):not([tabindex="-1"])', 'input:not([disabled]):not([type="hidden"]):not([tabindex="-1"])', 'select:not([disabled]):not([tabindex="-1"])',
    'textarea:not([disabled]):not([tabindex="-1"])', '[contenteditable]:not([tabindex="-1"])', '[tabindex]:not([tabindex="-1"])'
].join(', ');`

benouat:
When focus-trap needs to figure out what are the boundaries elements (first and last focusable elements) we need to perform an extra check to filter out elements with `tabindex="-1"`

Closes #2884

### DIFF
--- a/src/modal/modal.spec.ts
+++ b/src/modal/modal.spec.ts
@@ -14,6 +14,7 @@ import {async, ComponentFixture, TestBed} from '@angular/core/testing';
 import {NgbModalConfig} from './modal-config';
 import {NgbActiveModal, NgbModal, NgbModalModule, NgbModalRef} from './modal.module';
 
+
 const NOOP = () => {};
 
 @Injectable()
@@ -747,6 +748,16 @@ describe('ngb-modal', () => {
           fixture.detectChanges();
         });
 
+        it('should skip element with tabindex=-1 when finding the first focusable element', () => {
+          fixture.detectChanges();
+          const modal = fixture.componentInstance.openCmpt(WithSkipTabindexFirstFocusableModalCmpt);
+          fixture.detectChanges();
+
+          expect(document.activeElement).toBe(document.querySelector('button.other'));
+          modal.close();
+          fixture.detectChanges();
+        });
+
         it('should focus modal window as a default fallback option', () => {
           fixture.detectChanges();
           const modal = fixture.componentInstance.open('content');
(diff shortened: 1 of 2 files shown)


src/modal/modal.spec.ts
```diff
@@ -882,6 +893,16 @@ export class WithAutofocusModalCmpt {
...
 }
 
+@Component({
+  selector: 'modal-skip-tabindex-firstfocusable-cmpt',
+  template: `
+  <button tabindex="-1" class="firstFocusable close">Close</button>
+  <button class="other">Other button</button>
+`
+})
+export class WithSkipTabindexFirstFocusableModalCmpt {
```
maxokorokov: Good naming! Though, could you change to `WithNegativeTabIndexElementThatShouldBeSkippedWhileFindingFirstFocusableElementModalComponent`? 

Just kidding :)

PR #2891, opened 2018-11-28
fix(typeahead): input value properly reset when hint/ngModel are used together

### LINKED ISSUE #2816: it is not possible to reset input value for typeahead component

benouat:
I refactored a bit (very very little bit) the way `showHint` and especially how to backup and restore input value properly when:

+ model gets reset
+ window is dismissed (Esc)
+ window is dismissed (click outside)

fixes #2816
closes #2850 (this PR closes and replaces that one)

PS: on the long run `showHint` would probably need a complete overhaul

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -13,18 +13,21 @@ import {
   Output,
   Renderer2,
   TemplateRef,
-  ViewContainerRef
+  ViewContainerRef,
 } from '@angular/core';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
-import {Observable, BehaviorSubject, Subscription, fromEvent} from 'rxjs';
-import {positionElements, PlacementArray} from '../util/positioning';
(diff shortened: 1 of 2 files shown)


src/typeahead/typeahead.ts
```diff
@@ -206,7 +209,12 @@ export class NgbTypeahead implements ControlValueAccessor,
 
   registerOnTouched(fn: () => any): void { this._onTouched = fn; }
 
-  writeValue(value) { this._writeInputValue(this._formatItemForInput(value)); }
+  writeValue(value) {
+    this._writeInputValue(this._formatItemForInput(value));
+    if (this.showHint) {
```
maxokorokov: I don't think it is necessary, because the first thing you do on user input is already `this._inputValueBackup = value;`

benouat: I removed it.

PR #2893, opened 2018-11-29
chore: bump CLI to 7.1 + dev dependencies

maxokorokov:
Bumping dev dependencies:

- `@angular/cli` from 7.0 to 7.1 → this fixes the following error in TDD mode:
```
xxx.spec.ts is missing from the TypeScript compilation
```
- `typescript` from 3.1.1 to 3.1.6 →  this fixes some regressions in compilation performance
- `tsikle` from 0.33 to 0.34 → this fixes the following installation warning:
```
warning " > tsickle@0.33.1" has incorrect peer dependency "typescript@3.1.3"
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "rxjs": "6.3.0",
     "ts-node": "~7.0.0",
     "ts-node-dev": "^1.0.0-pre.30",
-    "tsickle": "^0.33.0",
+    "tsickle": "^0.34.0",
     "tslib": "^1.9.0",
     "tslint": "^5.11.0",
     "tslint-jasmine-rules": "^1.3.2",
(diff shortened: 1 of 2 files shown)


maxokorokov:
@pkozlowski-opensource not that it will answer your question, but the diff between what was built before (`cli@7.0.1`, `tsikle@0.33`) and after this PR (`cli@7.1.0`, `tsikle@0.34`) can be resumed in this screenshot below.

Otherwise apart these comment changes everything is the same. So I'd say at least _THIS_ particular change should be safe. 

screen shot 2018-12-04 at 16 30 05

maxokorokov:
Even though it won't break anything, will close this PR. Will update deps in `5.0`.

PR #2901, opened 2018-12-04
Refactor keyboard handling to avoid event hijacking

### LINKED ISSUE #2895: ngb-rating swallows traps user keyboard navigation once focused
### Bug description:

Once `ngbRating` gets focused using keyboard navigation, it seems there is no reliable way to move focus to the next control and user navigation capabilities are lost - unless user clicks outside of the control to release focus

https://stackblitz.com/edit/angular-ng-rating-controls

- click within app to allow keyboard navigation
- move to Button via TAB so it gets focus outline
- move focus via TAB to next controls: rating control
- control gets focused
- try to get focus to next control using TAB

Angular: 7.0.*

ng-bootstrap: 4.0.0

Bootstrap: 4.1.*

Thanks!

maxokorokov:
Before #2473 `Keys` corresponded ONLY to keys processed by the current widget
After the change `Keys` contain the list of ALL known Keys to ALL widgets.

Because of this some `preventDefault()` were called on keys not processed by current widget.

(2 commits)

Fixes #2895

P.S. Not sure what kind of tests to add for this

### DIFF
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -244,35 +244,32 @@ export class NgbTypeahead implements ControlValueAccessor,
     }
 
     // tslint:disable-next-line:deprecation
-    const {which} = event;
-    if (Key[toString(which)]) {
-      switch (which) {
-        case Key.ArrowDown:
+    switch (event.which) {
+      case Key.ArrowDown:
+        event.preventDefault();
+        this._windowRef.instance.next();
+        this._showHint();
+        break;
+      case Key.ArrowUp:
+        event.preventDefault();
+        this._windowRef.instance.prev();
+        this._showHint();
+        break;
+      case Key.Enter:
+      case Key.Tab:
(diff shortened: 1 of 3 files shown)


benouat:
Not sure any kind of test should be added for this.

src/typeahead/typeahead.ts
```diff
@@ -244,35 +244,34 @@ export class NgbTypeahead implements ControlValueAccessor,
...
+          this._selectResult(result);
+        }
+        this._closePopup();
+        break;
+      case Key.Escape:
+        event.preventDefault();
+        this._resubscribeTypeahead.next(null);
+        this.dismissPopup();
+        break;
+      default:
```
benouat: In this particular usage, adding the `default: return` seems to be extraneous

maxokorokov: agreed, removed it

PR #2903, opened 2018-12-05
fix(timepicker): setting hour/minute/second step to undefined

### LINKED ISSUE #2851: timepicker: Setting minuteStep undefined
### Bug description:

Setting input minuteStep  to undefined on timepicker will set time to null when decreasing minutes, and it will act normal on increasing minutes.

Expected: throw an error that minuteStep is undefined, or get notified somehow.

Angular: 7.0.1

ng-bootstrap: 4.0.0

Bootstrap: 4.0.0

avsorokin:
Closes #2851

Also added some tests for step setting via component input properties

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -1216,188 +1216,164 @@ describe('ngb-timepicker', () => {
 
     beforeEach(() => {
       TestBed.configureTestingModule(
-        {imports: [NgbTimepickerModule], providers: [{provide: NgbTimepickerConfig, useValue: config}]});
+          {imports: [NgbTimepickerModule], providers: [{provide: NgbTimepickerConfig, useValue: config}]});
     });
 
-    it('should increment / decrement hours by 6', async(() => {
-      const html = `<ngb-timepicker [(ngModel)]="model" [hourStep]="6"></ngb-timepicker>`;
+    it('should increment / decrement hours by 6', async(async() => {
+         const html = `<ngb-timepicker [(ngModel)]="model" [hourStep]="6"></ngb-timepicker>`;
 
-      const fixture = createTestComponent(html);
-      fixture.componentInstance.model = {hour: 10, minute: 30, second: 0};
-      fixture.detectChanges();
-      fixture.whenStable()
-        .then(() => {
-          fixture.detectChanges();
-          return fixture.whenStable();
-        })
-        .then(() => {
+         const fixture = createTestComponent(html);
+         fixture.componentInstance.model = {hour: 10, minute: 30, second: 0};
+         fixture.detectChanges();
+         await fixture.whenStable();
 
-          const buttons = getButtons(fixture.nativeElement);
+         fixture.detectChanges();
+         await fixture.whenStable();
+         const buttons = getButtons(fixture.nativeElement);
 
-          expectToDisplayTime(fixture.nativeElement, '10:30:00');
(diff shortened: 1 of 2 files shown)


src/timepicker/timepicker.ts
```diff
@@ -96,6 +96,10 @@ export class NgbTimepicker implements ControlValueAccessor,
   disabled: boolean;
   model: NgbTime;
 
+  private _hourStep;
```
maxokorokov: please add `: number` type

src/timepicker/timepicker.ts
```diff
@@ -136,7 +159,7 @@ export class NgbTimepicker implements ControlValueAccessor,
    */
   @Input() size: 'small' | 'medium' | 'large';
 
-  constructor(config: NgbTimepickerConfig, private _ngbTimeAdapter: NgbTimeAdapter<any>) {
+  constructor(private readonly config: NgbTimepickerConfig, private _ngbTimeAdapter: NgbTimeAdapter<any>) {
```
maxokorokov: As config becomes private, please rename to `_config` (see https://github.com/ng-bootstrap/ng-bootstrap/wiki/Contributions%3A-Code-conventions)

src/timepicker/timepicker.spec.ts
```diff
@@ -1207,7 +1207,198 @@ describe('ngb-timepicker', () => {
...
+          expect(fixture.componentInstance.model).toEqual({hour: 12, minute: 30, second: 0});
+
+          (<HTMLButtonElement>buttons[1]).click();  // H-
+          fixture.detectChanges();
+          expectToDisplayTime(fixture.nativeElement, '10:30:00');
+          expect(fixture.componentInstance.model).toEqual({hour: 10, minute: 30, second: 0});
+        });
+    }));
+
+    it('should increment / decrement minutes by 7', async(() => {
```
maxokorokov: Could you please simplify tests, as we should use `async/await` or `fakeAsync` (I know you were inspired by already exsiting ones, but now there is a shorter way to write) → 

Before:

```ts
it('should increment / decrement minutes by 7', async(() => {
  //...
  fixture.detectChanges();
  fixture.whenStable()
    .then(() => {
      fixture.detectChanges();
      return fixture.whenStable();
    })
    .then(() => {
      // ...
    });
}));
```

After:
```ts
it('should increment / decrement minutes by 7', async(async () => {
  // ...
  fixture.detectChanges();
  await fixture.whenStable();
  fixture.detectChanges();
  // ...
}));
```

src/timepicker/timepicker.ts
```diff
@@ -114,17 +118,36 @@ export class NgbTimepicker implements ControlValueAccessor,
   /**
    * Number of hours to increase or decrease when using a button.
    */
-  @Input() hourStep: number;
+  @Input() set hourStep(step: number) {
+    this._hourStep = isNumber(step) ? step : this.config.hourStep;
```
maxokorokov: Maybe better `isInteger()` ?

maxokorokov:
Oh, also the build failed because of improper code formatting: https://travis-ci.org/ng-bootstrap/ng-bootstrap/builds/463821660#L508

You could run `yarn ci` locally to run all things Traivs CI does ot just `yarn check-format` to check formatting. More details here: https://github.com/ng-bootstrap/ng-bootstrap/blob/master/DEVELOPER.md#useful-commands. Thanks for contributing!

Issue #459, opened 2016-07-22
timepicker: add support for arrowkeys and mousewheel


pkozlowski-opensource:
Ref: http://angular-ui.github.io/bootstrap/#/timepicker

It should be trivial to implement but needs tests (as always!)

sendilkumarn:
will take this one 👍

deeg:
@pkozlowski-opensource, just tested this, it seems to work on the demo. Can probably be closed.

jdhines:
Looks like this has been in the pipeline for months. Is the PR still in limbo?

jnizet:
@pkozlowski-opensource is there a chance at least keyboard support is added to the timepicker? You chose to close #1334 with the comment "We will pass on this one for now till we figure out what to do with the timepicker...", but I'm not sure what you meant by that.

Orodan:
Is there any reason to keep the issue opened ? From what I can see in the documentation the arrow keys support is working :)

maxokorokov:
Last piece closed in #2053. I don't think implementing mouse wheel is a good idea, for example because of accidental scrolls it might introduce.

PR #2912, opened 2018-12-07
feat(timepicker): use arrow keys to increment/decrement values

jnizet:
Mousewheel support is more controversial, and would probably
need an additional input + config to enable/disable it, as it's used by
users which are not proficient keyboard users to scroll the page.
But I think keyboard support is a must, and can be turned on always.

refs #459

### DIFF
--- a/src/timepicker/timepicker.spec.ts
+++ b/src/timepicker/timepicker.spec.ts
@@ -1,7 +1,7 @@
 import {async, ComponentFixture, inject, TestBed} from '@angular/core/testing';
 import {createGenericTestComponent} from '../test/common';
 
-import {ChangeDetectionStrategy, Component, Injectable} from '@angular/core';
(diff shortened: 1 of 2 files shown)


maxokorokov:
Looks like its a simpler alternative to #2053

jnizet:
Hi @maxokorokov Sorry I missed your comment, and also missed this previous PR.

Regarding your first comment, I agree that it looks a bit weird, but it doesn't look like a deal breaker to me either (and I'm not sure how to change it either). What does loo like a deal breaker is the lack of keyboard support :-)

Your second comment I also agree with, but I have the feeling that fixing it would require quite a lot of dirty browser-dependant code for a small benefit.

Since these two issues (IMHO) are cosmetic issues, but the lack of keyboard support is a big functional issue, can't we at least start with this and fix those issues later if necessary?

Issue #2847, opened 2018-11-06
Feature Request:  ngb-accordion card-header accessibility


k-spen:
It seems as though the rendered code looks like this:

```
<div role="tab" id="ngb-panel-13-header" class="card-header ">
<h5 class="mb-0"><button class="btn btn-link collapsed" type="button" aria-expanded="false" aria-controls="ngb-panel-13"> Simple
</button></h5></div>
```

Would it be possible to remove the h5?  This breaks some accessibility rules.  It's not good to skip heading levels and in the context of my page, it goes from h2 to an h5.  Screen readers expect there to be order.  Thanks.

Angular: 6

ng-bootstrap: 3.3.1

Bootstrap: 4

pkozlowski-opensource:
Hmm, interesting. This is actually the markup mandated by Bootstrap itself, see: https://getbootstrap.com/docs/4.1/components/collapse/#accordion-example

As such it should be reported to https://github.com/twbs/bootstrap/issues first. Could you please report the issue in there? We would update our markup as it is considered a bug and fixed on the Bootstrap's end

pkozlowski-opensource:
OK, this got clarified on the Bootstrap side: https://github.com/twbs/bootstrap/issues/27619

tl;dr; we should replace `<h5>` with sth else (probably `<div>`). PRs welcomed!

maxokorokov:
Fixed by #2915, but see #2927 for more accessibility issues

PR #2915, opened 2018-12-09
fix(accordion): do not use static header level for heading

peterblazejewicz:
As discussed in #2847, this PR:

- replaces `h5` header tag with `div` tag removing no longer necessary `mb-0` helper class
~~- add heading role attribute to the `div` tag
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role~~

Thanks!

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -115,13 +115,11 @@ export interface NgbPanelChangeEvent {
     <ng-template ngFor let-panel [ngForOf]="panels">
       <div class="card">
         <div role="tab" id="{{panel.id}}-header" [class]="'card-header ' + (panel.type ? 'bg-'+panel.type: type ? 'bg-'+type : '')">
-          <h5 class="mb-0">
-            <button type="button" class="btn btn-link"
-              (click)="toggle(panel.id)" [disabled]="panel.disabled" [class.collapsed]="!panel.isOpen"
-              [attr.aria-expanded]="panel.isOpen" [attr.aria-controls]="panel.id">
-              {{panel.title}}<ng-template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></ng-template>
-            </button>
-          </h5>
+          <button type="button" class="btn btn-link"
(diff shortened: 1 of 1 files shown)


peterblazejewicz:
@maxokorokov 
Done, squashed and rebased.
I've probably misread the docs and fully agree with you, cheers!

src/accordion/accordion.ts
```diff
@@ -115,13 +115,13 @@ export interface NgbPanelChangeEvent {
     <ng-template ngFor let-panel [ngForOf]="panels">
       <div class="card">
         <div role="tab" id="{{panel.id}}-header" [class]="'card-header ' + (panel.type ? 'bg-'+panel.type: type ? 'bg-'+type : '')">
-          <h5 class="mb-0">
+          <div>
```
maxokorokov: Don't you think that in this case we can remove this element altogether? It doesn't affect rendering, so I don't see any reasons in having it at all (that was the second part of my previous review).

On the bootstrap side they've changed example from `<h5>` → `<h2>`, but again it's just an example. Any kind of markup could be put into cards.

peterblazejewicz: agree, removed the wrapping `div` completely in the updated version.
Thanks!

troddick:
After updating to 4.0.2, I'm still seeing the header text running out of the <ng-template ngbPanelTitle> container. I thought it may be my code, so I tested on the StackBlitz on the components/accordion/examples page. It is still running over rather than word-wrapping there too. Am I missing something?
Thanks!
https://stackblitz.com/edit/angular-ukshlz?file=app/accordion-basic.html

peterblazejewicz:
@troddick 
Hey, you need to use BS in version at least 4.2.1 to have at least text that wraps in buttons used in accordions (collapse):
https://github.com/twbs/bootstrap/commit/24bdadf9d3246b8b7668bd442dccf1f2aa5fae37#diff-8a4fc4262c9197c73b80fb80d8863ae5

> After updating to 4.0.2

This is the BS version used in your SBlitz

troddick:
That did it! I didn't realize I needed to update BS also. It's working for me now.
Thanks!!

PR #2921, opened 2018-12-11
docs(home): add direct link to project code coverage

peterblazejewicz:
This wraps the statement about code coverage into the link,
to allow users see that coverage. Should enforce user trust.

Thanks!

image


peterblazejewicz:
> Thanks, good idea!

Thanks!

PR #2925, opened 2018-12-13
fix(datepicker): focus handling performance regression

maxokorokov:
Removes unnecessary model updates and change detections when focus moves
inside the datepicker

Before:

before

After:

after

Two commits, don't squash


maxokorokov:
Thanks for reviewing, @benouat !

Issue #2930, opened 2018-12-14
(tabset): ViewEncapsulation.None


AbakumovAlexandr:
### Bug description:
Maybe I'm missing something, but in [a25d5d2](https://github.com/ng-bootstrap/ng-bootstrap/commit/a25d5d2) `ViewEncapsulation` was switched to `None` for all widgets.

However, [`ngb-tabset`](https://github.com/ng-bootstrap/ng-bootstrap/blob/1d9a84ed4fd55c6c0f84b6569dac7a298290d011/src/tabset/tabset.ts#L90) still uses the `ViewEncapsulation.Emulated`.

Can we have `ngb-tabset` switched to `ViewEncapsulation.None` as well?

### Versions of Angular, ng-bootstrap and Bootstrap:

ng-bootstrap: 4.1.3

maxokorokov:
@AbakumovAlexandr, not only `tabset`, but basically all widgets that don't have `styles` or `styleUrls`: `accordion`, `rating`,  `buttons`, etc.

It should be `None` automatically if I understand the docs correctly: https://angular.io/api/core/Component#encapsulation

Are you having any specific issues?

P.S. On a side note, we've forgotten to add `ViewEncapsulation.None` to the `timepicker`...

AbakumovAlexandr:
@maxokorokov Oh, thank you! I was aware that `ViewEncapsulation.None` would be set automatically for components that don't have own styles.

Also, I've looked more into how `ViewEncapsulation.None` works and understand now that my issue isn't come from NGB, so this one could be closed I guess.

As for your side note, just a suggestion. Maybe we still need to explicitly declare `ViewEncapsulation.None` for all components? Otherwise during further maintenance when someone would need to introduce some styling to a component, there is a high likelihood that he'd forget to add that explicit declaration and miss the fact that the component became `ViewEncapsulation.Emulated`.

maxokorokov:
> As for your side note, just a suggestion. Maybe we still need to explicitly declare ViewEncapsulation.None for all components?

@AbakumovAlexandr that's a good suggestion, thanks!

PR #2932, opened 2018-12-17
fix(timepicker): set ViewEncapsulation.None forgotten in 4.0.0

maxokorokov:
Part of #2930

cc @fbasso


fbasso:
A class is required for the input styling inside the scss, otherwise the viewEncapsulation none will apply the rules to all input : https://github.com/ng-bootstrap/ng-bootstrap/blob/ff6dd0130788869bbd6fc68c59e90baf51dc5d69/src/timepicker/timepicker.scss#L50

Another remark : a refactor would be nice to group some rules under the same base, for example:

``` css
.ngb-tp {
    ...
    &-input-container {
       ...
    }
}
```

maxokorokov:
@fbasso thanks, amended!

maxokorokov:
@fbasso I've added `.ngb-tb-chevron`, `.ngb-tp-input` and kept `.chevron` so it would work for any users who tried to override `.chevron` class.

The only way they could override `.chevron` is with adding `!important`, so all should be fine.

fbasso:
LGTM

Issue #2904, opened 2018-12-05
Feature request: change NgbModalRef's options after the modal is opened


EpaXapate:
It would be nice to be able to change the properties of the opened modal such as `backdrop or` `beforeDismiss.`
I need to change the `backdrop` based on a conditional promise so that the user cannot close the modal when such condition is true.

benouat:
@EpaXapate I don't really see the added value to be able to update these values dynamically ....
Couldn't you simply consume your promise before calling `ngbModalService.open()` ?

EpaXapate:
@benouat Not for my specific use case. What I need is to open the modal like you normalli would and once inside said modal I would make a post request. While that promise is resolving I need to disable the `backdrop` and `keyboard` click events so the user can't close the modal.

EpaXapate:
@benouat Could this issue be addressed any time soon or would a pull request be more convenient?

benouat:
If you already have a PR ready, or if you are willing to invest some time on it, feel free to do so.

Any contribution is always welcome.

EpaXapate:
@Terrafire123 Yes that is one way to do it and considering the current codebase I believe there is no other way to do it without typescript complaining.

EpaXapate:
@benouat You misunderstand my use case. I need to change the beforeDismiss() property inside the modal component's class.

benouat:
@EpaXapate I know !! :) I was just commenting regarding the workaround that @Terrafire123 posted.

Nonetheless, if your post request could be done from outside the modal's component class (like maybe using observables and/or promises), it think it could work 👍

EpaXapate:
@benouat Yes indeed I believe it would and it is a nice way of doing it aswell, unfortunately my post request must be done inside the modal aswell.

benouat:
Another try :)

```typescript
import { Component, Input } from '@angular/core';
import { NgbActiveModal, NgbModal } from '@ng-bootstrap/ng-bootstrap';

@Component({
  selector: 'ngbd-modal-content',
  template: `
    <div class="modal-header">
      <h4 class="modal-title">Hi there!</h4>
      <button type="button" class="close" aria-label="Close" (click)="activeModal.dismiss('Cross click')">
        <span aria-hidden="true">&times;</span>
      </button>
    </div>
    <div class="modal-body">
      <p>Hello, {{name}}!</p>
    </div>
    <div class="modal-footer">
[...]

PR #2936, opened 2018-12-18
feat(modal): change opts after the modal is opened (#2904)

EpaXapate:


benouat:
I just had a brief look at your code, and I don't think it's a good/viable option to actually provide access to the instance of an internal component. This `windowComponent` thing is a _private_ thing that we probably don't want to expose...

EpaXapate:
@benouat Yes I can see how that could be troublesome if you want to make breaking changes in the future. I took this approach because of the way the properties are inherited through the modal's instantiation. Considering they are copied, you lose the hability to simply change the passed `options` object's properties to achieve the same purpose.

Issue #2883, opened 2018-11-21
Importing/referencing NgbAnything inside LazyModule breaks tree shaking


intellix:
### Bug description:
Trying to debug why my app is constantly growing with every new Angular version and one of the culprits (to name a few) is NgBootstrap.

If I create an App with NgbModalModule and use any of it's various things, it's 10kb in my bundle. Now if I create a LazyModule and have any sort of reference (even if indirect) to NgbModalModule, it's now 50kb in my bundle.

I've tried a couple of things:
 - AppModule using it's own modal (10kb)
 - SharedModule exporting a Modal in a LazyModule (50kb)
 - Defining lazily and opening it within LazyModule (50kb)
 - Creating a modal within App and using AppService to open it within LazyComponent (50kb).

### Link to minimally-working StackBlitz that reproduces the issue:
https://github.com/intellix/ng7-treeshake-break/blob/master/src/app/lazy/lazy.module.ts#L14 (sorry I tried to create a StackBlitz but it's pretty much unusable since the update)

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 7.0.4

ng-bootstrap: 4.0.0

Bootstrap: N/A

pkozlowski-opensource:
@intellix did you, by any chance, check what this "50kb" contains? We will try to reproduce with your repo but it would speed up things if we knew what gets pulled in.

intellix:
Was just checking source-map-explorer where it says ng-bootstrap-js.pre-build-optimizer.js (47.13KB) but looking into the JS file I'm seeing references to:
 - ngbDateAdapter
 - ngbTimeAdapter
 - ngbAccordion
 - ngbAlert
 - ngbButton
 - ngbRadio
- ngbCarousel

(I'm guessing everything)

maxokorokov:
@intellix thanks for the setting up the GiHub project.

I've been playing with the following:
- ng-bootstrap versions (2.x, 3.x, 4.x)
- various scenarios (lazy, no lazy)
- different ways to import (`NgbModule` vs modules separately)

For now I've found that indeed when using `fesm5` + lazy loading we end up with around ~60Kb in the main bundle **regardless** of which ng-bootstrap modules you import...

Switching to `esm5` fixes this issue and in your example I end up with going

from:
[...]

PR #2938, opened 2018-12-19
Use 'esm5' instead of 'fesm5' by default

maxokorokov:
This changes default library output for builds from `fesm5` to `esm5` to fix several tree-shaking issues. Changes in distributed `package.json`:

Before:
```json
  "module": "fesm5/ng-bootstrap.js",
  "es2015": "fesm2015/ng-bootstrap.js",
```

After:
```json
  "module": "esm5/ng-bootstrap.js",
  "es2015": "esm2015/ng-bootstrap.js",
```

Using `esm5` improves significantly what ends up in the final `main.js` bundle when using ng-bootstrap and importing modules individually like `NgbAlertModule` or `NgbDatepickerModule`, ex. see: #2883 and https://github.com/ng-bootstrap/ng-bootstrap/issues/2883#issuecomment-448650536

On the other head when importing `NgbModule` size increases ~10% compared to using `fesm5`.


aitboudad:
this should be fixed by https://github.com/angular/angular-cli/pull/13650, 
instead, I'd suggest to keep the fesm5 and provide a deep import similar to `material2` which doesn't require to wait until `V5`

maxokorokov:
@aitboudad could you please elaborate on your deep import suggestion? not sure what you mean.

aitboudad:
provide a `fesm` build for sub-package, for example to import the alert module users can do:

```patch
-- import {NgbAlertModule} from '@ng-bootstrap/ng-bootstrap';
++ import {NgbAlertModule} from '@ng-bootstrap/ng-bootstrap/alert';
```

`ng-packages` provide a such feature through the lib config see https://github.com/ngx-formly/ngx-formly/blob/master/src/material/ng-package.json#L5

PR #2950, opened 2018-12-27
fix(modal): support lazy loaded modules in an app-wide modal service

### LINKED ISSUE #2920: Unable to open lazy loaded modal's component from app wide singleton service
### Bug description:
I am trying to create a single entry point for handling all modal's related logic following DRY (don't repeat yourself) principle, which would help in improving modal handling through out our application, (which actually heavily relies on modals)

So instead of replicating modal handling logic inside each component (which was happening before), I have moved modal handling logic to a separate service (singleton throughout the app) provided in the main app module. This is what that service modal handling method looks like, (I have omitted some irrelevant configurations)

```
  openModal(content: any, size?: 'sm' | 'lg' | 'xl', windowClass: string = 'compact') {
    if (size === 'xl') {
      size = 'xl' as 'lg';
    }
    const ngbModalOptions: NgbModalOptions = {
      backdrop: 'static',
      keyboard: true,
      size: size
    };
    if (windowClass) {
      ngbModalOptions.windowClass = windowClass;
    }

    return this.ngbModal.open(content, ngbModalOptions);
  }
```

and this is how I am triggering it,

`this.modalService.openModal(LazyModalComponent);`

Just to make it clear, my lazy loaded modules / (shared module in my actual application) does import +export `NgbModule`, and I have also added my modal content component to `entryComponents`,

```
@NgModule({
  imports: [
    CommonModule,
    RouterModule.forChild(routes),
    NgbModalModule
  ],
  declarations: [LazyComponent, LazyModalComponent],
  entryComponents: [LazyModalComponent]
})
```

when I try to open any modal from lazy loaded modules I am getting following error,

> **Error: No component factory found for LazyModalComponent. Did you add it to @NgModule.entryComponents?**

If I call a local version of the same method from inside lazy loaded module, modal opens fine, this only happens when I try to use modals using a wrapper app wide singleton service.

P.S I did explored some closed issues regarding the subject but almost all of those are related to not importing `NgbModule` in their lazy loaded module which is not the case here.

https://stackblitz.com/edit/angular-kzh4wr

Angular: 7.1.2

ng-bootstrap: 4.0.0

Bootstrap: 4.1.3

jotwea:

Fixes #2920

With these little changes you are free to:
- still provide a ```ModalService``` for each lazy loaded module (like recommended in several previous support requests)
- provide a single ```ModalService``` in your app module and pass the child ```Injector``` of your lazy loaded module as an argument to the ```open``` method.

I use this little extension already since three months in an enterprise app and I just can say that it works perfectly. Now I found the time to contribute. It is not 100% testable in a unit test because I found no way to setup the two modules with different injectors using TestBed. If anyone has an idea how to test that, please leave a comment or feeling free to update the tests.

https://stackblitz.com/edit/angular-us1evx

### DIFF
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -26,8 +26,11 @@ export class NgbModal {
    * Also see the [`NgbModalOptions`](#/components/modal/api#NgbModalOptions) for the list of supported options.
    */
   open(content: any, options: NgbModalOptions = {}): NgbModalRef {
+    const injector = options.injector || this._injector;
+    const cfr = injector.get<ComponentFactoryResolver>(ComponentFactoryResolver as any, this._moduleCFR);
     const combinedOptions = Object.assign({}, this._config, options);
-    return this._modalStack.open(this._moduleCFR, this._injector, content, combinedOptions);
+
+    return this._modalStack.open(cfr, this._injector, content, combinedOptions);
   }
 
   /**
--- a/src/modal/modal.spec.ts
+++ b/src/modal/modal.spec.ts
@@ -1,5 +1,5 @@
 import {CommonModule} from '@angular/common';
-import {Component, Injectable, Injector, NgModule, OnDestroy, ViewChild} from '@angular/core';
+import {Component, ComponentFactoryResolver, Injectable, Injector, NgModule, OnDestroy, ViewChild} from '@angular/core';
 import {async, ComponentFixture, TestBed} from '@angular/core/testing';
 import {NgbModalConfig} from './modal-config';
 import {NgbActiveModal, NgbModal, NgbModalModule, NgbModalRef} from './modal.module';
@@ -93,7 +93,7 @@ describe('ngb-modal', () => {
   describe('default configuration', () => {
 
     beforeEach(() => {
-      TestBed.configureTestingModule({imports: [NgbModalTestModule]});
+      TestBed.configureTestingModule({imports: [NgbModalTestModule, NgbModalChildTestModule]});
       fixture = TestBed.createComponent(TestComponent);
     });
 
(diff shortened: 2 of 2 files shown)


src/modal/modal.ts
```diff
@@ -21,8 +21,11 @@ export class NgbModal {
    * NgbActiveModal class to close / dismiss modals from "inside" of a component.
    */
   open(content: any, options: NgbModalOptions = {}): NgbModalRef {
+    const injector = options.injector || this._injector;
+    const cfr = injector.get<ComponentFactoryResolver>(ComponentFactoryResolver as any, this._moduleCFR);
     const combinedOptions = Object.assign({}, this._config, options);
-    return this._modalStack.open(this._moduleCFR, this._injector, content, combinedOptions);
+
+    return this._modalStack.open(cfr, this._injector, content, combinedOptions);
```
benouat: I am not starting a proper review here, but just throwing one comment: shouldn't `this._injector` be `injector` instead as the second argument ?

jotwea: Yes and no!
With he current implementation of the ```ModalStack``` it does not make any difference. The ```ModalStack``` uses the injector from the ```options``` if given - if not it uses the one passed as 2nd argument. I decided to keep this as the ```ModalStack``` then still has access to both injectors (if they might be different). Although I do not see at the moment a use cases where the ```ModalStack``` needs both of them. My intention was not to change too much in this bugfix-PR. Of course I can change this if desired.

kamilchlebek:
This one is great!

I would go even further and add method: `openAsync(content: any, options?: NgbModalOptions): Promise<NgbModalRef>` to `NgbModal`.

To make it happen I'd make use of:
- `NgModuleFactoryLoader`,
- `lazyModules` inside `angular.json`

this will allow us to open modal with it's content being loaded lazily. Previously I had to do some hacking with named router outlets to acheieve such feature.

@pkozlowski-opensource is it worth doing it? You have mentioned here: #2920 that things will get better in the ivy version. I'm not sure if I should implement it before ivy stable release?

benouat:
> To make it happen I'd make use of:
> 
> * `NgModuleFactoryLoader`,
> * `lazyModules` inside `angular.json`

@kamilchlebek would you have any link to documentation about that ?

kamilchlebek:
I haven't found anything official about it, but here: https://netbasal.com/the-need-for-speed-lazy-load-non-routable-modules-in-angular-30c8f1c33093 you can find simple implementation of that concept.

jotwea:
@maxokorokov I see no progress anymore in reviewing that PR. It was sometime added to milestone 4.0.3 but removed few days later. Is there any doing on my side expected? Did I miss something? Please give me an advice if I have to do something - otherwise reviewing/approving this PR would be nice.

AndreaVitale:
@jotwea do you know a workaround to make it work?
I made a [stackblitz](https://stackblitz.com/edit/angular-ee2tma) to show the issue. Do you have any hint to point me in the right direction?

jotwea:
@AndreaVitale I had a look at your stackblitz. What you wanna do is to load a component in a modal directly that is declared in a lazy module. This is with the current implementation (and also with my bugfix) not possible and would be a new feature request that would be quite interesting (see also the comment of @kamilchlebek).
My bugfix is only to make it possible as an alternative to have one app-wide `modal.service` as a singleton that can be used within any lazy loaded module to open modals and no need to have such a service in every lazy loaded module. This is what you can see in the [stackblitz](https://stackblitz.com/edit/angular-us1evx) I provided.

Nevertheless your situation can also be solved easiliy by using auxiliary routes and `ng_bootstrap` containing my bugfix. You need to introduce a second `router-outlet` on the app-level with a name like i.e. `modal` or `overlay` and define a route for it to a wrapper component that is part of your lazy module. This wrapper component injects the global `modal.service` and the `injector` and uses the `openModal` function to open the desired component in a modal. You can see this in the `lazy.component` implementation of my stackblitz.

I hope that I was able to help you a little bit or at least give you a direction.

kfein:
@jotwea great feature i tried this for my use case and this is totally this what i need. @maxokorokov is there any plan to add this to the next release?

PR #2964, opened 2019-01-14
Pagination updates

liyokuna:
Hi !

My PR on pagination is a small improvement on chevrons button. Some of us use the pagination component by adding another css style and I realise on clicking next or previous button, the chevron of yours is displayed and visible.
You can see that right [here](http://ng-boosted.orange.com/#/documentation/pagination).
I added an option, that shows or not the chevrons.

BR

### DIFF
--- a/demo/src/app/components/pagination/demos/chevron/chevron-span.html
+++ b/demo/src/app/components/pagination/demos/chevron/chevron-span.html
@@ -1,0 +1,2 @@
+<p>Chevron can be hidden:</p>
+<ngb-pagination [collectionSize]="50" [(page)]="page" [spanChevron]="false"></ngb-pagination>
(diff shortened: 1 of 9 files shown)


liyokuna:
> @liyokuna thanks for the PR,
> 
> however I'd prefer to have a more generic approach, like in #2971.
> 
> You could remove chevrons by providing empty templates:
> 
> ```
> <ngb-pagination>
>   <ng-template ngbPaginationFirst></ng-template>
>   <ng-template ngbPaginationLast></ng-template>
>   <ng-template ngbPaginationPrevious></ng-template>
>   <ng-template ngbPaginationNext></ng-template>
> </ngb-pagination>
> ```
> 
> The only inconvenience would be not able to do it application-wide (no flags in `NgbPaginationConfig`). In this case it would be preferable to wrap the example above into `my-pagination` component and use it everywhere you need.
> 
> I don't believe adding flags to the API for any possible use case would make any good in the long run.

Ok thanks @maxokorokov . I try this and let you know if i have any other question.

maxokorokov:
@liyokuna as mentioned before, will close this in favour of more generic #2971. 
Thanks for contributing though!

PR #2965, opened 2019-01-14
fix(modal): add missing 'aria-modal' and 'aria-hidden' attributes

### LINKED ISSUE #2575: Can escape Modals to reach background using Screen Readers
### Bug description: 
Can escape Modals to reach background using Voice Over/JAWS
Steps,
1. Turn on Voice Over in System settings on Mac in Chrome/Firefox
2. Click Launch Demo Modal, click on input element
3. Press back arrow key until out of Modal and navigate to 'Launch Demo Modal' again. Hit enter
The Modal is launched on top of the original modal.

Note, I've only tested this in Voice Over, but other screen readers should have the same behaviour on different Operating Systems. For most greenfield desktop browsers, setting ```aria-modal="true"```, (introduced in aria 1.1), on the same div with the ```role="dialog"``` should fix the issue. For older browsers, the only solution is to set ``aria-disabled="true"`` on the root of the background content.

[Also, note this is different from the previous issue asking to add a tab trap](https://github.com/ng-bootstrap/ng-bootstrap/issues/642)

[More info on aria-modal](https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/)

[Here](https://gjvjmylgpjep.angular.stackblitz.io/)

Angular:
Angular 6.0.4

ng-bootstrap:
2.2.2

Bootstrap:
4.1.1

maxokorokov:
- add missing `aria-modal` on the modal window.
- add `aria-hidden` on all sibling DOM branches from the modal window up to the document body

Fixes #2575

### DIFF
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -21,11 +21,12 @@ import {NgbModalWindow} from './modal-window';
 
 @Injectable({providedIn: 'root'})
 export class NgbModalStack {
-  private _windowAttributes = ['ariaLabelledBy', 'backdrop', 'centered', 'keyboard', 'size', 'windowClass'];
+  private _activeWindowCmptHasChanged = new Subject();
+  private _ariaHiddenValues: Map<Element, string> = new Map();
   private _backdropAttributes = ['backdropClass'];
   private _modalRefs: NgbModalRef[] = [];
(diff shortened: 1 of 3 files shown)


maxokorokov:
@fbasso @divdavem @benouat, could anybody please check if that's what should be done for modals?

fbasso:
I did a code review. LGTM 👍

PR #2971, opened 2019-01-16
feat(pagination): allow customizing pagination links

### LINKED ISSUE #899: Custom pagination boundary and directional link text

maxokorokov:
Provides the way of customizing the markup of pagination links using child templates (in the cases where CSS is not enough):

```html
<ngb-pagination>
  <ng-template ngbPaginationFirst>First</ng-template>
  <ng-template ngbPaginationLast>Last</ng-template>
  <ng-template ngbPaginationPrevious>Prev</ng-template>
  <ng-template ngbPaginationNext>Next</ng-template>
  <ng-template ngbPaginationEllipsis>...</ng-template>
  <ng-template ngbPaginationNumber let-page>{{ page }}</ng-template>
</ngb-pagination>
```

Obviously one can customize only what's necessary, not everything.

`NgbPaginationNumber` template has the following context:
```ts
interface {
  $implicit: number;   // page displayed by current cell
  currentPage: number; // currently selected page
  disabled: boolean;   // true, if current cell is disabled
}
```

All other templates have the following context:
```ts
interface {
  currentPage: number; // currently selected page
  disabled: boolean;   // true, if current cell is disabled
}
```

P.S. Will open another PR with documentation / demo updates

Fixes #899

### DIFF
--- a/demo/src/app/components/pagination/demos/customization/pagination-customization.html
+++ b/demo/src/app/components/pagination/demos/customization/pagination-customization.html
@@ -1,0 +1,9 @@
+<p>A pagination with customized links:</p>
+<ngb-pagination [collectionSize]="70" [(page)]="page" aria-label="Custom pagination">
+  <ng-template ngbPaginationPrevious>Prev</ng-template>
+  <ng-template ngbPaginationNext>Next</ng-template>
+  <ng-template ngbPaginationNumber let-p>{{ getPageSymbol(p) }}</ng-template>
+</ngb-pagination>
+<hr>
+
+<pre>Current page: {{page}}</pre>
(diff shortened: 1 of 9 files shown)


benouat:
LGTM, merge granted 😄 !

maxokorokov:
Added commits with demo and overview update:

screen shot 2019-02-12 at 12 20 28

screen shot 2019-02-12 at 12 20 41

Issue #2924, opened 2018-12-12
Popover positioned partially off-screen and cropped


peterblazejewicz:
The `top` popover, when triggered near the page border, is being rendered in such way it gets cropped:
https://stackblitz.com/angular/rdokjdpkdoq
https://ng-bootstrap.github.io/#/components/popover/examples

![image](https://user-images.githubusercontent.com/14539/49836572-750bbc00-fda3-11e8-81cf-d5d794944a18.png)
                                                                              
The BS uses popper.js, rendering in 4.1.3:

![image](https://user-images.githubusercontent.com/14539/49836690-dcc20700-fda3-11e8-93be-236daf984b04.png)

ng-bootstrap: v4.0.0
Chrome, macOS

maxokorokov:
Ok, will reopen this as a feature request. 

In case of the `top` placement above, bootstrap/popper **SHIFT** the popover horizontally until it fits (note the arrow is not centered), we will just crop it if **ONLY** `top` is specified → we don't support this feature.

P.S. Workaround would be using `[position]="['top', 'auto']"`

Sachin4dotnet:
Any update on this ?

pelgro:
Pretty sure this is a bug. Secondary positioning should be applied to the arrow and not the whole popover window. At least that's how it works in popper.js

placement="top" results in top-center placement in positionElements. I believe the center placement should apply to the arrow and not the whole window. That way the popover will always be in the viewport as it's only moved above the reference object.

ZiluckMichael:
@maxokorokov Any update? This is an incredibly annoying shortcoming of using popovers in this library.

janwidmer:
Any Update on this bug?

dmrickey:
Yeah `'auto'` has issues on its own and this is pretty annoying..

Issue #2914, opened 2018-12-07
Tooltip placement incorrect when element is near edge


randspy:
### Bug description:
The tooltip is not positioned properly when button is located near the edge. It is also flickering(this problem does not appear in my project, just in StackBlitz) in chrome. In firefox I can see most of the time flickering pointer icon.

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-fneazi-fzgupq

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 7.1.1

ng-bootstrap: 4.0.0

Bootstrap: 4.1.3

pkozlowski-opensource:
What your stackblitz reproduces is a duplicate of #2775 - it flickers since a tooltip showing up covers the button which a browser interprets as mouse being moved out of tooltip, so it closes, but then the mouse is again over the button and the whole process repeats. 

When I change to `placement="bottom"` then the flickering goes away (again, this is a separate problem) but we can see the "wrong" positioning then: https://stackblitz.com/edit/angular-fneazi-fk9cqa?file=app/tooltip-basic.html

Is this the same issue that you see in your project? I would really like to see reproduce scenario that corresponds to your project, otherwise we might be fixing the wrong thing... In any case it seems like the issue in the stackblitz linked above is due to `float: right;`

randspy:
Hi,

Thanks to your response. I am not 100% sure as in my project the tooltip moves above the button, in the stackblitz I have sent bellow. I was not able to reproduce the exact situation. But when I change in my project placement to bottom then it is displayed as in the stackblitz you send. For the issue itself when I replace the ```float: right``` with ```margin-left: 130px``` the problem still occurs (https://stackblitz.com/edit/angular-fneazi-fh7q5e?file=app/tooltip-basic.scss)

BR

maxokorokov:
This should be fixed in `4.0.3`, but please see this → https://github.com/ng-bootstrap/ng-bootstrap/pull/2972#issuecomment-460641793. Feel free to reopen if you're still having issues.

Issue #2346, opened 2018-04-27
Popover positioning and sizing incorrect when inside of navbar with dynamic HTML content


velvok:
### Bug description:
1. Position of a popover incorrectly calculated. 
2. Position and size of a popover changes on click on the window.

Popover is placed inside of navbar and has a dynamic HTML content.

The issue appears to be related to popover's width changes due to dynamic nature of the content. Looks that ngbPopover calculates positioning before the internal content is fully displayed. And of course IE11/Edge acts differently.

#### IE11/Edge:
![image](https://user-images.githubusercontent.com/11169743/39380180-a5a54a1a-4a2b-11e8-9a15-329bcef91934.png)
#### Chrome/Firefox:
![image](https://user-images.githubusercontent.com/11169743/39380242-d9ac46f6-4a2b-11e8-8543-91e79732b2d9.png)
##### Behavior: After several clicks anywhere in the window the popover moves toward the right place. But still arrow is not completely centered as mentioned here: [#2089](https://github.com/ng-bootstrap/ng-bootstrap/issues/2089)
![image](https://user-images.githubusercontent.com/11169743/39380652-52380a32-4a2d-11e8-93a1-472029553122.png)

I have noticed that when the width is set explicitly such as
`<div class="row no-gutters" style="width: 194px;">`
then it looks more as expected but this doesn't seam to be a good solution when content is dynamic.

3. Another thing I have noticed that might be related to the same issue is position changes on resizing window:
![image](https://user-images.githubusercontent.com/11169743/39380878-1f84220a-4a2e-11e8-96ca-5c98201e6f60.png)

### Link to minimally-working plunker that reproduces the issue:
https://stackblitz.com/edit/angular-ipuuur

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 5.2.10

ng-bootstrap: 1.1.2

Bootstrap: 4.1.0

Arackow:
I have the same issue. Mine is just a simple line of text that has different sizes... for my specific use case I can probably just get away with using a fixed width but the solution is probably something to the effect of creating but not rendering the content, measure it, move the popover accordingly, then render...

yaitskov:
Explicit style value bothers me also. 
It is not even needed. If I disable left property layout fixes itself.

maxokorokov:
[...]

Issue #2069, opened 2018-01-03
Multiline tooltip at right edge gets resized on click


tzunghaor:
### Bug description:

I have a button with a tooltip so far right, that the tooltip reaches the right edge of the window / position:relative container.
Clicking the button calls a method (irrelevant what it does, in the plunkr example it does nothing), and then unexpectedly the tooltip becomes narrower.
(I experienced something similar with popovers too.)

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/dUKdE9lmtJ8fBbINagz4?p=preview

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 5.0.2

ng-bootstrap: 1.0.0-beta.8

Bootstrap: 4.0.0-beta

milasevicius:
Setting `container="body"` seems to solve this issue

tzunghaor:
@milasevicius - `container="body"` seems to solve it in the plunker for the centered button, but not for the button on the right.

maxokorokov:
This should be fixed in `4.0.3`, but please see this → https://github.com/ng-bootstrap/ng-bootstrap/pull/2972#issuecomment-460647837. Feel free to reopen if you're still having issues.

PR #2972, opened 2019-01-16
fix(positioning): use CSS translate for positioning

### LINKED ISSUE #2470: Tooltip placement incorrect when element is near edge and when using ng-template
### Bug description:

So I had reported #2465 earlier, but it turns out my issue was something else entirely. If you are using an ng-template to display the content of a tooltip, and the element that displays the tooltip is near the edge of the screen, the tooltip will be incorrectly placed. You can see in the below stackblitz that there are three buttons. All three pull from the same variable for it's content. But the last two, display the content in a ng-template. The second button works fine, but the third button which is at the edge of the screen you can see is positioned incorrectly, there is plenty of room for it all around, but its offset from the element incorrectly.

https://stackblitz.com/edit/angular-fneazi-ljhkng?file=app%2Ftooltip-basic.html

Angular: 6.0.0

ng-bootstrap: 2.1.1

Bootstrap: 4.1.0

fbasso:
fix #2470

### DIFF
--- a/e2e-app/src/app/tooltip/position/tooltip-position.component.html
+++ b/e2e-app/src/app/tooltip/position/tooltip-position.component.html
@@ -1,0 +1,31 @@
+<h3>Tooltip positioning tests</h3>
+
+<form id="default">
+  <div class="m-3">
+    <button class="btn btn-outline-secondary mx-1" id="flex-left" (click)="setPosition('start')">Left</button>
+    <button class="btn btn-outline-secondary mx-1" id="flex-center" (click)="setPosition('center')">Center</button>
+    <button class="btn btn-outline-secondary mx-1" id="flex-right" (click)="setPosition('end')">Right</button>
+  </div>
+  <div class="d-flex {{flexPosition}}">
+    <div class="d-flex flex-column mx-3">
+      <button id="btn-normal" type="button" container="body" triggers="click" class="btn btn-outline-secondary" placement="auto" [ngbTooltip]="content">
+        Normal tooltip
+      </button>
+
+      <button id="btn-innerHtml" type="button" container="body" triggers="click" class="btn btn-outline-secondary" [placement]="['top', 'top-left', 'top-right']" [ngbTooltip]="tooltipHtml">
+          InnerHtml
+      </button>
+
+      <button id="btn-body-off" type="button" container="body" triggers="click" class="btn btn-outline-secondary" placement="auto" [ngbTooltip]="content">
+        body off
+      </button>
+
+    </div>
+    <ng-template #tooltipHtml><div [innerHtml]="content"></div></ng-template>
+  </div>
+  <div class="m-3">
+    <button id="btn-fixed" type="button" triggers="click" class="btn btn-outline-secondary position-fixed" [placement]="['top', 'top-left', 'top-right']" [ngbTooltip]="tooltipHtml" [ngStyle]="{'left': fixedPositionLeft, 'right': fixedPositionRight}">
+      Fixed
+    </button>
+  </div>
+</form>
--- a/e2e-app/src/app/tooltip/position/tooltip-position.e2e-spec.ts
+++ b/e2e-app/src/app/tooltip/position/tooltip-position.e2e-spec.ts
@@ -1,0 +1,93 @@
+import {openUrl} from '../../tools.po';
+import {TooltipPositionPage as TooltipPositionPage} from './tooltip-position.po';
+
+const roundLocation = function(location) {
+  location.x = Math.round(location.x);
+  location.y = Math.round(location.y);
+
+  return location;
+};
+
+describe('Tooltip Position', () => {
+  let page: TooltipPositionPage;
+
+  const expectTooltipsPosition = async(type: string, expectedPlacement: string) => {
+
+
+    const btn = page.getTooltipButton(type);
+    await btn.click();
+    const btnLocation = roundLocation(await btn.getLocation());
+    const btnSize = await btn.getSize();
+    const tooltip = await page.getTooltip();
+    const tooltipLocation = roundLocation(await tooltip.getLocation());
+    const tooltipSize = await tooltip.getSize();
+
+    let [primary, secondary] = expectedPlacement.split('-');
+    const classname = await tooltip.getAttribute('class');
+    expect(classname).toContain(`bs-tooltip-${primary}`, 'Missing primary class');
+    if (secondary) {
+      expect(classname).toContain(`bs-tooltip-${primary}-${secondary}`, 'Missing secondary class');
+    }
+
+    let yDiff: number, xDiff: number;
+
+    if (primary === 'top') {
+      yDiff = (tooltipLocation.y + tooltipSize.height) - btnLocation.y;
+      if (secondary === 'left') {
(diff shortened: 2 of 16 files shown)


maxokorokov:
It breaks this case for me:

screen shot 2019-01-16 at 14 14 37

fbasso:
ok, it seems more complicated than expected. I'll try a complete refactor of the positioning utility.

fbasso:
This is a new version . It includes a refactor about how the positioning is calculated. 

The positioning was pre-calculated before applying the positions. In some edge case, the width or height of the element could change afterwards, leading to issues.

Now the positioning is really done for each wanted position (i.e. 'top', 'right', ...) until the element is inside the viewport or the last asked position.

src/tooltip/tooltip.spec.ts
```diff
@@ -28,15 +28,14 @@ describe('ngb-tooltip-window', () => {
     fixture.detectChanges();
 
     expect(fixture.nativeElement).toHaveCssClass('tooltip');
-    expect(fixture.nativeElement).toHaveCssClass('bs-tooltip-top');
+    expect(fixture.nativeElement).not.toHaveCssClass('bs-tooltip-top');
     expect(fixture.nativeElement.getAttribute('role')).toBe('tooltip');
   });
 
-  it('should position tooltips as requested', () => {
+  it(`shouldn't add 'bs-tooltip-left' class with binding`, () => {
```
maxokorokov: Why should it even try applying left if you removed `fixture.componentInstance.placement = 'left';` ?

this isn't testing anything now, you should put it back :)

See what you did in `popover.spec.ts`

src/popover/popover.ts
```diff
@@ -102,12 +84,12 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
...
-   *    "left", "left-top", "left-bottom", "right", "right-top", "right-bottom"
-   * and array of above values.
-   */
-  @Input() placement: PlacementArray;
+    * Placement of a tooltip accepts:
+    *  - Array or space separated string of following values:
+    *    "top", "top-left", "top-right", "bottom", "bottom-left", "bottom-right",
+    *    "left", "left-top", "left-bottom", "right", "right-top", "right-bottom",
+    */
+  @Input() placement: string | PlacementArray;
```
maxokorokov: Let's open a separate feature PR for this later for ALL widgets supporting positioning

src/tooltip/tooltip.ts
```diff
@@ -83,11 +62,11 @@ export class NgbTooltip implements OnInit, OnDestroy {
...
   /**
     * Placement of a tooltip accepts:
+    *  - Array or space separated string of following values:
     *    "top", "top-left", "top-right", "bottom", "bottom-left", "bottom-right",
-    *    "left", "left-top", "left-bottom", "right", "right-top", "right-bottom"
-    * and array of above values.
+    *    "left", "left-top", "left-bottom", "right", "right-top", "right-bottom",
     */
-  @Input() placement: PlacementArray;
+  @Input() placement: string | PlacementArray;
```
maxokorokov: Let's open a separate feature PR for this later for ALL widgets supporting positioning

src/tooltip/tooltip.ts
```diff
@@ -178,18 +156,14 @@ export class NgbTooltip implements OnInit, OnDestroy {
         this._document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
       }
 
-      this._windowRef.instance.placement = Array.isArray(this.placement) ? this.placement[0] : this.placement;
+      if (!Array.isArray(this.placement)) {
```
maxokorokov: Let's open a separate feature PR for this later for ALL widgets supporting positioning

src/tooltip/tooltip.spec.ts
```diff
@@ -292,6 +291,34 @@ describe('ngb-tooltip', () => {
         expect(windowEl.textContent.trim()).toBe('Great tip!');
       });
 
+      it('should accept placement in space seperated (second value should be applied)', () => {
```
maxokorokov: Let's open a separate feature PR for this later for ALL widgets supporting positioning

src/popover/popover.ts
```diff
@@ -61,19 +56,6 @@ export class NgbPopoverWindow {
   constructor(private _element: ElementRef<HTMLElement>, private _renderer: Renderer2) {}
```
maxokorokov: you should remove the constructor as well, not used anymore

e2e-app/src/app/tooltip/position/tooltip-position.component.html
```diff
@@ -0,0 +1,31 @@
+<h3>Tooltip autoclose tests</h3>
```
maxokorokov: Tooltip positioning tests

e2e-app/src/app/tooltip/position/tooltip-position.component.html
```diff
@@ -0,0 +1,31 @@
...
+
+<form id="default">
+  <div class="m-3">
+    <button class="btn btn-outline-secondary mx-1" id="flex-left" (click)="setPosition('start')">Left</button>
+    <button class="btn btn-outline-secondary mx-1" id="flex-center" (click)="setPosition('center')">Center</button>
+    <button class="btn btn-outline-secondary mx-1" id="flex-right" (click)="setPosition('end')">Right</button>
+  </div>
+  <div class="d-flex" [ngClass]="flexPosition">
```
maxokorokov: nitpick: could simply be 
```html
<div class="d-flex {{flexPosition}}">
```

e2e-app/src/app/tooltip/position/tooltip-position.e2e-spec.ts
```diff
@@ -0,0 +1,94 @@
+import {Key, browser} from 'protractor';
```
maxokorokov: please cleanup unused imports

e2e-app/src/app/tooltip/position/tooltip-position.po.ts
```diff
@@ -0,0 +1,13 @@
+import {$} from 'protractor';
+
+export class TooltipPositionPage {
+  async clickOutside() { await $('#outside-button').click(); }
```
maxokorokov: please cleanup unused methods

fbasso:
I think the code is almost ready now. I'll try to find out what are the issues this PR will fix.

fbasso:
fix #2914, but placement alternatives need to be specified (otherwise, the tooltip is not completely inside the viewport). i.e. : `[placement]="['top', 'auto']"`

fbasso:
close #2069 
The same, placements alternatives need to be specified to get it into the viewport. 

The main idea is if only one placement is specified, it will stick on it even if it's partly outside of the view, but it will never overlap the main element (from which the tooltip/popover is triggered).

fbasso:
close #2924

fbasso:
Partially fix #2346. The popover/dropdown content is no longer resized, but is still arrow positioning issue. For example, if the placement ends to be on bottom-right, the arrow will be 2rem from the right, like this:

image

I'm not sure ng-bootstrap will be able to manage this case, styling from the application will be necessary (for example, override .bs-popover-bottom-right .arrow)

PR #2981, opened 2019-01-23
feat(accordion): allow any content in the accordion card header

### LINKED ISSUE #717: Accordion Feature Request - Title Content that does not cause expand/collapse

maxokorokov:
A more genetic alternative to #2845 to add any content to accordion header.

The idea is to introduce two directives `ng-template[ngbPanelHeader]` and `button[ngbPanelToggle]` to be used as follows:

```html
<ngb-accordion>
  <ngb-panel>
    <ng-template ngbPanelHeader let-opened="opened">
      <!-- anything here -->
      <button ngbPanelToggle>Panel Title</button>
      <!-- anything here -->
    </ng-template>
     <ng-template ngbPanelContent> ... </ng-template>
  </ngb-panel>
</ngb-accordion>
```

Demo shows an example of adding some controls to headers:

screen shot 2019-02-14 at 16 02 56

This will compliment and take precedence over two existing ways of providing a title:

```html
<!-- 1. title as an attribute -->
<ngb-accordion>
  <ngb-panel title="Panel Title">
     <ng-template ngbPanelContent> ... </ng-template>
  </ngb-panel>
</ngb-accordion>

<!-- 2. title as template for the button content -->
<ngb-accordion>
  <ngb-panel>
     <ng-template ngbPanelTitle>Panel Title</ng-template>
     <ng-template ngbPanelContent> ... </ng-template>
  </ngb-panel>
</ngb-accordion>
```

I would even deprecate `ngbPanelTitle`

Fixes #717

### DIFF
--- a/demo/src/app/components/accordion/accordion.module.ts
+++ b/demo/src/app/components/accordion/accordion.module.ts
@@ -10,8 +10,11 @@ import { NgbdAccordionConfig } from './demos/config/accordion-config';
 import { NgbdAccordionPreventchange } from './demos/preventchange/accordion-preventchange';
 import { NgbdAccordionStatic } from './demos/static/accordion-static';
 import { NgbdAccordionToggle } from './demos/toggle/accordion-toggle';
+import { NgbdAccordionHeader } from './demos/header/accordion-header';
 
-const DEMO_DIRECTIVES = [NgbdAccordionBasic, NgbdAccordionPreventchange, NgbdAccordionStatic, NgbdAccordionToggle, NgbdAccordionConfig];
+const DEMO_DIRECTIVES = [
+  NgbdAccordionBasic, NgbdAccordionPreventchange, NgbdAccordionStatic, NgbdAccordionToggle, NgbdAccordionConfig, NgbdAccordionHeader
+];
 
 const DEMOS = {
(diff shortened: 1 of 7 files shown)


Devqon:
Is there a reason why `ngbPanelToggle` can only be attached on a `button` element?
In my use case I want the whole header to be clickable to open the accordion.

stguitar:
@Devqon I was looking for this too, but decided to use CSS to control the styling.

`<button class='btn btn-link'>Something</button>` should get you started.

Also, I doubt asking questions on a merged PR is going to get you much attention but you got lucky with me hah

maxokorokov:
@Devqon @stguitar please see #3104

PR #3004, opened 2019-02-07
fix(autoclose): fix popup automatic closing on touch on iOS

### LINKED ISSUE #2995: Dropdown does not autoClose on outside click (iOS only)
### Bug description:

Dropdown does not autoClose on outside click (iOS only). Tested with an iPad Air 2, iOS 12.1.2.

With your iOS device, visit
https://ng-bootstrap.github.io/#/components/dropdown/examples
Open the first Dropdown example and click outside of it. The Dropdown will not close.

See https://ng-bootstrap.github.io

maxokorokov:
Tested on Android, Windows 10 with touch and iOS.

Autoclose was broken only on iOS, other browsers work fine with `mousedown/mouseup` while `iOS` needs to have `touchstart/touchend`

Fixes #2995

cc @divdavem

### DIFF
--- a/src/util/autoclose.ts
+++ b/src/util/autoclose.ts
@@ -1,11 +1,15 @@
-import {Injectable, NgZone, Inject} from '@angular/core';
+import {Inject, Injectable, NgZone} from '@angular/core';
 import {fromEvent, Observable, race} from 'rxjs';
 import {DOCUMENT} from '@angular/common';
-import {takeUntil, filter, delay, withLatestFrom, map} from 'rxjs/operators';
+import {delay, filter, map, takeUntil, withLatestFrom} from 'rxjs/operators';
 import {Key} from './key';
 
 const isHTMLElementContainedIn = (element: HTMLElement, array?: HTMLElement[]) =>
     array ? array.some(item => item.contains(element)) : false;
+
+// we'll have to use 'touch' events instead of 'mouse' events on iOS and add a more significant delay
+// to avoid re-opening when handling (click) on a toggling element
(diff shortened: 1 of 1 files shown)


src/util/autoclose.ts
```diff
@@ -39,18 +45,18 @@ export class AutoClose {
...
+                                .pipe(map(shouldCloseOnClick), takeUntil(closed$));
 
-        const outsideClicks$ = fromEvent<MouseEvent>(this._document, 'mouseup')
-                                   .pipe(
-                                       withLatestFrom(mouseDowns$), filter(([_, shouldClose]) => shouldClose), delay(0),
-                                       takeUntil(closed$));
+        const closeableClicks$ = fromEvent<MouseEvent>(this._document, iOS ? 'touchend' : 'mouseup')
+                                     .pipe(
+                                         withLatestFrom(mouseDowns$), filter(([_, shouldClose]) => shouldClose),
+                                         delay(iOS ? 16 : 0), takeUntil(closed$));
```
maxokorokov: If anybody has a better solution for this, please share it :)

Problem is when there is a `click` handler on a triggering element (say, button): 
- to open: `click` → `open()` 
- to close: `touchdown` → `touchup` → `click (is still generated that leads to)` → `open() (popup reopening)`

by delaying the `touchup`, `click` and `open()` are processed while the popup is still open

maxokorokov:
As discussed, I'll merge the fix as-is and we'll investigate better ways for handling focus and closing things

cc @fbasso @benouat

Issue #2623, opened 2018-08-17
v3: Access Line class from "@ng-bootstrap/ng-bootstrap/util/accessibility/live"


dmostert:
For a number of reasons, I've subclassed ```NgbTypeahead```. And to get the the constructor chain to work, I need to include the ```Live``` (```@ng-bootstrap/ng-bootstrap/util/accessibility/live```) class as a parameter.

With v2.x this worked fine, but I can't get it to work with the new v3 packaging format.

Intellisense still knows about Live, but I can't compile:

```typescript
import {
  OnInit, Directive, ElementRef, ViewContainerRef, Renderer2, Injector, ComponentFactoryResolver, NgZone, forwardRef, Input
} from "@angular/core";
import { NgbTypeahead, NgbTypeaheadConfig, NgbTypeaheadWindow  } from "@ng-bootstrap/ng-bootstrap";
import { NG_VALUE_ACCESSOR } from "@angular/forms";
import { Observable } from "rxjs";

//THIS IS THIS PROBLEM
import { Live } from "@ng-bootstrap/ng-bootstrap/util/accessibility/live";
//THIS IS THE PROBLEM

export class MyTypeahead extends NgbTypeahead {
  constructor(private _elementRef2: ElementRef, _viewContainerRef: ViewContainerRef, _renderer: Renderer2,
    _injector: Injector, componentFactoryResolver: ComponentFactoryResolver, config: NgbTypeaheadConfig,
    ngZone: NgZone, live: Live) {
    super(_elementRef2, _viewContainerRef, _renderer, _injector, componentFactoryResolver, config, ngZone, live);
    }

  ...
}
```

Any ideas on how to access Live?

Angular:
6.1
ng-bootstrap:
3.0.0
Bootstrap:
4.1

pkozlowski-opensource:
@dmostert generally speaking we are not super-keen on people sub-classing widgets as it opens up the whole new set of challenges in terms of maintaining public API surface. Would you mind sharing your use-case and the need to sub-class?

@maxokorokov in this particular case we should probably export `Live` from `@ng-bootstrap/ng-bootstrap`, WDYT?

dmostert:
@pkozlowski-opensource 
- I've added logic to ensure than an item get selected from the drop down
- If no selection, clear the underlying model on blur
- I've styled the popup window to have a max height
- I try to find and scroll to the current value on popup
- Keep "input" when entering a value that is not in the dropdown (until blur)


[...]

PR #3011, opened 2019-02-11
refactor(autoclose): transform service into a function to avoid DI

maxokorokov:
The motivation for this PR is to rollback breaking changes for those who extend autocloseable components (`NgbDatepickerInput`, `NgbTypeahead`, `NgbTooltip`, `NgbDropdown`, `NgbPopover`):

```ts
  @Component()
  class NgbPopover {
    constructor(..., private _autoClose: AutoClose);
  }
```

It was introduced in `4.0.1` and will break in `NgbPopover` children classes, because `AutoClose` is a private service.

---

**However, we don't want people to subclass our components** and https://github.com/ng-bootstrap/ng-bootstrap/issues/2623#issuecomment-413793539 is still true, but adding an injectable`AutoClose` change should have not been a part of a minor release IMO

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -1,10 +1,12 @@
 import {
+  ChangeDetectorRef,
   ComponentFactoryResolver,
   ComponentRef,
   Directive,
   ElementRef,
   EventEmitter,
   forwardRef,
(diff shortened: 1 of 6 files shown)


pkozlowski-opensource:
LGTM

PR #3012, opened 2019-02-11
docs(typeahead): overview

ymeine:


maxokorokov:
Quick couple of notes, before looking into the actual content:

1. We should drop line hightlighting I guess, because 1.5 seconds to open a page on a fast desktop is not acceptable. It freezes noticeably:

screen shot 2019-02-12 at 13 58 48

2. I have this in the console, when I open overview:

screen shot 2019-02-12 at 14 00 35

maxokorokov:
Another idea is to think of an overview, as a marketing page targeting developers

benouat:
Just restarted the build

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
...
+<h5>Dismiss the popup</h5>
+
+<p>
+  Dismissing the popup consists in closing the popup without selecting an option.
+</p>
+
+<table class="table mt-4">
+  <tbody>
+    <tr>
+      <td><kbd>Escape</kbd></td>
```
maxokorokov: I'd leave only `Escape` case here. Remove others (outside, inside) and add a link to the `[autoClose]` API saying something like "for more options see the autoclose API"

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
...
+  The classic usage of the Typeahead is to use the input to edit the search term, and trigger a search process which will spawn a new options list.
+</p>
+
+<p>
+  Apart from creating an Observable based on the <code>input</code> event of the input, the Typeahead does nothing else specific, so to edit the search term just use it as any other input.
+</p>
+
+
+
+<h5>Navigate through options in the popup</h5>
```
maxokorokov: Also, this would be more than enough to list ONLY keyboard shourcuts as a simple table like here → https://ng-bootstrap.github.io/#/components/datepicker/overview#keyboard-shortcuts

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
```diff
@@ -0,0 +1,41 @@
+<p>
+  As said above, the options list can contain any kind of item, and this is handy since selecting an option will commit the actual item to the data model, allowing to store complex objects.
+</p>
+
+<p>
+  But the requirement is to be able to display those options, be it in the list of the popup or inside the input field.
```
maxokorokov: Two paragraphs too complex, maybe → 

Typeahead might use complex objects as options, so it would need to transform them to HTML to display in the results popup. 

If your typeahead uses strings there is nothing to do, otherwise you need to either provide `[resultFormatter]` function or a `[resultTemplate]`

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
```diff
@@ -0,0 +1,41 @@
...
+</p>
+
+
+
+<h5>Simple representation with a string</h5>
```
maxokorokov: Using result formatter ?

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
```diff
@@ -0,0 +1,41 @@
...
+
+<ngbd-code [lang]="snippets.resultFormatter.template.language" [code]="snippets.resultFormatter.template.code" [highlightedLines]="snippets.resultFormatter.template.highlightedLines"></ngbd-code>
+
+<p>And define it in the component: </p>
+
+<ngbd-code [lang]="snippets.resultFormatter.component.language" [code]="snippets.resultFormatter.component.code" [highlightedLines]="snippets.resultFormatter.component.highlightedLines"></ngbd-code>
+
+
+
+<h5>Complex representation with a template</h5>
```
maxokorokov: Providing a template ?

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
```diff
@@ -0,0 +1,41 @@
...
+
+<p>
+  Sometimes displaying a simple string is not enough. In that case, you can use a template by using the property <code>resultFormatter</code>, which expects a template reference.
```
maxokorokov: resultFormatter → resultTemplate ?

demo/src/app/components/typeahead/overview/sections/introduction/template.html
```diff
@@ -0,0 +1,31 @@
...
+  let-color="result"
+  let-term="term"
+>
+  <div class="middle">
+    <div class="color-chip" [ngStyle]="{'background-color': '#' + color.hexCode}"></div>
+    <ngb-highlight [result]="color.name" [term]="term"></ngb-highlight><span> ({{color.hexCode}})</span>
+  </div>
+</ng-template>
+
+<p>Typeahead example: </p>
```
maxokorokov: Pick a color: ?

demo/src/app/components/typeahead/overview/sections/introduction/template.html
```diff
@@ -0,0 +1,31 @@
...
+
+<input
```
maxokorokov: Maybe limit width and put in one line, like so:

```
Pick a color: [                    ]
```

Now it takes the whole width which is strange

demo/src/app/components/typeahead/overview/sections/customize-option-in-input/template.html
```diff
@@ -0,0 +1,19 @@
+<p>
+  It is also possible to customize the display of the option inside the input field, using the property <code>inputFormatter</code> which expects a function taking the option item as input and returning a string as output.
+</p>
+
+<p>
+  There are two limitations though:
```
maxokorokov: Why all this? Just say again (as for popup formatting):

Typeahead might use complex objects as options, so it would need to transform them to strings to display in the input when user selects an item (say with the mouse).

If your typeahead uses strings there is nothing to do, otherwise you would need to provide the `[inputFormatter]` function.

demo/src/app/components/typeahead/overview/sections/introduction/template.html
```diff
@@ -0,0 +1,31 @@
...
+  [inputFormatter]="inputFormatter"
+/>
+
+<p>The Typeahead is basically a component which:</p>
+<ul>
+  <li>takes a user input as a string: try typing <code>bl</code></li>
+  <li>pipes this input to an Observable chain to get a list of options</li>
+  <li>handles the display of the popup containing the options: you should see the colors <code>black</code> and <code>blue</code></li>
+  <li>handles the navigation through the options: try using the keyboard arrows up and down</li>
+  <li>handles the selection of one option: press <kbd>enter</kbd> to select the highlighted option, or click directly on it, and see it appear in the input field</li>
```
maxokorokov: <kbd>Enter</kbd>  (capital) ?

angular.json
```diff
@@ -75,6 +75,8 @@
             "styles": [
               "node_modules/bootstrap/dist/css/bootstrap.css",
               "node_modules/prismjs/themes/prism.css",
+              "node_modules/prismjs/plugins/line-highlight/prism-line-highlight.css",
```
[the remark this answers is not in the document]
ymeine: I will integrate a replacement for the line highlighting, which I think doesn't have these performance issues. I developed it separately for now.

demo/src/app/components/typeahead/overview/sections/introduction/template.html
```diff
@@ -0,0 +1,31 @@
...
+  type="text"
+  class="form-control"
+  [(ngModel)]="model"
+  [ngbTypeahead]="initializeTypeahead"
+
+  [resultTemplate]="displayResult"
+  [inputFormatter]="inputFormatter"
+/>
+
+<p>The Typeahead is basically a component which:</p>
```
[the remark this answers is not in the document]
ymeine: Agreed

angular.json
```diff
@@ -75,6 +75,8 @@
...
+              "node_modules/prismjs/plugins/line-numbers/prism-line-numbers.css",
```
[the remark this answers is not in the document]
ymeine: It was used only to make the line highlighter work properly, otherwise it was broken. A further reason to remove/replace it.

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,49 @@
+<h5>How to set up the Typeahead</h5>
```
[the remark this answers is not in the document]
ymeine: Agreed, I'll rather start the first sentence with "To set up a Typeahead instance, you need...."

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,49 @@
...
+
+<p>
+  You need to pass an initialization function to the Typeahead. This function must return an Observable spawning lists of options: this is how you control the list of displayed options.
+</p>
+
+<p>
+  Pass it that way on the template side:
+</p>
+
+<ngbd-code [lang]="snippets.observableSetup.template.language" [code]="snippets.observableSetup.template.code" [highlightedLines]="snippets.observableSetup.template.highlightedLines"></ngbd-code>
```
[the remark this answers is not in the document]
ymeine: The goal was to highlight the exact part relevant there, and I really want to keep line highlighting (with the replacement I mentioned in another comment). But it's _line_ highlighting only... I could improve it to make it a _range_ highlighting but that would look weird. For a simple example like that, a one liner could be fine too.

demo/src/app/components/typeahead/overview/sections/basic-usage/snippets.ts
```diff
@@ -0,0 +1,55 @@
...
+          [ngbTypeahead]="initializeTypeahead"
+        />
+      `,
+    }),
+    component: Snippet({
+      language: 'typescript',
+      highlightedLines: '2-3',
+      code: `
+        export class MyComponent {
+          initializeTypeahead = (input$: Observable<string>): Observable<string[]> =>
```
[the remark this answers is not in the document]
ymeine: I use the name `input$` because it comes from the `input` event of the Typeahead — to differentiate from `click` or `focus`. That can also be understood as "the content of the __input__ field". So I'd rather keep this name.

demo/src/app/components/typeahead/overview/sections/basic-usage/snippets.ts
```diff
@@ -0,0 +1,55 @@
...
+  observable: {
+    mapping: Snippet({
+      language: 'typescript',
+      highlightedLines: '1, 5-8',
+      code: `
+        import {map} from 'rxjs/operators';
+        // ...
+        initializeTypeahead = (input$: Observable<string>): Observable<string[]> =>
+          input$.pipe(
+            map(searchTerm => searchTerm.length === 0
```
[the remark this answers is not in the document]
ymeine: Also, `text` is quite generic, and in the documentation around I'm referring to this piece of text being edited through the input field as the _search term_. `searchTerm` has more semantic than the generic `text` name (which to me is like `data`, `object`, `value`... they are fine and sometimes really adapted, but very generic and sometimes overused)

demo/src/app/components/typeahead/overview/sections/observable-contract/template.html
```diff
@@ -0,0 +1,27 @@
...
+  The Typeahead is designed around a single Observable, to which it subscribes in order to get an updated list of options and update its popup accordingly.
+</p>
+
+<p>
+  The choice of the Observable is due to the nature of the data processing: often updated and potentially asynchronous. Observables are designed mainly for this use case, and bring a common API to handle data flow no matter if the processing is asynchronous or synchronous.
+</p>
+
+
+
+<h5>What is the Observable contract?</h5>
```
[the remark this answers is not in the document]
ymeine: It's a contract like an API, it means that Typehead expects developers to do something specific to use it, so developers kind of follow the contract established by Typeahead, which in this case is the use of an observable in a specific way (mapping text to lists, etc.)

I can change a bit the form of the section (the content will remain the same though)

demo/src/app/components/typeahead/overview/sections/observable-contract/template.html
```diff
@@ -0,0 +1,27 @@
...
+<p>
+  The elements of the list are formatted to be displayed in the popup, as well as in the input when an option is selected.
+</p>
+
+<p>
+  Selecting an option commits the original item to the data model of the Typeahead. If your option items can be converted to proper strings out of the box, you don't <em>need</em> to customize the formatting (but you still can). Otherwise, you will have to (more on that in below section).
+</p>
+
+<p>
+  The Typeahead updates the display of its popup accordingly to the length of the options list it receives: if it's empty it won't display the popup, and if it's not empty it will display the popup with that list of options.
```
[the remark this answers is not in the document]
ymeine: I agree I need to simplify this particular sentence.

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
...
+
+
+<h5>Select an option</h5>
+
+<p>Selecting means committing to the data model. It will display the currently selected option in the input.</p>
+
+<table class="table mt-4">
+  <tbody>
+    <tr>
+      <td><kbd>Enter</kbd></td>
```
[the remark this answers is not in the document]
ymeine: Oh, forgot that... Noted!

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
+<h5>Edit the search term</h5>
```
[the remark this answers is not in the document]
ymeine: I see that it might go too much into the obvious (for instance, I could go deeper and say that to edit they have to use anything that throws keys events to the user agent, bla bla bla that would be too much). Having an `input` element makes it quite obvious.

However, I could maybe remind and say more precisely that "with a classic usage" (I don't know how to turn that yet), the search is done anytime the input value changes, i.e. vs not only when blurring or using a special key.

That's part of the user interaction with Typeahead, so I would keep that.

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
...
+
+
+
+<h5>Navigate through options in the popup</h5>
+
+<table class="table mt-4">
+  <tbody>
+    <tr>
+      <td><kbd>ArrowUp</kbd></td>
+      <td>Activate previous (above) option</td>
```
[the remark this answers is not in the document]
ymeine: See comment below

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
...
+    </tr>
+    <tr>
+      <td><kbd>ArrowDown</kbd></td>
+      <td>Activate next (below) option</td>
```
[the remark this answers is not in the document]
ymeine: See comment below

angular.json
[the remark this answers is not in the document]
maxokorokov: @ymeine could you do it separately after this one, please? Let's focus on overview content here

maxokorokov:
@ymeine, also forgot to attach this one, sorry:

Screen Shot 2019-03-28 at 11 13 35

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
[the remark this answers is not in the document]
ymeine: Ha so here we go... I struggled finding proper naming, and I am still not satisfied.

Initially, I wanted to say _select_, but this is what we use to say that we choose the option / commit it to the model.

Then _highlight_, but I found it confusing with the `ngb-highlight` component, whose role is to highlight part of the option. Also, "highlight" is a word related to visual appearance only, so I find it's not obvious it is related to where the "cursor" (by cursor I mean a current position in the list). And if we ever had a feature to highlight the currently selected/committed option, this would even be more confusing, so I don't like this term.

So to sum up: 

- _activate_: not very clear, we both agree
- _select_: good but already taken to say we choose an option
- _highlight_: in the current case works because indeed putting the (virtual) cursor on an option highlights it, but to me it's like using a side effect to represent the thing (e.g. if we said "to outline" instead of "to focus", while the outline is a result of a focus)

Still thinking...

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,82 @@
...
+
+<h5>Select an option</h5>
+
+<p>Selecting means committing to the data model. It will display the currently selected option in the input.</p>
+
+<table class="table mt-4">
+  <tbody>
+    <tr>
+      <td><kbd>Enter</kbd></td>
+      <td>Select active option</td>
```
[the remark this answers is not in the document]
ymeine: See comment above

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
ymeine: I disagree with that. If we are talking about user interactions, a lot of people are using the mouse, and Typeahead itself implements things specifically for the mouse. So removing that would be providing half the information.

Regarding the "1 table" vs "multiple tables", I just wanted to separate by intent. As a user, I will first think about doing something specific (navigate, select, close, etc.) and then will want to know how I can do it (keyboard, mouse, API, etc.). So the sectioning totally makes sense. And it is not breaking the readability. I could add some specific styling to visually separate the sections a bit more though.

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
ymeine: I can simplify the paragraphs. But in your proposition, there are shortcuts which make the statement not exact: 

- transform to HTML: it's either transformed to a string or using a template, both being different from going through an HTML representation
- "there is nothing to do": a but confusing, if they want to customize, they still have to do something.

But these are details, I got the idea, I will simplify. However I am still willing to highlight the fact that the model (`ngModel`) is storing the actual, initial value provided by the developer, not any representation.

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
ymeine: Well, I say that very early in the first paragraph of the section. To me using `resultFormatter` in the title does not make it obvious they will use a string to represent the option.

With the two paragraphs names, I highlight the fact it's "string" vs "template" methods, no matter how the API is actually named ("formatter" is already a confusing name).

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
ymeine: I can however use a simpler phrase indeed, like the one you proposed. And for the other paragraph above, it would be "providing a string".

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
ymeine: Oh yeah, my bad 😅

demo/src/app/components/typeahead/overview/sections/introduction/template.html
ymeine: Yes, better!

demo/src/app/components/typeahead/overview/sections/introduction/template.html
ymeine: Totally agreed

demo/src/app/components/typeahead/overview/sections/customize-option-in-input/template.html
ymeine: All this is because it is not obvious that the input formatting might mess with future searches (depending on how it formats).

Some implementations might change the input element value when the user is about to edit the search term, exactly to avoid this. This is not the case in out implementation.

Some implementations might use complex layers to be able to display rich content inside what is considered to be the input, thus allowing more representations than just a simple string. This is not the case in out implementation.

This is for me a good point to highlight. I can simplify the wording though.

demo/src/app/components/typeahead/overview/sections/introduction/template.html
ymeine: Yes, good catch

ymeine:
> 
> 
> @ymeine, also forgot to attach this one, sorry:
> 
> Screen Shot 2019-03-28 at 11 13 35

Well, the empty space shocked me too, but this is due to the styling of the header.

I was also thinking about removing the header... And the anchor is not really needed, people can link using http://ng-bootstrap.github.io/#/components/datepicker/overview directly.

Removing it would fix both issues ;)

angular.json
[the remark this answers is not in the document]
ymeine: If I do it separately, I could do it _before_ this one instead. If it's a matter of PR code separation, it works too. For content matters, I consider line highlighting as part of the content (it's not just decoration).

ymeine:
> 
> 
> Hey,
> 
> it looks much more like an overview now, thanks for the work!
> 
>     * please remove line highlighting, nothing has changed since the last PR. It's still horribly slow :)
> 
>     * please do not capitalize "Typeahead" and "Observable" in the text
> 
>     * also please do not use "The Typeahead", just "typeahead"
> 
>     * left comments about the passages that could be simplified (in my point of view)
> 
> 
> Cheers.

Thanks for the review. For the several points: 

- line highlighting: agreed, alternative implementation PR planned
- replied to comments

However, for "(the) T·typeahead" and "Oꞏobservable" I have mixed feeling.

If I consider the widget as a tool, an application, a framework, anything, I would drop the "the" indeed. For instance I don't say "The Chrome" or "The Angular". However in this case it means we consider it as a proper noun, not a common noun, so it does use a capital letter.

Now, to me "observable" is a generic noun, and in theory it could be anything. Here we are talking about a specific kind of object, a known entity. It can't work with others. That's why I use a capital letter. I was also hesitating adding code block formatting but was afraid of visual clutter. An Observable is like a Promise, and [on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) they use "the", the capital letter, and the code block, which comforts me in what I was thinking.

So to sum up: 

- I can remove the "the" in front of "Typeahead", but in this case not the capital letter
- I would keep the "the" and the capital letter for "Observable", and I could add the code block formatting

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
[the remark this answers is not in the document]
maxokorokov: ok, agreed. I just find `(below)` and `(above)` redundant I guess

demo/src/app/shared/code/code.component.ts
```diff
@@ -6,19 +6,36 @@ import {CodeHighlightService} from './code-highlight.service';
...
   `
 })
 export class NgbdCodeComponent implements AfterViewInit {
 
-  @ViewChild('code') codeEl: ElementRef<HTMLElement>;
+  @ViewChild('codeEl') codeEl: ElementRef<HTMLElement>;
 
   @Input() code = '';
   @Input() lang = '';
+  @Input() showLineNumbers = false;
```
maxokorokov: please remove

demo/src/app/shared/code/code.component.ts
```diff
@@ -6,19 +6,36 @@ import {CodeHighlightService} from './code-highlight.service';
   selector: 'ngbd-code',
   changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <pre class="language-{{ lang }}"><code #code class="language-{{ lang }}"></code></pre>
+    <pre
```
maxokorokov: please rollback to the version without `[ngClass]`, this works perfectly fine:

```html
<pre class="language-{{ lang }}"><code #codeEl class="language-{{ lang }}">{{code.trim()}}</code></pre>
```

demo/src/app/shared/code/code.component.ts
```diff
@@ -6,19 +6,36 @@ import {CodeHighlightService} from './code-highlight.service';
...
+    };
+  }
+
+  get language(): string {
+    return `language-${this.lang}`;
+  }
+
   ngAfterViewInit() {
-    this.codeEl.nativeElement.innerHTML = this._service.highlight(this.code, this.lang);
+    this._service.highlightElement(this.codeEl.nativeElement);
```
maxokorokov: Good catch, thanks!

ymeine:
Was line numbers bringing a performance issue?

Regarding line highlighting, I heard your comment, and I proposed a solution, that's why I didn't remove anything yet. I disagree with the idea on focusing only on text. I am not putting line highlighting as pure decoration, it conveys information and ease of reading. Just like we add images or other rich elements to our documentations: they are not less important (also just like the tone of a voice is as important, if not actually more, than only words). I have a working, simple implementation ready and it's very concise, so I will add it because I don't see any reason why it would be complicated to take it. I don't see myself keeping in a lost branch the piece of code which has the information of which lines I want to be highlighted in the future.

Now I can agree for the capitalization thing, but you didn't say about the usage of "the". Given what you said for the capitalization, I think I would use normal rules for both, depending on if I refer to a particular instance ("the") or any ("a"/"an").

demo/src/app/components/typeahead/overview/common/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  language: 'html' | 'typescript' | 'css';
+  code: string;
+  highlightedLines?: string;
+}
+
+export function Snippet(spec: SnippetSpec): SnippetSpec {
+  return {
+    language: spec.language,
+    code: fixIndent(spec.code),
+    highlightedLines: spec.highlightedLines,
```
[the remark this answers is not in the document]
ymeine: See main comment below

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,47 @@
...
+<ul>
+  <li>what: strings corresponding to the current content of the input</li>
+  <li>when: on the <code>input</code> event of the input</li>
+</ul>
+
+<p>
+  You can therefore use this input as a search term and apply filtering for example:
+</p>
+
+<ngbd-code [lang]="snippets.observable.mapping.language" [code]="snippets.observable.mapping.code" [highlightedLines]="snippets.observable.mapping.highlightedLines"></ngbd-code>
```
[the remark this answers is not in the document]
ymeine: Because the initial abstraction is poorly designed according to me. So I don't break things, but I make it work in a way that is more logical.

See it that way: with the current abstraction, the code and the language of the snippet are defined in two separate locations, while they are intrinsically linked. So why defining it in two separate locations??

What I would have liked to do is just pass a snippet object which has all the information in one place. Even for line highlighting, it should rather sit along with the code of the snippet otherwise you'll jump between two files to update things.

Something like: 

```html
<ngbd-code [snippet]="path.to.my.snippet"></ngbd-code>
```

demo/src/app/components/typeahead/overview/sections/customize-option-in-input/template.html
ymeine: I will rewrite it (keeping the info but making it simpler)

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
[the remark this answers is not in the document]
ymeine: Oh that I can remove. Luckily with a vertical layout I guess we don't have the same issues as for "trimStart" vs "trimLeft", etc. due to RTL mode. So "previous" and "above" will always remain equivalent in this case and same for "next" and "below". I'll keep "below" and "above" since this is directional, like the arrow keys are.

demo/src/app/shared/code/code.component.ts
ymeine: See main comment below

demo/src/app/shared/code/code.component.ts
ymeine: Well, I think it is easier to use `ngClass` when the class is complex, and it when adding line numbering. Actually I started to remove line numbering too following your comments, but I did half of the job. However I maintain that I'd like to keep it if the numbering doesn't bring performance impact (also I don't know which order of magnitude we are talking about).

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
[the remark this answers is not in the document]
maxokorokov: > you'll jump between two files to update things

That was my point. Firstly, you never change snippet type. Secondly, without highlighted lines it is still fine the way it is.

- it is out of scope of typeahead overview
- there is a #3111 if you want to continue discussion about `<ngbd-code>` specifically

demo/src/app/shared/code/code.component.ts
maxokorokov: here we have a single class, binding was just fine

maxokorokov:
> Was line numbers bringing a performance issue?

no idea, both highlighting and numbering included in the PR and Chrome trace mentioned highlighting.

> ... they are not less important (also just like the tone of a voice is as important, if not actually more, than only words). I have a working, simple implementation ready and it's very concise, so I will add it because I don't see any reason why it would be complicated to take it

I don't have much to add to my previous comment on the matter, let's focus on the overview content in this PR. If you think you absolutely need the highlighting / numbering feature for your PR, please let's introduce it to the demo site first and update existing overview pages / snippets.

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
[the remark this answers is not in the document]
ymeine: It happened to me that I was inserting some snippet somewhere, and wanted to change the snippet afterwards, sometimes with different language. It was easier to blindly change the "path" to the snippet rather than having to check its type, and update the language property specifically. And splitting the two remains illogical.

Since changing the API of the code component is not part of this PR, I can at least write the documentation in a way that appears more logical? After all, it just means: "insert this snippet there", you don't care which language it is when reading the template. The, you check the snippet and there you have all the information.

demo/src/app/shared/code/code.component.ts
ymeine: We won't have a single class anymore if keeping the **line highlighting**, and this is a pending item, therefore I won't go back and forth in changing the code. We are not talking about a final code here as long as this item is pending.

ymeine:
Line highlighting and numbering *IS* content too. That's what I'm trying to explain. When you are telling me "let's focus on the content", I don't dissociate it. Therefore I am still complying to your request of focusing on content.

Therefore I will propose the line highlighting and numbering in a PR as you are suggesting. Moreover, I really don't want to backup somewhere in a lost branch some information that will be integrated later on, that don't often end well.

ymeine:
Changes done.

As asked, line highlighting and numbering (temporarily) removed.

However I based it on a pending PR: [chore(demo): use Snippet model for ngbd-code by ymeine · Pull Request #3129 · ng-bootstrap/ng-bootstrap](https://github.com/ng-bootstrap/ng-bootstrap/pull/3129), because reverting impacted code would make many changes (especially for the indentation of the snippets).

For review, please select only the last commit of the 3 ;)

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,47 @@
...
+<h5>The basic <code>Observable</code> chain</h5>
+
+<p>
+  In this basic example we will simply use a statically defined and synchronously available set of data — color names:
+</p>
+
+<ngbd-code [snippet]="snippets.data"></ngbd-code>
+
+<p>
+  The function used to provide the <code>Observable</code> to the typeahead receives a single argument: another <code>Observable</code>. It spawns:
```
maxokorokov: This paragraph is still hard to read/understand, please check it if you have time

demo/src/app/components/typeahead/overview/sections/observable-contract/template.html
```diff
@@ -0,0 +1,27 @@
...
+  Our typeahead is designed around a single <code>Observable</code>, to which it subscribes in order to get an updated list of options and update its popup accordingly.
+</p>
+
+<p>
+  The choice of an observable is due to the nature of the data processing: often updated and potentially asynchronous. Observables are designed mainly for this use case, and bring a common API to handle data flow no matter if the processing is asynchronous or synchronous.
+</p>
+
+
+
+<h5>What is the <code>Observable</code> contract?</h5>
```
maxokorokov: This section is still hard to read/understand

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,70 @@
...
+    </tr>
+    <tr>
+      <td><code>mouseenter</code></td>
+      <td>Highlight option under the mouse pointer</td>
+    </tr>
+  </tbody>
+</table>
+
+<p>
+  Note that starting keyboard navigation will always start from the currently activated option, no matter if activated from the keyboard or the mouse.
```
maxokorokov: Stating the obvious, also this is not the typeahead user documentation.

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,70 @@
...
+</p>
+
+
+
+<h5>Select an option</h5>
+
+<p>Selecting means committing to the data model. It will display the currently selected option in the input. It will also make the popup close.</p>
```
maxokorokov: Stating the obvious, once you play with the example you know this

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,70 @@
...
+    </tr>
+  </tbody>
+</table>
+
+
+
+<h5>Dismiss the popup</h5>
+
+<p>
+  Dismissing the popup consists in closing the popup without selecting an option.
```
maxokorokov: Stating the obvious

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
```diff
@@ -0,0 +1,41 @@
...
+<p>
+  But the requirement is to be able to <em>display</em> those options, be it in the list of the popup or inside the input field.
+</p>
+
+
+
+<h5>Providing a string</h5>
+
+<p>
+  The option item can be displayed from a string. In this case, use the property <code>resultFormatter</code> to pass a function which takes the option item (whose model is up to you) as input and returns a string as output.
```
maxokorokov: Describing this API in english, code is simpler, no ?

```ts
@Input() resultFormatter: (item: any) => string;
```

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,47 @@
...
+</ul>
+
+<p>
+  You can therefore use this input as a search term and apply filtering for example:
+</p>
+
+<ngbd-code [snippet]="snippets.observable.mapping"></ngbd-code>
+
+<p>
+  Here we return all the colors when the search term is empty, otherwise we return all the colors starting with it.
```
maxokorokov: Describing a ternary operator with words...

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,47 @@
+<p>
+  To set up a typeahead instance, you need to pass an initialization function to it. This function must return an <code>Observable</code> spawning lists of options: this is how you control the list of displayed options.
+</p>
+
+<p>
+  Pass it that way on the template side:
```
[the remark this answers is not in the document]
ymeine: Will change it

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,47 @@
...
+</p>
+
+<ngbd-code [snippet]="snippets.observableSetup.template"></ngbd-code>
+
+<p>
+  And define it that way on the component side:
```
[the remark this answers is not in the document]
ymeine: Will change it too (consistently)

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
```diff
@@ -0,0 +1,47 @@
...
+</p>
+
+<ngbd-code [snippet]="snippets.observable.mapping"></ngbd-code>
+
+<p>
+  Here we return all the colors when the search term is empty, otherwise we return all the colors starting with it.
+</p>
+
+<p>
+  A basic typeahead instance will still require basic processing like debouncing, search term length threshold, etc., while more advanced use cases will require even further usage of the Observables.
```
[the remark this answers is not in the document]
ymeine: Both sound correct to me, and I don't have enough fluency in English to detect the subtleties at work here. I can add "some".

demo/src/app/components/typeahead/overview/sections/observable-contract/template.html
```diff
@@ -0,0 +1,27 @@
...
+
+<p>
+  Out typeahead expects to receive an <code>Observable</code> which spawns lists of items of any type.
```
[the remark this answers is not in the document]
ymeine: Nice catch

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
ymeine: Can you state what you consider being hard to read/understand? This is a highly personal opinion so I can't guess what is a true good form.

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
ymeine: I've seen it fairly commonly, a documentation which repeats quickly what the code does, to give a more literate style. Here we're talking only about one sentence, not overwhelming in my opinion.

It's not about describing the operator though, it's about describing the intent of the code. If it had been more complex (for any reason) than a pipe, a map, a ternary operator and a filtering, I would still keep that phrase, as the statement of the contract this piece of code is trying to achieve.

I personally don't see any problem in "describing a ternary operator with words" (knowing that this is an incorrect shortcut besides). If there is a concrete issue with it, please state it clearly. You just described what I did with your own vision which gives me no useful information.

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
ymeine: Well, to know how to balance text vs code is not always easy. Sometimes text gives more information. In that case it seems equivalent. I'll think about it, good suggestion

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
ymeine: "Seeing" is entirely linked to the extent to which the user plays with the widget, and we can't know what they are going to do. So the statement "you see" can't be proved.

It is the same for "you know these things", we can't just consider what people know or not when it directly touches this specific implementation we provide. We can for pre-requisites such as Angular or RxJS, etc.

If I go with your logic, I perfectly know that <kbd>escape</kbd> will close all typeaheads, so no need to tell it either! Same for using <kbd>enter</kbd> and <kbd>space</kbd>. We can't have double standards.

You example with the button is not appropriate here. You are describing the behavior of a standard button on the web platform. We can consider that as an implicit pre-requisite and not describe it again.

But our typeahead implementation differs from the standard `select` element or any other similar widgets I've seen in any other application. There is a reason I described those behaviors, and that's because they belong to the minority as far as I personally know.

To quickly prove my point, pick a similar widget (standard `select`, a browser URL bar, a website with a search field) a,d try the following scenario: 

- make the popup open
- navigate with the keyboard
- hover another item with the mouse: did the highlighting change or are there two different elements highlighted?
- navigate again with the keyboard: which item became highlighted then? The one before/after the one hovered the mouse or the one before/after the one previously highlighted with he keyboard?

If we can consider a standard typeahead, our implementation is not matching it, hence a comprehensive documentation. And back to "it's obvious", "we know", you are yourself telling to describe things which can fall into the "obvious" category, so I don't see any partial limit here.

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
ymeine: For the "obvious" part, see my main comment above. Regarding the user documentation aspect, the whole "interactions" section IS end user documentation. So what do we do?

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
ymeine: For the "obvious" part, see my main comment above.

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
```diff
@@ -0,0 +1,70 @@
...
+
+
+<h5>Select an option</h5>
+
+<p>Selecting means committing to the data model. It will display the currently selected option in the input. It will also make the popup close.</p>
+
+<table class="table mt-4">
+  <tbody>
+    <tr>
+      <td><kbd>Enter</kbd> ● <kbd>Tab</kbd></td>
```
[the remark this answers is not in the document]
ymeine: Oh you are right, I didn't think about it ^^ Could be considered as combination or else. I prefer the word "or" in this case.

demo/src/app/components/typeahead/overview/sections/user-interaction/template.html
ymeine: Well, I guess I didn't pay attention enough to the fact that the word "dismiss" is self-explanatory in this case, I guess.

ymeine:
While I do agree there can be complexity in phrases, such as in code, it is however quite difficult to evaluate truly. Is this hemingwayapp a recognized reference? Because the red and yellow sentences I see in their example really don't bother me personally. So if it's a pure matter of personal opinion, I wouldn't trust it. I will try to simplify as much as I can though, since it seems to be the issue (I don't want to consider users as illiterate though ^^ )
 
I know writing doc is hard, however I would be quite irritated to infer from this sentence that I am a newbie and you have more experience and that it would grant your opinions a status of truth. (same for "it's a good exercise") If it's the case, I would suggest a bit more diplomacy in your comments. I've read tons of documentations, and I even got mocked because of that, it would be a pity now for me to be considered as not having enough experience in this. React docs yes. Many libraries, tool, applications, articles, tutorials, etc. When I debuted I was also reading languages official specifications. It goes from very complex, formal, exhaustive documents to lighter ones. I've also written many of them. Got mocked for that too.

Now, the main culprit I've seen in documentations was not the complexity of sentences, which is something we should be able to process eventually, but the lack of information. People considering things are obvious when they're not at all, being very elusive, and you end up having to try by yourself, inferring things which could be false, or checking the source code directly.

And most of all, personal opinions remain personal. We need proper rules, other than saying "it is" as a truth, when it's actually a "it is" with an implicit "in my opinion".

demo/src/app/components/typeahead/overview/sections/basic-usage/template.html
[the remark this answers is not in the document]
maxokorokov: Was pointing out double word "basic" in a sentence

demo/src/app/components/typeahead/overview/sections/customize-options-list/template.html
maxokorokov: > Well, to know how to balance text vs code is not always easy

So was my point about "writing docs is hard

maxokorokov:
> I would be quite irritated to infer from this sentence that I am a newbie
...
> I would suggest a bit more diplomacy in your comments
...
> even got mocked because of that

Was not my intention to be impolite or mock you in any way. On the opposite, was trying to be helpful.

> And most of all, personal opinions remain personal

So are code reviews, and at this point I've said all I had to say for this one.

benouat:
@ymeine I think we need a last rebase here. I wanted to merge it, and saw the conflict.

PR #3017, opened 2019-02-13
fix(positionning): set the first placement as the default one

fbasso:
The purpose is to have the first given placement of a popup as the default one.

If any position of a  popover/tooltip/dropdown/... can't enter in the viewport, it's not natural to keep the last one as the preferred one.

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -180,7 +180,21 @@ export function positionElements(
     'top', 'bottom', 'left', 'right', 'top-left', 'top-right', 'bottom-left', 'bottom-right', 'left-top', 'left-bottom',
     'right-top', 'right-bottom'
   ];
+
   const classList = targetElement.classList;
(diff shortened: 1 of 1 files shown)


src/util/positioning.ts
```diff
@@ -181,6 +181,16 @@ export function positionElements(
     'right-top', 'right-bottom'
   ];
   const classList = targetElement.classList;
+  const addClasses = (primary: string, secondary: string) => {
+    if (baseClass) {
+      classesToAdd.push(`${baseClass}-${primary}`);
```
maxokorokov: Using `classesToAdd` is super confusing here to me :)

My thought process:
- looking at line 186 → whoa, but where is `classesToAdd` defined ?
- oh, it's much later on line 221... but... it is undefined, how this should work?
- oh it's initialised in the loop on line 225

It should just be a function-local variable with this new implementation, no ?
Otherwise, I'd say, pass it as an argument

PR #3018, opened 2019-02-13
chore(positioning): provide scrolling subscribers

fbasso:
This is a PR to discuss about adding a scrolling subscription functions to the positioning utility. 

At the moment, it is not ready to be integrated (test page to complete or remove, as it doesn't test anything for now :) ), but the 2e2 page `#/position/scroll` show the issue met when popup are inside parents with scrollbars : when the scroll is done, the popup position is frozen to its original position (comment `export function subscribeScroll` function code to see the issue on the page).

This can be done the the tooltip and popover, but it's also planned to be used in the PR #2823, where this issue will be more visible (a dropdown in a scrolled parent with container="body" will not move while scrolling).

So the current questions are:

- Is it this relevant to have this in an utilty ?
- Do the tooltip and popover placement fixed with this PR (there is no issue related to this problem at the moment) ?
- When scrolling is done, the popup must be scrolled too, or simply closed ? (two valuable UX decisions I think)

### DIFF
--- a/e2e-app/src/app/position/scroll/scroll.component.html
+++ b/e2e-app/src/app/position/scroll/scroll.component.html
@@ -1,0 +1,21 @@
+<h3>Position with scroll tests</h3>
+
+<div>
+  <button class="btn btn-outline-secondary ml-1" id="container-null" (click)="container = null">Append to widget</button>
+  <button class="btn btn-outline-secondary ml-1" id="container-body" (click)="container = 'body'">Append to body</button>
+</div>
+<div id="outer-container" class="p-3" style="height: 400px; overflow: scroll">
+  <div id="inner-container" class="position-relative d-inline-block p-3" style="height: 200px; overflow: scroll;">
+    <div class="position-relative d-flex justify-content-center align-items-center" style="height: 250px;">
+      <button id="tooltip-target" ngbTooltip="Tooltip here" [autoClose]="false" triggers="click" type="button"
+        class="btn btn-outline-secondary ml-2" [container]="container">
+        Tooltip
+      </button>
+      <button id="popover-target" ngbPopover="Popover here" [container]="container" [autoClose]="false" triggers="click" type="button"
+        class="btn btn-outline-secondary ml-2" [container]="container">
+        Popover
+      </button>
+    </div>
+  </div>
+  <div style="height: 300px; width: 100%;"></div>
+</div>
--- a/e2e-app/src/app/position/scroll/scroll.e2e-spec.ts
+++ b/e2e-app/src/app/position/scroll/scroll.e2e-spec.ts
@@ -1,0 +1,148 @@
+import {openUrl} from '../../tools.po';
+import {ScrollPage} from './scroll.po';
+
+const roundLocation = function(location) {
+  location.x = Math.round(location.x);
+  location.y = Math.round(location.y);
+
+  return location;
+};
+
+const popoverMarginBottom = 8;
+
+describe('Scroll', () => {
+
+  let page: ScrollPage;
+
+  const expectPositionOnTop = async(btn, popupElement, offset = 0) => {
+    const btnLocation = roundLocation(await btn.getLocation());
+    const btnSize = await btn.getSize();
+    const popupLocation = roundLocation(await popupElement.getLocation());
+    const popupSize = await popupElement.getSize();
+
+    const yDiff = (popupLocation.y + popupSize.height) - btnLocation.y + offset;
+    const xDiff = (popupLocation.x + popupSize.width / 2) - (btnLocation.x + (btnSize.width / 2));
+
+    expect(Math.abs(yDiff)).toBeLessThanOrEqual(1, `Popup top positionning`);
+    expect(Math.abs(xDiff)).toBeLessThanOrEqual(1, `Popup left positionning`);
+
+  };
+
+  beforeAll(() => page = new ScrollPage());
+
+  beforeEach(async() => await openUrl('position/scroll'));
+
+  describe('Tooltip', () => {
+
+    it(`move with the target (first parent scrolled, container = null)`, async() => {
+
+      const tooltipButton = page.getButton('tooltip');
+      await tooltipButton.click();
+      const tooltip = page.getPopup('tooltip');
+
+      await expectPositionOnTop(tooltipButton, tooltip);
+      await page.scrollElement('inner-container', 50);
+      await expectPositionOnTop(tooltipButton, tooltip);
+
+    });
+
+    it(`move with the target (first parent scrolled, container = body)`, async() => {
+
+      await page.selectContainer('body');
+
+      const tooltipButton = page.getButton('tooltip');
+      await tooltipButton.click();
+      const tooltip = page.getPopup('tooltip');
+
+      await expectPositionOnTop(tooltipButton, tooltip);
+      await page.scrollElement('inner-container', 50);
+      await expectPositionOnTop(tooltipButton, tooltip);
+
+    });
+
+    it(`move with the target (second parent scrolled, container = null)`, async() => {
+
+      const tooltipButton = page.getButton('tooltip');
+      await tooltipButton.click();
+      const tooltip = page.getPopup('tooltip');
+
+      await expectPositionOnTop(tooltipButton, tooltip);
+      await page.scrollElement('outer-container', 50);
+      await expectPositionOnTop(tooltipButton, tooltip);
+
+    });
+
+    it(`move with the target (second parent scrolled, container = body)`, async() => {
+
+      await page.selectContainer('body');
+
+      const tooltipButton = page.getButton('tooltip');
+      await tooltipButton.click();
+      const tooltip = page.getPopup('tooltip');
+
+      await expectPositionOnTop(tooltipButton, tooltip);
+      await page.scrollElement('outer-container', 50);
+      await expectPositionOnTop(tooltipButton, tooltip);
+
+    });
+
+
+  });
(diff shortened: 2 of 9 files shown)


src/util/positioning.ts
```diff
@@ -89,6 +93,42 @@ export class Positioning {
...
+        if (overflowRegex.test(style.overflow + style.overflowY + style.overflowX)) {
+          return parent;
+        }
+      }
+    }
+
+    return null;
+  }
+
+  subscribeScroll(hostElement: HTMLElement, callback: Function): Function {
```
maxokorokov: Could you refactor it to be like to the `ngbAutoClose` and `ngbFocusTrap` we have ?

```ts
// here is a sketch:

export function ngbScroll(hostElement: HTMLElement, stopped$: Observable<any>, callback: Function) {

   runOutsideAngular(() => {

        // to be used like
        fromEvent(...).pipe(takeUntil($stopped));

   });

}
```

- rename to `ngbScrollProvider` or something similar
- move `runOutsideAngular` inside the function
- pass `stop$` observable to avoid unsubscribing

src/tooltip/tooltip.ts
```diff
@@ -210,4 +216,21 @@ export class NgbTooltip implements OnInit, OnDestroy {
...
+    if (this._scrollSubscription) {
+      this._scrollSubscription();
+      this._scrollSubscription = null;
+    }
+  }
+
+  private _subscribeScroll() {
+    this._ngZone.runOutsideAngular(() => {
+      this._scrollSubscription = subscribeScroll(this._elementRef.nativeElement, () => {
+        positionElements(
```
maxokorokov: And I guess we should close it, not reposition, so it really would be like `ngbAutoClose`

src/util/positioning.ts
```diff
@@ -89,6 +93,42 @@ export class Positioning {
     return elOffset;
   }
 
+  private _getScrolledParent(element, includeHidden = false) {
```
maxokorokov: this could be static in I guess and missing types.
Also maybe `getScrollParent`, not `scrolled`?

src/util/scroll.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (overflowRegex.test(style.overflow + style.overflowY + style.overflowX)) {
+        return parent;
+      }
+    }
+  }
+
+  return null;
+};
+
+export function ngbScroll(zone: NgZone, hostElement: HTMLElement, close: () => void, closed$: Observable<any>) {
```
maxokorokov: let's rename `close: () => void` to `callback: () => void`

maxokorokov:
Also opened #3043

fbasso:
fix #3046

e2e-app/src/app/position/scroll/scroll.component.html
```diff
@@ -0,0 +1,21 @@
...
+  <button class="btn btn-outline-secondary ml-1" id="container-body" (click)="container = 'body'">Append to body</button>
+</div>
+<div id="outer-container" class="p-3" style="height: 400px; overflow: scroll">
+  <div id="inner-container" class="position-relative d-inline-block p-3" style="width: 600px; height: 200px; overflow: scroll;">
+    <div class="position-relative d-flex justify-content-center align-items-center" style="height: 250px;">
+      <button id="tooltip-target" ngbTooltip="Tooltip here" [autoClose]="false" triggers="click" type="button"
+        class="btn btn-outline-secondary ml-2" [container]="container">
+        Tooltip
+      </button>
+      <button id="popover-target" ngbPopover="Popover here" [container]="'body'" [autoClose]="false" triggers="click" type="button"
```
maxokorokov: `[container]="container"`, otherwise your tests are not really valid :)

e2e-app/src/app/position/scroll/scroll.component.html
```diff
@@ -0,0 +1,21 @@
+<h3>Position with scroll tests</h3>
+
+<div>
+  <button class="btn btn-outline-secondary ml-1" id="container-null" (click)="container = null">Append to widget</button>
+  <button class="btn btn-outline-secondary ml-1" id="container-body" (click)="container = 'body'">Append to body</button>
+</div>
+<div id="outer-container" class="p-3" style="height: 400px; overflow: scroll">
+  <div id="inner-container" class="position-relative d-inline-block p-3" style="width: 600px; height: 200px; overflow: scroll;">
```
maxokorokov: just a nitpick, but is there a reason for setting width to `600px` for these tests?

e2e-app/src/app/position/scroll/scroll.e2e-spec.ts
```diff
@@ -0,0 +1,149 @@
...
+
+      await expectPositionOnTop(tooltipButton, tooltip);
+      await page.scrollElement('inner-container', 50);
+      await expectPositionOnTop(tooltipButton, tooltip);
+
+    });
+
+    it(`move with the target (first parent scrolled, container = body)`, async() => {
+
+      page.selectContainer('body');
```
maxokorokov: I'm not sure this works, because `selectContainer` returns a `Promise` you're not waiting for

Same comment for all `page.selectContainer()` calls in this file

e2e-app/src/app/position/scroll/scroll.po.ts
```diff
@@ -0,0 +1,20 @@
+import {$, browser} from 'protractor';
+
+export class ScrollPage {
+  getButton(type: 'tooltip' | 'popover') { return $(`#${type}-target`); }
+
+  getPopup(type: 'tooltip' | 'popover') { return $(`ngb-${type}-window`); }
+
+  // getTooltipContent() { return this.getTooltip().$('div.tooltip-inner'); }
```
maxokorokov: please remove

e2e-app/src/app/position/scroll/scroll.po.ts
```diff
@@ -0,0 +1,20 @@
...
+  async scrollElement(elementId: string, elementScrollY) {
+    await browser.executeScript((id, scrollY) => {
+      // Scroll to given position
+      document.getElementById(id).scrollTo(0, scrollY);
+    }, elementId, elementScrollY);
+  }
+
+  async selectContainer(container: null | 'body') { await $(`#container-${container}`).click(); }
+
+  async openTooltip() { await $('button[ngbTooltip]').click(); }
```
maxokorokov: seems unused

src/popover/popover.ts
```diff
@@ -208,13 +207,17 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
...
-      if (this.container === 'body') {
+      const isBodyContainer = this.container === 'body';
+      if (isBodyContainer) {
         this._document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
       }
 
       // apply styling to set basic css-classes on target element, before going for positioning
       this._windowRef.changeDetectorRef.markForCheck();
 
+      if (isBodyContainer) {
```
maxokorokov: could we simply merge this condition into the previous one? (same for the tooltip)

```ts
if (this.container === 'body') {
    this._document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
    ngbScroll(this._ngZone, this._elementRef.nativeElement, () => this.positionElements(), this.hidden);
}
```

src/popover/popover.ts
```diff
@@ -185,9 +186,7 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
 
     this._zoneSubscription = _ngZone.onStable.subscribe(() => {
       if (this._windowRef) {
```
maxokorokov: I think it might be safer to have `if (this._windowRef)` inside the `this.positionElements()` function.
Now you're checking it only here, but not before the other call to `positionElements()`

src/tooltip/tooltip.ts
```diff
@@ -249,4 +252,10 @@ export class NgbTooltip implements OnInit, OnDestroy {
     }
     this._zoneSubscription.unsubscribe();
   }
+
+  private positionElements() {
```
maxokorokov: style: should be `_positionElements` as it is a private function

src/util/scroll.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  return null;
+};
+
+export function ngbScroll(zone: NgZone, hostElement: HTMLElement, callback: () => void, closed$: Observable<any>) {
+  zone.runOutsideAngular(() => {
+    let scrolledParent = getScrolledParent(hostElement);
+    while (scrolledParent) {
+      fromEvent(scrolledParent, 'scroll').pipe(takeUntil(closed$), sampleTime(10)).subscribe(() => { callback(); });
```
maxokorokov: I think you're using the wrong operator here, should be `throttleTime()` I guess. 
`sampleTime()` emits every 10ms regardless if there was scrolling or not:

Screen Shot 2019-05-13 at 12 09 13

src/util/scroll.ts
```diff
@@ -0,0 +1,34 @@
+import {NgZone} from '@angular/core';
+import {fromEvent, Observable, race} from 'rxjs';
```
maxokorokov: `race` is unused

src/util/scroll.ts
maxokorokov: Another comment is that we should probably make `10ms` configurable by the user via the dependency injection, similar to what we're doing for the live announcer.

We can do it separately later of course as things are private for now anyway.

fbasso:
Close this one as now we use popper to position elements, which manages this use-case.

PR #3019, opened 2019-02-13
fix(positionning): auto positionning for tooltip and popover

fbasso:
The commit #2972  introduced a non backward changes on dropdown placement for some widgets. This is the revert for their default configurations.


fbasso:
Finally, auto positioning popups in the viewport seems to be the right way. 

I've added this behavior for tooltip and popover

PR #3034, opened 2019-02-20
feat(position): accept placement values with a space separated string

fbasso:

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -159,6 +159,7 @@ export class Positioning {
   }
 }
 
+const placementSeparator = /\s+/;
 const positionService = new Positioning();
 
 /*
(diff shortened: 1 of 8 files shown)


src/util/positioning.ts
```diff
@@ -174,7 +174,7 @@ const positionService = new Positioning();
 export function positionElements(
     hostElement: HTMLElement, targetElement: HTMLElement, placement: string | Placement | PlacementArray,
     appendToBody?: boolean, baseClass?: string): Placement {
-  let placementVals: Array<Placement> = Array.isArray(placement) ? placement : [placement as Placement];
+  let placementVals: Array<Placement> = Array.isArray(placement) ? placement : placement.split(' ') as Array<Placement>;
```
ymeine: If we are opening the API that way, we might as well be more flexible regarding the white spacing. Using a regexp for instance: 

```suggestion
  let placementVals: Array<Placement> = Array.isArray(placement) ? placement : placement.split(/\s+/) as Array<Placement>;
```

fbasso: Thanks, @ymeine, good finding ! I'll just set the RegExp in a constant to avoid to get a new one each time.

PR #3042, opened 2019-02-26
fix(position): wrong placement class applies in some cases

### LINKED ISSUE #3040: Popover arrow placement can be incorrect when multiple placements are used
### Bug description:

When using multiple `Placement` values to position a popover, if none of them result in the popover being on screen, the first `Placement` is used to position the popover, but the _last_ `Placement` is used to control where the popovers arrow appears.

image

https://stackblitz.com/edit/angular-ducwpd

Click on the button to open the popover. The popover will appear centered over the button, but the arrow will be to the right edge of the popover. 

If you change the `placement` order in `app.component.html` to `['top', 'top-right', 'top-left']`, the arrow will appear towards the left edge of the popover. If you change the placements to just contain `top`, the arrow is centered as you would expect.

Angular: 7.2.6

ng-bootstrap: 4.0.4

Bootstrap: 4.3.1

fbasso:
fix #3040

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -223,7 +223,6 @@ export function positionElements(
   // The translate3d/gpu acceleration render a blurry text on chrome, the next line is commented until a browser fix
   // style['will-change'] = 'transform';
 
-  const lastPlacement = placementVals[placementVals.length - 1];
   let testPlacement: Placement;
   let isInViewport = false;
   for (testPlacement of placementVals) {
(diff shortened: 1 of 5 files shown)


src/util/positioning.ts
```diff
@@ -235,7 +235,7 @@ export function positionElements(
     }
 
     // Remove the baseClasses for further calculation, except for the last one
```
benouat: You should also adapt the comment here to get rid of _execpt for the last one_

fbasso:
Changes done :)

PR #3049, opened 2019-03-04
fix(tests): remove unexpected elements between tests

fbasso:


src/util/util.spec.ts
```diff
@@ -1,5 +1,15 @@
 import {toInteger, toString, getValueInRange, isInteger, isString, hasClassName} from './util';
 
+afterAll(() => {
```
benouat: What is the reasoning behind putting this global `afterAll` here inside `util/util.spec.ts` ?

src/util/util.spec.ts
```diff
@@ -1,5 +1,15 @@
...
+  // Check that only the last test element is here, all previous ones must have been removed
+  const divs = Array.from(document.body.children).filter((element: HTMLElement) => {
+    return element.tagName.toLocaleLowerCase() === 'div';
+  });
+
+  expect(divs.length).toBe(1, 'Tests cleaning issues');
```
benouat: This being a global check, I would not use an expectation here.

`expect` is usually trying to re-use the _it_ description with the _describe_ description. We don't have none of them  here.

So I would myself change this `expect()` by a `fail()` (see [fail()](https://jasmine.github.io/2.3/introduction#section-Manually_failing_a_spec_with_%3Ccode%3Efail%3C/code%3E) on jasmine documentation)

PR #3051, opened 2019-03-05
fix(positioning): wrong position if the relative parent is scrolled

fbasso:


fbasso:
This PR is linked to #3018, where the intermediate div between the relative parent and the toggling buttons (in scroll.component.html ) is necessary, otherwise the position is incorrect.

This PR fixes this (intermediate div not necessary).

fbasso:

PR #3055, opened 2019-03-06
feat(datepicker): add 'closed' event and fire after closing dp window

nOstap:
This pull request adds "closed" event to NgbInputDatepicker, which is fired after closing datepicker window.
Example:
`<input ngbDatepicker (closed)="onClose()">`
Very useful for handling whenever user closing date-picker by clicking outside of dp box.

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -186,9 +186,7 @@ export class NgbInputDatepicker implements OnChanges,
   @Output() navigate = new EventEmitter<NgbDatepickerNavigateEvent>();
 
   /**
-   * An event fired after datepicker closing the datepicker.
-   *
-   * @memberof NgbInputDatepicker
+   * An event fired after closing datepicker window.
    */
   @Output() closed = new EventEmitter();
 
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -185,6 +185,13 @@ export class NgbInputDatepicker implements OnChanges,
    */
   @Output() navigate = new EventEmitter<NgbDatepickerNavigateEvent>();
 
+  /**
+   * An event fired after datepicker closing the datepicker.
+   *
+   * @memberof NgbInputDatepicker
+   */
+  @Output() closed = new EventEmitter();
+
   @Input()
   get disabled() {
     return this._disabled;
@@ -314,6 +321,7 @@ export class NgbInputDatepicker implements OnChanges,
       this._vcRef.remove(this._vcRef.indexOf(this._cRef.hostView));
       this._cRef = null;
       this._closed$.next();
+      this.closed.emit();
       this._changeDetector.markForCheck();
     }
   }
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker-input.ts
```diff
@@ -185,6 +185,13 @@ export class NgbInputDatepicker implements OnChanges,
    */
   @Output() navigate = new EventEmitter<NgbDatepickerNavigateEvent>();
 
+  /**
+   * An event fired after datepicker closing the datepicker.
```
benouat: The message here is not that clear

src/datepicker/datepicker-input.ts
```diff
@@ -185,6 +185,13 @@ export class NgbInputDatepicker implements OnChanges,
...
+   *
+   * @memberof NgbInputDatepicker
```
benouat: We usually don't use this kind of jsdoc comments. Could you please remove it ?

src/datepicker/datepicker-input.ts
nOstap: My bad, msg changed to: "An event fired after closing datepicker window."

src/datepicker/datepicker-input.ts
nOstap: Thanks for advice, removed.

ronjb04:
hi guys, is this already available?

im using:
    "@ng-bootstrap/ng-bootstrap": "^4.1.0",

however it seems to be not working. :(

nOstap:
> hi guys, is this already available?
> 
> im using:
> "@ng-bootstrap/ng-bootstrap": "^4.1.0",
> 
> however it seems to be not working. :(

Pull request is still open (needs review) I wouldn't expect this to work before merge.

benouat:
This one should land in `4.2` as it is a new feature. We can't land it in `4.1.1`

nOstap:
> This one should land in `4.2` as it is a new feature. We can't land it in `4.1.1`

Do I need to do something with my pull request to get this merged? Or just wait until next release?

src/datepicker/datepicker-input.ts
```diff
@@ -185,6 +185,11 @@ export class NgbInputDatepicker implements OnChanges,
    */
   @Output() navigate = new EventEmitter<NgbDatepickerNavigateEvent>();
 
+  /**
+   * An event fired after closing datepicker window.
+   */
+  @Output() closed = new EventEmitter();
```
maxokorokov: Please add type `new EventEmitter<void>()`

nOstap: Thanks for advice, changes has been done.
Actually I've got a question, is
`closed: EventEmitter<void> = new EventEmitter()`
equal to
`closed = new EventEmitter<void>()`
will Generic Type be forced to void in second case?

maxokorokov: Basic type inference is described in → https://www.typescriptlang.org/docs/handbook/type-inference.html. No generics there though...

Anyway looks like it works, at least autocompletion in IDE understands that it is specifically `EventEmitter<void>` and if you try `this.close.emit(true);`, it will complain.

There is also a tslint rule called "no-inferrable-types" → https://palantir.github.io/tslint/rules/no-inferrable-types/. And we have it active, but it works only for simple types like `const b: boolean = false`

PR #3058, opened 2019-03-08
fix(dropdown): wrong positioning calculation

### LINKED ISSUE #3056: Dropdown uses only the first specified placement.
### Bug description:
In the Stackblitz example there is a dropdown in the top-left corner of the screen. 
The HTML body is set to "overflow: hidden". 
The placement of the Dropdown is specified as "top bottom".
Since there is no space over the dropdown, I would expect that the "bottom" placement is used.
Instead the "top" placement is used.

A little digging through the @ng-bootstrap code led me to the positionElements() function in the "positioning.ts" file. Just for clarity, there are two such functions, one of which answers the question: "Does the menu with a given placement fit in the viewport, or not?". Now the problem I stumbled upon is that the BCR (Bounding Client Rect) is not yet initialized, when the check for the first placement is made and the function always returns true.

`return targetElBCR.left >= 0 && targetElBCR.top >= 0 && targetElBCR.right <= windowWidth &&
        targetElBCR.bottom <= windowHeight;`

always returns true on the first check, because targetElBCR looks like this:

`{ bottom: 0, height: 0, left: 0, right: 0, top: 0, width: 0, x: 0, y: 0 }`

[https://stackblitz.com/edit/angular-2xpppu](url)

Angular: 7.2.8

ng-bootstrap: 4.1.0

Bootstrap: 4.3.1

fbasso:
fix #3056 

The fix only adds the 'show' class before the positioning, but I don't know if it's the best solution as the class attribute is bound. I tried to 'unbind' the class, but it's much more painfull.

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -562,10 +562,35 @@ describe('ngb-dropdown-toggle', () => {
     const compiled = fixture.nativeElement;
     const dropdown = fixture.debugElement.query(By.directive(NgbDropdown)).injector.get(NgbDropdown);
     dropdown.open();
+    fixture.detectChanges();
     const dropdownElement = document.querySelector('div[ngbDropdownMenu]');
     const parentContainer = dropdownElement.parentNode;
     expect(parentContainer).toHaveCssClass('dropdown');
     expect(parentContainer.parentNode).toBe(document.body, 'The dropdown should be attached to the body');
+
+  });
+
+  it(`should second placement if the first one doesn't fit`, () => {
+    const html = `
+      <div ngbDropdown placement="left right">
+          <button ngbDropdownToggle>
+            <span class="toggle">Toggle dropdown</span>
+          </button>
+          <div ngbDropdownMenu>
+            <a ngbDropdownItem>dropDown item</a>
+            <a ngbDropdownItem>dropDown item</a>
+        </div>
+      </div>`;
+
+    const fixture = createTestComponent(html);
+    const compiled = fixture.nativeElement;
+    const dropdown = fixture.debugElement.query(By.directive(NgbDropdown)).injector.get(NgbDropdown);
+    dropdown.open();
+    fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


src/dropdown/dropdown.spec.ts
```diff
@@ -562,13 +562,39 @@ describe('ngb-dropdown-toggle', () => {
...
     dropdown.open();
+    fixture.detectChanges();
     const dropdownElement = document.querySelector('div[ngbDropdownMenu]');
     const parentContainer = dropdownElement.parentNode;
     expect(parentContainer).toHaveCssClass('dropdown');
     expect(parentContainer.parentNode).toBe(document.body, 'The dropdown should be attached to the body');
 
   });
 
+  it('should toggle at the second placement if needed', () => {
```
maxokorokov: `should second placement if the first one doesn't fit` ?

src/dropdown/dropdown.spec.ts
```diff
@@ -562,13 +562,39 @@ describe('ngb-dropdown-toggle', () => {
...
+      </div>`;
+
+    const fixture = createTestComponent(html);
+    const compiled = fixture.nativeElement;
+    const dropdown = fixture.debugElement.query(By.directive(NgbDropdown)).injector.get(NgbDropdown);
+    dropdown.open();
+    fixture.detectChanges();
+    const dropdownEl = compiled.querySelector('[ngbdropdownmenu]');
+    const targetElement = compiled.querySelector('button');
+    // expect(dropdownElement.getAttribute('x-placement')).toBe('right');
```
maxokorokov: please remove or uncomment

src/dropdown/dropdown.ts
```diff
@@ -57,31 +57,7 @@ export class NgbDropdownMenu {
 
   @ContentChildren(NgbDropdownItem) menuItems: QueryList<NgbDropdownItem>;
 
-  constructor(
```
maxokorokov: This is unused code left by chance from previous refactoring

PR #3065, opened 2019-03-13
chore: cleanup unused imports in 'index.ts'

maxokorokov:


src/index.ts
```diff
@@ -1,28 +1,21 @@
...
-  NgbModalOptions,
-  NgbModalRef,
-  ModalDismissReasons
-} from './modal/modal.module';
+import {NgbModalModule} from './modal/modal.module';
 import {NgbPaginationModule} from './pagination/pagination.module';
 import {NgbPopoverModule} from './popover/popover.module';
 import {NgbProgressbarModule} from './progressbar/progressbar.module';
 import {NgbRatingModule} from './rating/rating.module';
-import {NgbTabsetModule, NgbTabChangeEvent} from './tabset/tabset.module';
```
pkozlowski-opensource: But people might use it for type checking, no? Applies to other places as well.

maxokorokov: These are just imports, we have all the exports a bit lower in the file: https://github.com/ng-bootstrap/ng-bootstrap/blob/8dc0513fa630df22faf16fdaae26515206644b05/src/index.ts#L88-L96

PR #3072, opened 2019-03-15
test: add the Bootstrap css to unit tests

fbasso:
This is a proposal. The purpose of this is:

- Check that everything is alright with the bootstrap css applied. By the way, there were `fixture.detectChanges();` to add in some tests, which were failing with Bootstrap,
- Open some cases to unit tests, for instance for positioning where Bootstrap classes can be required,
- Beautifying the unit tests :)

### DIFF
--- a/src/karma.conf.js
+++ b/src/karma.conf.js
@@ -6,6 +6,7 @@ const browsers = process.env.TRAVIS ? ['ChromeHeadlessNoSandbox'] : ['ChromeNoEx
 module.exports = function (config) {
   config.set({
     basePath: '',
+    files: ['../node_modules/bootstrap/dist/css/bootstrap.min.css'],
     frameworks: ['jasmine', '@angular-devkit/build-angular'],
     plugins: [
       require('karma-jasmine'),
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -359,7 +359,7 @@ describe('ngb-dropdown-toggle', () => {
 
   it(`should second placement if the first one doesn't fit`, () => {
     const html = `
-      <div ngbDropdown placement="left right">
+      <div ngbDropdown placement="left-top right-top">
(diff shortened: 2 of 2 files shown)


pkozlowski-opensource:
Why? I can see the value of Bootstrap CSS in e2e tests but not in unit tests... Sound like we just need to move more to e2e tests instead of "Open some cases to unit tests, for instance for positioning where Bootstrap classes can be required,"....

fbasso:
Thanks for your comment, @pkozlowski-opensource. As I said, it's a proposal, just to discuss about it.

Right, I have no real new use case in view at the moment, the main point for me was "Check that everything is alright with the bootstrap css applied", so that ensure that upgrading the Bootstrap version has no impact on the widgets. 

I had to adapt some tests here, meaning that bootstrap style can have an impact on the unit tests.

Another thing is it's visually better, which render things easier to inspect/debug, especially when it's not our own code.

Last point, there is no impact on performance, the tests run the same speed, so why deprive the tests of a better rendering ?

src/karma.conf.js
```diff
@@ -3,38 +3,29 @@
 const reporters = process.env.TRAVIS ? ['dots'] : ['progress'];
 const browsers = process.env.TRAVIS ? ['ChromeHeadlessNoSandbox'] : ['ChromeNoExtensions'];
 
-module.exports = function (config) {
+module.exports = function(config) {
   config.set({
     basePath: '',
+    files: ['../node_modules/bootstrap/dist/css/bootstrap.min.css'],
```
maxokorokov: please leave only this line in changes, don't reformat the file unnecesary

maxokorokov:
Sill code formatting for `karma.conf.js` pending

fbasso:
@maxokorokov Done ! I also add to adapt one unit test, as the dropdown items are now vertical aligned instead of being on the same line.

PR #3074, opened 2019-03-15
fix(positioning): round translated positions

### LINKED ISSUE #3071: Please take rounded/floor values of popovers translated positions to prevent fuzzy borders. (fractional pixels artifacts)
### Bug description:
Popovers get their positions with a "translate". Their calculations often result on non-rounded (fractional) values such as "transform: translate(810.75px, 130.25px);" This makes anything that has high-contrast one-pixel wide borders (or other sharp and thin graphic elements) appear fuzzy and buggy. (some times the resulting position is very close to an integer value and it makes the result appear sharp and clear but its not consistent and depends on positions and widths of elements.) This should be vey easy to fix.

fbasso:
fix #3071


boltex:
Thank you! 🥇 and sorry again that I did not give a full bug report with plunker and all other details. You rock!

PR #3093, opened 2019-03-28
chore: deactivating clang for /demo, activating prettier for /demo

benouat:
It can help a bit to keep the formatting _sane_.

`clang` will be deactivated for all `/demo`.

For `prettier` I suggest that people configure their IDE to only activate it when it sees a config file.


benouat:
After sleeping on it for a couple weeks, I think this is not such a great idea.
We should instead move to a newer version of `clang`, which seems to be much faster, and really close to what `Prettier` would generate.

Issue #2711, opened 2018-09-13
Typeahead dropdown not closed automatically when long clicking another typeahead and hovering an option


m-da:
### Bug description:

You need to have 2 typeaheads that are set to expand on focus, like the "Open on focus" from the demo page: https://ng-bootstrap.github.io/#/components/typeahead/examples#focus

Click on the 1st one, it will expand the list of options, then click AND HOLD on the 2nd one, it will expand the list of options, move your mouse to hover any option of the 2nd one, then release the click. Both typeaheads are now open at the same time, and you can hover any option of both. Clicking anywhere will close both of them. Works with any number of typeaheads that open on focus (3 on my plunkr below).

![2018-09-13 10_33_57-plunker](https://user-images.githubusercontent.com/43234178/45476934-91667100-b740-11e8-92e9-59a47a411bc6.png)

### Link to minimally-working plunker that reproduces the issue:

http://plnkr.co/edit/x8m6PzxW1MN2eZHr9NBC

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 6.1.0

ng-bootstrap: 3.2.0

Bootstrap: 4.0.0

maxokorokov:
Sorry for letting this slip through, the workaround here would be adding

```html
<input #t="ngbTypeahead" (blur)="t.dismissPopup()" />
```

A better fix would be using `mouseup` instead of `click`. Will open a PR to fix this and perf issues caused by using `(document)` handler when there are many typeaheads on the page.

PR #3113, opened 2019-04-08
fix(dropdown): close dropdown only on dropdown-item click

### LINKED ISSUE #3063: ngbDropdown closes when scrollbars are used
### Bug description:
If you have an ngbDropdownMenu with a scrollbar and you click on the scrollbar or drag it, the dropdown closes.

https://stackblitz.com/edit/angular-khpeym

Angular: 7.2.5

ng-bootstrap: 4.0.4 but also tried 4.1.0

Bootstrap: 4.3.1

fbasso:
fix #3063

### DIFF
--- a/demo/src/app/components/dropdown/demos/form/dropdown-form.html
+++ b/demo/src/app/components/dropdown/demos/form/dropdown-form.html
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col">
-    <div ngbDropdown class="d-inline-block" autoClose="outside">
+    <div ngbDropdown class="d-inline-block">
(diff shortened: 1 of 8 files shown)


fbasso:
After a closer investigation, the difference between ng-bootstrap and bootstrap is that we manage the dropdown close with mousedown/mouseup instead of click. We did it this way to manage use cases not done with Bootstrap (see #2711).

When clicking on a scrollbar, the dom click event is not triggered, but the mousedown/mouseup is.

I didn't find a clean/easy solution yet, so for now, I just included .dropdown-item and .dropdown-divider in the list of elements allowed to close the dropdown. 

Maybe it's the safer fix for now.

PR #3129, opened 2019-04-15
chore(demo): use Snippet model for ngbd-code

ymeine:
To define snippet information all in one place for clarity, avoid the multiplication of input properties on the `ngbd-code` component, and have common processing such as a handy indentation cleaner (common issue when defining multi-line strings with template strings).

### DIFF
--- a/demo/src/app/shared/code/snippet.ts
+++ b/demo/src/app/shared/code/snippet.ts
@@ -1,0 +1,36 @@
+export interface ISnippet {
+  lang: 'html' | 'typescript' | 'css' | 'bash';
+  code: string;
+}
+
+function removeEmptyLineAtIndex(lines: string[], index: number) {
+  if (lines[index].trim().length === 0) {
+    lines.splice(index, 1);
+  }
+}
+
+function findIndentLevel(lines): number {
+  return Math.min(...lines
+    .map(line => {
+      const result = /( *)[^ ]+/g.exec(line);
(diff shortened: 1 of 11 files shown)


demo/src/app/shared/code/snippet.ts
```diff
@@ -0,0 +1,41 @@
...
+  return Math.min(...lines
+    .map(line => {
+      const result = /( *)[^ ]+/g.exec(line);
+      return result == null ? null : result[1].length;
+    })
+    .filter(value => value != null)
+  );
+}
+
+export function fixIndent(lines: string[]): string[] {
```
maxokorokov: should not be exported

demo/src/app/shared/code/snippet.ts
```diff
@@ -0,0 +1,41 @@
+export interface ISnippet {
+  lang: 'html' | 'typescript' | 'css' | 'bash';
+  code: string;
+}
+
+export interface SnippetInput extends ISnippet {
```
maxokorokov: Let's remove fix indent option completely and simply always fix it

benouat:
Build is **KO** because linting is failing due to a missing whitespace. @ymeine could you PTAL ?

👍 also to **always** fix indentation. I can't think of a use-case where we would keep the thing unindented.

Once green, I'll proceed with merge.

ymeine:
Changes done and commits squashed

PR #3138, opened 2019-04-24
Open tooltip and popover correctly from hooks

### LINKED ISSUE #3130: Adding a ngbtooltip to a ngbmodal
### Bug description:
Adding a ngbtooltip to the first focus-able element will result in an error when the modal is opened. 

Error thrown is 
```
Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'className: undefined'. Current value: 'className: tooltip show'. It seems like the view has been created after its parent and its children have been dirty checked. Has it been created in a change detection hook ?
```

[Link to stack blitz](https://stackblitz.com/edit/angular-fbghyx)

Angular:
7.0.0
ng-bootstrap:
4.1.1

maxokorokov:
Forces change detection run on `.open()` to avoid having `ExpressionHasChanged` exception when opened after change detection was run (ex. from hooks)

Fixes #3130

### DIFF
--- a/src/popover/popover.spec.ts
+++ b/src/popover/popover.spec.ts
@@ -9,7 +9,8 @@ import {
   Injectable,
   OnDestroy,
   TemplateRef,
-  ViewContainerRef
+  ViewContainerRef,
+  AfterViewInit
 } from '@angular/core';
 
 import {Key} from '../util/key';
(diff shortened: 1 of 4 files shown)


maxokorokov:
@benouat could you PTAL?

PR #3165, opened 2019-04-30
test(modal): flag TestComponent queries as static

marclaval:
In Ivy, static queries must be flagged correctly.


benouat:
Closing as landed via #3220

PR #3173, opened 2019-05-06
fix(datepicker): don't change month content if navigation is prevented

maxokorokov:
Previous attempt in navigation cancellation changed was buggy → #3044

After cancellation year and month select boxes stayed unchanged, but the actual month content changed. 

Test expectations were incomplete.

### DIFF
--- a/src/datepicker/datepicker.spec.ts
+++ b/src/datepicker/datepicker.spec.ts
@@ -540,10 +540,11 @@ describe('ngb-datepicker', () => {
   it('should prevent navigation when calling preventDefault()', () => {
     const fixture = createTestComponent(
         `<ngb-datepicker #dp [startDate]="date" (navigate)="onPreventableNavigate($event)"></ngb-datepicker>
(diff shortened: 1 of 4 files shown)


src/datepicker/datepicker.spec.ts
```diff
@@ -540,17 +540,19 @@ describe('ngb-datepicker', () => {
...
     expect(getYearSelect(fixture.nativeElement).value).toBe('2016');
+    expect(getDay(fixture.nativeElement, 0).innerText).toBe('1');
 
     const button = fixture.nativeElement.querySelector('button#btn');
     button.click();
     fixture.detectChanges();
 
     expect(getMonthSelect(fixture.nativeElement).value).toBe('8');
     expect(getYearSelect(fixture.nativeElement).value).toBe('2016');
+    expect(getDay(fixture.nativeElement, 0).innerText).toBe('1');
```
maxokorokov: This was failing with previous version, as the month was actually changed

PR #3176, opened 2019-05-06
fix(alert): project content before the closing button

### LINKED ISSUE #3171: Change the order of the alert component template content to correct screen readers output
### Bug description:
With the current content rendering order within alert component template:

```html
    <button *ngIf="dismissible" type="button" class="close" aria-label="Close" i18n-aria-label="@@ngb.alert.close"
      (click)="closeHandler()">
      <span aria-hidden="true">&times;</span>
    </button>
    <ng-content></ng-content>
```
the content of the alert when read by screen reader always starts with `Close`, followed by the `ng-template` slot content. As the users of the `ng-bootstrap` cannot swap that rendering order (as opposed to BS users), this seems to impact the user experience greatly, while it can be easily fixed:

```html
    <ng-content></ng-content>
    <button *ngIf="dismissible" type="button" class="close" aria-label="Close" i18n-aria-label="@@ngb.alert.close"
      (click)="closeHandler()">
      <span aria-hidden="true">&times;</span>
    </button>
```

https://ng-bootstrap.github.io/#/components/alert/examples

Angular:
n/a
ng-bootstrap:
4.1.2
Bootstrap:
n/a

maxokorokov:
Fixes #3171

@benouat could you PTAL / merge if ok ?

### DIFF
--- a/src/alert/alert.spec.ts
+++ b/src/alert/alert.spec.ts
@@ -112,6 +112,14 @@ describe('ngb-alert', () => {
     expect(alertEl.textContent).toContain('Cool!');
   });
 
+  it('should project content before the closing button for a11y/screen readers', () => {
+    const fixture = createTestComponent('<ngb-alert [dismissible]="true"><span>Cool!</span></ngb-alert>');
+    const alertEl = getAlertElement(fixture.nativeElement);
+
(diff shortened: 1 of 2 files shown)


maxokorokov:
@benouat, thanks!

PR #3196, opened 2019-05-17
refactor(datepicker): move <select> value binding to <option> selected

marclaval:
Another approach to solve the issue from #3163, but without using lifecycle hooks.
This one works in Ivy too.


marclaval:
This approach does not work in all cases in ie10 and IE11.
See https://travis-ci.org/ng-bootstrap/ng-bootstrap/builds/533723056#L844

maxokorokov:
Thanks, closing in favour of #3163

Issue #3199, opened 2019-05-19
_ngZone.onStable.subscribe


cwayfinder:
### Bug description:
There are a few components in this library with a line
```
_ngZone.onStable.subscribe(() => this._updatePopupPosition());
``` 
in the constructor (dropdown, tooltip, popover, typeahead, datepicker).

`_ngZone.onStable` is fired when the microtask queue is empty and Angular does not plan to enqueue any more.
In my application, this happens too many times and too frequent and leads to freezes in Safari.

**Let's think about a more sophisticated approach to reposition a popup to optimise performance.**

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 7.2.15
ng-bootstrap: 4.1.3

cwayfinder:
I found how the same thing is done in Angular Material https://github.com/angular/components/blob/master/src/cdk/overlay/overlay-ref.ts#L136
They use `take(1)`.

benouat:
Hey @cwayfinder !

Would you mind providing a reproducible example via Stackblitz that demonstrate the performance issue ?

cwayfinder:
Hey @benouat, I added an example. It's quite easy to reproduce.

maxokorokov:
@cwayfinder I think Material has a dedicated `PositionStrategy` for repositioning (not 100% sure), what you're referencing is only for initial positioning. We have a more naive, but generic implementation at the moment, exactly as you're saying.

Firstly, regardless of this issue, I guess, you shouldn't be triggering change detection that often.

#3206 might improve things a little in your case for Safari and Firefox, I've seen quite a lot of improvement.

To position the popup we're trying things from `[placement]` one-by-one, something like:

```
for each placement 'p' in [top, left, right, bottom, etc] {
   1. apply appropriate css class for 'p' to the popup
   2. calculate position, apply 'translate'
   3. check positioned popup fits in the viewport ← this forces style recalculation (and might trigger layout)
   4. stop if it fits
}
```

I'm not aware of another way of checking positioning at the moment. Also in my simple tests repositioning is pretty fast, and I don't see this as a bottleneck at the moment. It never takes longer than a millisecond, **unless you're trying many positions before you finding one that fits and doing style recalculation several times**

---

So, could you please provide a performance timeline from Chrome as JSON for your use case (if possible to share for your application) to see the effects of positining ? If not at least a screenshot would be helpful.

---

Ideally we should allow customizing it:
- position only ONCE when opening
[...]

PR #3206, opened 2019-05-23
fix(positioning): put the popup in a separate layer

maxokorokov:
Without forcing the popup in a separate layer, Safari and Firefox are doing a lot more work than necessary.

Example: trying to reposition datepicker 3 times, before finding an optimal position.

Quite significant changes in Safari → removes full page layout and compositing.

Safari Before:

Screen Shot 2019-05-23 at 15 11 59

Safari After:

Screen Shot 2019-05-23 at 15 13 32

In Firefox it removes unnecessary paints as well.

---

Concerning the `The translate3d/gpu acceleration render a blurry text on chrome, the next line is commented until a browser fix` → we're not using `translate3d`, but `translate`. And it looks fine in Chrome.

Related to #3199

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -220,8 +220,7 @@ export function positionElements(
   style.position = 'absolute';
   style.top = '0';
   style.left = '0';
-  // The translate3d/gpu acceleration render a blurry text on chrome, the next line is commented until a browser fix
-  // style['will-change'] = 'transform';
+  style['will-change'] = 'transform';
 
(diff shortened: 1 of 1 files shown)


fbasso:
Unfortunately, the text is blurry on Chrome/Windows with will-change activated:

Tooltip blurry

maxokorokov:
Ok, so we've figured out that it happens only on Win10 with some font zoom levels in Chrome. Same issue in Bootstrap itself, so will merge.

PR #3214, opened 2019-05-29
fix(positioning): cache management

### LINKED ISSUE #3199: _ngZone.onStable.subscribe

fbasso:
The purpose of this PR is to increase performance of re-positioning popup by caching the popup dimension values.

It can partially fixed #3199 

This is not intended to be straightly merged, as we need to investigate the implications. For example when the popup content changes while it's open (for example, the typeahead).

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -525,6 +525,8 @@ export class NgbInputDatepicker implements OnChanges,
       throw new Error('ngbDatepicker could not find element declared in [positionTarget] to position against.');
     }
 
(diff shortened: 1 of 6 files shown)


fbasso:

PR #3126, opened 2019-04-12
feat(datepicker): expose validators minDate, maxDate, invalidDate

03byron:
Follow up for #3115.

PR #3219, opened 2019-06-03
Revert "feat(datepicker): expose validators minDate, maxDate, invalid Date (#3126)"

### LINKED ISSUE #3215: regression: datepicker marked invalid if custom adapter is used, since 4.2.0
### Bug description:

When a custom adapter is set on a datepicker popup, the form control is marked invalid by the `invalidDate` validator.
My debugging showed that every time we pick a date, the `calendar.isValid()`method is called several times with the NgbDate, but ends up being called with the control value, which is not an NgbDate, but is the adapted value, which thus fails the validation.

https://stackblitz.com/edit/angular-c54t21?file=app%2Fdatepicker-adapter.html

This is just a trimmed down version of the "custom adapter" demo example. Select a date, and you'll see that it's always marked invalid. 

Angular: 7.x, 8.0

ng-bootstrap: 4.2.0

Bootstrap: 4.x

maxokorokov:
This reverts commit 509bba3d486b001f51eb7eddaad0123d8378663f.

Closes #3215 

cc @03byron


meness:
This really fixes #3215. Thanks

PR #3225, opened 2019-06-07
feat(carousel): stay paused after calling `pause()` until `cycle()` is called

### LINKED ISSUE #3188: After setting on pause and clicking on arrows navigation, the cycle restart Carousel
Hello guys,

I created and defined a function to set on pause or play the carousel.
When setting on pause and clicking on one of the arrows navigation, the cycle restated by it own.

Here is a link to a demo, i fork one of your [demo](https://stackblitz.com/edit/angular-sgw6iw).

Angular: 7.0.0

ng-bootstrap: 4.1.2

Bootstrap: 4.2.3

divdavem:
After setting the carousel on pause through its `pause()` method and clicking on navigation arrows, the carousel no longer automatically restarts. Calling the `cycle()` method is needed to undo the effect of the `pause()` method.
When `pauseOnHover` is true, hovering on the carousel no longer uses the `pause()` and `cycle()` methods. The hover status is stored separately, so that calling `pause()` and removing the mouse from the carousel no longer restarts it automatically.

The slide event now has a "source" property allowing to know what triggered the event: "timer", "arrowLeft", "arrowRight" or "indicator".
The "pause" boolean status is now also included in the slide event.

Closes #3188

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -10,19 +10,19 @@ import {
   Inject,
   Input,
   NgZone,
-  OnChanges,
   OnDestroy,
   Output,
   PLATFORM_ID,
   QueryList,
-  TemplateRef
+  TemplateRef,
+  HostListener
 } from '@angular/core';
 import {isPlatformBrowser} from '@angular/common';
 
 import {NgbCarouselConfig} from './carousel-config';
 
-import {merge, Subject, timer} from 'rxjs';
-import {filter, map, switchMap, takeUntil} from 'rxjs/operators';
+import {Subject, timer, BehaviorSubject, combineLatest, NEVER} from 'rxjs';
+import {startWith, map, switchMap, takeUntil, distinctUntilChanged} from 'rxjs/operators';
 
 let nextId = 0;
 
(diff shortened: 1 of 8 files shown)


src/carousel/carousel.ts
```diff
@@ -147,18 +165,39 @@ export class NgbCarousel implements AfterContentChecked,
     this.showNavigationIndicators = config.showNavigationIndicators;
   }
 
+  @HostListener('mouseenter')
```
benouat: Could you please move back these 2 `@HostListener` to the decorator's metadata ? When there is only one statement in the function, we usually put them directly up there.
→ `'(mouseenter)': '_mouseHover$.next(true)'`

src/carousel/carousel.ts
```diff
@@ -147,18 +165,39 @@ export class NgbCarousel implements AfterContentChecked,
...
+  mouseEnter() {
+    this._mouseHover$.next(true);
+  }
+
+  @HostListener('mouseleave')
```
benouat: → `'(mouseleave)': '_mouseHover$.next(false)'`

src/carousel/carousel.ts
```diff
@@ -267,6 +306,14 @@ export interface NgbSlideEvent {
    * Possible values are `'left' | 'right'`.
    */
   direction: NgbSlideEventDirection;
+
+  /**
+   * Whether the pause() method was called (and no cycle() call was done afterwards).
+   */
+  pause: boolean;
```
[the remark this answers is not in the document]
benouat: As @ymeine said. 👍 for renaming to `paused`

src/carousel/carousel.ts
```diff
@@ -277,4 +324,11 @@ export enum NgbSlideEventDirection {
   RIGHT = <any>'right'
 }
 
+export enum NgbSlideEventSource {
+  TIMER = 'timer',
```
benouat: Do we need to add `<any>` in here ? Such as in `NgbSlideEventDirection`

src/carousel/carousel.ts
```diff
@@ -267,6 +306,14 @@ export interface NgbSlideEvent {
    * Possible values are `'left' | 'right'`.
    */
   direction: NgbSlideEventDirection;
+
+  /**
+   * Whether the pause() method was called (and no cycle() call was done afterwards).
+   */
+  pause: boolean;
+
+  /** Source triggering the slide change event */
```
benouat: I would make the comment a bit more explicative. It's our only way to document `NgbSlideEventSource` as we don't take into account enum in the doc generator.

divdavem:
@ymeine and @benouat Thank you for your review.

I have renamed the `pause` property to `paused` in `NgbSlideEvent`.

I have also added the list of possible values in the documentation for the `source` property (as it is done for the `direction` property).

I did not move the content of the `mouseEnter` and `mouseLeave` functions to the `host` property because it would not compile properly in AOT mode because `_mouseHover$` is private.

I did not add `<any>` in the `NgbSlideEventSource` enum because I don't know why it was done for the other enum in the file. For consistency, as you suggested @benouat, it would probably be better to remove `<any>` from the `NgbSlideEventDirection` enum in another PR.

PR #3228, opened 2019-06-08
docs: update installation section

peterblazejewicz:
This PR:

- adds support for `json` PrismJS language support in highlighter service used in generated content
- updates content of the installation section with the content already used on the landing page

Screenshot 2019-06-08 at 22 46 46

Thanks!

### DIFF
--- a/demo/src/app/pages/getting-started/getting-started.component.html
+++ b/demo/src/app/pages/getting-started/getting-started.component.html
@@ -60,6 +60,14 @@
 
   <ngbd-page-header title="Installation" fragment="installation"></ngbd-page-header>
 
+  <p>You need to have an Angular project with the supported Angular version. We strongly recommend using <a href="https://cli.angular.io" rel="nofollow" target="_blank">Angular CLI</a> for this.</p>
+
+  <p>You also need to add Bootstrap 4 CSS to your application by using your preferred way (it really depends on the setup you're using). Ex. for Angular CLI you can <a href="https://www.npmjs.com/package/bootstrap" rel="nofollow" target="_blank">get Bootstrap from npm</a> and update your <code>angular.json</code> with something like:</p>
+
+  <ngbd-code [snippet]="bootstrapInstall"></ngbd-code>
+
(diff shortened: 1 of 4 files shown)


demo/src/app/pages/getting-started/getting-started.component.html
```diff
@@ -60,6 +60,14 @@
 
   <ngbd-page-header title="Installation" fragment="installation"></ngbd-page-header>
 
+  <p>You need to have an Angular project with the supported Angular version. We strongly recommend using <a href="https://cli.angular.io" rel="nofollow" target="_blank">Angular CLI</a> for this.</p>
+
+  <p>You also need to add Bootstrap 4 CSS to your application by using your preferred way (it really depends on the setup you're using). Ex. for Angular CLI you can <a href="https://www.npmjs.com/package/bootstrap" rel="nofollow" target="_blank">get Bootstrap from npm</a> and update your <code>angular.json</code> with something like:</p>
+
+  <ngbd-code [snippet]="bootstrapInstall"></ngbd-code>
+
+  <p>Please note that you need only CSS and <strong>should not</strong> add other JavaScript dependencies like <code>bootstrap.js</code>, <code>jQuery</code> or <code>popper.js</code> as ng-bootstrap's goal is to completely replace them.</p>
```
benouat: Could you replace this `<p>` section by actually moving the `<ngb-alert>` one already present at the top of the file?
It sounds confusing to have 2 different disclaimers in the same files for the same thing with different wordings.

We could re-write a bit the content also, something like:

```html
<ngb-alert type="warning" [dismissible]="false">
    <p>Should I add <b>bootstrap.js</b> or <b>bootstrap.min.js</b> to my project?</p>

    No, the goal of ng-bootstrap is to <em>completely replace</em> JavaScript implementation for components. You <strong>should not</strong>  include other dependencies like <code>jQuery</code> or <code>popper.js</code>. It is not necessary and might interfere with ng-bootstrap code. You <strong>only</strong> need Bootstrap CSS
</ngb-alert>
```

benouat:
@peterblazejewicz Thanks for this PR ! Would you mind having a look at my comment left in the review?

PR #3220, opened 2019-06-03
CI with Angular 8

maxokorokov:
The goal of this PR is to make CI pass with Angular 8 (without ivy)

- Update dependencies: Angular 8.0, Bootstrap 4.3.1, latest CLI deps
- add `{static: true/false}` for queries
- fix flaky tests → execution order changed with latest dependencies, so some tests were leaving DOM nodes and some tests were using ids like `popover-1` depending on test execution order
- fix TS compilation issues

Please see separate commits and don't squash when merging

Fixes #3213

PR #3234, opened 2019-06-13
test: positioning check relative to its host

fbasso:
The purpose is to render the test more stable. The position test with the target element was done in absolute positions. It is now done with offsets compared to the host element.

It should help for #3220


fbasso:

PR #3236, opened 2019-06-14
feat(timepicker): activate type="number" for inputs

### LINKED ISSUE #2334: TimePicker - input type number
Hi,

As timepicker should never accept characters, if I'm not mistaken, why not set the input being used to [type="number"], so we could use the advantage of browsers to block non-numerical characters input?

Is it a feasible request?

fbasso:
fix #2334 

input type as number is more appropriate for the timepicker, although the behavior is different depending on the browser:

- Chrome filter characters to only accept numbers,
- FF does not filter characters, but put a red border on blur if the value is not valid
- IE set the field empty on blur if the value is not valid
- Edge do nothing

The inner arrows have been removed by css, as the timepicker already provides

Even with these discrepancies, the big advantage is to get the numeric keypad on mobile. For this, the property "inputmode" is more appropriate (as we don't need the arrows), but it's not fully supported on mobile: https://caniuse.com/#feat=input-inputmode

In this PR, the tests on navigation had to be tweak a little, as it's not possible to use selectionStart and selectionEnd on inputs with type="number". I didn't find another way to change the type before testing the navigation.

### DIFF
--- a/e2e-app/src/app/timepicker/navigation/timepicker-navigation.e2e-spec.ts
+++ b/e2e-app/src/app/timepicker/navigation/timepicker-navigation.e2e-spec.ts
@@ -1,4 +1,4 @@
-import {Key, ElementFinder} from 'protractor';
+import {Key, ElementFinder, browser} from 'protractor';
 
 import {openUrl, expectFocused, sendKey, getCaretPosition} from '../../tools.po';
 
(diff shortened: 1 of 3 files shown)


fbasso:
However, there are lots of issues declared with type="number" declared on internet. This type seems to be not stable through browser implementations. 

So I don't know if it's worth to take the risk to change it, just to get the numeric keypad on mobile.

benouat:
I don't know what to do with this one...

It's true that for mobile device keyboard type _pre-selection_ it would be perfect !

But ...

the nasty side-effect is that it implies to many _non-working_ edge cases on the desktop. All browsers vendors have pretty much implemented `input[type=number]` differently in a non consistent way.

I would not go down this road....

---

What if we provide a directive `input[ngbMobileNumberKeyboard]` that hooks itself on the `focus/focusin` to dynamically swap the type just before the keyboard is opened, and change it back on next tick (only on mobile platform) ? Would it work to have the proper keyboard type ?

benouat:
Let's close that one in favour of #3247.

Issue #3241, opened 2019-06-18
Popover classes reapplied on ngb-popover-window after timer tick (_ngZone.onStable)


frankrafael:
### Bug description:
I have an Angular application that has several timers running (`setInterval`). Whenever these timers are done the `ngZone.onStable()` subscription is triggered. NgbPopover listens to this event stream and tries to reposition the open Popovers, this causes the Popover classes to be removed and reapplied for each `ngZone.onStable()` tick, which is causing some performance issues and making the popovers move up and down (not always reproducible). Please see the StackBlitz for further info.

Possible solutions I see so far:
1. We don't change the classes if the new "repositioned" ones are the same as the current ones
2. Maybe we can look for a different trigger to reposition the popover? Does it need to be the `ngZone.onStable()`?

### Link to minimally-working StackBlitz that reproduces the issue:

https://angular-2xdqnt.stackblitz.io/

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 6.1.7

ng-bootstrap: 4.2.1

Bootstrap: 4.1.2

fbasso:
Sounds duplicate of #3199

Could you please tell more about you're real use-case ? What is launching this setInterval ?

frankrafael:
It think this is indeed a duplicate.

My app is a collaboration tool, I have to pull small chunks of data from the backend from time to time. E.g application version, document steps, and other proactive status controls. All the intervals are running from 1 to 30 seconds of interval, however, since there are around 8 intervals running they overlap and make the popover bounce.

PR #3245, opened 2019-06-20
fix(positioning): window resize listener and peformance optimisation

### LINKED ISSUE #3382: ngbpopover floats on page size change
When using container="body" , open the popover and change the screen size then popover don't stick to the link which has opened the popover.

This behavior can be replicated on the https://ngbootstrap.github.io/#/components/popover/examples as well.

I have tested this in chrome.
steps to replicate:

click to open
change the screen size.
popover will start to float on the page.

fbasso:
The purpose of this PR is a proposal for fixing #3199 and fixing #3241 

 - It's a partial fix as it comes complementary to #3018
 - ngZone.onStable is no longer use (the root cause of performance issue)
 - The 'position' method in tooltip and popover has been publicly  open in case of content change, in charge of the app to call it when needed.
 - window resize is listen to reposition the tooltip and popover

fix #3382

### DIFF
--- a/src/util/resize.ts
+++ b/src/util/resize.ts
@@ -1,0 +1,10 @@
+import {NgZone} from '@angular/core';
+import {fromEvent, Observable} from 'rxjs';
+import {takeUntil, throttleTime} from 'rxjs/operators';
+
+
+export function ngbWindowResize(zone: NgZone, callback: () => void, unsubscribe$: Observable<any>) {
+  zone.runOutsideAngular(() => {
+    fromEvent(window, 'resize').pipe(takeUntil(unsubscribe$), throttleTime(10)).subscribe(() => callback());
+  });
+}
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -26,10 +26,12 @@ import {DOCUMENT} from '@angular/common';
 
 import {listenToTriggers} from '../util/triggers';
 import {ngbAutoClose} from '../util/autoclose';
+import {ngbWindowResize} from '../util/resize';
 import {positionElements, PlacementArray} from '../util/positioning';
 import {PopupService} from '../util/popup';
 
 import {NgbPopoverConfig} from './popover-config';
+import {take} from 'rxjs/operators';
 
 let nextId = 0;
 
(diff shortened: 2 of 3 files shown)


tan9:
I applied the same technique @fbasso proposed to `typeahed` and `datepicker` at https://github.com/fbasso/ng-bootstrap/pull/2 . And gain a great performance boost on my application with tons of dependencies and polyfills under IE 11.

src/util/resize.ts
```diff
@@ -0,0 +1,10 @@
+import {NgZone} from '@angular/core';
+import {fromEvent, Observable} from 'rxjs';
+import {takeUntil, throttleTime} from 'rxjs/operators';
+
+
+export function ngbWindowResize(zone: NgZone, callback: () => void, unsubscribe$: Observable<any>) {
+  zone.runOutsideAngular(() => {
+    fromEvent(window, 'resize').pipe(takeUntil(unsubscribe$), throttleTime(10)).subscribe(() => callback());
+  });
+}
```
peterblazejewicz: 1) can we take advantage of the passive subscribers here, detecting browser support, and doing subscription based on that detection?
2) probably 'orientationchange' should be also involved into that positioning strategy for overlays:
e.g. (just sample, not the best code):
```diff
diff --git a/src/util/resize.ts b/src/util/resize.ts
index 59fd14b..d58e352 100644
--- a/src/util/resize.ts
+++ b/src/util/resize.ts
@@ -1,10 +1,33 @@
 import {NgZone} from '@angular/core';
-import {fromEvent, Observable} from 'rxjs';
+import {merge, fromEvent, Observable} from 'rxjs';
 import {takeUntil, throttleTime} from 'rxjs/operators';
 
+const passiveSupported: boolean = (() => {
+  let passiveSupported = false;
+  try {
+    const options = {
+      get passive() {
+        passiveSupported = true;
+        return true;
+      }
+    };
+    const listener = () => {};
+    window.addEventListener("test", listener, options);
+    window.removeEventListener("test", listener);
+  } catch (err) {
+    passiveSupported = false;
+  }
+  return passiveSupported;
+})();
 
 export function ngbWindowResize(zone: NgZone, callback: () => void, unsubscribe$: Observable<any>) {
   zone.runOutsideAngular(() => {
-    fromEvent(window, 'resize').pipe(takeUntil(unsubscribe$), throttleTime(10)).subscribe(() => callback());
+    const resizeObservable =
+        passiveSupported ? fromEvent(window, 'resize', {passive: true}) : fromEvent(window, 'resize');
+    const orientationChangeObservable = passiveSupported ? fromEvent(window, 'orientationchange', {passive: true}) :
+                                                           fromEvent(window, 'orientationchange');
+    merge(resizeObservable, orientationChangeObservable)
+        .pipe(takeUntil(unsubscribe$), throttleTime(10))
+        .subscribe(() => callback());
   });
 }
```

fbasso:

PR #3247, opened 2019-06-21
feat(timepicker): input filter to accept only numbers

### LINKED ISSUE #2334: TimePicker - input type number

fbasso:
Finally a better proposal than #3236 to fix #2334

- The timepicker inputs are filtered (so a better approach than type="number" which supported differently depending on the browsers), 
- inputmode="numeric" is used to open the numeric keypad on mobile device, which is basically the purpose of this attribute, for the supported browsers (https://caniuse.com/#feat=input-inputmode : Chrome, Opera, iOs safari, Android, Samsung Internet)

### DIFF
--- a/src/timepicker/timepicker.ts
+++ b/src/timepicker/timepicker.ts
@@ -15,6 +15,8 @@ import {NgbTimepickerConfig} from './timepicker-config';
 import {NgbTimeAdapter} from './ngb-time-adapter';
 import {NgbTimepickerI18n} from './timepicker-i18n';
 
+const timepickerFilterRegExp = /[^0-9]/g;
+
 const NGB_TIMEPICKER_VALUE_ACCESSOR = {
   provide: NG_VALUE_ACCESSOR,
   useExisting: forwardRef(() => NgbTimepicker),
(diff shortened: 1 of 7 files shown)


src/timepicker/timepicker.ts
```diff
@@ -269,6 +273,32 @@ export class NgbTimepicker implements ControlValueAccessor,
...
 
+  _filter(e: KeyboardEvent) {
+    // tslint:disable-next-line
+    const keyCode = e.keyCode;
+    const ctrlKey = e.ctrlKey;
+    const metaKey = e.metaKey;
+    const shiftKey = e.shiftKey;
+    if (
+        // Allow: Delete, Backspace, Tab, Escape, Enter
+        [46, 8, 9, 27, 13].indexOf(keyCode) !== -1 || (keyCode === 65 && ctrlKey === true) ||  // Allow: Ctrl+A
```
IAfanasovMob: would be nice split lines
```
[46, 8, 9, 27, 13].indexOf(keyCode) !== -1 || // Allow: Delete, Backspace, Tab, Escape, Enter
(keyCode === 65 && ctrlKey === true) ||  // Allow: Ctrl+A
```

benouat:
@IAfanasovMob Thanks for the review. It is much appreciated here!

src/timepicker/timepicker.ts
```diff
@@ -269,6 +273,31 @@ export class NgbTimepicker implements ControlValueAccessor,
...
   }
 
+  _filter(e: KeyboardEvent) {
+    // tslint:disable-next-line
+    const keyCode = e.keyCode;
+    const ctrlKey = e.ctrlKey;
+    const metaKey = e.metaKey;
+    const shiftKey = e.shiftKey;
+    if ([46, 8, 9, 27, 13].indexOf(keyCode) !== -1 ||  // Allow: Delete, Backspace, Tab, Escape, Enter
+        (keyCode === 65 && ctrlKey === true) ||        // Allow: Ctrl+A
```
msosa: not sure if this is any better but can also do something like
```typescript
([65, 67, 86, 88].indexOf(keyCode) !== -1  && (ctrlKey === true || metaKey === true)) ||    // Allow: Ctrl+A
```

fbasso: Thanks @msosa, yes you're right. I took this code somewhere on internet, but never had a closer look at the "if" part (it was the more complete I found, taking account of such cases). I'm going to change it.

fbasso:
Done !

PR #3256, opened 2019-06-26
fix(datepicker): NgbCalendarIslamicUmalqura.fromGregorian with date in the afternoon

### LINKED ISSUE #3237: NgbCalendarIslamicUmalqura's getToday() returns one day after the original date
Today Hijri date is 11 Shawal 1440 and Gregorian date is 14 June 2019

image

When I select Today in the calendar its display wrong date. today should be 11 Shawal 1440 instead of 12 Shawal 1440. This issue produces on the live official documentation https://ng-bootstrap.github.io/#/components/datepicker/calendars and also on the local machine. 
 
image

The problem is in this method `fromGregorian(new Date)`. Its return wrong date. When I invoke `this.calendar.getToday()` then it returns one day after the original date. I have tested it in chrome, IE and opera and faced the same issue. Maybe this is due to calculations in this method `fromGregorian(jsDate)`

**Reason of Issue**
Suppose today is 14 June 2019.  When I pass date like this `fromGregorian(new Date('Sat Jun 14 2019 11:00:00'))` then it's converting as expected But when I pass date like this `fromGregorian(new Date('Sat Jun 14 2019 13:00:00'))` then it's return with adding one day. It's mean this method change date after 12:00 pm (12 noon). 

Bootstrap Version 4.1.3
Angular Version 6.0.6

divdavem:
`NgbCalendarIslamicUmalqura.fromGregorian` returned a date one day after the correct date for dates containing a time in the afternoon.

Fixes #3237

### DIFF
--- a/src/datepicker/hijri/ngb-calendar-islamic-umalqura.spec.ts
+++ b/src/datepicker/hijri/ngb-calendar-islamic-umalqura.spec.ts
@@ -868,6 +868,16 @@ describe('ngb-calendar-islamic-umalqura', () => {
             .toBeTruthy(`Gregorian ${gDate} should be Hijri ${iDate.year}-${iDate.month}-${iDate.day}`);
       });
     });
+
+    it('should convert correctly from Gregorian to Hijri with time 23:59:59.999', () => {
+      DATE_TABLE.forEach(element => {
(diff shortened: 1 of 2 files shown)


divdavem:
@benouat Thank you for your review!

PR #3263, opened 2019-06-28
fix(dropdown): allow template directive on the menu

### LINKED ISSUE #3246: Can't use a template directive as a dropdown menu
### Bug description:

When using a template directive on a ngbDropdownMenu, the menu is not placed correctly on first click. When clicking again, the menu is at the right place.

This is due to the fact that `offsetParent` is looking for a parent with a position attribute. This is normally solved by `_applyPlacementClasses` in `dropdown.ts` with the `dropdown` class. 

If you try to add a template to it, ngOnInit is called before the full template rendering. This causes  `NgbDropdown._menu` to be null and `dropdown` class is never applied.

A possible solution (not tested) could be ngAfterViewInit()

https://stackblitz.com/edit/problem-template-dropdown

Angular: 7

ng-bootstrap: 4.2.1

Bootstrap: 4

fbasso:
fix #3246

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -9,8 +9,8 @@ import {
   Inject,
   Input,
   NgZone,
+  AfterContentInit,
   OnDestroy,
-  OnInit,
   Output,
   QueryList,
   Renderer2,
@@ -19,6 +19,7 @@ import {
 } from '@angular/core';
 import {DOCUMENT} from '@angular/common';
 import {Subject, Subscription} from 'rxjs';
+import {take} from 'rxjs/operators';
 
 import {Placement, PlacementArray, positionElements} from '../util/positioning';
 import {ngbAutoClose} from '../util/autoclose';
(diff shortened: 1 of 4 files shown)


src/dropdown/dropdown.ts
```diff
@@ -240,9 +242,10 @@ export class NgbDropdown implements OnInit, OnDestroy {
   }
 
   private _setCloseHandlers() {
+    const anchor = this._anchor;
     ngbAutoClose(
         this._ngZone, this._document, this.autoClose, () => this.close(), this._closed$,
-        this._menu ? [this._menuElement.nativeElement] : [], this._anchor ? [this._anchor.getNativeElement()] : [],
+        this._menu ? [this._menuElement.nativeElement] : [], anchor ? [anchor.getNativeElement()] : [],
```
benouat: This refactor is not strictly related to the underlying issue.

src/dropdown/dropdown.ts
```diff
@@ -341,14 +344,16 @@ export class NgbDropdown implements OnInit, OnDestroy {
   }
 
   private _getMenuElements(): HTMLElement[] {
-    if (this._menu == null) {
+    const menu = this._menu;
```
benouat: Same here

src/dropdown/dropdown.ts
```diff
@@ -341,14 +344,16 @@ export class NgbDropdown implements OnInit, OnDestroy {
...
+    if (menu == null) {
       return [];
     }
-    return this._menu.menuItems.filter(item => !item.disabled).map(item => item.elementRef.nativeElement);
+    return menu.menuItems.filter(item => !item.disabled).map(item => item.elementRef.nativeElement);
   }
 
   private _positionMenu() {
-    if (this.isOpen() && this._menu) {
+    const menu = this._menu;
```
benouat: Same here

src/dropdown/dropdown.ts
```diff
@@ -364,9 +369,10 @@ export class NgbDropdown implements OnInit, OnDestroy {
 
   private _resetContainer() {
     const renderer = this._renderer;
-    if (this._menuElement) {
+    const menuElement = this._menuElement;
```
benouat: Same here.

src/dropdown/dropdown.ts
```diff
@@ -396,7 +402,8 @@ export class NgbDropdown implements OnInit, OnDestroy {
   }
 
   private _applyPlacementClasses(placement?: Placement) {
-    if (this._menu) {
+    const menu = this._menu;
```
benouat: Same here.

benouat:
@fbasso could you please rebase ?

Issue #3030, opened 2019-02-18
ngbDropdownMenu wrong position on small device in navbar


mebibou:
### Bug description:

When using a `ngbDropdownMenu` inside a `navbar`, opening the menu on a large device will make it render as `position: absolute` and will display below the opening button. But when reducing the screen and the dropdown should appear in the hamburger menu, when opening it is still displayed using `position: absolute` which makes the menu look weird (not taking the full width).
Note: this issue does NOT happen on `4.0.2` but does on `4.0.3`. On `4.0.2` the `position` is not overridden by the library and it works. `4.0.3` is somehow adding `position: absolute` which is completely unnecessary and breaks the layout

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-gwm69f

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 7.0.0

ng-bootstrap: 4.0.3

Bootstrap: 4.2.0

maxkorz:
I think `ng-bootstrap` should behave like the original Bootstrap.

In [Bootstrap Dropdown](https://getbootstrap.com/docs/4.3/components/dropdowns/#single-button) JS adds `position` and `transform` properties to `.dropdown-menu`. But when [Dropdown is in a Navbar](https://getbootstrap.com/docs/4.3/components/navbar/#supported-content), it does not add anything.

![image](https://user-images.githubusercontent.com/25089646/53085841-d58e4680-3514-11e9-8dea-f62dde36c0a3.png)

![image](https://user-images.githubusercontent.com/25089646/53085896-f8b8f600-3514-11e9-9958-ba8c6c9db8d8.png)

ahmadmu:
I'm not sure it's related, but I'm currently having an issue where in dev mode the ngbDropdownMenu is positioned using `top: 38px`, and in prod mode it's positioned using `transform: translate(0, 38px)`, which is breaking my animations in production. Does someone know what might be the cause?

PR #3268, opened 2019-06-29
demo: add demo for dropdown positioning in navbar

jnizet:
also improve the documentation of the display input.

The rationale for this PR is that the change regarding the custom positioning of dropdowns in the navbar is documented in a location that isn't easy to find. And this change hasn't been marked as a breaking change although it should have been IMHO. I was myself hit by a regression regarding this change, and [I'm not alone](https://stackoverflow.com/q/56818511/571407).


demo/src/app/components/dropdown/demos/navbar/dropdown-navbar.html
```diff
@@ -0,0 +1,47 @@
+<p>
+  Custom placement of a dropdown in a navbar is only possible if the <code>display</code> property is explicitly
```
benouat: Even though the demo title is explicitly `Dynamic positioning in a navbar` I would start this documentation section by saying something like:

> By design dropdowns are always positioned dynamically via [placement](/positioning#dropdown) expect when used inside navbar elements. To gracefully display them properly on small screens, they are rendered in the html as block elements.

Then I would continue with your `Custom placement of a dropdown ...` and finally I would remove ~`Beware however that this breaks the positioning of the dropdown when the navbar is expanded on small displays`~

jnizet:
@benouat thanks for the review (and sorry for noticing late).
I have added the preamble you suggested, and also made some tiny improvements, in a last commit

fbasso:
Thanks for your PR @jnizet. You're right, the dropdown inside a navbar is a special case which needs clarifications.

This is my 2 cents on it. We introduced the display property in the dropdown menu to fix the issue #3030. We noticed at the time that we missed the Bootstrap specs on it, as on their side they don't use popper in this case : 

https://getbootstrap.com/docs/4.3/components/dropdowns/ : "Popper.js isn't used to position dropdowns in navbars though as dynamic positioning isn’t required."

The PR #3098 has been done to stick the Bootstrap specs (maybe we would have to merge it in a major release, as it can be considered as a breaking change, even if it was a fix regarding Bootstrap specs).

In my opinion, what is missing in the doc you provided here is the fact that "dynamic positioning isn't required", in most the cases. The main problem met by people today is that the menu align to the right have their dropdown menu going out of the screen. 

The first thing to consider is to think as Bootstrap : adding class="dropdown-menu-right" on their menu should be enough to fix the alignment (https://getbootstrap.com/docs/4.3/components/dropdowns/#responsive-alignment). Unfortunately, and to add to the confusion (positioning is hard ! :) ), this technique doesn't work without the PR #3270, because the x-placement attribute is set in static positioning (there are css rules this this attribute), so we have to merge it before this one.

The advantage is to keep a better compatibility with smaller screens, thanks to responsive alignment of Bootstrap. The dynamic positioning stay, of course, an option to explain in the doc.

jnizet:
@fbasso if it's indeed possible to have a right-aligned menu on large screen size, while still having the correct behavior on small screen sizes when the navbar is collapsed, then it's great news, and I should definitely change this PR to show how to use that CSS trick.

I'll try rebasing on #3270 and change my PR.

jnizet:
@benouat @fbasso I've added a third commit explaining how to right-align a dropdown in a navbar, and rebased it on #3270 which was a prerequisite for this to work. And it indeed works :tada:
Can you please review the changes in the last commit?

If it's OK, don't hesitate to ping me when #3270 is merged so that I can rebase this on master and make it easier for you to merge it.

benouat:
@jnizet #3270 is merged. Please do proceed with rebase thanks

jnizet:
@benouat rebase done. I squashed the 3 commits into a single one, too.

fbasso:
Thanks @jnizet, perfect for me !

PR #3272, opened 2019-07-03
fix(demo): direct link to specific carousel sample not working

divdavem:
The carousel samples in the demo site contain images that are loaded asynchronously. But as their size is not reserved beforehand, they shift the remaining content when loaded. As a consequence, the scrolling position that was set correctly to target a specific sample before the images are loaded is no longer correct after the images are loaded.

This PR reserves the space for the images by using the trick documented in the following link to also keep the responsive behavior:
https://itnext.io/how-to-stop-content-jumping-when-images-load-7c915e47f576

### DIFF
--- a/demo/src/style/demos.css
+++ b/demo/src/style/demos.css
@@ -41,3 +41,17 @@ ngbd-table-filtering span.ngb-highlight {
 ngbd-table-complete span.ngb-highlight {
   background-color: yellow;
 }
+
+ngb-carousel .picsum-img-wrapper {
+  position: relative;
+  height: 0;
+  padding-top: 55%; /* Keep ratio for 900x500 images */
+}
+
+ngb-carousel .picsum-img-wrapper>img {
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
+  right: 0;
+}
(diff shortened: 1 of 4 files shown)


divdavem:

demo/src/style/demos.css
```diff
@@ -41,3 +41,17 @@ ngbd-table-filtering span.ngb-highlight {
 ngbd-table-complete span.ngb-highlight {
   background-color: yellow;
 }
+
+.picsum-img-wrapper {
```
benouat: Some nitpicking here, but I would actually have prefixed that selector with `ngb-caroussel` so that we know it's used inside caroussel demos just by reading here.

```css
ngb-caroussel .picsum-img-wrapper {
}
```

demo/src/style/demos.css
```diff
@@ -41,3 +41,17 @@ ngbd-table-filtering span.ngb-highlight {
...
+  position: relative;
+  height: 0;
+  padding-top: 55%; /* Keep ratio for 900x500 images */
+}
+
+.picsum-img-wrapper>img {
```
benouat: Some remark as above.

divdavem:
@benouat Thank you for your review! I have prefixed the two css rules with `ngb-carousel` as you suggested.

divdavem:
@benouat Thank you for merging my PR!

PR #3275, opened 2019-07-07
feat(datepicker): input datepicker global config

### LINKED ISSUE #3273: Datepicker placement global settings
how to set datepicker placement for all elements.
like this:
config.placement = 'bottom-right'

IAfanasov:
Closes #3273

datepicker

### DIFF
--- a/demo/src/app/components/datepicker/demos/config/datepicker-config.html
+++ b/demo/src/app/components/datepicker/demos/config/datepicker-config.html
@@ -1,3 +1,13 @@
-<p>This datepicker uses customized default values.</p>
+<p>This datepicker input uses customized default values.</p>
 
-<ngb-datepicker [(ngModel)]="model"></ngb-datepicker>
+<form class="form-inline">
+  <div class="form-group">
+    <div class="input-group">
+      <input class="form-control" placeholder="yyyy-mm-dd"
+            name="dp" [(ngModel)]="model" ngbDatepicker #d="ngbDatepicker">
+      <div class="input-group-append">
+        <button class="btn btn-outline-secondary calendar" (click)="d.toggle()" type="button"></button>
+      </div>
+    </div>
+  </div>
+</form>
--- a/demo/src/app/components/datepicker/demos/config/datepicker-config.ts
+++ b/demo/src/app/components/datepicker/demos/config/datepicker-config.ts
@@ -1,5 +1,11 @@
 import {Component} from '@angular/core';
-import {NgbDatepickerConfig, NgbCalendar, NgbDate, NgbDateStruct} from '@ng-bootstrap/ng-bootstrap';
+import {
+  NgbDatepickerConfig,
+  NgbCalendar,
+  NgbDate,
+  NgbDateStruct,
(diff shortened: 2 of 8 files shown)


benouat:
Hey @IAfanasov! Sorry for the late review start.

Even before I start the review, I think I have one overall point to raise
I agree that for us to tackle #3273 we need to introduce a new config file. ngb-datepicker is not only 1 single widget, but more 2 widgets: `Calendar` and `InputDatepicker`.

So this placement config definitely belongs to a new `NgbInputDatepickerConfig` (like you did in this PR)

**Nonetheless** I think it would make sense for that `NgbInputDatepickerConfig` inherits from `NgbDatepickerConfig` no ? Or if inheritance does not make sense (usually extends does not fit very well with DI) we could at least have `NgbInputDatepickerConfig` to inject `NgbDatepickerConfig` and have both of them implements a common interface.

I don't really like that internally we have to inject both config inside the constructor of `NgbIputDatepicjer`.
On top, I think that people might want to be able to override config for calendar and inputdatepicker in a kind of independent way.

It's just a quick thought on the topic, I a not really sure nor sold about it.

WDYT ? cc @pkozlowski-opensource

IAfanasov:
> Hey @IAfanasov! Sorry for the late review start.
Hi @benouat! No worries for late review (: Glad to receive review and make valuable contribution! (:

> **Nonetheless** I think it would make sense for that `NgbInputDatepickerConfig` inherits from `NgbDatepickerConfig` no ? Or if inheritance does not make sense (usually extends does not fit very well with DI) we could at least have `NgbInputDatepickerConfig` to inject `NgbDatepickerConfig` and have both of them implements a common interface.
> 
> I don't really like that internally we have to inject both config inside the constructor of `NgbIputDatepicjer`.
> On top, I think that people might want to be able to override config for calendar and inputdatepicker in a kind of independent way.

Don't see clear solution here as well (: My impression - `NgbDatepickerInput` reuses `NgbDatepicker`. It's not one more `NgbDatepicker` that should be anyhow different from other `NgbDatepicker`s in the app. Thats why I see configs for this components as 2 independent instances without any relations between each other. It's only my opinion. Totally open for suggestions (:

To prevent injections of 2 services injection we can for sure inject config of `NgbDatepickerConfig` into `NgbInputDatepickerConfig`. When developer would use it as 

``` ts
constructor(inputDatepickerConfig: NgbInputDatepickerConfig) {
    // 1. approach with getters and setters for injected into `NgbInputDatepickerConfig` `datepickerConfig`
    inputDatepickerConfig.minDate = {year: 1900, month: 1, day: 1};
    // 2. approach with access to `datepickerConfig`
    inputDatepickerConfig.datepickerConfig.minDate = {year: 1900, month: 1, day: 1};
}
```
Feel like option 2 reveals to developer nature of configs right away while 1 can fool developer into thinking it's independent config.

IAfanasovMob:
Hey @benouat ! (:
seems like @pkozlowski-opensource doesn't have time for this. 
is there other option to make a decision and do this functionality? (:

PR #3283, opened 2019-07-10
fix(toast): proper handling of `autohide` toggling

### LINKED ISSUE #3280: Changing the autoHide of a Toast not working
### Bug description:

If i create a loop of objects to create Toast's and change the autohide input from true to false, the timer event that hide the Toast still fires.
I was trying to make a mouseenter event to stop the Toast from closing, but it still closed if if a set the autohide input to false. Looking into the Toast code i realized that the ngAfterContentInit event was beeing called two times:
```
  ngAfterContentInit() {
    if (this.autohide) {
      this._timeoutID = setTimeout(() => this.hide(), this.delay);
    }
}

```
With that there was two TimeOut events beeing created and only one ref beeing saved in the timeoutID variable, so even if i changed the autohide input to false and the code removing the timeout refered, it still had 1 TimeOut event active.
To fix this, simple add the clearTimeOut function every time the ngAfterContentInit fired, preventing multiple TimeOut events from being created.

```
  ngAfterContentInit() {
    if (this.autohide) {
      this._clearTimeout();
      this._timeoutID = setTimeout(() => this.hide(), this.delay);
    }
}
```

Angular: 8.0.1

ng-bootstrap: 5.0.0-rc.0

Bootstrap: 4.3.1

Edit:
### Stackblitz :
https://angular-toast-bug.stackblitz.io
https://stackblitz.com/edit/angular-toast-bug

benouat:
Toasts initially setup with `[autohide]="true"`were wrongly still emitting `(hide)` output after toggling `autohide` to `false`.

A new demo has also been created to illustrate the `autohide` toggling support, 'Prevent autohide on mouseover'.

Fixes #3280

### DIFF
--- a/src/toast/toast.ts
+++ b/src/toast/toast.ts
@@ -108,22 +108,24 @@ export class NgbToast implements AfterContentInit,
     this.autohide = config.autohide;
   }
 
-  ngAfterContentInit() {
-    if (this.autohide) {
-      this._timeoutID = setTimeout(() => this.hide(), this.delay);
-    }
-  }
+  ngAfterContentInit() { this._init(); }
 
   ngOnChanges(changes: SimpleChanges) {
     if ('autohide' in changes) {
       this._clearTimeout();
-      this.ngAfterContentInit();
+      this._init();
     }
   }
 
   hide() {
     this._clearTimeout();
     this.hideOutput.emit();
+  }
+
+  private _init() {
+    if (this.autohide && !this._timeoutID) {
+      this._timeoutID = setTimeout(() => this.hide(), this.delay);
+    }
   }
 
   private _clearTimeout() {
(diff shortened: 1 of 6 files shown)


src/toast/toast.ts
```diff
@@ -126,10 +122,16 @@ export class NgbToast implements AfterContentInit,
...
+    if (this.autohide && !this._timeoutID) {
+      this._timeoutID = setTimeout(() => this.hide(), this.delay);
+    }
+  }
+
   private _clearTimeout() {
     if (this._timeoutID) {
       clearTimeout(this._timeoutID);
-      this._timeoutID = null;
+      this._timeoutID = undefined;
```
fbasso: It's a question of preference for sure, but I would keep null here, as it's not necessary for the fix. Moreover undefined can be overridden in js, leading to issues, that's I usually prefers null usage.

benouat: done!

PR #3286, opened 2019-07-10
fix(modal): component based content was not scrollable

### LINKED ISSUE #3281: scrollable doesn't work on modals with component as content
### Bug description:

When using a component as the content of a modal rather than a template, with the scrollable option set to true, the modal body doesn't scroll.

The CSS magic apparently doesn't work when an additional element (the host element of the component) is inserted between the `.modal-content` and the `.modal-body`, but I'm so bad at CSS that I can't even find a workaround.

https://stackblitz.com/edit/angular-rekfnj?file=app/modal-component.ts

Angular: 8.1
ng-bootstrap: 5.0.0-rc.1
Bootstrap: 4.3.1

benouat:
In `5.0.0` we introduced the bootstrap scrollable content feature.
This behaviour appears to be broken with modal that have content generated
from a component. Overall styling is wrong, resulting in a broken modal
not scrollable.

~~The fix is duplicating the `.modal-content` css class on the component
host itself that will thus act as the
original `<div class="modal-content">`~~

~~**EDIT** after 32c400f : The fix is adding some extra styling in order to force the component
host to fit inside the parent `.model-content` flex display.~~

**EDIT** after 222752d: The fix is now adding some styling via a class to benefit from autoprefixer as IE10 does not understand `display: flex`

Fixes #3281

### DIFF
--- a/src/modal/modal.spec.ts
+++ b/src/modal/modal.spec.ts
@@ -10,10 +10,8 @@ import {
   ViewChild
 } from '@angular/core';
 import {async, ComponentFixture, TestBed} from '@angular/core/testing';
-
 import {NgbModalConfig} from './modal-config';
 import {NgbActiveModal, NgbModal, NgbModalModule, NgbModalRef} from './modal.module';
-
 
 const NOOP = () => {};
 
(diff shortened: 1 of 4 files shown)


src/modal/modal.spec.ts
```diff
@@ -36,7 +25,8 @@ describe('ngb-modal', () => {
       toHaveModal: function(util, customEqualityTests) {
         return {
           compare: function(actual, content?, selector?) {
-            const allModalsContent = document.querySelector(selector || 'body').querySelectorAll('.modal-content');
+            const allModalsContent =
+                document.querySelector(selector || 'body').querySelectorAll('.modal-dialog > .modal-content');
```
benouat: The selector here needed to be a bit strengthen as we might end-up with my changes with several nodes with the same `.modal-content` class.

benouat:
@fbasso thanks for the review. Your points are valid ones. I'll update my PR to generate inline styles directly

PR #3288, opened 2019-07-12
fix(popover): apply arrow styles to direct descendant

fbasso:
To stick with Bootstrap, since https://github.com/twbs/bootstrap/pull/27750

It hasn't been done for the tooltip, I just do the change for the popover.


leye0:
🤦‍♂ Why wasn't this change done in 4.x.x? Can't nest popovers.

PR #3314, opened 2019-07-25
feat(datepicker): place active day above its siblings

peterblazejewicz:
This is minor change that moves currently focus day within month view
above its siblings to let its outline be correctly rendered.
This is similar to minor stylistic corrections found in the BS:
https://git.io/fjyfh

The before rendering (Chrome/Mac OS):

Screenshot 2019-07-25 at 22 13 59

And after the change:
Screenshot 2019-07-25 at 22 14 23

Thanks!


benouat:
Nice catch! Thanks for this contribution 👍

Issue #3306, opened 2019-07-19
datepicker icon not showing


Balagii:
### Bug description:

I followed the documented install instructions adding _bootstrap_ and _ng-bootstrap_ to my Angular project, specifically to use the date-picker component. The icon shown on the documents page is not showing for the dropdown button in my project. After some digging I came across the following pages:

https://stackoverflow.com/questions/47049383/ng-bootstrap-datepicker-calender-icon-is-not-displaying

https://github.com/ng-bootstrap/ng-bootstrap/issues/3023

https://github.com/ng-bootstrap/ng-bootstrap/issues/1862

The issue and the solution is revealed by a comment found in #1826 : "it is by design that you need [...] add calendar image yourself"

Could you please mention this in the documentation? Like many consumers of this awesome library, I expected the out-of-the-box Bootstrap CSS file to work the same as the documented example, and when it does not, that can be misleading and cause confusion. Thanks.

PR #3315, opened 2019-07-27
docs(datepicker): example svg symbol icon with states

peterblazejewicz:
This commit introduces SVG based calendar icon that follows colors of
the current theme by using `fill` with `currentColor`. This provides
usage scenario of the icon similar to real life ones.

Calendar icon svg picked as replacement from png file content comes from icon library referenced in the project: https://iconmonstr.com/calendar-4-svg/

Closes 3313

If accepted, it could be used as the base of the sample note on how to  provide custom icons for datepicker component as didscussed e.g. in #3306 and also to move other icons reused in wrapper template into definition section and reuse them in the page.
Relying on the current color allows to use any theme from Bootstrap palette, e.g.:

Screenshot 2019-07-27 at 22 57 10

### DIFF
--- a/demo/src/app/components/datepicker/demos/adapter/datepicker-adapter.html
+++ b/demo/src/app/components/datepicker/demos/adapter/datepicker-adapter.html
@@ -20,7 +20,11 @@ In this example we are converting from and to a JS native Date object</p>
           <input class="form-control" placeholder="yyyy-mm-dd"
                  name="d2" #c2="ngModel" [(ngModel)]="model2" ngbDatepicker #d2="ngbDatepicker">
           <div class="input-group-append">
-            <button class="btn btn-outline-secondary calendar" (click)="d2.toggle()" type="button"></button>
+            <button class="btn btn-outline-secondary calendar" (click)="d2.toggle()" type="button">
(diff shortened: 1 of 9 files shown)


benouat:
Hey @peterblazejewicz!

This is an interesting approach. Using svg refs is actually very good.
We already started experimenting with svg icons in the demo, in a [slightly different way](https://github.com/ng-bootstrap/ng-bootstrap/tree/826019819a11c5241fa27eaa1765c943e9c79298/demo/src/app/shared/icons).

I think we might actually transform what we already have to use refs. 🤔

benouat:
So I had a deeper look at your PR, and my conclusion might not be the one you are waiting for.

Despite the fact I find your idea very clever, I am afraid we won't be able to use it like that in our demos. Reason for that is simple → Stackblitz 😞
Each single demo is actually a standalone angular module, with no dependency other than the ones from Angular.

So your PR is actually working perfectly locally, it would also work on github.io, but would fail on Stackblitz

We need to think about something else...

peterblazejewicz:
Fair, thanks for a solid write-up!
Closing.

PR #3333, opened 2019-08-20
fix(generic): Protractor Webdriver Combatibility

gpolychronisAmadeus:
Updated Webdriver versions, so that Protractor's Webdriver is compatible with Chrome.

@benouat Please review

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9312,9 +9312,9 @@ webdriver-js-extender@2.1.0:
     selenium-webdriver "^3.0.1"
 
 webdriver-manager@^12.0.6:
-  version "12.1.4"
-  resolved "https://registry.yarnpkg.com/webdriver-manager/-/webdriver-manager-12.1.4.tgz#d737ab845fb131e4d66c0eaf8ac374c2bc3bfe22"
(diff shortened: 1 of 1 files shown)


gpolychronisAmadeus:
Hello @benouat 

> 
> 
> Hey @gpolychronisAmadeus, thanks for your first contribution!
> 
> Nonetheless, I have a few change requests.
> 
>     * Adding the `.iml` to `.gitignore` does not sound to be related to this protractor fix. Could you please revert it, and eventually open another small PR to add it back ?

Done.

> 
>     * I assume you have bump the protractor version in here... but I can't see the change in `package.json`. Actually it's not even part of the file changes. Could you have a look ?

The issue was the Selenium Webdriver. So I updated its version. I removed irrelevant version updating.

> 
>     * The project is using [Conventional Commits](https://www.conventionalcommits.org/en/). Could change your commit message to use the type: **chore** ? Just like `chore: protactor version update`

Done.

pkozlowski-opensource:
Hint: git ignores related to a local dev environment should go to your global gitignore, not to a project specific one: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

Otherwise you will be sending PRs like this for each end every project you are working on making your (and project maintainers) live harder...

PR #3344, opened 2019-08-29
chore: update dependencies

fbasso:
The purpose of this CR is to update dependencies to have tests running with Chrome 76. This led to several upgrades :

- Webdriver-manager (just by removing yarn.lock),
- As it produced some tests issues, the version of Angular had to be updated too (8.2)
- As I was investigating and upgrading versions, I tried to upgrade node to the latest version (12).

I found out that gulp 3 was not compatible with node 12. There were two solutions : either upgrade to gulp 4, or write a little script for clang-format as it's only used fr that purpose (the 'check only' with clang-format is not straightforward). As we tried to removed the gulp dependency before, I choose to the new script solution.

It's maybe too much thing for one PR, but at least it presents all the things to be done in order to not be 'blocked' in our dev environment. Don't hesitate to tell if several PRs are better !


benouat:
@fbasso thanks for the work! Could please nonetheless double check Angular changelog and validate what has been delivered in between `8.0.0` and `8.2.0` ?

fbasso:
I checked the change log of Angular, only two changes are talking to me, although I think there is no impact (in terms of issues):

core: TypeScript 3.5 support (#31615) (6ece7db)
core: add automatic migration from Renderer to Renderer2 (#30936) (c095597)

For the other ones, it's mostly bug fixes, performance improvements, bazel fixes, ... I don't think they'll have bad impacts, as far as the ng-bootstrap tests are passing !

benouat:
@fbasso 

> * As it produced some tests issues, the version of Angular had to be updated too

Do you remember what was actually failing ?

fbasso:
I'm not sure (as there were so many issues to get the PR validated !), but as far as I remember, it had errors on e2e tests. I can't give you the messages, as one of my first tries was to update Angular and it becomes fine with the tests.

Maybe it would be easy (I hope !) to check it by doing locally a fallback of the Angular version.

fbasso:
Updated the commit message and the title, as the issue for Chrome 16 has been managed in a separated PR.

misc/check-format.js
```diff
@@ -0,0 +1,35 @@
+
+var path = require("path");
```
maxokorokov: why JS, not TS? :/

fbasso: I saw that some others were in js, so I thought we had the choice. I didn't noticed that all the main ones were in ts.

I changed it for typescript.

package.json
```diff
@@ -57,19 +57,19 @@
...
+    "@angular/common": "8.2.4",
+    "@angular/compiler": "8.2.4",
+    "@angular/compiler-cli": "8.2.4",
+    "@angular/core": "8.2.4",
+    "@angular/forms": "8.2.4",
+    "@angular/platform-browser": "8.2.4",
+    "@angular/platform-browser-dynamic": "8.2.4",
+    "@angular/platform-server": "8.2.4",
+    "@angular/router": "8.2.4",
     "@nguniversal/express-engine": "8.0.0-rc.1",
```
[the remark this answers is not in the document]
maxokorokov: `8.1.1`

PR #3345, opened 2019-08-30
Datepicker enhachment/5.1.0/event callback

gpolychronisAmadeus:
Additional API to allow customization of Keydown event.

Rebased upon cc3b755.

@benouat Please review


gpolychronisAmadeus:
> 
> 
> Am I correct that we've decided not to implmement this change in favour of making keyboard mapping service public ?

Yes you are right.

PR #3351, opened 2019-09-05
fix(datepicker): remove hardcoded bg-light

gpolychronisAmadeus:
@benouat 
@maxokorokov 

Please review

### DIFF
--- a/src/datepicker/datepicker-month-view.scss
+++ b/src/datepicker/datepicker-month-view.scss
@@ -20,6 +20,8 @@ ngb-datepicker-month-view {
   &-weekdays {
     border-bottom: 1px solid rgba(0, 0, 0, 0.125);
     border-radius: 0;
+    background-color: #f8f9fa;
+    background-color: var(--light);
   }
   &-day,
   &-weekday,
--- a/src/datepicker/datepicker-month-view.ts
+++ b/src/datepicker/datepicker-month-view.ts
@@ -10,7 +10,7 @@ import {DayTemplateContext} from './datepicker-day-template-context';
   encapsulation: ViewEncapsulation.None,
   styleUrls: ['./datepicker-month-view.scss'],
   template: `
-    <div *ngIf="showWeekdays" class="ngb-dp-week ngb-dp-weekdays bg-light">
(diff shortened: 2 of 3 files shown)


src/datepicker/datepicker-month-view.scss
```diff
@@ -20,6 +20,7 @@ ngb-datepicker-month-view {
   &-weekdays {
     border-bottom: 1px solid rgba(0, 0, 0, 0.125);
     border-radius: 0;
+    background-color: var(--light);
```
fbasso: Using 'var' was a good idea, but unfortunately we can't use this functionality as ng-bootstrap supports IE >= 10

gpolychronisAmadeus: var is already in use:

  &-weekday {
    color: #5bc0de;
    color: var(--info);
  }

Do it like this maybe better?

fbasso:
Thanks for your PR @gpolychronisAmadeus !

You'll also need to rename the commit message (and the title by the way) with `fix(datepicker): ...` (instead of chore), and add a comment `fix #...` to reference the original issue.

fbasso:
I just changed the title with fix (less important than the commit message, but just to be coherent). Otherwise, LGTM !

src/datepicker/datepicker-month-view.scss
```diff
@@ -20,6 +20,8 @@ ngb-datepicker-month-view {
   &-weekdays {
     border-bottom: 1px solid rgba(0, 0, 0, 0.125);
     border-radius: 0;
+    background-color: #f8f9fa;
```
thirdknife: Removing `bg-light` class but you are also introducing a hard coded background color. Why not set on parent background color and in custom css it can be overridden?

benouat: `.ngb-datepicker-month-view-weekdays` is a custom-class publicly available.
You can override the background-color by defining your own. Isn't it enough ?

thirdknife: I can but I have to override for each including header, week, weekday, day using `!important`

PR #3354, opened 2019-09-05
refactor(demos): improve icons usage with svg symbols

benouat:
Inspired by #3315 I refactored a bit the way we display icons in demos structure by introducing the usage of svg symbols.

Unfortunately we just can't use that in the demos themselves, as it would simply not work in Stackblitz.


fbasso:
Thanks for the PR @benouat !

LGTM. 

I have however one remark about SVG and symbol (with no impact for this PR though, but it can feed some future improvements):

I would have preferred implement something with an external link instead of having them inline in the page. It would have taken the advantage of the browser cache, and the svg containing the symbols would have been in the public directory instead of being in the icons component (this would have been more generic by working with the svg you want, no need to embed everything in icons.component.html).

But unfortunately, it comes with a drawback: it doesn't work with IE, so a workaround is necessary, rendering the solution more complex. 

So for now, your PR is fine for me, as it improves the SVG usage in the demo !

PR #3355, opened 2019-09-06
Calendar get prev get next focus on same day instead of first

gpolychronisAmadeus:
@benouat 
@maxokorokov 

Rebased and updated tests.
Please review.

### DIFF
--- a/src/datepicker/datepicker-tools.ts
+++ b/src/datepicker/datepicker-tools.ts
@@ -78,11 +78,12 @@ export function generateSelectBoxYears(date: NgbDate, minDate: NgbDate, maxDate:
 }
 
 export function nextMonthDisabled(calendar: NgbCalendar, date: NgbDate, maxDate: NgbDate) {
-  return maxDate && calendar.getNext(date, 'm').after(maxDate);
+  const nextDate = Object.assign(calendar.getNext(date, 'm'), {day: 1});
+  return maxDate && nextDate.after(maxDate);
 }
 
 export function prevMonthDisabled(calendar: NgbCalendar, date: NgbDate, minDate: NgbDate) {
-  const prevDate = calendar.getPrev(date, 'm');
+  const prevDate = Object.assign(calendar.getPrev(date, 'm'), {day: 1});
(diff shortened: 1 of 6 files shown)


gpolychronisAmadeus:
@benouat @maxokorokov 
Please review

gpolychronisAmadeus:
@ExFlo

src/datepicker/datepicker-tools.ts
```diff
@@ -78,11 +78,14 @@ export function generateSelectBoxYears(date: NgbDate, minDate: NgbDate, maxDate:
 }
 
 export function nextMonthDisabled(calendar: NgbCalendar, date: NgbDate, maxDate: NgbDate) {
-  return maxDate && calendar.getNext(date, 'm').after(maxDate);
+  const nextDate = calendar.getNext(date, 'm');
```
maxokorokov: FYI, can be a bit shorter, but I have no preference really. Change if you like it
```ts
  const nextDate = Object.assign(calendar.getNext(date, 'm'), {day: 1});
```

src/datepicker/datepicker-tools.spec.ts
```diff
@@ -305,16 +305,12 @@ describe(`datepicker-tools`, () => {
       months = buildMonths(calendar, new NgbDate(2018, 6, 5), state, i18n, false);
       expect(months).toBe(state.months);
       expect(months.length).toBe(2);
-      // the structures should be swapped:
```
maxokorokov: As discussed, this if not acceptable, because clearly the:
- behaviour is broken
- you're adapting the test to the broken behaviour

Could fix this:

https://github.com/ng-bootstrap/ng-bootstrap/blob/cca081b13c7bfe36bf09d85fc7c96e2a71dc7aa3/src/datepicker/datepicker-tools.ts#L102

I guess we should make sure that this date also always have the `{day: 1}`...

Second option would be to make the tests always use dates with day 1 → `buildMonths(..., new NgbDate(2018, 7, 1), ...)`, but I think I prefer the first one...

PR #3360, opened 2019-09-09
feat(checkbox): created the checkbox component

Villiard94:

### DIFF
--- a/src/checkbox/checkbox.spec.ts
+++ b/src/checkbox/checkbox.spec.ts
@@ -1,0 +1,240 @@
+import {Component, DebugElement} from '@angular/core';
+import {ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import {FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators} from '@angular/forms';
+import {By} from '@angular/platform-browser';
+import {createGenericTestComponent} from '../test/common';
+import {NgbCheckbox} from './checkbox';
+import {NgbCheckboxModule} from './checkbox.module';
+
+const createTestComponent = (html: string) =>
+    createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
+
+function createKeyEvent(key: number) {
+  return {which: key, preventDefault: () => {}};
+}
+
+function getCheckboxLabel(fixture: ComponentFixture<TestComponent>): DebugElement {
+  return fixture.debugElement.query(By.directive(NgbCheckbox)).query(By.css('label'));
+}
+
+function getCheckboxInput(fixture: ComponentFixture<TestComponent>): DebugElement {
+  return fixture.debugElement.query(By.directive(NgbCheckbox)).query(By.css('input'));
+}
+
+function getCheckboxElement(fixture: ComponentFixture<TestComponent>): DebugElement {
+  return fixture.debugElement.query(By.directive(NgbCheckbox));
+}
+
+function getCheckboxComponent(fixture: ComponentFixture<TestComponent>): NgbCheckbox {
+  return getCheckboxElement(fixture).componentInstance;
+}
+
+describe('ngb-checkbox', () => {
+  beforeEach(() => {
+    TestBed.configureTestingModule(
+        {declarations: [TestComponent], imports: [NgbCheckboxModule, FormsModule, ReactiveFormsModule]});
+  });
+
+  it('should initialize inputs with default values', () => {
+    const checkbox = new NgbCheckbox(null);
+    expect(checkbox.disabled).toBe(false);
+    expect(checkbox.readonly).toBe(false);
+    expect(checkbox.value).toBe(false);
+  });
+
+  it('handles correctly the click event', fakeAsync(() => {
+       const fixture = createTestComponent('<ngb-checkbox [(value)]="value"></ngb-checkbox>');
+       const checkbox = getCheckboxComponent(fixture);
+       const checkboxLabel = getCheckboxLabel(fixture).nativeElement;
+
+       expect(checkbox.value).toEqual(false);
+
+       // click on label
+       checkboxLabel.click();
+       fixture.detectChanges();
+       expect(checkbox.value).toEqual(true);
+       expect(fixture.componentInstance.value).toBe(true);
+
+       // second click on label
+       checkboxLabel.click();
+       fixture.detectChanges();
+       expect(checkbox.value).toEqual(false);
+       expect(fixture.componentInstance.value).toBe(false);
+     }));
+
+  it('ignores the click event on a readonly checkbox', () => {
+    const fixture = createTestComponent('<ngb-checkbox [(value)]="value" [readonly]="true"></ngb-checkbox>');
+    const checkbox = getCheckboxComponent(fixture);
+    const checkboxLabel = getCheckboxLabel(fixture).nativeElement;
+
+    expect(checkbox.value).toEqual(false);
+
+    // click on label
+    checkboxLabel.click();
+    fixture.detectChanges();
+    expect(checkbox.value).toEqual(false);
+    expect(fixture.componentInstance.value).toBe(false);
+  });
+
+  it('ignores the click event on a disabled checkbox', () => {
+    const fixture = createTestComponent('<ngb-checkbox [(value)]="value" [disabled]="true"></ngb-checkbox>');
+    const checkbox = getCheckboxComponent(fixture);
+    const checkboxLabel = getCheckboxLabel(fixture).nativeElement;
+
+    expect(checkbox.value).toEqual(false);
+
+    // click on label
+    checkboxLabel.click();
+    fixture.detectChanges();
+    expect(checkbox.value).toEqual(false);
+    expect(fixture.componentInstance.value).toBe(false);
+  });
+
+  describe('forms', () => {
+    it('should work with reactive form validation', () => {
+      const html = `
+        <form [formGroup]="form">
+          <ngb-checkbox formControlName="checkbox"></ngb-checkbox>
+        </form>`;
+
+      const fixture = createTestComponent(html);
+      const element = getCheckboxElement(fixture);
+      const checkbox = getCheckboxComponent(fixture);
+
+      expect(checkbox.value).toEqual(false);
+      expect(element.nativeElement).toHaveCssClass('ng-invalid');
+      expect(element.nativeElement).toHaveCssClass('ng-untouched');
+
+      fixture.componentInstance.form.patchValue({checkbox: true});
+      fixture.detectChanges();
+      expect(checkbox.value).toEqual(true);
+      expect(element.nativeElement).toHaveCssClass('ng-valid');
+      expect(element.nativeElement).toHaveCssClass('ng-untouched');
+
+      fixture.componentInstance.form.patchValue({checkbox: false});
+      fixture.detectChanges();
+      expect(checkbox.value).toEqual(false);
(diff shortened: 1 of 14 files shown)


Villiard94:
Build is failing because of IE 10 and 11. Not sure why it's failing, the stack trace is not really helpful.

IE 10.0.0 (Windows 8.0.0) ngb-checkbox forms should not mark control as touched when blurred FAILED
	TypeError: Object doesn't support this action
	error properties: Object({ number: -2146827843 })
	    at <Jasmine>
	   at Anonymous function (http://localhost:9876/_karma_webpack_/main.js:4090:25)
	   at step (http://localhost:9876/_karma_webpack_/main.js:3907:13)
	   at Anonymous function (http://localhost:9876/_karma_webpack_/main.js:3888:46)
	   at Anonymous function (http://localhost:9876/_karma_webpack_/main.js:3882:9)
	   at ZoneAwarePromise (http://localhost:9876/_karma_webpack_/vendor.js:95784:29)
	   at Anonymous function (http://localhost:9876/_karma_webpack_/main.js:3878:5)
	   at Anonymous function (http://localhost:9876/_karma_webpack_/main.js:4081:77)
	   at invoke (http://localhost:9876/_karma_webpack_/vendor.js:95243:13)
	   at onInvoke (http://localhost:9876/_karma_webpack_/vendor.js:93502:13)
	   at invoke (http://localhost:9876/_karma_webpack_/vendor.js:95243:13)

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * If `true`, the control will be disabled and the value can't be changed.
+   */
+  @Input() disabled = false;
+  /**
+   * The identifier of the input control.
+   * The default value is an auto-generated ID.
+   */
+  @Input() id: string = '_checkbox_' + checkbox_id_generator++;
```
peterblazejewicz: this should be probably:
```ts
@Input() id = `ngb-checkbox-${nextId++}`;
```

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.value = input.checked;
+  }
+
+  _blur(): void {
+    /**
+     * Checks if the focus is still inside the component itself.
+     * Need to use `setTimeout` to put the execution on top of the call stack.
+     */
+    setTimeout(() => {
+      if (!this._elementRef.nativeElement.querySelector(':focus')) {
```
peterblazejewicz: if anything, this should probably check if `activeElement` is same a child:
```ts
const activeElement = this.document.activeElement;
if (activeElement !== element && !element.contains(activeElement)) {
```
this would require `@Inject(DOCUMENT)` DI addition
or in `ng-bootstrap` way (modals, no DI required):
```ts
if (!this._elRef.nativeElement.contains(document.activeElement)) {
```

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,102 @@
...
+      [attr.id]="id"
+    />
+    <label [attr.tabindex]="-1" class="custom-control-label" [attr.for]="id">
+      <ng-content></ng-content>
+    </label>
+  `,
+  providers: [NGB_CHECKBOX_VALUE_ACCESSOR]
+})
+export class NgbCheckbox implements ControlValueAccessor {
+  private _onChange: (v: any) => void;
```
peterblazejewicz: this can be set as in NG inputs, e.g. `datepicker-input.ts`:
```ts
 private _onChange = (_: any) => {};
 private _onTouched = () => {};
```
this would remove the need of constructor implementation:
```ts
constructor(private _elementRef: ElementRef) {}
```

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,102 @@
...
+};
+
+let checkbox_id_generator = 1;
+
+/**
+ * A component that allows the user to edit boolean values.
+ */
+@Component({
+  selector: 'ngb-checkbox',
+  host: {'[class.custom-control]': 'true', '[class.custom-checkbox]': 'true'},
```
peterblazejewicz:
```ts
host: {'class': 'custom-control custom-checkbox'},
```
probably would be cleaner, shorter. As it never changes it does not require bracket notation for bindings.

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,102 @@
...
+  template: `
+    <input
+      type="checkbox"
+      class="custom-control-input"
+      [disabled]="disabled || readonly"
+      [checked]="value"
+      (change)="_valueChange($event)"
+      (blur)="_blur()"
+      [attr.id]="id"
```
peterblazejewicz: `id` is mapped, so we could use just:
```html
[id]="id"
```
https://angular.io/guide/template-syntax#html-attribute-vs-dom-property

Villiard94:
@peterblazejewicz I'm done with my changes

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,100 @@
...
+    <input
+      type="checkbox"
+      class="custom-control-input"
+      [disabled]="disabled || readonly"
+      [checked]="value"
+      (change)="_valueChange($event)"
+      (blur)="_blur()"
+      [id]="id"
+    />
+    <label [attr.tabindex]="-1" class="custom-control-label" [attr.for]="id">
```
peterblazejewicz: This is static, so:
```html
<label tabindex="-1" class="custom-control-label" [attr.for]="id">
```

src/checkbox/checkbox.spec.ts
```diff
@@ -0,0 +1,223 @@
...
+import {FormControl, FormGroup, FormsModule, ReactiveFormsModule, Validators} from '@angular/forms';
+import {By} from '@angular/platform-browser';
+import {createGenericTestComponent} from '../test/common';
+import {NgbCheckbox} from './checkbox';
+import {NgbCheckboxModule} from './checkbox.module';
+
+const createTestComponent = (html: string) =>
+    createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
+
+function createKeyEvent(key: number) {
```
peterblazejewicz: This block is not used

peterblazejewicz:
@Villiard94 

> Build is failing because of IE 10 and 11. Not sure why it's failing, the stack trace is not really helpful.
> 

This comes from tests, so you could use the `xit` (skip test) to figure out which one, e.g.:
```ts
// tslint:disable-next-line: no-disabled-tests
xit('should mark input as checked / unchecked based
```

Villiard94:
@peterblazejewicz I know it's been a while but now everything builds and there are no conflicts anymore. Do you have anything else to review? What would be the next step?

peterblazejewicz:
@Villiard94 
let me push this further
/cc @maxokorokov

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,100 @@
...
+    const input = event.target as HTMLInputElement;
+    this.value = input.checked;
+  }
+
+  _blur(): void {
+    /**
+     * Checks if the focus is still inside the component itself.
+     * Need to use `setTimeout` to put the execution on top of the call stack.
+     */
+    setTimeout(() => {
```
peterblazejewicz: I guess this is for the same reason as described in NG Material:
https://github.com/angular/components/blob/master/src/material/checkbox/checkbox.ts#L223-L231
Am i correct?

peterblazejewicz:
@maxokorokov 
Should the `checkbox` implement `ideterminate` feature support?
https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate
https://getbootstrap.com/docs/4.4/components/forms/#checkboxes
It cannot be bound via HostBinding, only with plain JS

src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,100 @@
...
+ * A component that allows the user to edit boolean values.
+ */
+@Component({
+  selector: 'ngb-checkbox',
+  host: {'class': 'custom-control custom-checkbox'},
+  template: `
+    <input
+      type="checkbox"
+      class="custom-control-input"
+      [disabled]="disabled || readonly"
```
peterblazejewicz: Not sure here, the `readonly` is really a different beast from `disabled`. 
in NG Material it's used to control the 'can one change value if control is readonly':
https://github.com/angular/components/blob/8f62ffe7359d929ab28bfc6fa8bf4fafcebc3dc6/src/material/input/input.ts#L82
What's your stance on this?

src/checkbox/checkbox.ts
Villiard94: Yes that's correct

benouat:
Hey @Villiard94 !
I am sorry to be late to the party 😞 Thanks for the contribution, nonetheless could you describe briefly what would be the added value of your new component as opposed to _just_ using

+ an `input[type=checkbox]` with either `formControl` or `[(ngModel)]`
+ a label

Advantages of your solution are not clear here. I know that from our [wiki page](https://github.com/ng-bootstrap/ng-bootstrap/wiki/What-qualifies-as-a-ng-bootstrap-directive-or--component%3F) to qualify what makes a new component or not we are saying that for forms it could make sense

> inputs / forms - markup + CSS is quite verbose so encapsulating it with proper Angular forms support (data binding, validation) would be huge win;

I am just not that confident here that the _win_ would be so _huge_

Villiard94:
> Hey @Villiard94 !
> I am sorry to be late to the party 😞 Thanks for the contribution, nonetheless could you describe briefly what would be the added value of your new component as opposed to _just_ using
> 
> * an `input[type=checkbox]` with either `formControl` or `[(ngModel)]`
> * a label
> 
> Advantages of your solution are not clear here. I know that from our [wiki page](https://github.com/ng-bootstrap/ng-bootstrap/wiki/What-qualifies-as-a-ng-bootstrap-directive-or--component%3F) to qualify what makes a new component or not we are saying that for forms it could make sense
> 
> > inputs / forms - markup + CSS is quite verbose so encapsulating it with proper Angular forms support (data binding, validation) would be huge win;
> 
> I am just not that confident here that the _win_ would be so _huge_

Hi @benouat !
Sorry for the late reply. The benefits are the following : 
- Simplifies the markup when using a bootstrap style checkbox
- Handles all the Angular form states

Example without component : 
````
<div class="custom-control custom-checkbox">
  <input type="checkbox" class="custom-control-input" id="customCheck1">
  <label class="custom-control-label" for="customCheck1">Check this custom checkbox</label>
</div>
````

Example with component :
````
<ngb-checkbox>Check this custom checkbox</ngb-checkbox>
````

I see your point about the _huge_ win, but at least I think it's a win. I'm pretty sure it will come handy for certain users of the library. Let me know if you'd like to see more benefits.

src/checkbox/checkbox.ts
Villiard94: I agree with you, in our case the readonly flag is only used for "disabled" state of the underlying input which in turn will properly style it.
I guess it would be more semantically correct to treat readonly as its own. I'll try something

Villiard94:
> @maxokorokov
> Should the `checkbox` implement `ideterminate` feature support?
> https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate
> https://getbootstrap.com/docs/4.4/components/forms/#checkboxes
> It cannot be bound via HostBinding, only with plain JS

I don't know about this one though. The problem is that when used in a form, a checkbox represents a boolean. If we need to support the indeterminate value, then I wonder what the value of the checkbox would be or what value will be sent to the form? Should it be `null` or `undefined`? I honestly don't know.

PR #3362, opened 2019-09-10
test: introduce e2e tests for the demo site

fbasso:
The purpose is to check that the demo site keeps working on each PR. 
This first version just checks basic things.


demo/protractor.conf.js
```diff
@@ -0,0 +1,51 @@
...
+  async onPrepare() {
+    require('ts-node').register({
+      project: require('path').join(__dirname, './tsconfig.spec.json')
+    });
+
+    if (!process.env.TRAVIS) {
+      jasmine.getEnv().addReporter(new SpecReporter({spec: {displayStacktrace: true}}));
+    }
+
+    browser.waitForAngularEnabled(false);
```
maxokorokov: could you please add a comment that we need this because we have timeouts in demos?

demo/protractor.conf.js
```diff
@@ -0,0 +1,51 @@
...
+    await browser.get(`/`);
+
+  },
+  async onComplete() {
```
maxokorokov: no need in onComplete, I guess

demo/src/app/components/components.e2e-spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  ];
+  components.forEach((component) => {
+    describe(`${component} page`, () => {
+
+
+      beforeAll(async() => {
+        await getLinkElement(`components/${component}`).click();
+      });
+
+      it (`should be displayed without errors`, async () => {
```
maxokorokov: Maybe check this after each test instead of doing a separate test case?

demo/src/app/components/components.e2e-spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        const browserLog = await browser.manage().logs().get('browser');
+        if (browserLog.length > 0) {
+          console.error(browserLog);
+          fail(`Unexpected console messages found`);
+        }
+      });
+
+      it(`should display the mains tabs`, async () => {
```
maxokorokov: Should be better if we
- get all the tabs from DOM of a component page, ex: `Overview`, `Examples`, `API`
- loop over them and click to open each tab

package.json
```diff
@@ -29,6 +29,7 @@
     "ngb:test": "ng test ng-bootstrap --code-coverage --source-map true --progress false --watch false",
     "ngb:tdd": "ng test ng-bootstrap --source-map false",
     "ngb:e2e": "ng e2e e2e-app",
+    "ngb:e2e-demo": "yarn demo:docs && ng e2e demo",
```
maxokorokov: please rename to `demo:e2e` / move lower to all other `demo:**` tasks for consistency with others

fbasso:
Thanks for the review @maxokorokov !

Changes done.

PR #3368, opened 2019-09-16
Fix link to stackblitz in README

loicgasser:
Simple fix in `README`


peterblazejewicz:
@loicgasser thx for the fix!

PR #3369, opened 2019-09-18
fix(datepicker): replace shift with alt

gpolychronisAmadeus:


gpolychronisAmadeus:
@benouat 
@maxokorokov 
Please review

PR #3371, opened 2019-09-20
fix(datepicker): restore focus after closing popup

### LINKED ISSUE #3317: Reselect element with focus when datepicker closes
This is request for change.

When:
- user selects a value in datepicker
- user dismisses datepicker clicking outside or using ESC key
previously select element should be probably reselected, as it's done for modal with element to focus:
https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/modal/modal-window.ts#L68

This would allow users to quickly reselect value in the case of e.g. minor error that could be quickly corrected
.
Other reference implementation supporting this behaviour:
https://material.angular.io/components/datepicker/overview
https://clarity.design/documentation/datepicker#examples

This would be non breaking change (unless some tests rely on keyboard navigation with presumed current focus behaviour).

Thanks!

NgBootstrap: 5.1

### LINKED ISSUE #3358: Closing modal with ESC is interrupted by datepicker
### Bug description:

If there is a datepicker component with an overlay, which uses ESC key within parent modal, also with ESC key feature enabled, opening the datepicker and then closing it with ESC, disabled effectively the ESC feature on the container modal, requiring user to perform additional action to refocus any content within modal.

This is probably related to: #3317 

This could be repeated with default demo content:

https://stackblitz.com/run?file=app/modal-basic.ts
https://ng-bootstrap.github.io/#/components/modal/examples#basic

- open basic demo modal, it comes with the datepicker component
- use keyboard to focus the datepicker button
- open datepicker but do not pick a date
- instead close the picker using the ESC key (works)
- now you should be back to modal container
- being in the modal, try to use ESC key again

As stated, this is repeatable with demo content:
https://ng-bootstrap.github.io/#/components/modal/examples#basic

Angular: n/a

ng-bootstrap:
5.1.1

Bootstrap:
n/a

maxokorokov:
At the moment the focus is lost, so it will be restored to the
previously focused element (before datepicker was opened)

Fixes #3317
Fixes #3358

### DIFF
--- a/e2e-app/src/app/datepicker/focus/datepicker-focus.e2e-spec.ts
+++ b/e2e-app/src/app/datepicker/focus/datepicker-focus.e2e-spec.ts
@@ -59,34 +59,26 @@ describe('Datepicker', () => {
     await expectFocused(page.getToggle(), `Toggle element should stay focused after datepicker is closed`);
   });
 
-  it(`should be closed on Escape and focus nothing`, async() => {
+  it(`should be closed on Escape and re-focus toggle element`, async() => {
     await page.openDatepicker();
 
     // close
     await sendKey(Key.ESCAPE);
     expect(await page.getDatepicker().isPresent()).toBeFalsy(`Datepicker should not be present on the page`);
 
-    // check nothing is focused
-    await expectFocused($('body'), `Nothing should be focused after datepicker is closed`);
-
-    // tab should focus toggling element
-    await sendKey(Key.TAB);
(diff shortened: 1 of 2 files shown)


maxokorokov:
@peterblazejewicz thanks for the review, you're right. Moreover it's the problem not only with the datepicker, but with all our components that are displayed inside the modal, ex. dropdown, typeahead. So I'd like to find a more generic solution for this problem

maxokorokov:
Will merge this one as is and open another one to fix closing modal

Issue #3376, opened 2019-09-21
NgbCheckBox autocomplete attribute support


peterblazejewicz:
The 'autocomplete' attribute applies to somehow closed list of interactive controls types derived from the specs group of types (text, urls, passwords, multilines, etc) and it is not yet there for checkbox, though there is a request to incorporate this:
https://github.com/whatwg/html/issues/1894

https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/buttons/checkbox.ts#L22

```ts
@Directive({
  selector: '[ngbButton][type=checkbox]',
  host: {
    'autocomplete': 'off',
    '[checked]': 'checked',
    '[disabled]': 'disabled',
    '(change)': 'onInputChange($event)',
    '(focus)': 'focused = true',
    '(blur)': 'focused = false'
  },
  providers: [NGB_CHECKBOX_VALUE_ACCESSOR]
})
```
for the moment this attribute is a no-op here, offering no support and could mislead I think.

Thanks!

Kissaki:
Note that just because it is not in the standard does not mean it is a no-op. Browsers are not consistently and completely following the HTML standard.

I am not using ng-bootstrap but I stumbled over this ticket through the referenced ticket after experiencing restoration/autocomplete behavior in Firefox (96) besides it not being in the HTML spec. autocomplete=off works despite not being listed.

So heads up, if this is still removed, and depending on context, this may lead to unintended behavior.

PR #3377, opened 2019-09-21
fix(checkbox): remove autocomplete attribute

peterblazejewicz:
This attribute is yet to be supported by specs.

See: https://git.io/JesTb and #3376

### DIFF
--- a/src/buttons/checkbox.spec.ts
+++ b/src/buttons/checkbox.spec.ts
@@ -172,6 +172,22 @@ describe('NgbCheckBox', () => {
        }));
   });
 
+  describe('auto attributes', () => {
+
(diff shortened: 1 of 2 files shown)


peterblazejewicz:
Hey, yes, you're correct. I didn't have the context of the FF problem in mind. I can't repeat that issue on my local setup (MacOS) at the moment. I was concerned about non-standard use. I see folks from BS ditched the idea completely.
Up to you!
Thanks!

benouat:
I checked the way they fixed it in Bootstrap twbs/bootstrap#28952
As @maxokorokov said, they needed a fix in JS that we probably don't need at all because of our usage of Angular with bindings.

I will merge.

PR #1676, opened 2017-07-11
fix(datepicker): use noon instead of midnight for date conversion

maxokorokov:
Using noon should avoid having time saving and leap second issues
Tested with the issue below.

Fixes #1615

Issue #1615, opened 2017-06-12
Datepicker bug in firefox browser


Tolitech:
Hello.
I found a bug when using the datepicker component.

Using the firefox version, the month of October is displayed with the incorrect days.

I've tested on Edge and Chrome and both are working properly.
My version of Firefox is Developer Edition.

I created a Plunker with the reproduction that occurs in Firefox and also has the image of the problem.
[https://embed.plnkr.co/4WvupnKyZQ8t5oFn1G4l/](https://embed.plnkr.co/4WvupnKyZQ8t5oFn1G4l/)

Image:
[http://www.codegenerator.com.br/images/error.png](http://www.codegenerator.com.br/images/error.png)

maxokorokov:
Hey, @Tolitech.

Unfortunately you're plunkr is not working (anymore?) for me

However the demos from the site https://ng-bootstrap.github.io/#/components/datepicker are opening fine in plunker for me with Firefox Developer 55.0b1 on MacOS

Could you please check your plunker?

Tolitech:
Hi, @maxokorokov.
Thanks for the answer.

Unfortunately the bug also occurs on the demo site.
I am using windows 10.

I downloaded now the latest version of firefox (54.0) and the problem is the same.
In October the days are incorrect.

Image of the demo site:
[http://www.codegenerator.com.br/images/ff-error.png](http://www.codegenerator.com.br/images/ff-error.png)

Thank you.

pkozlowski-opensource:
I can't reproduce despite trying several OSX machines:
<img width="1440" alt="screen shot 2017-06-15 at 22 16 21" src="https://user-images.githubusercontent.com/973550/27200023-718fd53a-5218-11e7-876a-b2f117e933d5.png">

@Tolitech are you sure that you are not having any FFox extensions installed? Did you try to reproduce in the incognito mode?

Also, what is your timezone setting?

Tolitech:
@pkozlowski-opensource
 
I believe it's only in Firefox versions of Windows.
I have a virtual machine on Azure and I just installed Firefox again.
There was only Chrome.

I installed 64-bit version.
Same problem as 32 bit version on my machine.

I'm in São Paulo, Brazil, UTC -3:00
The virtual machine I tested is in the United States, but I can not remember exactly which region.

I'm trying to figure out some website that will emulate firefox in MacOS and Windows to know if the problem only occurs in Windows.

I do not have any extensions installed.

[...]

PR #3378, opened 2019-09-23
chore(datepicker): native date adapter defaults to midnight

### LINKED ISSUE #3340: Bug: Datetime picker defaults to 12 PM instead of 12 AM
### Bug description:
When one selects a date through `ngb-datepicker` it automatically takes the time as 12PM when it should take 12 AM by default. 

As seen in the demo below. Given I am in IST timezone (GMT+5:30) when I select say for example 15th Aug, 2019. I get the following output: `2019-08-15T06:30:00.000Z` which is 15th Aug, 2019, 12 PM instead of expected `2019-08-14T18:30:00.000Z` which is 15th Aug, 2019, 12 AM.

image

You can also check in `Material` it selects 12 AM: https://stackblitz.com/edit/material-datepicker-12am

https://stackblitz.com/edit/ngb-datepicker-12pm-issue

Angular: 82.3

ng-bootstrap: 5.0.0

Bootstrap: 4+

peterblazejewicz:
This changes the default time hour computed by native date adapter to be
a midnight, which is default for default Date object when no hours are
set and aligns the behaviour with the UTC date adapter one.

The current behaviour (being changed by this PR), has been introduced without a comment here, changing the default `Date` object constructor defaulting to midnight, instead defaulting to noon:
https://github.com/ng-bootstrap/ng-bootstrap/commit/83930cd84190a077f0305b4f450b1485546bd874#diff-9f982ce59e7ca1c194d2a43bd90a2e4fR13

Fixes: #3340

### DIFF
--- a/src/datepicker/adapters/ngb-date-native-adapter.spec.ts
+++ b/src/datepicker/adapters/ngb-date-native-adapter.spec.ts
@@ -35,12 +35,12 @@ describe('ngb-date-native model adapter', () => {
     });
 
     it('should convert a valid date',
-       () => { expect(adapter.toModel({year: 2016, month: 10, day: 15})).toEqual(new Date(2016, 9, 15, 12)); });
+       () => { expect(adapter.toModel({year: 2016, month: 10, day: 15})).toEqual(new Date(2016, 9, 15)); });
 
     it('should convert years between 0 and 99 correctly', () => {
 
       function jsDate(jsYear: number, jsMonth: number, jsDay: number): Date {
-        const date = new Date(jsYear, jsMonth, jsDay, 12);
+        const date = new Date(jsYear, jsMonth, jsDay);
         if (jsYear >= 0 && jsYear <= 99) {
           date.setFullYear(jsYear);
         }
--- a/src/datepicker/adapters/ngb-date-native-adapter.ts
+++ b/src/datepicker/adapters/ngb-date-native-adapter.ts
@@ -29,7 +29,7 @@ export class NgbDateNativeAdapter extends NgbDateAdapter<Date> {
   }
 
   protected _toNativeDate(date: NgbDateStruct): Date {
-    const jsDate = new Date(date.year, date.month - 1, date.day, 12);
+    const jsDate = new Date(date.year, date.month - 1, date.day);
     // avoid 30 -> 1930 conversion
     jsDate.setFullYear(date.year);
     return jsDate;


shyamal890:
@maxokorokov can you please merge this change.

maxokorokov:
Might be related to this → https://github.com/ng-bootstrap/ng-bootstrap/pull/1676, need to check in more details.

peterblazejewicz:
@maxokorokov 
Can you describe the steps required to repeat the problem?
https://github.com/ng-bootstrap/ng-bootstrap/issues/1615#issuecomment-314033891
I've looked briefly into MD native date adapter on how to handle the hours (they default to midnight).

maxokorokov:
@peterblazejewicz nope, not exactly, sorry. But it was something like resetting current timezone to Brazil.

shyamal890:
Guys if there is no underlying reason for this change why not revert back to the standard followed by most UI libraries? 

Current flow seems like a hack. Moreover, which developer would want to have 12 pm as default, seems really odd.

shyamal890:
@maxokorokov Can you please include this in the upcoming datepicker updates.

shyamal890:
Can we have some clarity on whether this feature would be included or not?

shyamal890:
Kindly share an update on this pull request

shyamal890:
Dear Team, kindly merge this pull request to the master repo, or give an attribute through which this can be activated.

benouat:
@shyamal890 I can't tell you the details because I need to dig up into my memory... But for sure I can tell you (with more than 10 years of JS datepickers implementation expertise) that **we must keep noon** instead of midnight.

While I am performing the check in my memory, could you please explain me why you absolutely need it ?

shyamal890:
@benouat 

1. Most (Almost all) including `Angular Material` sets date to midnight instead of noon. 
2. Any type of filtering needs you to have a `startDate` and `endDate`. Having it automatically set to noon is an anti-pattern (Against the norm, developers have to be trained to add one more step to correct it)

benouat:
@shyamal890 Thanks for your highlights!

I now remember one use case for example, leap second adjustment. If date model is set to midnight, it happens that you are changing day as compared to noon where you don't

And, yes, this is not made up example, I had this issue each single time, and as I told you I have some experience dealing with Javascript Framework ~10 years of various datepicker development)

terencehonles:
@benouat do you mind explaining where/how the leap second adjustments come into play?

If I were to write

```javascript
const year = 2020;
const month = 0;
const day = 1;
const date = new Date(year, month, day);

expect(date.getFullYear()).toBe(year); 
expect(date.getMonth()).toBe(month);
expect(date.getDate()).toBe(day);
```

Aside from years < 100 what days would fail that test case? I would expect any day created to return the same day. Is that not the case? Or are you suggesting doing **math** with said day might cause issues since the day is not guaranteed to be 24 * 60 * 60 * 1000 msec long?

Issue #3335, opened 2019-08-22
Tooltip does not react on changes of `tooltipClass`


joergplewe:
### Bug description:

When changing `tooltipClass` programatically while tooltip is open, it does not reflect the change.

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-3azkna

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular:
8.2.3

ng-bootstrap:
5.0.0

Bootstrap:
4

joergplewe:
I found a workaround, doing a close()/open() in a setTimeout():

https://stackblitz.com/edit/angular-3azkna-rbbglf

```
  toggleError() {
    this.isError = !this.isError;
    setTimeout( () => {
      if ( this.tooltip.isOpen() ) {
        this.tooltip.close();
        this.tooltip.open();
      }
    }, 0);
  }
```

peterblazejewicz:
@joergplewe
Covered by #3380

PR #3380, opened 2019-09-24
fix(tooltip): implement change detection

peterblazejewicz:
This fixes issues with custom tooltip class changes  not being applied after
tooltip creation.

Fixes: #3335

### DIFF
--- a/src/tooltip/tooltip.spec.ts
+++ b/src/tooltip/tooltip.spec.ts
@@ -5,7 +5,6 @@ import {By} from '@angular/platform-browser';
 import {
   AfterViewInit,
   ChangeDetectionStrategy,
-  ChangeDetectorRef,
   Component,
   TemplateRef,
   ViewChild,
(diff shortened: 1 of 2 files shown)


joergplewe:
Great news. Will it be backported to the v4 release of ng-bootstrap? (I'm stuck with Angular7)

src/tooltip/tooltip.ts
```diff
@@ -253,6 +264,13 @@ export class NgbTooltip implements OnInit, OnDestroy {
         this.close.bind(this), +this.openDelay, +this.closeDelay);
   }
 
+  ngOnChanges(changes: SimpleChanges) {
```
maxokorokov: just a shorter variant:

```ts
ngOnChanges({tooltipClass}: SimpleChanges) { }
```

src/tooltip/tooltip.spec.ts
```diff
@@ -60,6 +61,78 @@ describe('ngb-tooltip-window', () => {
...
+    fixture.detectChanges();
+    fixture.componentInstance.tooltipClass = 'second-custom-class';
+    fixture.detectChanges();
+    fixture.componentInstance.tooltipClass = 'third-custom-class';
+    fixture.detectChanges();
+
+    expect(fixture.nativeElement).toHaveCssClass('third-custom-class');
+  });
+
+  describe('ngb-tooltip-window tooltipClass change detection', () => {
```
maxokorokov: Check my other comment, it's Angular who will call change detection if any of the inputs change regardless of what you do. So this `describe` looks useless to me.

src/tooltip/tooltip.spec.ts
```diff
@@ -60,6 +61,78 @@ describe('ngb-tooltip-window', () => {
     expect(fixture.nativeElement).toHaveCssClass('my-custom-class');
   });
 
+  it('should allow to change custom class to new one', () => {
```
maxokorokov: This is kinda covered by the previous test in this file...

What you should actually test is that:
- when you change `classThatChanges` → `<div [ngbTooltip]="hi" [tooltipClass]="classThatChanges">`
- it actually changes class on the `NgbTooltipWindow`

src/tooltip/tooltip.ts
peterblazejewicz: redacted

src/tooltip/tooltip.spec.ts
peterblazejewicz: redacted, thanks!

src/tooltip/tooltip.spec.ts
peterblazejewicz: redacted, thanks!

src/tooltip/tooltip.ts
```diff
@@ -39,16 +41,25 @@ let nextId = 0;
...
   styleUrls: ['./tooltip.scss']
 })
-export class NgbTooltipWindow {
+export class NgbTooltipWindow implements OnChanges {
   @Input() id: string;
   @Input() tooltipClass: string;
+
+  constructor(private _changeDetector: ChangeDetectorRef) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
```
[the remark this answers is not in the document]
peterblazejewicz: redacted, thanks!

peterblazejewicz:
reworked, updated. Thanks!

Issue #2780, opened 2018-10-10
Datepicker next/prev month arrows lose focus


maxokorokov:
In some browsers (ex. Firefox and Safari on Mac OS), clicking on the `previous month` or `next month` buttons loses focus and brings it to the `<body>`.

Steps to reproduce:
- open https://stackblitz.com/edit/angular-dlxsfr
- open popup → today becomes focused
- click on `next month` → focus is lost

Notes:
- works fine in Chrome
- if `next month` is already focused, clicking on it doesn't lose focus

Korla:
Hi, I am wondering about this. Why was it removed from the milestone? Would you accept a PR which fixes this issue?

maxokorokov:
@Korla, sure, it would be nice to fix it. If I remember correctly this really varies from OS and on Windows should work fine. 

I think `<button>` simply can't be focused on click in mentioned browsers, but needs investigation

PR #3381, opened 2019-09-25
fix(datepicker): retain focus on navigation links

peterblazejewicz:
This fixes an issue on some clients due to browser being inconsistent on
handling focus on the button clicks. After the change the navigation
button within the datepicker will retain focus to allow subsequent
keyboard based navigation within datepicker instance.

Fixes: #2780

This is quite established problem and here is most recent article on the topic:
https://zellwk.com/blog/inconsistent-button-behavior/

Thanks!

### DIFF
--- a/src/datepicker/datepicker-navigation.spec.ts
+++ b/src/datepicker/datepicker-navigation.spec.ts
@@ -92,17 +92,35 @@ describe('ngb-datepicker-navigation', () => {
   it('should send navigation events', () => {
     const fixture =
         createTestComponent(`<ngb-datepicker-navigation (navigate)="onNavigate($event)"></ngb-datepicker-navigation>`);
-
-    const links = getNavigationLinks(fixture.nativeElement);
+    const[previousButton, nextButton] = getNavigationLinks(fixture.nativeElement);
+    const previousButtonSpan = previousButton.querySelector<HTMLElement>('span');
+    const nextButtonSpan = nextButton.querySelector<HTMLElement>('span');
     spyOn(fixture.componentInstance, 'onNavigate');
 
     // prev
-    links[0].click();
+    previousButton.click();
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker-navigation.spec.ts
```diff
@@ -92,19 +92,31 @@ describe('ngb-datepicker-navigation', () => {
   it('should send navigation events', () => {
     const fixture =
         createTestComponent(`<ngb-datepicker-navigation (navigate)="onNavigate($event)"></ngb-datepicker-navigation>`);
-
-    const links = getNavigationLinks(fixture.nativeElement);
+    const[previousButton, nextButton] = getNavigationLinks(fixture.nativeElement);
```
maxokorokov: 👍

src/datepicker/datepicker-navigation.ts
```diff
@@ -55,4 +55,14 @@ export class NgbDatepickerNavigation {
   @Output() select = new EventEmitter<NgbDate>();
 
   constructor(public i18n: NgbDatepickerI18n) {}
+
+  onClickPrev(event: MouseEvent) {
+    (event.target as HTMLElement).focus();
```
maxokorokov: This won't work if you click on the internal `<span>` element, unfortunately. Could we change the implemnetation to something like this ?

```html
<button #prev (click)="onNavigate(prev, navigation.PREV)"></button>
```

```ts
onNavigate(button: HTMLButtonElement, direction: NavigateEvent) {}
```

benouat: @peterblazejewicz @maxokorokov this is a bit overkill, I would suggest to switch from `event.target` to `event.currentTarget`

```typescript
onClickPrev(event: MouseEvent) {
    (event.currentTarget as HTMLElement).focus();
}
```

peterblazejewicz: thanks folks for the comments, the detail's has been reworked!

PR #3384, opened 2019-09-27
fix(modal): don't close popup on ESC when closing nested components

### LINKED ISSUE #3358: Closing modal with ESC is interrupted by datepicker

maxokorokov:
Fixes an issue for datepicker, typeahead, dropdown inside the modal.
When these components are closed on ESC, modal stays open.

Nested

The idea is to:
- use the same `keyup` for all ESC autoclose handling
- make modal `keyup` check run after all other autoclose checks
- don't close the modal if event was default prevented

Fixes #3358

### DIFF
--- a/e2e-app/src/app/app.module.ts
+++ b/e2e-app/src/app/app.module.ts
@@ -14,6 +14,7 @@ import {DropdownAutoCloseComponent} from './dropdown/autoclose/dropdown-autoclos
 import {DropdownFocusComponent} from './dropdown/focus/dropdown-focus.component';
 import {DropdownPositionComponent} from './dropdown/position/dropdown-position.component';
 import {ModalFocusComponent} from './modal/focus/modal-focus.component';
+import {ModalNestingComponent} from './modal/nesting/modal-nesting.component';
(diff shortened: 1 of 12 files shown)


fbasso:
Thanks @maxokorokov 

I fail to find a simpler solution for this, so LGTM :)

maxokorokov:
@fbasso thanks, merging!

PR #3332, opened 2019-08-20
feat(datepicker): export NgbDatepickerKeyMapperService

gpolychronisAmadeus:

PR #3396, opened 2019-10-04
feat(datepicker): export NgbDatepickerMonthView component

gpolychronis:
Export NgbDatepickerMonthView component.
This component corresponds to the view of the days of a single month of the datepicker.
This export is an attempt to allow:
- Use of the month view without the rest datepicker visual elements.
- Customization of how a month is displayed inside datepicker i.e. add a header title per month.

Code has been refactored, so that:
- NgbDatepickerMonthView's only input is the year and month to be displayed.
- NgbDatepicker can have a custom content. Combined with the export of NgbDatepickerMonthView it allows to customize datepicker with new DOM elements per month.

NgbDatepickerMonthView remains linked to an instance of a NgbDatepicker, to support the various functionalities it is accompanied with (i.e. i18n, focus, keyboard navigation etc.)


gpolychronisAmadeus:
@benouat @maxokorokov @ExFlo 
Please review

maxokorokov:
@gpolychronisAmadeus haven't looked at it yet, but the build is failing. 

Also some description of what you're trying to do, why and providing some examples is always helpful to review.

src/datepicker/datepicker-month-view.ts
```diff
@@ -1,51 +1,51 @@
...
+  @Input() month: NgbMonthStruct;
 
-  @Output() select = new EventEmitter<NgbDate>();
+  viewModel: MonthViewModel;
 
-  constructor(public i18n: NgbDatepickerI18n) {}
+  constructor(
+      public i18n: NgbDatepickerI18n, public datepicker: NgbDatepicker, private _elementRef: ElementRef<HTMLElement>) {}
+
+  ngOnChanges() { this.viewModel = this.datepicker.registerMonthView(this, this._elementRef.nativeElement); }
```
maxokorokov: Will it work if we inject the datepicker service here and remove `registerMonthView` from the datepicker completely?

So we'll just do these things right here inside the datepicker month view:
- handle focus events
- handle clicks
- inject a service

src/datepicker/datepicker-service.ts
```diff
@@ -155,6 +155,15 @@ export class NgbDatepickerService {
     return this._calendar.isValid(ngbDate) ? ngbDate : defaultValue;
   }
 
+  registerMonth(struct: NgbMonthStruct) {
```
maxokorokov: maybe `getMonth` ?
It doesn't actually register anything

src/datepicker/datepicker-service.ts
```diff
@@ -163,7 +172,8 @@ export class NgbDatepickerService {
   }
 
   private _patchContexts(state: DatepickerViewModel) {
-    const {months, displayMonths, selectedDate, focusDate, focusVisible, disabled, outsideDays} = state;
+    const {months,   displayMonths, selectedDate, focusDate, focusVisible,
```
maxokorokov: you're added stuff here and all of it is unused, please rollback

src/datepicker/datepicker.ts
```diff
@@ -83,47 +94,48 @@ export interface NgbDatepickerNavigateEvent {
...
+          [showSelect]="model.navigation === 'select'"
+          [prevDisabled]="model.prevDisabled"
+          [nextDisabled]="model.nextDisabled"
+          [selectBoxes]="model.selectBoxes"
+          (navigate)="onNavigateEvent($event)"
+          (select)="onNavigateDateSelect($event)">
+        </ngb-datepicker-navigation>
+      </div>
+
+      <div #months class="ngb-dp-months" (keydown)="onKeyDown($event)">
```
maxokorokov: - keyboard handling should be moved to the `datepikcerMonthView` as well I guess
- also `#months` is not used anymore

demo/src/app/components/datepicker/overview/datepicker-overview.component.ts
```diff
@@ -117,6 +117,17 @@ export class NgbdDatepickerOverviewComponent {
         <ngbDatepicker [dayTemplate]=“t”/>
       `,
     }),
+    contentTemplate: Snippet({
+      lang: 'html',
+      code: `
+        <ngb-datepicker>
```
maxokorokov: This example is not working if I copy-paste the code

src/datepicker/ngb-date-struct.ts
```diff
@@ -5,7 +5,14 @@
...
  */
-export interface NgbDateStruct {
+export interface NgbDateStruct extends NgbMonthStruct {
+  /**
+   * The day of month, starting at 1
+   */
+  day: number;
+}
+
+export interface NgbMonthStruct {
```
maxokorokov: Let's remove this new structure. People are already not very happy with `NgbDateStruct` :)

Let's use `month: {month: number, year: number, day?: number}` for public APIs, it will consume everything

gpolychronisAmadeus:
@maxokorokov 
I will rebase current PR upon https://github.com/ng-bootstrap/ng-bootstrap/pull/3332 to use the state.
However, this means that this PR effectively is dependent on 3332, and cannot be reviewed before it has been merged.

maxokorokov:
@gpolychronisAmadeus I think it should be safe to rebase this on master now

gpolychronisAmadeus:
> 
> 
> @gpolychronisAmadeus I think it should be safe to rebase this on master now

@maxokorokov 
keydown event in broken in my local rebased branch. After I figure out why and fix it I will push again.

gpolychronisAmadeus:
> 
>     * [ ]  we also have `NgbInputDatepicker` that should support this functionality, don't forget about it. We could do it in a separate PR, will just need to make sure it will work

Will create a different PR or provide an alternative via demo example. Let's discuss on this.

demo/src/app/components/datepicker/demos/keyboard/datepicker-keyboard.ts
```diff
@@ -10,14 +10,14 @@ const Key = {
 
 @Injectable()
 export class CustomKeyboardService extends NgbDatepickerKeyboardService {
-  processKey(event: KeyboardEvent, dp: NgbDatepicker, calendar: NgbCalendar) {
+  processKey(event: KeyboardEvent, dp: NgbDatepicker) {
     const state = dp.state;
     switch (event.code) {
       case Key.PageUp:
-        dp.focusDate(calendar.getPrev(state.focusedDate, event.altKey ? 'y' : 'm'));
+        dp.focusDate(this.calendar.getPrev(state.focusedDate, event.altKey ? 'y' : 'm'));
```
maxokorokov: Typescript is pretty unhappy with `this.calendar` for some reason. Not sure why

src/datepicker/datepicker-keyboard-service.ts
```diff
@@ -10,18 +10,22 @@ import {Key} from '../util/key';
  */
 @Injectable({providedIn: 'root'})
 export class NgbDatepickerKeyboardService {
+  constructor(private _calendar: NgbCalendar) {}
```
maxokorokov: As `KeyboardService` provided in `root`, I wonder what will get injected when the `calendar` is overridden in the datepicker level ?

I guess it should be left as an argument to the `processKey` function

src/datepicker/datepicker-service.ts
```diff
@@ -151,6 +151,15 @@ export class NgbDatepickerService {
     return this._calendar.isValid(ngbDate) ? ngbDate : defaultValue;
   }
 
+  getMonth(struct: NgbDateStruct) {
```
maxokorokov: should be `NgbDate`, no ?

src/datepicker/datepicker-service.ts
```diff
@@ -151,6 +151,15 @@ export class NgbDatepickerService {
...
+    for (let viewModel of this._state.months) {
```
maxokorokov: nit: maybe `month` instead of `viewModel`?

src/datepicker/datepicker.ts
```diff
@@ -33,7 +34,7 @@ import {NgbDatepickerI18n} from './datepicker-i18n';
 import {isChangedDate, isChangedMonth} from './datepicker-tools';
 import {hasClassName} from '../util/util';
 
-const NGB_DATEPICKER_VALUE_ACCESSOR = {
+export const NGB_DATEPICKER_VALUE_ACCESSOR = {
```
maxokorokov: why `export` ?

src/datepicker/datepicker.ts
```diff
@@ -91,6 +92,19 @@ export interface NgbDatepickerState {
    * The date currently focused by the datepicker
    */
   readonly focusedDate: NgbDate;
+
+  /**
+   * The months in range of the datepicker
```
maxokorokov: need more details here, not clear that dates are first dates for month

src/datepicker/datepicker-service.ts
gpolychronisAmadeus: It is the input of NgbDatepickerMonthView. User might provide a an object with just the year and the month and not a proper NgbDate object.

src/datepicker/datepicker.ts
gpolychronisAmadeus: It is required in datepicker-month-view.spec.ts:228

src/datepicker/datepicker-service.ts
maxokorokov: it was a nitpick and I was talking purely about the `viewModel` variable name

demo/src/app/components/datepicker/demos/keyboard/datepicker-keyboard.ts
```diff
@@ -10,14 +10,14 @@ const Key = {
 
 @Injectable()
 export class CustomKeyboardService extends NgbDatepickerKeyboardService {
-  processKey(event: KeyboardEvent, dp: NgbDatepicker, calendar: NgbCalendar) {
+  processKey(event: KeyboardEvent, dp: NgbDatepicker) {
```
maxokorokov: Can't remove `calendar: NgbCalendar` here, because`dp.calendar` is not a public API

gpolychronisAmadeus:
@maxokorokov 
I wait until https://github.com/ng-bootstrap/ng-bootstrap/pull/3530#issuecomment-572940113 is closed.

I think this might help with performance.

src/datepicker/datepicker-service.ts
gpolychronisAmadeus: It's called from NgbDatepickerMonthView with NgbDateStruct argument. So I would like to keep it as it is.

demo/src/app/components/datepicker/demos/keyboard/datepicker-keyboard.ts
gpolychronisAmadeus: Generally, each datepicker instance is tied with a specific calendar instance. So maybe it's better to export the calendar instance via a getter.

gpolychronis:
As discussed the following changes will have to be done:

- Replace **month** of NgbDatepickerMonthView with setter.
- Use language or (existing's) library find method in getMonth. Check if supported by all browsers first
- Make **contentTemplate** static = true

Documentation issue will be fixed by an other commit.

src/datepicker/datepicker.ts
```diff
@@ -293,10 +309,9 @@ export class NgbDatepicker implements OnDestroy,
   onTouched = () => {};
 
   constructor(
-      private _service: NgbDatepickerService, private _calendar: NgbCalendar, public i18n: NgbDatepickerI18n,
-      config: NgbDatepickerConfig, private _keyboardService: NgbDatepickerKeyboardService, cd: ChangeDetectorRef,
-      private _elementRef: ElementRef<HTMLElement>, private _ngbDateAdapter: NgbDateAdapter<any>,
-      private _ngZone: NgZone) {
+      public _service: NgbDatepickerService, private readonly _calendar: NgbCalendar, public i18n: NgbDatepickerI18n,
```
maxokorokov: - `_service` → no reason to make it public, should stay as before
- `cd` → no reason to make it private, should stay as before

src/datepicker/datepicker-keyboard-service.ts
```diff
@@ -15,8 +14,9 @@ export class NgbDatepickerKeyboardService {
   /**
    * Processes a keyboard event.
    */
-  processKey(event: KeyboardEvent, datepicker: NgbDatepicker, calendar: NgbCalendar) {
+  processKey(event: KeyboardEvent, datepicker: NgbDatepicker) {
     const state = datepicker.state;
+    const calendar = datepicker.calendar;
```
maxokorokov: Just a cosmetic thing. Could be:
```ts
const {state, calendar} = datepicker;
```

maxokorokov:
As we've discussed:
- removed the additional div element that was added for months
- renamed `ngb-datepicker-month-view` to `ngb-datepicker-month`
- updated the custom month layout demo with navigation example

PR #3400, opened 2019-10-04
fix(progressbar): display progressbar correctly for invalid max value

### LINKED ISSUE #3386: Better handling for progress bar with max=0
### Bug description:

Currently progress bar with the max value of 0 fills for the width necessary for the label to display. This is not consistent behavior and it will be better to make max=0 a special case and always fill either 0% or 100%. 

I think 100% will be a more sensible behavior because of cases like value=42 and max=0. In this case progress bar will [trim value to 0](https://github.com/ng-bootstrap/ng-bootstrap/blob/75f09660671577cd7988e6d93332b7749f76db25/src/progressbar/progressbar.ts#L74), but 100% will make some sense in this situation unlike 0%.

https://stackblitz.com/edit/angular-mc1mw9

Angular: 8.2.0

ng-bootstrap: 5.1.0

Bootstrap: 4.3.1

chrstnbrn:
closes #3386

Currently the following is displayed when max is set to 0, null or undefined:
image

After this change the max value will be set to 100 if it is invalid so that a progressbar with a valid value can be displayed.

### DIFF
--- a/src/progressbar/progressbar.spec.ts
+++ b/src/progressbar/progressbar.spec.ts
@@ -58,6 +58,48 @@ describe('ngb-progressbar', () => {
       expect(progressCmp.getPercentValue()).toBe(20);
     });
 
+    it('should calculate the percentage (custom max size of null)', () => {
+      progressCmp.max = null;
+
+      progressCmp.value = 25;
+      expect(progressCmp.getPercentValue()).toBe(25);
+    });
+
+    it('should calculate the percentage (custom max size of undefined)', () => {
+      progressCmp.max = undefined;
+
+      progressCmp.value = 25;
+      expect(progressCmp.getPercentValue()).toBe(25);
+    });
+
+    it('should calculate the percentage (custom max size of zero)', () => {
+      progressCmp.max = 0;
+
+      progressCmp.value = 25;
+      expect(progressCmp.getPercentValue()).toBe(25);
+    });
+
+    it('should calculate the percentage (custom negative max size)', () => {
+      progressCmp.max = -10;
+
+      progressCmp.value = 25;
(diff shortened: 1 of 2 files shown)


peterblazejewicz:
@chrstnbrn 
I believe this guard should also catch negative values and negative infinite value?
This is how the native progress handless max value (it requires positive, non-0 and finite values)
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress
just for consideration

chrstnbrn:
@peterblazejewicz I changed the code to handle negative values. I also updates the docs to say that max should be a positive number.

src/progressbar/progressbar.spec.ts
```diff
@@ -58,6 +58,48 @@ describe('ngb-progressbar', () => {
       expect(progressCmp.getPercentValue()).toBe(20);
     });
 
+    it('should calculate the percentage (custom max size of null)', () => {
+      progressCmp.max = null;
+
+      progressCmp.value = 25;
+      expect(progressCmp.getPercentValue()).toBe(100);
```
maxokorokov: All tests should return `25`

src/progressbar/progressbar.ts
```diff
@@ -21,6 +21,8 @@ import {NgbProgressbarConfig} from './progressbar-config';
 export class NgbProgressbar {
   /**
    * The maximal value to be displayed in the progress bar.
+   *
+   * Should be a positive number.
```
maxokorokov: Should be a positive number. Will default to 100 otherwise.

chrstnbrn:
Hi @maxokorokov,

thank you for your feedback! I agree that adjusting the invalid max value makes more sense and made the changes you suggested.

PR #3409, opened 2019-10-12
test(rating): add missing coverage block

peterblazejewicz:
This should cover missing block in the coverage for `rating.ts`.

Thanks!


peterblazejewicz:
well, completely unexpected coverage change :)

PR #3411, opened 2019-10-14
fix(pagination): pagination for the dirffrent size of the screens

BehnooshShiva:

### DIFF
--- a/src/pagination/pagination.scss
+++ b/src/pagination/pagination.scss
@@ -1,0 +1,9 @@
+.pagination{
+    display: block;
+    .page-item{
+        display: inline-block;
+        a{
+            display: table;
+        }
+    }
+}
--- a/src/pagination/pagination.ts
+++ b/src/pagination/pagination.ts
@@ -166,7 +166,8 @@ export class NgbPaginationPrevious {
         </a>
       </li>
     </ul>
-  `
+  `,
+  styleUrls: ['./pagination.scss']
 })
 export class NgbPagination implements OnChanges {
   pageCount = 0;


benouat:
Hey @BehnooshShiva 

Thanks for your first-time contribution to the project!
Would you mind show me some documentation (maybe Bootstrap) that would be related to the change you propose ?

I don't see anything related to that on Bootstrap side.

BehnooshShiva:
Hello @benouat 
Here is the behavior of component pagination in a small screen. It comes out of the screen instead of going in a new line.
image
The pagination class come from Bootstrap

benouat:
Thank you for the clarifications!

> I don't see anything related to that on Bootstrap side.

What I meant by that is that Bootstrap don't handle this use-case either.
Here is a screenshot from their demo website
image

Hence to implement it you actually don't need
```scss
.pagination{
     display: block;
     .page-item{
         display: inline-block;
         a{
             display: table;
         }
     }
 } 
```
but could use instead (simply declare that in your application)
```css
.pagination {
    flex-wrap: wrap
}
``` 
---
That discussion leads me to another point I could highlight:

If you ever need to flex-wrap your pagination, in term of UX it means that maybe it would be much better to use the combination of  `ellipses/maxSize` [inputs](https://ng-bootstrap.github.io/#/components/pagination/api#NgbPagination) as suggested in [this demo](https://ng-bootstrap.github.io/#/components/pagination/examples#advanced) instead.

---
Tell me if you could use the `flex-wrap` solution, in which case we could simply close that PR

BehnooshShiva:
Ok, yes i can use flex-wrap

PR #3421, opened 2019-10-21
chore: refactor karma IE configuration

maxokorokov:


angular.json
```diff
@@ -34,6 +34,19 @@
           "configurations": {
             "ie": {
               "tsConfig": "src/tsconfig-ie.spec.json"
+            },
+            "saucelabs": {
+              "sourceMap": false,
+              "progress": false,
+              "karmaConfig": "src/karma.sauce.conf.js",
+              "browsers": "CHROME,FIREFOX,EDGE16,EDGE17,SAFARI11,SAFARI12"
```
benouat: nitpicking → I would have keep that one the `src/karma.sauce.conf.js` file.

angular.json
```diff
@@ -34,6 +34,19 @@
           "configurations": {
             "ie": {
               "tsConfig": "src/tsconfig-ie.spec.json"
```
benouat: Can be removed!

maxokorokov: You still need it if you want to run unit tests with `ie`

angular.json
maxokorokov: Good idea actually! Will update

PR #3422, opened 2019-10-23
fix(modal): keep focus trap between stack modals

### LINKED ISSUE #3392: Stacked modal lost keyboard trap navigation
### Bug description:
The focus doesn't stay within the modal window after interaction with a nested modal.

### Step to reproduce:

[Stackblitz env](https://stackblitz.com/edit/angular-8a3sor-xd3eqt) with bug.
(fork from `stacked modals` documentation)

1. Open the first modal (`launch initial modal` button).
2. Navigate via keyboard to `launch nested modal` button.
3. Open nested modal via pressing space.
4. Click the cross button or close button with the keyboard.
5. Use the keyboard for navigation inside the initial modal.

Important:
- While investigating the issue, discover, that it is reproducing only if the element that triggers the nested modal is the last interactive element in the initial modal.

Actual result:
- The focus has moved beyond the modal window.

Expected result:
- The focus moving within the modal window.

Angular: `8.0.0`

ng-bootstrap: `5.1.1`

fbasso:
fix #3392

### DIFF
--- a/e2e-app/src/app/modal/stack/modal-stack.component.ts
+++ b/e2e-app/src/app/modal/stack/modal-stack.component.ts
@@ -1,0 +1,9 @@
+import {Component, TemplateRef} from '@angular/core';
+import {NgbModal} from '@ng-bootstrap/ng-bootstrap';
+
+@Component({templateUrl: './modal-stack.component.html'})
+export class ModalStackComponent {
+  constructor(private modalService: NgbModal) {}
+
+  openModal(content: TemplateRef<any>) { this.modalService.open(content); }
+}
--- a/e2e-app/src/app/modal/stack/modal-stack.e2e-spec.ts
+++ b/e2e-app/src/app/modal/stack/modal-stack.e2e-spec.ts
@@ -1,0 +1,32 @@
+import {Key} from 'protractor';
+import {expectFocused, expectNoOpenModals, openUrl, sendKey} from '../../tools.po';
+import {ModalStackPage} from './modal-stack.po';
+
+describe('Modal stacked', () => {
+  let page: ModalStackPage;
+
+  beforeAll(() => { page = new ModalStackPage(); });
+
+  beforeEach(async() => await openUrl('modal/stack'));
+
+  afterEach(async() => { await expectNoOpenModals(); });
+
+  it('should keep tab on the first modal after the second modal has closed', async() => {
+    await page.openModal();
+    await page.openStackModal();
+
+    // close the stack modal
+    await sendKey(Key.ESCAPE);
+
+    // Check that the button is focused again
+    await expectFocused(page.getStackModalButton(), 'Button element not focused');
+    await sendKey(Key.TAB);
+
+    await expectFocused(page.getCoseIcon(), 'Close icon not focused');
+
+    // close the main modal
+    await sendKey(Key.ESCAPE);
+
+  });
+
+});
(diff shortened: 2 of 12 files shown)


src/modal/modal-window.ts
```diff
@@ -97,7 +97,7 @@ export class NgbModalWindow implements OnInit,
     } else {
       elementToFocus = body;
     }
-    elementToFocus.focus();
+    setTimeout(() => elementToFocus.focus());
```
pkozlowski-opensource: do we want to run it in the Angular zone and trigger additional CD after destroy?

e2e-app/src/app/tools.po.ts
```diff
@@ -34,8 +34,8 @@ export const offsetClick = async(el: ElementFinder, offset) => {
  * @param message to display in case of error
  */
 export const expectFocused = async(el: ElementFinder, message: string) => {
-  const focused = await browser.driver.switchTo().activeElement();
-  expect(await WebElement.equals(el.getWebElement(), focused)).toBeTruthy(message);
+  await browser.wait(
+      () => { return WebElement.equals(el.getWebElement(), browser.driver.switchTo().activeElement()); }, 500, message);
```
maxokorokov: I see no reason for it to be `500`. It should be `0`, because, this `wait` will always be scheduled after the `.focus()`

e2e-app/src/app/modal/stack/modal-stack.e2e-spec.ts
```diff
@@ -0,0 +1,43 @@
+import {Key, browser} from 'protractor';
+import {expectFocused, expectNoOpenModals, openUrl, sendKey} from '../../tools.po';
+import {ModalStackPage} from './modal-stack.po';
+import {DatepickerPage} from '../../datepicker/datepicker.po';
+import {DropdownPage} from '../../dropdown/dropdown.po';
+import {TypeaheadPage} from '../../typeahead/typeahead.po';
+
+describe('Modal stacked', () => {
+  let page: ModalStackPage;
```
maxokorokov: please leave only `ModalStackPage` and remove the rest

e2e-app/src/app/modal/stack/modal-stack.component.ts
```diff
@@ -0,0 +1,33 @@
...
+import {Observable} from 'rxjs';
+import {map} from 'rxjs/operators';
+
+@Component({templateUrl: './modal-stack.component.html'})
+export class ModalStackComponent {
+  constructor(private modalService: NgbModal) {}
+
+  openModal(content: TemplateRef<any>) { this.modalService.open(content); }
+
+  search = (text$: Observable<string>) => text$.pipe(map(() => ['one', 'two', 'three']));
```
maxokorokov: please remove unused functions

e2e-app/src/app/modal/stack/modal-stack.component.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  openSecondModal() { this.modalService.open(NgbdStackedModal); }
```
maxokorokov: can't you simply reuse `openModal()` and remove `NgbdStackedModal` component (use templates for both first and second modals) ? It would be much simpler.

src/modal/modal-window.ts
maxokorokov: No a side note it wont change anything for this particular PR, because the whole focus trap runs inside Angular zone oO. I'll fix that in a separate PR.

Issue #2773, opened 2018-10-09
ReadonlyArray<T> instead of T[]


AbakumovAlexandr:
### Feature description:
A proposal is to use `ReadonlyArray<T>` instead of `T[]` for _input_ (as seen from NGB side) arrays throughout the publicly visible API, for instance, as a return value type for `ngbTypeahead` function in `NgbTypeahead`:
```
(text: Observable<string>) => Observable<ReadonlyArray<any>>
```
instead of:
```
(text: Observable<string>) => Observable<any[]>
```

It explicitly declares by NGB the immutability of arrays as well as ensures that both groups of developers who use/don't use  `ReadonlyArray<T>` would be able to communicate with NGB without extra effort like `array.slice()` before passing to\returning from NGB methods.

The above part doesn't introduce a breaking change. We could also consider go further and do the same for _output_ arrays as well to promote immutability in the client code bases.

### Versions of Angular, ng-bootstrap and Bootstrap:

ng-bootstrap: 3.3.0

pkozlowski-opensource:
Sounds like a sensible request. Small, focused (separate one per widget) PRs welcomed.

benouat:
> would be able to communicate with NGB without extra effort like `array.slice()` before passing to\returning from NGB methods

Could you illustrate this _extra effort_ you're mentioning? I am not sure to get what you are describing here.

AbakumovAlexandr:
@benouat Sure. Let's take the above example with `ngbTypeahead`. Its signature requires a function to return `any[]` instead of `ReadonlyArray<any>`.

Now, suppose our app conforms to immutability ideology and uses `ReadonlyArray`s only. To implement the `ngbTypeahead` function, we thus have to somehow convert our `ReadonlyArray<any>` to `any[]` before returning it to NGB.

The first thought is just to perform a type cast:
```
public typeaheadListFn = (text: Observable<string>): Observable<any[]> => {
        const list: ReadonlyArray<any> = this.getTypeageadListForText(text);
        return list as any[];
    };
```
But we shouldn't do this, since if NGB actually calls any mutating method on that array which is missing in our `ReadonlyArray<any>`, there would be a run-time exception.

So, the simplest way I'm aware of is to make a copy of our immutable array with `slice()` before returning it:
```
public typeaheadListFn = (text: Observable<string>): Observable<any[]> => {
        const list: ReadonlyArray<any> = this.getTypeageadListForText(text);
        return list.slice();
    };
```
[...]

PR #3423, opened 2019-10-23
feat(typeahead): use readonly arrays for inputs (#2773)

reduckted:

Part of #2773.

### DIFF
--- a/src/typeahead/highlight.ts
+++ b/src/typeahead/highlight.ts
@@ -38,7 +38,7 @@ export class NgbHighlight implements OnChanges {
    * The term or array of terms to be highlighted.
    * Since version `v4.2.0` term could be a `string[]`
    */
-  @Input() term: string | string[];
+  @Input() term: string | readonly string[];
 
   ngOnChanges(changes: SimpleChanges) {
     const result = toString(this.result);
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -135,7 +135,7 @@ export class NgbTypeahead implements ControlValueAccessor,
    *
    * Note that the `this` argument is `undefined` so you need to explicitly bind it to a desired "this" target.
    */
-  @Input() ngbTypeahead: (text: Observable<string>) => Observable<any[]>;
+  @Input() ngbTypeahead: (text: Observable<string>) => Observable<readonly any[]>;
 
   /**
    * The function that converts an item from the result list to a `string` to display in the popup.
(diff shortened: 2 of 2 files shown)


src/typeahead/highlight.ts
```diff
@@ -38,7 +38,7 @@ export class NgbHighlight implements OnChanges {
    * The term or array of terms to be highlighted.
    * Since version `v4.2.0` term could be a `string[]`
    */
-  @Input() term: string | string[];
+  @Input() term: string | ReadonlyArray<string>;
```
maxokorokov: Could we use `string | readonly string[]` syntax please ?

src/typeahead/typeahead.ts
```diff
@@ -135,7 +135,7 @@ export class NgbTypeahead implements ControlValueAccessor,
    *
    * Note that the `this` argument is `undefined` so you need to explicitly bind it to a desired "this" target.
    */
-  @Input() ngbTypeahead: (text: Observable<string>) => Observable<any[]>;
+  @Input() ngbTypeahead: (text: Observable<string>) => Observable<ReadonlyArray<any>>;
```
maxokorokov: Could we use `readonly any[]` syntax here, please?

src/test/typeahead/common.ts
```diff
@@ -9,7 +9,7 @@ export function getWindowLinks(element: DebugElement): DebugElement[] {
   return Array.from(element.queryAll(By.css('button.dropdown-item')));
 }
 
-export function expectResults(nativeEl: HTMLElement, resultsDef: string[]): void {
+export function expectResults(nativeEl: HTMLElement, resultsDef: ReadonlyArray<string>): void {
```
maxokorokov: Let's drop this as a non-public API

src/typeahead/typeahead.ts
```diff
@@ -367,7 +367,7 @@ export class NgbTypeahead implements ControlValueAccessor,
     this._renderer.setProperty(this._elementRef.nativeElement, 'value', toString(value));
   }
 
-  private _subscribeToUserInput(userInput$: Observable<any[]>): Subscription {
+  private _subscribeToUserInput(userInput$: Observable<ReadonlyArray<any>>): Subscription {
```
maxokorokov: Let's drop this as a non-public API

reduckted: This one will need to stay as read-only (but I'll change it to `readonly any[]`) because the value comes from result of `ngbTypeahead`.

```ts
const results$ = inputValues$.pipe(this.ngbTypeahead);
const userInput$ = this._resubscribeTypeahead.pipe(switchMap(() => results$));
this._subscription = this._subscribeToUserInput(userInput$);
```
```
Argument of type 'Observable<readonly any[]>' is not assignable to parameter of type 'Observable<any[]>'.
  The type 'readonly any[]' is 'readonly' and cannot be assigned to the mutable type 'any[]'.ts(2345)
```

PR #3428, opened 2019-10-23
chore: update yarn.lock to update selenium

fbasso:
As tests are failing with teh new version of Chrome


fbasso:
Thanks @maxokorokov, it avoids further investigation on my side :)

PR #3431, opened 2019-10-24
test(dropdown): use 'ng-template' instead of 'ng-container'

maxokorokov:
`ngbDropdownMenu` uses `@ContentChildren` query to get `ngbDropdownItem`'s that wont't work with NG9


pkozlowski-opensource:
LGTM! We might relax the `<ng-container>` constraints in the end in v9 but `<ng-template>` is a "better" pattern here, anyway!

PR #3440, opened 2019-10-28
feat(datepicker): Add example of datepicker range selection in popup.

gpolychronisAmadeus:
Added an example of datepicker range selector in popup.

Probably many users of the framework would like to use a datepicker range selector in popup, so hosting such example would be useful.

### DIFF
--- a/demo/src/app/components/datepicker/datepicker.module.ts
+++ b/demo/src/app/components/datepicker/datepicker.module.ts
@@ -29,6 +29,8 @@ import { NgbdDatepickerPopup } from './demos/popup/datepicker-popup';
 import { NgbdDatepickerPopupModule } from './demos/popup/datepicker-popup.module';
 import { NgbdDatepickerRange } from './demos/range/datepicker-range';
 import { NgbdDatepickerRangeModule } from './demos/range/datepicker-range.module';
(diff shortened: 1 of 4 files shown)


gpolychronisAmadeus:
@maxokorokov 

Please review.

benouat:
I did not yet start a review, but out of curiosity what is the main difference as compared to the already existing example we have which is not in a popup ?

gpolychronisAmadeus:
> 
> 
> I did not yet start a review, but out of curiosity what is the main difference as compared to the already [existing example](https://ng-bootstrap.github.io/#/components/datepicker/examples#range) we have which is not in a popup ?

This is the difference. It's in popup.

gpolychronisAmadeus:
> 
> 
> Hey, @gpolychronisAmadeus!
> 
> While I haven't look at the implementation details, I think we can do better than this visually:
> 
> Screen Shot 2019-11-04 at 16 57 43
> 
> A quick google search couldn't give me any datepickers that use two inputs for the range selection in the popup, ex. booking, airbnb don't use inputs.
> 
> Would you have any examples of what you're trying to achieve visually?
> 
> Let's play with it a little more.

There are sites that use two datepickers, one for from and one for to date. In some of them after selecting from date the to date datepicker opens automatically. Replacing two datepickers with a single date range picker might be a good alternative.

I updated the example.

PR #3119, opened 2019-04-09
feat(datepicker): add positionTarget property

dzonatan:
Partially closes #3092.

PR #3452, opened 2019-11-06
feat(popover): allow specifying target for popover

jbarrus:
This is incomplete, looking for feedback on whether this would get merged if it were completed. My use case is around providing popovers inside dynamic content. In my case I want to add popovers for user mentions inside content produced from [quill](https://quilljs.com/) and [quill-mention](https://github.com/afconsult/quill-mention). This is similar to the popover github uses when mentioning issues/prs/users in content. If there's an easier way to do this, let me know.


maxokorokov (changes requested):
Hey,

Ok for the feature, but could be implement it the same way as it is done for the datepicker already here → https://github.com/ng-bootstrap/ng-bootstrap/pull/3119/files

```ts
@Input() positionTarget: string | HTMLElement;
```

Cheers,
Max

PR #3461, opened 2019-11-09
feat(datepicker): change 'select' event to 'dateSelect'

### LINKED ISSUE #3444: Focus on datepicker footer triggers extra selects
### Bug description:
When using a datepicker with a custom footer, if that footer contains focusable elements (e.g. inputs), navigating between those elements with keyboard tab causes select events which are invalid. The selection payload is not an `NgbDate` instance, but a JS `Event` with `type: 'select'`.

https://stackblitz.com/edit/ng-bootstrap-footer-focus

Angular: 8.2.12

ng-bootstrap: 5.1.2

Bootstrap: 4.3.0

peterblazejewicz:
This commit changes name of the component custom event to avoid
conflicts with native 'select' event.
Existing event name is marked as deprecated for backward compatibility.

Fixes #3444

Thanks!

### DIFF
--- a/src/datepicker/datepicker-service.ts
+++ b/src/datepicker/datepicker-service.ts
@@ -45,9 +45,6 @@ export class NgbDatepickerService {
 
   get dateSelect$(): Observable<NgbDate> { return this._dateSelect$.pipe(filter(date => date !== null)); }
 
-  /** @deprecated please use `dateSelect$` */
-  get select$(): Observable<NgbDate> { return this.dateSelect$; }
-
   set dayTemplateData(dayTemplateData: NgbDayTemplateData) {
     if (this._state.dayTemplateData !== dayTemplateData) {
       this._nextState({dayTemplateData});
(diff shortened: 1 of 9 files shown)


src/datepicker/datepicker-service.ts
```diff
@@ -43,7 +43,10 @@ export class NgbDatepickerService {
 
   get model$(): Observable<DatepickerViewModel> { return this._model$.pipe(filter(model => model.months.length > 0)); }
 
-  get select$(): Observable<NgbDate> { return this._select$.pipe(filter(date => date !== null)); }
+  get dateSelect$(): Observable<NgbDate> { return this._dateSelect$.pipe(filter(date => date !== null)); }
+
+  /** @deprecated please use `dateSelect$` */
+  get select$(): Observable<NgbDate> { return this.dateSelect$; }
```
maxokorokov: This should be removed, not deprecated because `NgbDatepickerService` is the internal implementation. I'd say either:
- don't rename `_select$`  and `get select()` at all, because it's internal (I think I prefer this personally)
- or remove this deprecation completely, because it should not be used anywhere anyway

PR #3468, opened 2019-11-15
fix(pagination): disallow tab focus on disabled pagination

Smoggy:
The purpose of this PR i to fix focus on disabled pagination control using TAB key

### DIFF
--- a/src/pagination/pagination.spec.ts
+++ b/src/pagination/pagination.spec.ts
@@ -630,17 +630,27 @@ describe('ngb-pagination', () => {
          expect(fixture.componentInstance.onPageChange).not.toHaveBeenCalled();
        }));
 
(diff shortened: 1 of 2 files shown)


src/pagination/pagination.spec.ts
```diff
@@ -641,6 +641,18 @@ describe('ngb-pagination', () => {
            expect(buttons[i]).toHaveCssClass('disabled');
          }
        }));
+
+    it('should set tabindex for links correctly for disabled state', fakeAsync(() => {
```
maxokorokov: This test doesn't need to be `fakeAsync` and `tick()` is also not necessary.
The `should set classes correctly for disabled state` you used as a base is also wrong and doesn't need them.

Do you mind fixing both please, while you're at it ?

PR #3470, opened 2019-11-15
feat(pagination): `aria-current` attribute support

peterblazejewicz:
This commit adds support for aria-current attribute as per:
https://getbootstrap.com/docs/4.3/components/pagination/#disabled-and-active-states
https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Pagination

Thanks!


ihorzenich:
@peterblazejewicz Keeping `<span class="sr-only">(current)</span>` together with `aria-current="page"` on parent cause double announcement of "current" page in NVDA screen-reader.
image

It is necessary to remove `<span class="sr-only">(current)</span>` from `ngb-pagination`.

Issue #2877, opened 2018-11-20
Pagination component is not responsive (doesn't wrap)


talynone:
### Bug description:
When pagination has enough pages to extend beyond the available width of the screen it doesn't wrap. I tried adding 

`class="d-flex flex-wrap justify-content-center"`

to the ngb-pagination html tag but that had no effect. It needs to be added the ul tag that has the pagination class applied to it to have an effect. As a hacky work-around in my style.scss Sass I use an extend on the .pagination class to include those other classes:
```
@import "~bootstrap/scss/bootstrap";

.pagination
{
  @extend .d-flex, .flex-wrap, .justify-content-center;
}
```
If you remove the above class extend in the stackblitz's style.scss example linked below you can see the default behavior.

For those who wish to not implement Sass, the following in style.css should also work:

```
.pagination
{
  display: -webkit-box;      /* OLD - iOS 6-, Safari 3.1-6 */
  display: -moz-box;         /* OLD - Firefox 19- (buggy but mostly works) */
  display: -ms-flexbox;      /* TWEENER - IE 10 */
  display: -webkit-flex;     /* NEW - Chrome */
  display: flex;             /* NEW, Spec - Opera 12.1, Firefox 20+ */
  justify-content: center;
  -webkit-justify-content: center;
  flex-wrap : wrap;
  -webkit-flex-wrap: wrap;
}
```

### Link to minimally-working StackBlitz that reproduces the issue:
https://stackblitz.com/edit/bootstrap4paginationflex

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 7.0.4

ng-bootstrap: 4.0.0

Bootstrap: 4.1.3

pkozlowski-opensource:
> When pagination has enough pages to extend beyond the available width of the screen it doesn't wrap.

@talynone this is not what I'm seeing... What is the browser that you are using? Could you please share screen-shoots of the expected vs. actual rendering?

talynone:
@pkozlowski-opensource  I'm using Chrome x64 Version 70.0.3538.102 on Windows 10. You have to comment out the ".pagination" override in style.scss to see it without my fix. 

Expected Rendering:
<img width="960" alt="ng bs4 pagination - way it should look" src="https://user-images.githubusercontent.com/2554644/48964533-5f526600-ef5e-11e8-8bb3-c32de6791d16.png">

Actual Rendering:
[...]

PR #3477, opened 2019-11-20
feat(pagination): accept an "align" attribute

terencehonles:
Currently because there is no validation on "size" you can stuff additional classes into the "size" attribute to perform Bootstrap alignment on the pagination component. This change allows a user to easily, and in a supported manner, pass flexbox utility classes such as "justify-content-center" to the pagination component.

### DIFF
--- a/src/pagination/pagination.ts
+++ b/src/pagination/pagination.ts
@@ -128,7 +128,7 @@ export class NgbPaginationPrevious {
       {{ page }}
       <span *ngIf="page === currentPage" class="sr-only">(current)</span>
     </ng-template>
-    <ul [class]="'pagination' + (size ? ' pagination-' + size : '')">
+    <ul [class]="'pagination' + (size ? ' pagination-' + size : '') + (align ? ' ' + align : '')">
       <li *ngIf="boundaryLinks" class="page-item"
         [class.disabled]="previousDisabled()">
         <a aria-label="First" i18n-aria-label="@@ngb.pagination.first-aria" class="page-link" href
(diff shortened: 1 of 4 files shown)


pkozlowski-opensource:
But this is already possible without any additional `@Input()`s: https://ng-bootstrap.github.io/#/components/pagination/examples#justify

terencehonles:
@pkozlowski-opensource I didn't try that. That will work with simple justification, but I don't think it will work with `flex-wrap` since what it is doing is adding another layer of flex rather than actually adjusting the layout of the children. I could be wrong :thinking:

src/pagination/pagination.ts
```diff
@@ -191,6 +191,13 @@ export class NgbPagination implements OnChanges {
   @ContentChild(NgbPaginationNumber, {static: false}) tplNumber: NgbPaginationNumber;
   @ContentChild(NgbPaginationPrevious, {static: false}) tplPrevious: NgbPaginationPrevious;
 
+  /**
+   * The pagination alignment.
+   *
+   * Any of the Bootstrap flex utilties (i.e. "justify-content-center").
+   */
+  @Input() align: string;
```
maxokorokov: IMO you should have done either:
- `@Input() classNames: string` or
- `@Input() align: 'start' | 'center' | 'end' | etc... '`

terencehonles:
@maxokorokov based on #2877 (and your comment) I would suggest the classNames property be considered and we just change this PR to do that. I know you *can* select the pagination component that way, but that won't allow you to use the bootstrap flexbox utilities which is *why* they are there.

I currently have a working solution to stuff them in via the size attribute, and this PR was merely to address the fact that the work around _might_ break on us. After writing this PR and looking at the tests on the size attribute it is actually unlikely that will happen, but I'm guessing you'll get comments about this from people that are more familiar with bootstrap and are less familiar with this library and how it differs from bootstrap itself.

src/pagination/pagination.ts
terencehonles: I didn't call it classNames since I didn't think that was clear where the class names were going, and there was already a size attribute it seemed like it would make sense to call it align. I wasn't going to enumerate all the values because there are much more than are used for the size parameter, and unless you have ivy and strict type checking enabled you're not going to know you're passing an "invalid" type (I haven't seen the warning myself, and I'm in the processes of testing ivy with our application so I've not had a chance to ratchet up the type checking)

PR #3482, opened 2019-11-21
Update modal-config.ts to allow multiple bs modal sizes

alecsvaldez:
Accept all values for modal sizes, even accept user-defined custom class names

### DIFF
--- a/src/modal/modal-config.ts
+++ b/src/modal/modal-config.ts
@@ -69,8 +69,9 @@ export interface NgbModalOptions {
 
   /**
    * Size of a new modal window.
+   * 'sm' | 'md' | 'lg' | 'xl' | 'custom-class-name'
    */
-  size?: 'sm' | 'lg' | 'xl';
+  size?: string;
 
   /**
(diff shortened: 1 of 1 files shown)


src/modal/modal-config.ts
```diff
@@ -69,8 +69,9 @@ export interface NgbModalOptions {
 
   /**
    * Size of a new modal window.
+   * 'sm' | 'md' | 'lg' | 'xl' | 'custom-class-name'
    */
-  size?: 'sm' | 'lg' | 'xl';
+  size?: string;
```
benouat: Couldn't we keep `'sm' | 'md' | 'lg' | 'xl' | string` here ?

@maxokorokov don't we have something similar somewhere already ?

peterblazejewicz: 2c, the `md` would have no effect, it is implicit one and there is no class for that size:
https://getbootstrap.com/docs/4.4/components/modal/#optional-sizes

PR #3491, opened 2019-11-28
feat(all): codelyzer accessibility linting

gpolychronisAmadeus:
Add accessibility rules in linting phase.

In current PR the accessibility rules of Codelyzer are used.
Ideally, we should find a linter that tests proper use of WCAG attributes like tslint-react-a11y does for React.

Codelyzer is proposed by blog.angular.io:
https://blog.angular.io/accessibility-in-angular-e84f73a223f

If the linter is approved, I will proceed in fixing all related errors, so that it can be included in CI.

### DIFF
--- a/src/modal/modal.spec.ts
+++ b/src/modal/modal.spec.ts
@@ -876,12 +876,6 @@ export class WithSkipTabindexFirstFocusableModalCmpt {
       </ng-template>
     </ng-template>
     <button id="open" (click)="open('from button')">Open</button>
-    <div id="open-no-focus" (click)="open('from non focusable element')">Open</div>
-    <div
-      id="open-no-focus-ie"
-      (click)="open('from non focusable element but stored as activeElement on IE')"
-      style="display: inline-block;"
-    >Open</div>
   `
 })
 class TestComponent {
--- a/src/tslint.json
+++ b/src/tslint.json
@@ -1,3 +1,18 @@
 {
-  "extends": "../tslint.json"
+  "extends": "../tslint.json",
+  "rulesDirectory": [
+    "../node_modules/codelyzer"
+  ],
+  "rules": {
+    "template-accessibility-alt-text": true,
+    "template-accessibility-elements-content": true,
+    "template-accessibility-label-for": true,
+    "template-accessibility-tabindex-no-positive": true,
+    "template-accessibility-table-scope": true,
+    "template-accessibility-valid-aria": true,
+    "template-click-events-have-key-events": true,
+    "template-mouse-events-have-key-events": true,
+    "template-no-autofocus": true,
+    "template-no-distracting-elements": true
+  }
 }
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "@types/prismjs": "1.16.0",
     "bootstrap": "4.3.1",
     "clang-format": "1.0.35",
+    "codelyzer": "^5.2.0",
     "concurrently": "^4.1.0",
     "conventional-changelog-cli": "^2.0.12",
     "core-js": "^2",
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -45,6 +45,7 @@ export class NgbSlide {
  *
  * Allows to set intervals, change the way user interacts with the slides and provides a programmatic API.
  */
+/* tslint:disable:template-click-events-have-key-events */
 @Component({
   selector: 'ngb-carousel',
   exportAs: 'ngbCarousel',
--- a/src/datepicker/datepicker-month-view.ts
+++ b/src/datepicker/datepicker-month-view.ts
@@ -4,6 +4,7 @@ import {NgbDate} from './ngb-date';
 import {NgbDatepickerI18n} from './datepicker-i18n';
 import {DayTemplateContext} from './datepicker-day-template-context';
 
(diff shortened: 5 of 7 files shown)


gpolychronisAmadeus:
@maxokorokov @ExFlo @benouat 
Please provide feedback

benouat:
`codelyzer` is already part of any Angular CLI project. Thus I think we should be able to declare the rules directly, but I am not sure if we still should declare it in the `rulesDirectory`, to be double checked with @maxokorokov 

Then I have one question about such a tool, how do you locally disable a rule ? Because here we are not in a `.ts` file, but I assume the lint will be executed on the compiled version of the html template.

gpolychronisAmadeus:
> 
> 
> `codelyzer` is already part of any Angular CLI project. Thus I think we should be able to declare the rules directly, but I am not sure if we still should declare it in the `rulesDirectory`, to be double checked with @maxokorokov
> 
> Then I have one question about such a tool, how do you locally disable a rule ? Because here we are not in a `.ts` file, but I assume the lint will be executed on the compiled version of the html template.

In codelyzer's accessibility guide (https://web.dev/accessible-angular-with-codelyzer/) the codelyzer path is included in relusDirectory. I tried locally unsuccessfully to avoid adding it explicitly.

Also, codelyzer produces messages that refer to *.ts files and runs upon tslint, so probably it evaluates *.ts files themselves.
I haven't found how to deactivate it locally.

Codelyzer is proposed by blog.angular.io:
https://blog.angular.io/accessibility-in-angular-e84f73a223f

tslint.json
```diff
@@ -80,6 +81,16 @@
       true,
       "always"
     ],
+    "template-accessibility-alt-text": true,
```
maxokorokov: I'd say we should **not** do this for all applications (`ng-bootstrap`, `demo`, `e2e-app`, `ssr-app`), this should be only for the `ng-bootstrap` in `src/tslint.json`

gpolychronisAmadeus:
> Then I have one question about such a tool, how do you locally disable a rule ? Because here we are not in a `.ts` file, but I assume the lint will be executed on the compiled version of the html template.

Add comment:
`/* tslint:disable:template-click-events-have-key-events */`
Just before the `@Component` annotation. Also `<!-- tslint:disable:template-click-events-have-key-events -->` comment just before the element in template might work.

It does not interfere with the documentation mechanism.

tslint.json
gpolychronisAmadeus: > 
> 
> I'd say we should **not** do this for all applications (`ng-bootstrap`, `demo`, `e2e-app`, `ssr-app`), this should be only for the `ng-bootstrap` in `src/tslint.json`

Done.

src/modal/modal.spec.ts
```diff
@@ -876,12 +876,6 @@ export class WithSkipTabindexFirstFocusableModalCmpt {
       </ng-template>
     </ng-template>
     <button id="open" (click)="open('from button')">Open</button>
-    <div id="open-no-focus" (click)="open('from non focusable element')">Open</div>
```
maxokorokov: Any reason for removing this part of the test?

src/tslint.json
```diff
@@ -1,3 +1,18 @@
...
+    "../node_modules/codelyzer"
+  ],
+  "rules": {
+    "template-accessibility-alt-text": true,
+    "template-accessibility-elements-content": true,
+    "template-accessibility-label-for": true,
+    "template-accessibility-tabindex-no-positive": true,
+    "template-accessibility-table-scope": true,
+    "template-accessibility-valid-aria": true,
+    "template-click-events-have-key-events": true,
```
maxokorokov: I was more thinking about deactivating / removing `template-click-events-have-key-events` and `template-mouse-events-have-key-events` completely from our configuration, not disabling them on per-widget basis.

Reasoning being that rule is too simplistic for us, we don't add event handlers using the `(keyup)`, but manually with typescript. Also I don't really understand why enforce this rule on links and buttons (aslo often the case in ng-bootstrap)...

WDYT?

src/modal/modal.spec.ts
gpolychronisAmadeus: This part does not pass linter.

It does not affect CI, so removing it completely seems better than fixing it.

src/tslint.json
gpolychronisAmadeus: > 
> 
> I was more thinking about deactivating / removing `template-click-events-have-key-events` and `template-mouse-events-have-key-events` completely from our configuration, not disabling them on per-widget basis.
> 
> Reasoning being that rule is too simplistic for us, we don't add event handlers using the `(keyup)`, but manually with typescript. Also I don't really understand why enforce this rule on links and buttons (aslo often the case in ng-bootstrap)...
> 
> WDYT?

The rule is too simplistic, that's true. However, it's still a reminder that for every click handler there should be a keyboard handler. Personally, I prefer disabling the rule per case than accidentally creating an inaccessible component. WDYT?

maxokorokov: Well we don't create components without keyboard support here anyway :) To me, all 4 cases it reported in ng-bootstrap were completely invalid and the less case-by-case exceptions we have - the better.

maxokorokov: After going through all the rules in codelyzer, I'm not at all convinced they're very useful in our library scope :/ It's more for the end apps from what I feel...

PR #3493, opened 2019-11-29
chore: add strict unused imports/variables check at TS level

maxokorokov:
Adds TS level checking for unused imports as a part of CI

Just adding `noUnusedLocals` to the `tsconfig.json` might be too restrictive during development and ng-packagr doesn't build the library correctly when it is activated.

### DIFF
--- a/misc/check-ts.ts
+++ b/misc/check-ts.ts
@@ -1,0 +1,39 @@
+import {join} from 'path';
+import {spawn} from 'cross-spawn';
+import {ChildProcess, SpawnOptions} from 'child_process';
+
+const options: SpawnOptions = {
+  cwd: join(__dirname, '..')
+};
+
+const tsConfigs = [
+  'src/tsconfig.json', 'src/tsconfig.spec.json', 'demo/tsconfig.json', 'e2e-app/tsconfig.json',
+  'e2e-app/tsconfig.spec.json'
+];
+
+const args = ['--pretty', '--noEmit', '--noUnusedLocals'];
+
+const messages = new Set<string>();
+const running = new Set<ChildProcess>();
+
+tsConfigs.forEach(tsConfig => {
+  const child = spawn('npx', ['tsc', ...args, `--project`, tsConfig], options);
+  running.add(child);
+
+  child.stdout.on('data', chunk => {
+    chunk.toString().split('\n').filter(message => message.length > 0).forEach(message => messages.add(message));
+  });
+
+  child.on('exit', () => {
+    running.delete(child);
+    if (running.size === 0 && messages.size > 0) {
+      messages.forEach(message => console.error(`\x1b[31m${message}`, '\x1b[0m'));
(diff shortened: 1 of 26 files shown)


maxokorokov:
@benouat could you please check if https://github.com/ng-bootstrap/ng-bootstrap/pull/3493/files#diff-50d3ac6bdf4783d9637b987cda7b85e9 is fine ?

All other changes are removal of unused code

src/datepicker/datepicker-input.ts
```diff
@@ -263,7 +263,7 @@ export class NgbInputDatepicker implements OnChanges,
   constructor(
       private _parserFormatter: NgbDateParserFormatter, private _elRef: ElementRef<HTMLInputElement>,
       private _vcRef: ViewContainerRef, private _renderer: Renderer2, private _cfr: ComponentFactoryResolver,
-      private _ngZone: NgZone, private _service: NgbDatepickerService, private _calendar: NgbCalendar,
+      private _ngZone: NgZone, service: NgbDatepickerService, private _calendar: NgbCalendar,
```
peterblazejewicz: this change marks this as unused (`'service' is declared but its value is never read.ts(6133)`), should be probably just removed

maxokorokov: Good catch, @peterblazejewicz, thanks! Will update the PR

misc/check-ts.ts
```diff
@@ -0,0 +1,35 @@
...
+  'e2e-app/tsconfig.spec.json'
+];
+
+const args = ['--noEmit', '--noUnusedLocals'];
+
+const messages = new Set<string>();
+const running = new Set<ChildProcess>();
+
+tsConfigs.forEach(tsConfig => {
+  const child = spawn('npx', ['tsc', ...args, `--project`, tsConfig], options);
```
gpolychronisAmadeus: Could you please add error handling?
i.e.
`.on('error', function( err ){ throw err })`
as mentioned in https://stackoverflow.com/questions/27688804/how-do-i-debug-error-spawn-enoent-on-node-js

This might be useful when it doesn't work on a local environment.

Also this file is added to avoid adding `noUnusedLocals `and `noUnusedParameters `in .tsconfig.json, right?

maxokorokov: @gpolychronisAmadeus good idea, will update it.

> Also this file is added to avoid adding noUnusedLocals and noUnusedParameters in .tsconfig.json, right?

Yes, it's the one that is executed with `yarn check-ts`

gpolychronisAmadeus: @maxokorokov 
As discussed:

- In windows environment it doesn't work. Please add `shell: true` in options.
- Also please reset the color.

maxokorokov: @gpolychronisAmadeus thanks for checking it on Windows, fixed both in the update.

PR #3496, opened 2019-11-29
chore(all): enable "noUnusedLocals" and "noUnusedParameters"

gpolychronisAmadeus:
Enable flags noUnusedLocals and noUnusedLocals.
These flags raise errors when there are unused imports. At the same time the raise error if there is an unused local, instance variable or an unused injection.

Also, it enforced _<name> naming for unused arguments in functions.

The aim of this commit is to make development independent of the IDE used.

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -839,6 +839,6 @@ class TestComponent {
     {id: 'two', disabled: false, title: 'Panel 2', content: 'bar', type: ''},
     {id: 'three', disabled: false, title: 'Panel 3', content: 'baz', type: ''}
   ];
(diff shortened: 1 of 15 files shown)


gpolychronisAmadeus:
@maxokorokov @benouat @ExFlo 
Please provide feedback. Shall I continue with the rest errors, so that the commit can be merged into CI?

gpolychronisAmadeus:
Similar issue: feat(all): codelyzer accessibility linting #3491

gpolychronisAmadeus:
> 
> 
> Hey, @gpolychronisAmadeus!
> 
> I've actually opened a similar PR today → #3493, PTAL.
> 

It seems we had the same idea after all :D I closed the issue as duplicate.

PR #3498, opened 2019-12-02
fix(modal): fix scrollbar compensation in some scaling situations

### LINKED ISSUE #3448: Bad scrollbar replacement on non-integer width windows
### Bug description:

When a modal is shown on a scrollable page, the scrollbar is hidden and the page body is padded to compensate for the lost width. Currently the decision, _whether_ padding should be applied for that reason, is sometimes taken incorrectly if the browser window has a non-integer width. In an extreme case this can lead to the same padding being applied over and over again (cf. StackBlitz).

Non-integer widths occur when zooming in on the page with the browser, or when scaling your display in system settings like [these in Windows](https://www.windowscentral.com/how-set-custom-display-scaling-setting-windows-10).

https://stackblitz.com/edit/ngb-scrollbar-handling

Please make sure to follow the instructions on screen.

Angular: 8.2.13

ng-bootstrap: 5.1.2

Bootstrap: 4.3.1

### Resolution sketch:

Some rounding needs to be applied to [util/scrollbar.ts](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/util/scrollbar.ts#L53).

ymeine:
Now the actual width of the scrollbar is used to detect if it is present or not, instead of using the wrong hypothesis that if body's rect's width is smaller than viewport's width then there is a scrollbar. Now the difference in width is compared to the width of the scrollbar.

Closes #3448

----

I am sorry I am mixing the fix along with a bit of variables renaming and comments rewriting (but I felt separating commits/PRs would be too cumbersome, and at the same time the code was not self-explanatory enough).

The core of the fix really resides: 

- in `compensate` which now passes the scrollbar width not only to `_adjustBody` but also to `_isPresent`
- and in `_isPresent` which now uses this width to refine widths comparison and more accurately determine if a scrollbar is present or not

### DIFF
--- a/src/util/scrollbar.ts
+++ b/src/util/scrollbar.ts
@@ -22,25 +22,31 @@ export class ScrollBar {
   constructor(@Inject(DOCUMENT) private _document: any) {}
 
   /**
-   * Detects if a scrollbar is present and if yes, already compensates for its
-   * removal by adding an equivalent padding on the right of the body.
+   * To be called right before a potential vertical scrollbar would be removed:
+   *
+   * - if there was a scrollbar, adds some compensation padding to the body
+   * to keep the same layout as when the scrollbar is there
+   * - if there was none, there is nothing to do
    *
    * @return a callback used to revert the compensation (noop if there was none,
    * otherwise a function removing the padding)
    */
-  compensate(): CompensationReverter { return !this._isPresent() ? noop : this._adjustBody(this._getWidth()); }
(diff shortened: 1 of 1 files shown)


fbasso:
Just an idea, but maybe a pure css solution would be simplier:

https://stackoverflow.com/questions/1417934/how-to-prevent-scrollbar-from-repositioning-web-page

The point is to apply : padding-left: calc(100vw - 100%); on the body.

PR #3499, opened 2019-12-03
Migration to @commitlint with husky integration

benouat:
Lets use `@commitlint` with `husky` and `@commitlint/config-angular` on commit msg

Be aware that with this change, `chore` type is not allowed anymore.

I also implemented a dynamic computation of the scopes that can be used based on the name of our widgets + `demo` & `demos`

```
- demo
- demos
- accordion
- alert
- buttons
- carousel
- collapse
- datepicker
- dropdown
- modal
- pagination
- popover
- progressbar
- rating
- tabset
- timepicker
- toast
- tooltip
- typeahead
```


benouat:
Ok people interested into reviewing this PR, go ahead!

But you can't not test it live :( as apparently **husky** is having difficulties installing properly with `yarn` ... git hooks are not installed. (work though with `npm` or by manually calling husky install

peterblazejewicz:
What about the blank line spec? It defaults to warning in importent config, so what about config change:

```js
rules: {
    "body-leading-blank": [2, "always"],
    "scope-enum": [2, "always", scopes]
  }
```
?
This will change warning `⚠   body must have leading blank line [body-leading-blank]` (and pass) to actual error `✖   body must have leading blank line [body-leading-blank]`

benouat:
> We could add something like this in `preinstall.js`, works on my machine, but should be x-platform :)

I don't think we should do that. We should document in `DEVELOPER.md` how to setup properly everything, and that should be enough.

maxokorokov:
@benouat, LGTM. Could you please just update documentation as you've suggested?

PR #3500, opened 2019-12-03
feat(accordion): Adding animations to accordion

andreRBarata:
Some work on #295, adding animations for both ngIf and CSS implementations of collapsing behaviour.

Also my first pull request :) I can make changes if need.

### DIFF
--- a/src/accordion/accordion.module.ts
+++ b/src/accordion/accordion.module.ts
@@ -1,6 +1,6 @@
 import {NgModule} from '@angular/core';
 import {CommonModule} from '@angular/common';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
+import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
 
 import {NgbAccordion, NgbPanel, NgbPanelTitle, NgbPanelContent, NgbPanelHeader, NgbPanelToggle} from './accordion';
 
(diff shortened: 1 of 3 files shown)


edthenet:
When will this functionality be available for production release?
Collapsing is default functionality of Bootstrap (https://getbootstrap.com/docs/4.0/components/collapse/)

benouat:
Hey @andreRBarata !

Thank you for the effort! Unfortunately I don't think we will ever be able to merge your PR. So I will decline it and close it, **but** ...

#### Let me explain why 💡
Animation is what we struggled the most with for a very long time 😔. Being based on Bootstrap it was clear to us that **we did not** want to replicate their animations using Angular animations system. (hence why your PR is out...)

We want to have a kind a transparent way to rely on Bootstrap CSS animations on our Angular widgets/directives. The most obvious reason is that people could have their own custom theme, in which they could change any of the transitions, and **we still want** their custom animations/transition code to work in their app.

With Angular animations we would have hardcoded default Bootstrap animations, with no simple (even maybe none) way to override them nicely.

The "_official_" PR to follow is #2817. We are almost done. If you want to help on the topic, what you could do here is two folds:

- you could review the PR is you feel so
- you could checkout  #2817 branch, build `ng-bootstrap` locally and test in **your** codebase that everything is working.

PR #3502, opened 2019-12-05
feat(datepicker): add datepicker popover example without using the di…

gpolychronisAmadeus:
…rective

An example of using component datepicker in popup.


benouat:
I'll just had a quick look at live code, trying to play with it, here is my first points:

- [ ] It did not open in the right place the first time I opened it. (it opened on top centered, I can't reproduce it... weird)
- [ ] focustrap is not working (maybe you did not used it...)
- [ ] Initial binding is broken (input contains undefined)
- [ ] popup should not open when input is focused → it feels weird to have the datepicker reacting visually to live inputs
- [ ] the popover is too low below the input. Should be closer to it
- [ ] There is an issue with the popover itself, on my test, the next demo title is overlapping the popover 


Nitpick 💄
- [ ] We should keep the same settings as before the change → display a single month

gpolychronisAmadeus:
> 
> 
> I'll just had a quick look at live code, trying to play with it, here is my first points:
> 
>     * [ ]  It did not open in the right place the first time I opened it. (it opened on top centered, I can't reproduce it... weird)
> 
>     * [ ]  focustrap is not working (maybe you did not used it...)
> 
>     * [ ]  Initial binding is broken (input contains undefined)
> 
>     * [ ]  popup should not open when input is focused → it feels weird to have the datepicker reacting visually to live inputs
> 
>     * [ ]  the popover is too low below the input. Should be closer to it
> 
>     * [ ]  There is an issue with the popover itself, on my test, the next demo title is overlapping the popover
> 
> 
> Nitpick 💄
> 
>     * [ ]  We should keep the same settings as before the change → display a single month

I updated the code.
In order to use focus trap I had to add a flag in Popover. This probably requires tests and maybe examples of other use cases, or even a new issue.

gpolychronis:
Use dropdown/dropup instead of popover.

PR #3503, opened 2019-12-05
test: activate e2e for other browsers than Chrome

fbasso:
At the moment, just a try with FF.


benouat:
I am closing this PR as it seems that the time spent trying to make it work it definitely not worth it.

Issue #3506, opened 2019-12-06
Dynamic change of minDate and maxDate doesn't reflect on DatePicker popover


arvindm95:
### Bug description:
When datepicker is used in a popup, then dynamic changes to minDate and maxDate values doesn't reflect on the popup. But it works when the popup is closed and opened again.

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-zpsmfq
1. Click on the toggle button as soon as the page loads.
2. After 5 seconds, notice the maxDate/minDate change in the Datepicker on the right, but not in the one on the left side of the page.
3. Close the toggle and open again, maxDate/minDate will be reflected now.

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 8.2.4

ng-bootstrap: 5.1.4

Bootstrap: 4.3.1

PR #3507, opened 2019-12-06
fix(datepicker): reflect changes on popover when minDate, maxDate val…

arvindm95:
…ues change dynamically (#3506)

Fixes: #3506

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -410,6 +410,16 @@ export class NgbInputDatepicker implements OnChanges,
   ngOnChanges(changes: SimpleChanges) {
     if (changes['minDate'] || changes['maxDate']) {
       this._validatorChange();
+
+      if (this.isOpen()) {
+        if (changes['minDate']) {
+          this._cRef.instance.minDate = this._dateAdapter.toModel(changes.minDate.currentValue);
+        }
+        if (changes['maxDate']) {
+          this._cRef.instance.maxDate = this._dateAdapter.toModel(changes.maxDate.currentValue);
+        }
+        this._cRef.instance.ngOnChanges(changes);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/datepicker/datepicker-input.ts
```diff
@@ -410,6 +410,16 @@ export class NgbInputDatepicker implements OnChanges,
...
       this._validatorChange();
+
+      if (this.isOpen()) {
+        if (changes['minDate']) {
+          this._cRef.instance.minDate = this._dateAdapter.toModel(changes.minDate.currentValue);
+        }
+        if (changes['maxDate']) {
+          this._cRef.instance.maxDate = this._dateAdapter.toModel(changes.maxDate.currentValue);
+        }
+        this._cRef.instance.ngOnChanges(changes);
```
peterblazejewicz:
```ts
this._cRef.instance.ngOnChanges(changes);
```
I believe this bit is not required,  or at least should not directly use `NgOnChanges` and be changed to use of CDR:
```ts
this._cRef.changeDetectorRef.detectChanges();
```

arvindm95: Dynamic components loading will not trigger `ngOnChanges` in child component even after calling `detectChanges` because this event is reserved only for template bindings. 
Here, `datepicker-input` dynamically uses `datepicker` component and hence this will not trigger ngOnChanges. So, It has to be done manually. 
This is the reason behind the above case.

arvindm95:
@maxokorokov Can you please review this? Thanks!

arvindm95:
@maxokorokov Can you please look into this?

arvindm95:
> LGTM

Can you please merge if this is fine? @peterblazejewicz

peterblazejewicz:
I can't
/cc @maxokorokov

PR #3530, opened 2020-01-06
fix(datepicker): refine focus state checks

### LINKED ISSUE #3494: bug(datepicker): Day focus doesn't work if more than 1 datepicker in same page
### Bug description:
In a page that contains more than one datepicker instances (i.e. https://ng-bootstrap.github.io/#/components/datepicker/examples):
* Tap on a day on one datepicker (i.e. on "Basic datepicker" example).
* Check with keyboard that focusing a day works properly (i.e. press left arrow key).
* Tap on a day on the other datepicker (i.e. on "Multiple months" example).
* Check with keyboard that focusing day doesn't work (i.e. press left arrow key).

peterblazejewicz:
This adds additional condition to restrict focus in/out to instance of
the datepicker by verifyng if focus event target and related targets are
descendants of the datepicker element. This removes false results based
only on the class name checks, which fail when focus is switched between
datepicker instances.

Closes #3494

/cc
@gpolychronisAmadeus

### DIFF
--- a/src/datepicker/datepicker.ts
+++ b/src/datepicker/datepicker.ts
@@ -388,7 +388,9 @@ export class NgbDatepicker implements OnDestroy,
           .pipe(
               filter(
                   ({target, relatedTarget}) =>
-                      !(hasClassName(target, 'ngb-dp-day') && hasClassName(relatedTarget, 'ngb-dp-day'))),
+                      !(hasClassName(target, 'ngb-dp-day') && hasClassName(relatedTarget, 'ngb-dp-day') &&
+                        this._elementRef.nativeElement.contains(target as Node) &&
+                        this._elementRef.nativeElement.contains(relatedTarget as Node))),
               takeUntil(this._destroyed$))
           .subscribe(({type}) => this._ngZone.run(() => this._service.focusVisible = type === 'focusin'));
     });


src/datepicker/datepicker.ts
```diff
@@ -388,7 +388,9 @@ export class NgbDatepicker implements OnDestroy,
           .pipe(
               filter(
                   ({target, relatedTarget}) =>
-                      !(hasClassName(target, 'ngb-dp-day') && hasClassName(relatedTarget, 'ngb-dp-day'))),
+                      !(hasClassName(target, 'ngb-dp-day') && hasClassName(relatedTarget, 'ngb-dp-day') &&
+                        this._elementRef.nativeElement.contains(target as Node) &&
+                        this._elementRef.nativeElement.contains(relatedTarget as Node))),
```
benouat: Please create a local reference `const {nativeElement} = this._elementRef;` outside of the Observable chain.

maxokorokov:
@peterblazejewicz or actually even a unit test **might** be fine, I think, if you simulate `.click()` between two datepicker dates and then press any arrow key. It will lose focus now.

maxokorokov:
Closing in favour of #3549 that is based on this commit + has tests

PR #3532, opened 2020-01-07
fix(modal): remove key/mouse handlers only when modal is destroyed

### LINKED ISSUE #3515: Dismissing one of stacked modals dismisses keyboard/backdrop events for all modals
## Bug description:

When using stacked modals to prevent dismissal of first level modal, dismissing the second level modal by keyboard/backdrop disables the keyboard/backdrop for first modal too.

https://stackblitz.com/edit/angular-dhs8oz

Angular: 8.2.14

ng-bootstrap: 5.1.4

Bootstrap: 4.4.1

maxokorokov:
At the moment key/mouse handlers are removed when modal triggers dismiss event. The problem is that modal closure could be cancelled with `beforeDismiss` and modal will stay opened, but handlers will be removed.

This commit introduces `closed$` event triggered inside `onDestroy` to clean up handlers.

Fixes #3515


maxokorokov:
@benouat could you PTAL?

PR #3535, opened 2020-01-08
feat: add `ViewEncapsulation.None` to the rest of components

### LINKED ISSUE #3479: Tabset ViewEncapsulation
Hey guys, hope you are doing well

what do you think about changing the View Encapsuation of NgbTabset to None? as we really needed to change it quite often

Thanks in advance!

maxokorokov:
Previously it was added only to components with custom styles

Fixes #3479

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -9,7 +9,8 @@ import {
   Optional,
   Output,
   QueryList,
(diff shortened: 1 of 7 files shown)


maxokorokov:
@benouat @IAfanasov thanks for checking this

PR #3539, opened 2020-01-09
feat(datepicker): add 'restoreFocus' input

### LINKED ISSUE #3483: bug(datepicker): regression with #3371
### Bug description:

I'm opening this as it's own issue so it doesn't get lost. The issue #3317 and its associated PR #3371 caused what probably should be considered a regression. The scenario is with the following code:

```html
<input #picker ngbDatepicker (focus)="picker.open()" />
```

you will never be able to close the datepicker because closing it will call focus which will then in turn open it.

This was also reported by @chumbalum and @zm3d3ni who provided the stackblitz https://stackblitz.com/edit/angular-phaczz as a more complete example of what should continue to work. 

There are some comments on #3317 and suggestions for a possible solution.

maxokorokov:
`restoreFocus` input will define re-focusing strategy when closing the input datepicker.

```ts
@Input() restoreFocus: true | string | HTMLElement; // defaults to true
```

By default it re-focuses element focused at the moment of datepicker opening.
Otherwise accepts either selector or element reference.

Fixes #3483

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -33,6 +33,7 @@ import {NgbDateParserFormatter} from './ngb-date-parser-formatter';
 import {NgbDateStruct} from './ngb-date-struct';
 import {NgbInputDatepickerConfig} from './datepicker-input-config';
 import {NgbDatepickerConfig} from './datepicker-config';
+import {isString} from '../util/util';
 
 const NGB_DATEPICKER_VALUE_ACCESSOR = {
   provide: NG_VALUE_ACCESSOR,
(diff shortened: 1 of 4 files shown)


peterblazejewicz:
How one could just disable focus behaviour?
```ts
<input ngbDatepicker [restoreFocus]="false" [startDate]="{year: 2018, month: 3}"/>
<input ngbDatepicker restoreFocus="false" [startDate]="{year: 2018, month: 3}"/>
```
(original use-case). Would that be supported?

maxokorokov:
@peterblazejewicz this is intentional force focus something, as focus should not be lost ideally.

Though you can always do:
- `restoreFocus="body"` → will focus document body
- `[restoreFocus]="false"` or any invalid thing → will focus previously focused element

src/datepicker/datepicker-input.ts
```diff
@@ -373,7 +382,16 @@ export class NgbInputDatepicker implements OnChanges,
...
+        const testElement = this._document.querySelector(this.restoreFocus);
+        if (isFocusable(testElement)) {
+          elementToFocus = testElement;
+        }
+      } else if (isFocusable(this.restoreFocus)) {
+        elementToFocus = this.restoreFocus;
+      }
+
       elementToFocus.focus();
     }
```
fbasso: Usage of isFocusable is problably not useful here, as we know that html elements are handled (just a check for a not null should be enough)

maxokorokov: Thanks, it makes sense. Will remove `isFocusable()` as discussed

PR #3547, opened 2020-01-20
fix(datepicker): add additional checks for 'restoreFocus' in IE

maxokorokov:
As focusing is asynchronous and unstable in IE it is simpler to disable it

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -925,9 +925,9 @@ describe('NgbInputDatepicker', () => {
 
     it('should focus previously focused element', () => {
       const fixture = createTestCmpt(`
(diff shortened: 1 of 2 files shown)


peterblazejewicz:
@maxokorokov 
Can we take those failing CI builds for IE10 as a sign to ditch IE10 from CI tests? The end of Jan. is looming just before us:
https://support.microsoft.com/en-us/help/4488955/support-ending-for-internet-explorer-10

maxokorokov:
@peterblazejewicz, I was actually wrong in where the problem comes from, can't reproduce it locally, so playing here. 

Problem is `document.activeElement` can contain strange things in IE at the moment of opening the popup, so disabling tests won't help :/

Also we'll have to support IE until either Bootstrap or Angular drop it.

PR #3549, opened 2020-01-21
fix(datepicker): focus on multiple instances (#3494)

gpolychronisAmadeus:

 - [ ] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) guide.

### DIFF
--- a/e2e-app/src/app/datepicker/datepicker.po.ts
+++ b/e2e-app/src/app/datepicker/datepicker.po.ts
@@ -7,9 +7,9 @@ export class DatepickerPage {
 
   getToggle() { return $('#toggle'); }
 
-  getDayElement(date: Date) {
+  getDayElement(date: Date, datepicker = this.getDatepicker()) {
     const ariaLabel = date.toLocaleString('en', {weekday: 'long', year: 'numeric', month: 'long', day: 'numeric'});
-    return this.getDatepicker().$(`div.ngb-dp-day[aria-label="${ariaLabel}"]`);
+    return datepicker.$(`div.ngb-dp-day[aria-label="${ariaLabel}"]`);
   }
 
   getWeekdayElements() { return this.getDatepicker().$$('div.ngb-dp-weekday'); }
--- a/e2e-app/src/app/datepicker/multiple/datepicker-multiple.component.html
+++ b/e2e-app/src/app/datepicker/multiple/datepicker-multiple.component.html
@@ -1,0 +1,14 @@
+<h3>Datepicker focus tests</h3>
+
+<form>
+  <div>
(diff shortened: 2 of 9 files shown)


maxokorokov:
FYI tests are failing because of an unrelated issue. I just pushed the fix to the master

gpolychronisAmadeus:
Added tests.
@maxokorokov Please review

Also, it seems to fix the performance issue of #3396

e2e-app/src/app/datepicker/multiple/datepicker-multiple.e2e-spec.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  beforeEach(async() => await openUrl('datepicker/multiple'));
+
+  it('should lose focus on tapping on a different datepicker instance', async() => {
+    await page.getDayElement(new Date(2016, 7, 1), 0).click();
+    await sendKey(Key.ARROW_DOWN);
+    await page.expectActive(new Date(2016, 7, 8), 0);
+    await page.getDayElement(new Date(2016, 7, 1), 1).click();
+    await sendKey(Key.ARROW_DOWN);
+    await page.expectActive(new Date(2016, 7, 8), 1);
```
maxokorokov: could you please check here that the day is actually focused? there is an `expectFocused()` utility for this
It still can be active, but not focused

e2e-app/src/app/datepicker/datepicker.po.ts
```diff
@@ -7,9 +7,9 @@ export class DatepickerPage {
 
   getToggle() { return $('#toggle'); }
 
-  getDayElement(date: Date) {
+  getDayElement(date: Date, index = 0) {
```
maxokorokov: This change is very confusing from the readability point of view to me.

Could we replace index with the datepicker, so the test would look something like this (instead of using indexes):

```
const datepicker1 = page.getDatepicker('myDatepicker1');
const element = page.getDayElement(date, datepicker1);
```

e2e-app/src/app/datepicker/multiple/datepicker-multiple.component.html
```diff
@@ -0,0 +1,14 @@
+<h3>Datepicker focus tests</h3>
+
+<form>
+  <div>
+    <ngb-datepicker name="dp1" placeholder="yyyy-mm-dd" #d1 [startDate]="startDate"></ngb-datepicker>
```
maxokorokov: also `name`, `placeholder` and `#d1` are not necessary. Same for the second datepicker

PR #3554, opened 2020-01-22
fix(nav): allow importing 'NgbNavModule' independently

### LINKED ISSUE #3553: NgbNav module is not exported
`NgbNavModule` is missing in https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/index.ts#L92

maxokorokov:
Closes #3553


maxokorokov:
@benouat could you PTAL? and I'll do the 5.2.1 afterwards

PR #3559, opened 2020-01-24
demos: npm downloads count

benouat:
Adding NPM download stats to the header of demos website.
Code is only retrieving the live count when in production mode.

When in dev mode
image

When in production mode
image

### DIFF
--- a/demo/src/app/app.component.ts
+++ b/demo/src/app/app.component.ts
@@ -1,32 +1,34 @@
-import { ViewportScroller } from '@angular/common';
-import { Component, NgZone, OnInit } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
-import { filter } from 'rxjs/operators';
+import {ViewportScroller} from '@angular/common';
+import {HttpClient} from '@angular/common/http';
+import {Component, NgZone, OnInit} from '@angular/core';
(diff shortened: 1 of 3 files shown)


demo/src/app/app.component.ts
```diff
@@ -1,32 +1,33 @@
...
-  }
+      private _analytics: Analytics, route: ActivatedRoute, vps: ViewportScroller, zone: NgZone,
+      httpClient: HttpClient) {
+    route.fragment.pipe(filter(fragment => !!fragment))
+        .subscribe(fragment => zone.runOutsideAngular(() => requestAnimationFrame(() => vps.scrollToAnchor(fragment))));
 
-  ngOnInit(): void {
-    this._analytics.trackPageViews();
+    if (environment.production) {
+      httpClient.get('https://api.npmjs.org/downloads/point/last-month/@ng-bootstrap/ng-bootstrap')
```
maxokorokov: maybe add pipe → `catchError(() => of(''))`or the error handler in subscribe → `() => this.downloadCount = ''`

So it would:
- not shout in the console in case of issues
- display just the `npm` logo without dashes

Screen Shot 2020-01-27 at 11 31 26

benouat: done

PR #3572, opened 2020-01-31
fix(timepicker): When We Enter 60 in Seconds... are not setting to 0

gpolychronis:
Attempt to fix 3546 and 3548.

In this case of binding, origin value (**model.second**) remains the same and destination value (**#seconds.input.value**) changes. Angular does not correct the destination value, because it detects no change in origin value.

Attempted to fix by forcing the update of destination value.....

@maxokorokov @benouat 
Please review


gpolychronis:
As discussed:
- remove "[value]"
- refactor to avoid code duplication

Issue #3584, opened 2020-02-12
NgbModalStack doesn't support container selector existing within a shadowdom


drobins29:
### Bug description:
When providing the selector for the container option when opening a modal, it is impossible to specify an element that exists within a shadow dom. It would be nice to be able to pass a DOM object with the NgbModalOptions to the NgbModal open method. 

My use case is that we are embedding an angular app within an existing application and we do not want external styling to affect the styling of the embedded app. So the main component is using `encapsulation: ViewEncapsulation.ShadowDom` to get around the styling issues and importing the bootstrap css within the main app component's sass file in a parent selector.

If we could pass a DOM element instead of a selector that would allow the modal to pick up the correct styling.

### Link to minimally-working StackBlitz that reproduces the issue:
https://stackblitz.com/edit/angular-wymgqp

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: >=8.2.4

ng-bootstrap: 5.2.2

PabloPollono-Nexplore:
hey I have the same issue and the container option is not working for me, can you show me the fix for the shadow dom + container option?

PR #3585, opened 2020-02-12
feat(modal): pass an `HTMLElement` to modal container option (#3584)

drobins29:

### DIFF
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -46,8 +46,9 @@ export class NgbModalStack {
   }
 
   open(moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any, options): NgbModalRef {
-    const containerEl =
-        isDefined(options.container) ? this._document.querySelector(options.container) : this._document.body;
+    const containerEl = options.container instanceof HTMLElement ? options.container : isDefined(options.container) ?
+                                                                   this._document.querySelector(options.container) :
+                                                                   this._document.body;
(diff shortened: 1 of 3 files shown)


src/util/util.ts
```diff
@@ -26,6 +26,10 @@ export function isDefined(value: any): boolean {
   return value !== undefined && value !== null;
 }
 
+export function isDomObject(value: any): boolean {
```
peterblazejewicz: I think this one is not needed, simple `sth instanceof HTMLElement` works ok even in E11, and is already used in the library for same purpose:
https://github.com/ng-bootstrap/ng-bootstrap/blob/4f61496fff898996e92bd8928d83d5536382aa90/src/datepicker/datepicker-input.ts#L506

src/modal/modal-stack.ts
```diff
@@ -46,8 +46,9 @@ export class NgbModalStack {
   }
 
   open(moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any, options): NgbModalRef {
-    const containerEl =
-        isDefined(options.container) ? this._document.querySelector(options.container) : this._document.body;
+    const containerEl = isDomObject(options.container) ? options.container : isDefined(options.container) ?
```
peterblazejewicz: so the `sth instanceof HTMLElement` woudl return falsy for anything but HTMLElement, allowing to simplify these lines

src/util/util.ts
drobins29: I removed this method and used instanceof check as suggested

PR #3587, opened 2020-02-12
fix(tooltip): prevent autoclose with default triggers on Android

### LINKED ISSUE #3582: Tooltip disappears immediately on Android
### Bug description:

NgbTooltip disappears immediately after opening on mobile devices. Same happens when using DevTools with responsive mode enabled. 

I got it fixed by enabling `wrapAsyncForiOS` for everyone. https://github.com/ng-bootstrap/ng-bootstrap/blob/59581c775181c1c3b811bcce281a5c1daf9fee5f/src/util/autoclose.ts#L23

Of course I could disable `hover` trigger, but I'd rather keep if for desktops. 

edit: I've found a workaround. Add `click` at the beginning of default `triggers` (=> `click hover focus`)

https://angular-n4vyua.stackblitz.io (forked `Quick and easy tooltips` demo from docs)

Angular: 8.2.4
ng-bootstrap: 5.2.2

maxokorokov:
Couldn't find a better alternative at the moment, autoclose needs rethinking...

Fixes #3582

### DIFF
--- a/src/util/autoclose.ts
+++ b/src/util/autoclose.ts
@@ -10,24 +10,22 @@ const isContainedIn = (element: HTMLElement, array?: HTMLElement[]) =>
 const matchesSelectorIfAny = (element: HTMLElement, selector?: string) =>
     !selector || closest(element, selector) != null;
 
-// we'll have to use 'touch' events instead of 'mouse' events on iOS and add a more significant delay
-// to avoid re-opening when handling (click) on a toggling element
+// we have to add a more significant delay to avoid re-opening when handling (click) on a toggling element
 // TODO: use proper Angular platform detection when NgbAutoClose becomes a service and we can inject PLATFORM_ID
-let iOS = false;
-if (typeof navigator !== 'undefined') {
-  iOS = !!navigator.userAgent && /iPad|iPhone|iPod/.test(navigator.userAgent);
-}
+const isMobile = () => typeof navigator !== 'undefined' ?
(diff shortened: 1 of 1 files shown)


src/util/autoclose.ts
```diff
@@ -10,24 +10,22 @@ const isContainedIn = (element: HTMLElement, array?: HTMLElement[]) =>
...
-// we'll have to use 'touch' events instead of 'mouse' events on iOS and add a more significant delay
-// to avoid re-opening when handling (click) on a toggling element
+// we have to add a more significant delay to avoid re-opening when handling (click) on a toggling element
 // TODO: use proper Angular platform detection when NgbAutoClose becomes a service and we can inject PLATFORM_ID
-let iOS = false;
-if (typeof navigator !== 'undefined') {
-  iOS = !!navigator.userAgent && /iPad|iPhone|iPod/.test(navigator.userAgent);
-}
+const isMobile = () => typeof navigator !== 'undefined' ?
+    !!navigator.userAgent && /iPad|iPhone|iPod|Android/.test(navigator.userAgent) :
```
kamilchlebek: Unfortunately this won't work for iPads with iOS >= 13. Since this version `userAgent` does not contain `iPad`.

https://stackoverflow.com/questions/2153877/what-is-the-ipad-user-agent/56923008#56923008

I checked it on browserstack and I can confirm that the issue still persists. This commit did the fix for Android though

maxokorokov: Yikes, thanks for the info. Of course we should not use `userAgent` at all for this...

PR #3590, opened 2020-02-13
ci: add 'demo' type to commit messages

maxokorokov:
Currently the `'demo'` is part of the scopes, not types in commit linting.

So if I want to update one of the accordion demos, it is only possible to have commits like:

```
fix(demo): ...
docs(accordion): ...
```

Also `'fix'` commits would end up in release notes.

I would expect to have commits like:

```
demo(accordion): ...
demo(datepicker): ...
```

That would have a demo site as the scope.

WDYT?

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,17 +1,22 @@
 const fs = require('fs');
+const angularTypes = require('@commitlint/config-angular-type-enum').value();
 
 /*
   Let's get widget names as valid list of scopes.
-  Adding also demo & demos as valid too.
 */
 const scopes = [
-  "demo", "demos", ...fs.readdirSync("./src", {withFileTypes: true})
-                       .filter(d => d.isDirectory())
-                       .filter(d => !["test", "util"].includes(d.name))
-                       .map(d => d.name)
+  ...fs.readdirSync("./src", {withFileTypes: true})
+    .filter(d => d.isDirectory())
+    .filter(d => !["test", "util"].includes(d.name))
+    .map(d => d.name)
 ];
+
+const types = ['demo', ...angularTypes];
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


benouat:
I don't know actually ...
I don't find weird to have `fix(demo): accordion ..` it's actually saying what it does, fixing a demo for a given component

I would nonetheless keep demos in the scope in order to say `feat(demos): stackblitz button ...` or something equivalent

For me scope `demo` meant touching a specific demo with the component as part of the message, and then `demos` meant fixing something on the demo website

commitlint.config.js
```diff
@@ -1,17 +1,22 @@
...
 /*
   Let's get widget names as valid list of scopes.
-  Adding also demo & demos as valid too.
 */
 const scopes = [
-  "demo", "demos", ...fs.readdirSync("./src", {withFileTypes: true})
-                       .filter(d => d.isDirectory())
-                       .filter(d => !["test", "util"].includes(d.name))
-                       .map(d => d.name)
+  ...fs.readdirSync("./src", {withFileTypes: true})
```
benouat: let's keep `demos` here to be able to target changes on the website.

maxokorokov:
@benouat I just would not want `feat/fix(demo): ...` to end up in release notes / changelog, because it's really has nothing to do with the library itself.

benouat:
> I just would not want `feat/fix(demo): ...` to end up in release notes / changelog

valid point, or we could also filter them from changelog

Proceed with your changes, we can always do something else later.

maxokorokov:
@benouat pushed the documentation update with some examples. Do you think you're ok with 

```
demo(alert): → alert demo changes
demo: → any other demo site changes
```

?

Issue #3327, opened 2019-08-10
Tabs are not accessibly via Left Arrow Right Arrow


constantincechina:
### Bug Description
Accordingly with [accessibility best practices](https://www.w3.org/TR/wai-aria-practices/#tabpanel) in regards to the Tabs - Keyboard Interactions
![image](https://user-images.githubusercontent.com/17467157/62827703-9e831e80-bba2-11e9-97d0-11998f6dc3bf.png)

### Solution
Implement listener for keydown.ArrowLeft and keydown.ArrowRight and perform the select operation to set the active tab. The activate automatically is a nice to have but not a must, hence we can have a flag exposed for it.

maxokorokov:
We should implement this for the `NgbNav` directives

maxokorokov:
Closed in #3612

PR #3612, opened 2020-02-20
feat(nav): add keyboard interaction according to accessibility support

Smoggy:
issue [#3327](https://github.com/ng-bootstrap/ng-bootstrap/issues/3327)

 - [ ] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) guide.

### DIFF
--- a/src/nav/nav.spec.ts
+++ b/src/nav/nav.spec.ts
@@ -4,6 +4,7 @@ import {By} from '@angular/platform-browser';
 import {NgbNav, NgbNavConfig, NgbNavItem, NgbNavLink, NgbNavModule, NgbNavOutlet} from './nav.module';
 import {createGenericTestComponent} from '../test/common';
 import {isDefined} from '../util/util';
+import {Key} from 'src/util/key';
 
 function createTestComponent(html: string, detectChanges = true) {
   return createGenericTestComponent(html, TestComponent, detectChanges) as ComponentFixture<TestComponent>;
@@ -37,6 +38,13 @@ function getLinks(fixture: ComponentFixture<any>): HTMLElement[] {
   return fixture.debugElement.queryAll(By.directive(NgbNavLink)).map(debugElement => debugElement.nativeElement);
 }
 
+function createKeyDownEvent(key: number) {
+  const event = {which: key, preventDefault: () => {}, stopPropagation: () => {}};
+  spyOn(event, 'preventDefault');
+  spyOn(event, 'stopPropagation');
+  return event;
+}
+
 function expectLinks(fixture: ComponentFixture<any>, expected: boolean[], shouldHaveNavItemClass = false) {
   const links = getLinks(fixture);
   expect(links.length).toBe(expected.length, `expected to find ${expected.length} links, but found ${links.length}`);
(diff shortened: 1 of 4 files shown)


src/nav/nav.ts
```diff
@@ -119,6 +121,8 @@ export class NgbNavItem implements AfterContentChecked, OnInit {
   isPanelInDom() {
     return (isDefined(this.destroyOnHide) ? !this.destroyOnHide : !this._nav.destroyOnHide) || this.active;
   }
+
+  isCommentNode() { return this.elementRef.nativeElement.nodeType === Node.COMMENT_NODE; }
```
maxokorokov: this will not be necessary

src/nav/nav.ts
```diff
@@ -173,12 +183,20 @@ export class NgbNav implements AfterContentInit {
    */
   @Input() roles: 'tablist' | false;
 
+  /**
+ * Determine if the tabs will be automaticaly changed when user pushes arrow keys(by default false)
+ */
+  @Input() selectOnKeyboardEvent: boolean;
```
maxokorokov: Let's change the API similar to the once we have in carousel:

```ts
@Input() keyboard: boolean | 'changeWithArrows';
```

- `false` → no keyboard support whatsoever
- `true` → keyboards support if `role=tab`
- `'changeWithArrows'` → changes the tab automatically with keyboard arrow (same as `selectOnKeyboardEvent = true` before

If you have better naming suggestions for `changeWithArrows`, don't hesitate.

src/nav/nav.ts
```diff
@@ -196,6 +214,68 @@ export class NgbNav implements AfterContentInit {
...
 
+  onKeyDown(event: KeyboardEvent) {
+    if (this.roles !== 'tablist') {
+      return;
+    }
+    // tslint:disable-next-line: deprecation
+    const key = event.which;
+    const tabs: any[] = this.items.filter(x => !x.disabled);
+
+    const navLinkElements = tabs.map(
```
maxokorokov: I suggest to simplify this significantly using `QueryList<NgbNavLink>`:

1. Need to inject the `public elRef: ElementRef` in the `NgbLavLink` constructor.
2. Query links same as items in the `NgbNav`:

```ts
  @ContentChildren(NgbNavItem) items: QueryList<NgbNavItem>;
  @ContentChildren(forwardRef(() => NgbNavLink), {descendants: true}) links: QueryList<NgbNavLink>;
```

3. Use only enabled links (note that link knows its navItem):

```ts
    const enabledLinks = this.links.filter(link => !link.navItem.disabled);
    const {length} = enabledLinks;

    let position = -1;
    enabledLinks.forEach((link, index) => {
      if (link.elRef.nativeElement === this._document.activeElement) {
        position = index;
      }
    });
```

4. Just do this to focus/select:

```ts
      if (this.selectOnKeyboardEvent) {
        this.select(enabledLinks[position].navItem.id);
      }
      enabledLinks[position].elRef.nativeElement.focus();
```

src/nav/nav.ts
```diff
@@ -196,6 +214,68 @@ export class NgbNav implements AfterContentInit {
...
+
+    let position = -1;
+
+    navLinkElements.forEach((element, index) => {
+      if (element === this._document.activeElement) {
+        position = index;
+      }
+    });
+
+    if (navLinkElements.length) {
```
maxokorokov: Let's maybe factor this out as `const {length} = enabledLinks;` as I suggested in another comment

src/nav/nav.spec.ts
```diff
@@ -861,6 +874,140 @@ describe('nav', () => {
     fixture.detectChanges();
     expectContents(fixture, ['1-false', '2-true'], 1);
   });
+
+  describe(`nav keyboard navigation`, () => {
```
maxokorokov: Code coverage went down with this PR, here is what's not covered (you can see it in `/coverage` after running `yarn test`):

Screen Shot 2020-03-19 at 13 30 16

So, basically at least:

- [ ] vertical keyboard navigation
- [ ] it would also be nice to have at least one test with disabled navs

maxokorokov:
Note to self: CI is a fluke with datepicker tests, check if this will need rebasing when merging

PR #3641, opened 2020-03-10
Enable strict null checks

maxokorokov:
This PR enables `strictNullChecks` for all applications. Works together with `strictTemplates`

Split in multiple commits to simplify navigation. Let's focus on the library code first when reviewing.

### Main library changes
- account for possibility of internal implementation methods and fields returning `null`
- account for specific APIs of standard methods like `parseInt()`, `Array.indexOf()`, `ViewContainerRef.indexOf()`, `QueryList.first`, etc. Null/types handling might be specific for them.
- add `!` in places where we're 100% sure, example:

```ts
// popover.ts
private _windowRef: ComponentRef<NgbPopoverWindow>| null = null;

if (this.isOpen() && ... ) {
  this._windowRef !.instance.class = ...
}
```

- make **some** of the public APIs accept `method(t: T | null | undefined)` instead of `method(t: T)` for convenience, ex, when working with dates:

```ts
class NgbDate {
  equals(date?: NgbDate | null);
}

class NgbCalendar {
  isValid(date?: NgbDate | null);
}

```

- use `<any>null` in places where a variable can't be null, but we still want to nullify the reference when destroying:

```ts
// modal-ref.ts
this._backdropCmptRef = <any>null;
```

- pass `null` and not `undefined` to `[ngTemplateOutlet]`

```html
// nav.ts
[ngTemplateOutlet]="item.contentTpl?.templateRef || null
```

- **Timepicker**: for backward compatibility `NgbTimeStruct` might contain `null` for hours, minutes and seconds despite `NgbTimeStruct` using only `number`s for type definition
- **Datepciker**: for backward compatibility `NgbDateStruct/NgbDate` might contain `null` for years, months and days despite `NgbDateStruct/NgbDate` using only `number`s for type definition

### Unit test changes
- use `<any>null` when constructing manually and don't care about DI
- use `<any>...` casting in places where boundary values like `''` and 'undefined' are tested
- use `textContent!`, `querySelector()!`, `parentElement!`, etc

### Other changes

Demo, e2e and ssr app changes are just to compile with `strictNullChecks` correctly

cc @pkozlowski-opensource

### DIFF
--- a/demo/src/app/components/alert/demos/selfclosing/alert-selfclosing.ts
+++ b/demo/src/app/components/alert/demos/selfclosing/alert-selfclosing.ts
@@ -10,15 +10,15 @@ export class NgbdAlertSelfclosing implements OnInit {
   private _success = new Subject<string>();
 
   staticAlertClosed = false;
-  successMessage: string;
+  successMessage = '';
 
   ngOnInit(): void {
     setTimeout(() => this.staticAlertClosed = true, 20000);
 
-    this._success.subscribe((message) => this.successMessage = message);
+    this._success.subscribe(message => this.successMessage = message);
     this._success.pipe(
       debounceTime(5000)
-    ).subscribe(() => this.successMessage = null);
+    ).subscribe(() => this.successMessage = '');
   }
 
   public changeSuccessMessage() {
--- a/demo/src/app/components/datepicker/demos/customday/datepicker-customday.ts
+++ b/demo/src/app/components/datepicker/demos/customday/datepicker-customday.ts
@@ -31,6 +31,6 @@ export class NgbdDatepickerCustomday {
   constructor(private calendar: NgbCalendar) {
   }
 
-  isDisabled = (date: NgbDate, current: {month: number}) => date.month !== current.month;
+  isDisabled = (date: NgbDate, current: {month: number, year: number}) => date.month !== current.month;
   isWeekend = (date: NgbDate) =>  this.calendar.getWeekday(date) >= 6;
 }
--- a/demo/src/app/components/datepicker/demos/range-popup/datepicker-range-popup.ts
+++ b/demo/src/app/components/datepicker/demos/range-popup/datepicker-range-popup.ts
@@ -32,10 +32,10 @@ import {NgbDate, NgbCalendar, NgbDateParserFormatter} from '@ng-bootstrap/ng-boo
 })
 export class NgbdDatepickerRangePopup {
 
-  hoveredDate: NgbDate;
+  hoveredDate: NgbDate | null = null;
 
-  fromDate: NgbDate;
-  toDate: NgbDate;
+  fromDate: NgbDate | null;
+  toDate: NgbDate | null;
 
   constructor(private calendar: NgbCalendar, public formatter: NgbDateParserFormatter) {
     this.fromDate = calendar.getToday();
@@ -45,7 +45,7 @@ export class NgbdDatepickerRangePopup {
   onDateSelection(date: NgbDate) {
     if (!this.fromDate && !this.toDate) {
       this.fromDate = date;
-    } else if (this.fromDate && !this.toDate && date.after(this.fromDate)) {
+    } else if (this.fromDate && !this.toDate && date && date.after(this.fromDate)) {
       this.toDate = date;
     } else {
       this.toDate = null;
@@ -58,14 +58,14 @@ export class NgbdDatepickerRangePopup {
   }
 
   isInside(date: NgbDate) {
-    return date.after(this.fromDate) && date.before(this.toDate);
+    return this.toDate && date.after(this.fromDate) && date.before(this.toDate);
   }
 
   isRange(date: NgbDate) {
-    return date.equals(this.fromDate) || date.equals(this.toDate) || this.isInside(date) || this.isHovered(date);
+    return date.equals(this.fromDate) || (this.toDate && date.equals(this.toDate)) || this.isInside(date) || this.isHovered(date);
   }
 
-  validateInput(currentValue: NgbDate, input: string): NgbDate {
+  validateInput(currentValue: NgbDate | null, input: string): NgbDate | null {
     const parsed = this.formatter.parse(input);
     return parsed && this.calendar.isValid(NgbDate.from(parsed)) ? NgbDate.from(parsed) : currentValue;
   }
--- a/demo/src/app/components/shared/api-docs/api-docs.component.ts
+++ b/demo/src/app/components/shared/api-docs/api-docs.component.ts
@@ -55,7 +55,7 @@ export class NgbdApiDocs {
    */
   defaultInputValue(input: InputDesc): string {
     const configProperty = this._configProperties[input.name];
-    return configProperty ? configProperty.defaultValue : input.defaultValue;
+    return (configProperty ? configProperty.defaultValue : input.defaultValue) || '';
   }
 
   /**
--- a/demo/src/app/components/shared/examples-page/demo.component.ts
+++ b/demo/src/app/components/shared/examples-page/demo.component.ts
@@ -1,7 +1,14 @@
 import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
 
 import {Analytics} from '../../../shared/analytics/analytics';
-import {Snippet} from '../../../shared/code/snippet';
+import {ISnippet, Snippet} from '../../../shared/code/snippet';
+
+const TYPES: {[name: string]: string} = {
+  html: 'HTML',
+  scss: 'Style (SCSS)',
+  css: 'Style (CSS)',
+  ts: 'Typescript'
+};
 
 @Component({
   selector: 'ngbd-widget-demo',
@@ -20,8 +27,8 @@ export class NgbdWidgetDemoComponent {
   get markupSnippet() { return Snippet({lang: 'html', code: this.markup}); }
   get codeSnippet() { return Snippet({lang: 'typescript', code: this.code}); }
 
-  getFileSnippet({name, source}) {
-    return Snippet({code: source, lang: name.split('.').pop()});
+  getFileSnippet({name, source}): ISnippet {
+    return Snippet({code: source, lang: name.split('.').pop() || ''});
   }
 
   get hasManyFiles() {
@@ -31,15 +38,7 @@ export class NgbdWidgetDemoComponent {
   constructor(private _analytics: Analytics) {}
 
   tabType(name: string) {
-    const ext = name.split('.').pop();
-    return (
-      {
-        html: 'HTML',
-        scss: 'Style (SCSS)',
-        css: 'Style (CSS)',
-        ts: 'Typescript'
-      }[ext] || 'Code'
-    );
+    return TYPES[(name.split('.').pop() || '')] || 'Code';
   }
 
   trackStackBlitzClick() {
--- a/demo/src/app/components/table/demos/complete/country.service.ts
+++ b/demo/src/app/components/table/demos/complete/country.service.ts
@@ -6,7 +6,7 @@ import {Country} from './country';
 import {COUNTRIES} from './countries';
 import {DecimalPipe} from '@angular/common';
 import {debounceTime, delay, switchMap, tap} from 'rxjs/operators';
-import {SortDirection} from './sortable.directive';
+import {SortColumn, SortDirection} from './sortable.directive';
 
 interface SearchResult {
   countries: Country[];
@@ -17,20 +17,18 @@ interface State {
   page: number;
   pageSize: number;
   searchTerm: string;
-  sortColumn: string;
+  sortColumn: SortColumn;
   sortDirection: SortDirection;
 }
 
-function compare(v1, v2) {
-  return v1 < v2 ? -1 : v1 > v2 ? 1 : 0;
-}
+const compare = (v1: string, v2: string) => v1 < v2 ? -1 : v1 > v2 ? 1 : 0;
 
-function sort(countries: Country[], column: string, direction: string): Country[] {
-  if (direction === '') {
+function sort(countries: Country[], column: SortColumn, direction: string): Country[] {
+  if (direction === '' || column === '') {
     return countries;
   } else {
     return [...countries].sort((a, b) => {
-      const res = compare(a[column], b[column]);
+      const res = compare(`${a[column]}`, `${b[column]}`);
       return direction === 'asc' ? res : -res;
     });
   }
@@ -82,7 +80,7 @@ export class CountryService {
   set page(page: number) { this._set({page}); }
   set pageSize(pageSize: number) { this._set({pageSize}); }
   set searchTerm(searchTerm: string) { this._set({searchTerm}); }
-  set sortColumn(sortColumn: string) { this._set({sortColumn}); }
+  set sortColumn(sortColumn: SortColumn) { this._set({sortColumn}); }
   set sortDirection(sortDirection: SortDirection) { this._set({sortDirection}); }
 
   private _set(patch: Partial<State>) {
--- a/demo/src/app/components/table/demos/complete/table-complete.html
+++ b/demo/src/app/components/table/demos/complete/table-complete.html
@@ -37,7 +37,7 @@
 
   <div class="d-flex justify-content-between p-2">
     <ngb-pagination
-      [collectionSize]="total$ | async" [(page)]="service.page" [pageSize]="service.pageSize">
+      [collectionSize]="(total$ | async)!" [(page)]="service.page" [pageSize]="service.pageSize">
     </ngb-pagination>
 
     <select class="custom-select" style="width: auto" name="pageSize" [(ngModel)]="service.pageSize">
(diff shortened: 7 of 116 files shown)


src/accordion/accordion.ts
```diff
@@ -316,7 +316,7 @@ export class NgbAccordion implements AfterContentChecked {
     });
   }
 
-  private _findPanelById(panelId: string): NgbPanel | null { return this.panels.find(p => p.id === panelId); }
+  private _findPanelById(panelId: string): NgbPanel | null { return this.panels.find(p => p.id === panelId) || null; }
```
fbasso: Just to open a discussion on this pattern (I put the comment here, but I found the same in other locations).

This code sounds strange to me, as `| null` has been added in order to resolve types issue. In javascript, null and undefined are often considered the same (that's why a special comparison rule for `== null` or `!= null` is usually enabled for the linter).

On the other hand, I understand that it's not easy to managed different functions which can accept null or undefined. 

So the question : would it be possible (and convenient) to define a custom type (for example : NotDefined: null | undefined) and use it when necessary ?

src/accordion/accordion.spec.ts
```diff
@@ -370,9 +370,9 @@ describe('ngb-accordion', () => {
     const contents = getPanelsContent(compiled);
     expect(contents.length).not.toBe(0);
 
-    contents.forEach((content: HTMLElement, idx: number) => {
+    contents.forEach((content, idx) => {
```
fbasso: Why did you remove the types here ? A little bit furthe, the same kind of code(i.e. : `contents.forEach((content: HTMLElement)`) has been left with the types. I think that we should keep the consistency in the whole file.

src/carousel/carousel.ts
```diff
@@ -184,7 +184,7 @@ export class NgbCarousel implements AfterContentChecked,
       this._ngZone.runOutsideAngular(() => {
         const hasNextSlide$ = combineLatest([
                                 this.slide.pipe(map(slideEvent => slideEvent.current), startWith(this.activeId)),
-                                this._wrap$, this.slides.changes.pipe(startWith(<{}>null))
+                                this._wrap$, this.slides.changes.pipe(startWith(<null>null))
```
fbasso: Is it necessary to cast here ? I noticed that removing the type (just `startWith(null)`compiles fine).

src/dropdown/dropdown.ts
```diff
@@ -312,7 +312,7 @@ export class NgbDropdown implements AfterContentInit, OnDestroy {
 
     // closing on Enter / Space
     if (key === Key.Space || key === Key.Enter) {
-      if (isEventFromItems && (this.autoClose === true || this.autoClose === 'inside')) {
+      if (itemElement && isEventFromItems && (this.autoClose === true || this.autoClose === 'inside')) {
```
fbasso: The check with itemElement and isEventFromItems is redundant, as they are set only if an element is found. isEventFromItems can be removed completely, and the checks can by relied on the existence of itemElement.

src/accordion/accordion.spec.ts
maxokorokov: thanks, will restore

src/carousel/carousel.ts
maxokorokov: Indeed, I haven't checked it it compiles without casting. There were lining issues related to `rxjs` version before this. Will remove casting.

src/dropdown/dropdown.ts
maxokorokov: That's one of the cases, where typesctipt can't make the assumption you've just made

I could change this:
```ts
if (itemElement && isEventFromItems && (this.autoClose === true || this.autoClose === 'inside')) {
  fromEvent(itemElement, 'click').pipe(take(1)).subscribe(() => this.close());
}
```

to this:

```ts
if (isEventFromItems && (this.autoClose === true || this.autoClose === 'inside')) {
  fromEvent(itemElement !, 'click').pipe(take(1)).subscribe(() => this.close());
}
```

WDYT ?

maxokorokov: I think will change to `// 2` as you're suggesting. Thanks.

src/progressbar/progressbar.spec.ts
```diff
@@ -19,7 +19,7 @@ function getBarHeight(nativeEl): string {
 }
 
 function getBarValue(nativeEl): number {
-  return parseInt(getProgressbar(nativeEl).getAttribute('aria-valuenow'), 10);
+  return parseInt(getProgressbar(nativeEl).getAttribute('aria-valuenow') || '', 10);
```
fbasso: `parseInt(null, 10)` or `parseInt('', 10)` returns undefined. 

`parseInt(<any>getProgressbar...` seems to me a better option, since it doesn't result to extra code.

src/dropdown/dropdown.ts
fbasso: Yes, I personally prefer 2, as there is only one variable to manage the flag and the element, so no redundancy.

src/timepicker/timepicker.ts
```diff
@@ -267,7 +268,7 @@ export class NgbTimepicker implements ControlValueAccessor,
     }
   }
 
-  formatMinSec(value: number) { return padNumber(value); }
+  formatMinSec(value?: number) { return padNumber(isNumber(value) ? value : NaN); }
```
fbasso: I don't how exactly how to manage this case, but I'm not comfortable with this code. The point is that I think that the isNumber is used too many times : `formatMinSec` checks `value` to send value or NaN, then padNumber checks `value` (isNumber again !) to send back the padded value or an empty string. 

Maybe one solution is to change padNumber in order to accept `value: any` (as isNumber) and remove the extra check ?

src/progressbar/progressbar.spec.ts
maxokorokov: @fbasso good catch, haven't thought about extra runtime code

src/timepicker/timepicker.ts
maxokorokov: Agreed

src/util/accessibility/live.ts
```diff
@@ -39,7 +39,7 @@ export class Live implements OnDestroy {
 
   ngOnDestroy() {
     const element = getLiveElement(this._document);
-    if (element) {
+    if (element != null && element.parentElement != null) {
```
fbasso: Is there a case where element is not null and element.parentElement is null ? If not, I would probably add a `element.parentElement !.removeChild(...`  on the next line.

src/util/popup.ts
```diff
@@ -38,15 +38,15 @@ export class PopupService<T> {
       this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._windowRef.hostView));
       this._windowRef = null;
 
-      if (this._contentRef.viewRef) {
+      if (this._contentRef && this._contentRef.viewRef) {
```
fbasso: We can use the optional chaining operator:
``` 
if (this._contentRef?.viewRef) {
```
(I thought about this right now, I maybe forgot this idea in previous comments, to be checked !)

maxokorokov: Oh yes, apparently, it is available starting from typescript `3.7.x` (which we got with Angular 9) and looks like this:

```ts
if (a.b?.c) {
}

// compiles to

if ((_a = a.b) === null || _a === void 0 ? void 0 : _a.c) {
}
```

Will do an additional pass through the code for this one.

src/util/accessibility/live.ts
maxokorokov: we're always attaching it to body, so parent element is always body if I'm not mistaken

fbasso: Except if a developer already created a `<div id="ngb-live">` in his app, no ?

demo/src/app/components/table/demos/complete/table-complete.html
```diff
@@ -37,7 +37,7 @@
 
   <div class="d-flex justify-content-between p-2">
     <ngb-pagination
-      [collectionSize]="total$ | async" [(page)]="service.page" [pageSize]="service.pageSize">
+      [collectionSize]="(total$ | async)!" [(page)]="service.page" [pageSize]="service.pageSize">
```
benouat: This use case needs to be reported to Angular team.

demo/src/app/components/table/demos/sortable/table-sortable.ts
```diff
@@ -39,12 +39,13 @@ const COUNTRIES: Country[] = [
   }
 ];
 
+export type SortColumn = keyof Country | '';
 export type SortDirection = 'asc' | 'desc' | '';
 const rotate: {[key: string]: SortDirection} = { 'asc': 'desc', 'desc': '', '': 'asc' };
-export const compare = (v1, v2) => v1 < v2 ? -1 : v1 > v2 ? 1 : 0;
+export const compare = (v1: string | number, v2: string | number) => v1 < v2 ? -1 : v1 > v2 ? 1 : 0;
```
benouat: Same comment as above.

demo/src/app/components/shared/examples-page/demo.component.ts
```diff
@@ -20,8 +27,8 @@ export class NgbdWidgetDemoComponent {
   get markupSnippet() { return Snippet({lang: 'html', code: this.markup}); }
   get codeSnippet() { return Snippet({lang: 'typescript', code: this.code}); }
 
-  getFileSnippet({name, source}) {
-    return Snippet({code: source, lang: name.split('.').pop()});
+  getFileSnippet({name, source}: {name: string, source: string}) {
```
benouat: Can't we use `ISnippet` interface here as a return type ?

src/util/accessibility/live.ts
benouat: > Is there a case where element is not null and element.parentElement is null

I don't know if it's the case here, but typically _detached node_ would be an example of that.

fbasso: Yes, sure, I said it because of the specific getLiveElement implementation : querySelector is done on the body, and attached to the body if it doesn't exist.

demo/src/app/components/alert/demos/selfclosing/alert-selfclosing.ts
```diff
@@ -10,7 +10,7 @@ export class NgbdAlertSelfclosing implements OnInit {
   private _success = new Subject<string>();
 
   staticAlertClosed = false;
-  successMessage: string;
+  successMessage: string | null = null;
```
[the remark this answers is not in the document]
fbasso: Yes, it can be a good trick. 

At least, it's quite safe the demo part. We need to check case by case to see if a difference between empty string and null. But the code visibility/complexity seems better to me with empty string.

demo/src/app/components/datepicker/demos/range-popup/datepicker-range-popup.ts
```diff
@@ -32,20 +32,20 @@ import {NgbDate, NgbCalendar, NgbDateParserFormatter} from '@ng-bootstrap/ng-boo
...
+  fromDate: NgbDate | null;
+  toDate: NgbDate | null;
 
   constructor(private calendar: NgbCalendar, public formatter: NgbDateParserFormatter) {
     this.fromDate = calendar.getToday();
     this.toDate = calendar.getNext(calendar.getToday(), 'd', 10);
   }
 
-  onDateSelection(date: NgbDate) {
+  onDateSelection(date: NgbDate | null) {
```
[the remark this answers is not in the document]
maxokorokov: I don't think we can, but I guess the reason is:
- `selectedDate` in the model is `null` initially, so `NgbDate | null`
- it is internal method

I'd better check where it is emitted and put `!` there.

demo/src/app/components/datepicker/demos/customday/datepicker-customday.ts
```diff
@@ -31,6 +31,6 @@ export class NgbdDatepickerCustomday {
   constructor(private calendar: NgbCalendar) {
   }
 
-  isDisabled = (date: NgbDate, current: {month: number}) => date.month !== current.month;
+  isDisabled = (date: NgbDate, current: {month: number, year: number}) => date.month !== current.month;
```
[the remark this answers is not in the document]
maxokorokov: It would be a breaking change, but why not. would be the first date of the month. Initially it was just a `month`, but later there was a feature request, where people asked to add a `year`. In 90% of the cases you would need this to compare `date.month === month`

demo/src/app/components/alert/demos/selfclosing/alert-selfclosing.ts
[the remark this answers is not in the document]
maxokorokov: +1 from me

demo/src/app/components/shared/api-docs/api-docs.component.ts
```diff
@@ -53,7 +53,7 @@ export class NgbdApiDocs {
    * Returns the default value of the given directive input by first looking for it in the matching config service
    * property. If there is no matching config property, it reads it from the input.
    */
-  defaultInputValue(input: InputDesc): string {
+  defaultInputValue(input: InputDesc): string | undefined {
     const configProperty = this._configProperties[input.name];
     return configProperty ? configProperty.defaultValue : input.defaultValue;
   }
```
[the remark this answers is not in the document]
maxokorokov: Agreed

demo/src/app/components/table/demos/complete/country.service.ts
```diff
@@ -17,16 +17,16 @@ interface State {
   page: number;
   pageSize: number;
   searchTerm: string;
-  sortColumn: string;
+  sortColumn: SortColumn;
   sortDirection: SortDirection;
 }
 
-function compare(v1, v2) {
+function compare(v1: string | number, v2: string | number) {
```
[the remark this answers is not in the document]
maxokorokov: Will check that, thanks

demo/src/app/components/shared/examples-page/demo.component.ts
maxokorokov: Will check

maxokorokov:
Cool stuff, thanks @fbasso and @benouat. Please leave any more comments if you have, I won't push anything for now. 

I'm going to leave this comment to sum up your reviews:
- [ ] use conditional chaining `this._foo?.bar` where possible
- [ ] prefer `s = ''` to `s: string | null = null`
- [ ] prefer using `any` in cases like `parseInt(a || '')` to avoid generating stupid code
- [ ] consistency across demos (ex. some might use `date?: NgbDate`, some `date: NgbDate | null`)

And of course + all specific cases you've pointed out.

demo/src/app/components/table/demos/complete/country.service.ts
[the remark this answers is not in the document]
fbasso: By the way, the string literal usage is not required here, `compare(a[column], b[column])` will do the job ;)

src/datepicker/datepicker-input.ts
```diff
@@ -293,26 +300,26 @@ export class NgbInputDatepicker implements OnChanges,
...
-      return {'ngbDate': {minDate: {minDate: this.minDate, actual: value}}};
+      if (this.maxDate && ngbDate !.after(NgbDate.from(this.maxDate))) {
+        return {'ngbDate': {maxDate: {maxDate: this.maxDate, actual: value}}};
+      }
     }
 
-    if (this.maxDate && ngbDate.after(NgbDate.from(this.maxDate))) {
-      return {'ngbDate': {maxDate: {maxDate: this.maxDate, actual: value}}};
-    }
+    return null;
```
fbasso: Assuming that this._fromDateStruct returns a valid NgbDateStruct (isValid is used inside _fromDateStruct) or null, we can just check that ngbDate is not null (otherwise, it's invalid).

Extra validation is done here. Instead, I would write:

```
      const ngbDate = this._fromDateStruct(this._dateAdapter.fromModel(value));

      if (!ngbDate) {
        return {'ngbDate': {invalid: value}};
      }

      if (this.minDate && ngbDate.before(NgbDate.from(this.minDate))) {
        return {'ngbDate': {minDate: {minDate: this.minDate, actual: value}}};
      }

      if (this.maxDate && ngbDate.after(NgbDate.from(this.maxDate))) {
        return {'ngbDate': {maxDate: {maxDate: this.maxDate, actual: value}}};
      }
```

benouat: but then you would get an exit way without any return, pretty sure TypeScript will yell 😄

fbasso: Yes you're right, we still need the `return null;`, I just posted the changed code inside the if ;)

src/timepicker/timepicker.ts
maxokorokov: Actually in this case I just aligned the implementation with the `formatHour` function, so I think I'll keep it as is

demo/src/app/components/shared/examples-page/demo.component.ts
maxokorokov: We can indeed!

src/datepicker/datepicker-input.ts
maxokorokov: @fbasso good catch, fixed it

maxokorokov:
Updated → fixed all remarks

fbasso:
Thanks @maxokorokov . I reviewed the remarks I did, and it's fine for me.

Issue #3640, opened 2020-03-09
Bootstrap dropdown transform causes overflow


jchapelle:
### Bug description:
I'm using a dropdown inside the last cell of a table. The dropdow menu is large. For that reason, I'm using the bottom-right placement of the menu. The menu is properly positionned but the css transform property seems to introduce some strange overflow in the width of the window.
![dropdown-issue](https://user-images.githubusercontent.com/6162013/76260376-f4629680-6257-11ea-8595-4de1571bd5b0.png)

When applying the exact same value of the transform css property to the top and left css properties, it seems to work fine ... 
![dropdown-issue-2](https://user-images.githubusercontent.com/6162013/76260687-aac67b80-6258-11ea-81d0-b2fc5392afa8.png)

### Link to minimally-working StackBlitz that reproduces the issue:
You have to click at least 2 times for the bug to occurs
Also, the bug depends of the size of your window. Maybe you'll have to add some cell in the table row. You'll have to get the dropdow at the very right of your window (cfr screenshiot) for the bug to occur
https://stackblitz.com/edit/angular-74szwx

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 9

ng-bootstrap: 6

Bootstrap: 4.4.1

benouat:
Closed in #3644

maxokorokov:
This fix broke popover in `6.0.1`, so this is reverted in `6.0.2`

jchapelle:
In addition, the fix wasn't really working fine. A blink effect on horizontal scrollbar was introduced as well

fbasso:
I tested it with pure bootstrap stuff. It seems that it's the same kind of issue. By default (with Bootstrap), the tooltip is set on the body. When I set it explicitly inside the table, the scrollbars appear:

![image](https://user-images.githubusercontent.com/1152930/77519896-99988400-6e80-11ea-9413-6d4833b1fd22.png)

To solve the issue, I think the best way is to set it explicitly on the body with container="body": https://ng-bootstrap.github.io/#/components/tooltip/examples#container

benouat:
Closing as Bootstrap also has the same issue, and we do have a fix on our side to prevent it with previous comment.

PR #3644, opened 2020-03-12
fix(dropdown): ngbDropdown adds vertical scrollbar ... (#3580)

### LINKED ISSUE #3580: NgbDropdown adds vertical scrollbar when opened in table.
### Bug description:

`NgbDropdown` adds horizontal scrollbar when opened in table. Problem occurs in Chromium browsers (Chrome & New Edge) when `NgbDropdown` is in table row. In Safari & Firefox it works as expected.

It seems that problem is due to `transform` CSS property added to `NgbDropdownMenu` on dropdown open where negative values for translate (e.g. `transform: translate(-132px, 45px)`) are causing this problem. When `NgbDropdown` has position `static` (default is `relative`) then translate of `NgbDropdownMenu` has positive values (e.g. `transform: translate(82px, 37px)`) and everything works as expected. 

https://stackblitz.com/edit/angular-xhgz4p-biuuss

Angular: 8.2.14

ng-bootstrap: 5.2.2

Bootstrap: 4.4.1

gpolychronis:
Fixes #3580 and #3640
Adds `overflow:hidden` on dropdown.
Issue affects all derrative components i.e. datepicker upon popup, tooltip, typeahead etc.

CC: @maxokorokov @benouat Please review

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -390,6 +390,7 @@ export class NgbDropdown implements AfterContentInit, OnDestroy {
       renderer.appendChild(dropdownElement, dropdownMenuElement);
       renderer.removeStyle(dropdownMenuElement, 'position');
       renderer.removeStyle(dropdownMenuElement, 'transform');
+      renderer.removeStyle(dropdownMenuElement, 'overflow');
     }
     if (this._bodyContainer) {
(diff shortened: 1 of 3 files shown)


fbasso:
Same comment than the #3640, using container="body" seems to be the best solution.

Issue #3630, opened 2020-02-28
Tooltip does not appear when placed on element different than button (ios)


kamilchlebek:
### Bug description:
When `ngbTooltip` is placed on element element different than `button` (ex. `div`, `span`) - tooltip does not appear after tapping an element.

This happens only on iOS. 

This and another possibly related issue is also described here: https://github.com/ng-bootstrap/ng-bootstrap/issues/3591#issuecomment-586973588 (tooltip will not show again after closing it by tapping an element that triggered the tooltip) 

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-gbi8wc

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 9.0.4 (same issue on v8)
ng-bootstrap: 6.0.0

MrAlexand0r:
This bug still occurs
Would it be possible to look into it again?

PR #3645, opened 2020-03-13
fix(tooltip): tooltip does not appear when ... (ios) #3630

gpolychronis:
An attempt to fix all mobile related tooltip issues without using OS specific method.

Note: Test upon iOS device is required.

### DIFF
--- a/src/util/autoclose.ts
+++ b/src/util/autoclose.ts
@@ -10,26 +10,12 @@ const isContainedIn = (element: HTMLElement, array?: HTMLElement[]) =>
 const matchesSelectorIfAny = (element: HTMLElement, selector?: string) =>
     !selector || closest(element, selector) != null;
 
-// we have to add a more significant delay to avoid re-opening when handling (click) on a toggling element
-// TODO: use proper Angular platform detection when NgbAutoClose becomes a service and we can inject PLATFORM_ID
-const isMobile = (() => {
-  const isIOS = () => /iPad|iPhone|iPod/.test(navigator.userAgent) ||
-      (/Macintosh/.test(navigator.userAgent) && navigator.maxTouchPoints && navigator.maxTouchPoints > 2);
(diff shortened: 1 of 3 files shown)


gpolychronis:
CC: @maxokorokov @benouat 
Please review

benouat:
@gpolychronis there is something weird with your commit messages (I also noticed it on #3644)

Could you please fix them, like removing this `...` in between ?

src/util/autoclose.ts
```diff
@@ -10,26 +10,12 @@ const isContainedIn = (element: HTMLElement, array?: HTMLElement[]) =>
 const matchesSelectorIfAny = (element: HTMLElement, selector?: string) =>
     !selector || closest(element, selector) != null;
 
-// we have to add a more significant delay to avoid re-opening when handling (click) on a toggling element
-// TODO: use proper Angular platform detection when NgbAutoClose becomes a service and we can inject PLATFORM_ID
-const isMobile = (() => {
```
maxokorokov: Hmm, are you aware that autoClose is used not only by the tooltip, but also datepicker, popover, dropdown and typeahead?

Removing this will break all others.

Also changes to this code should be tested on Desktop, Android, iOS and iPadOS

gpolychronis: Currently I don't have access to these browsers. I will have to find a solution on that and update this PR in the future.

kamilchlebek: @maxokorokov Seems that browserstack.com is free for open source projects. `ng-bootstrap` has some e2e tests. Can't we setup `protractor.conf.js` to run against selenium grid provided by `browserstack` ?

arsenasatryan: > Removing this will break all others.
Seems this already is broken on IPad  IOS  > 13, I tested some of the IPads though available on BrowserStack.
You can't close datepickers there.

Issue #3648, opened 2020-03-17
Disabled date picker is clickable/focusable


benouat:
![Kapture 2020-03-17 at 10 28 46](https://user-images.githubusercontent.com/1152740/76842164-1fda1800-683a-11ea-82e9-554d80a64fb5.gif)

benouat:
It's reproducible on the [demo](https://ng-bootstrap.github.io/#/components/datepicker/examples#disabled) directly

PR #3649, opened 2020-03-18
fix(datepicker): disabled date picker is clickable/focusable #3648

Smoggy:
Issue #3648[](https://github.com/ng-bootstrap/ng-bootstrap/issues/3648)

### DIFF
--- a/e2e-app/src/app/datepicker/focus/datepicker-focus.component.html
+++ b/e2e-app/src/app/datepicker/focus/datepicker-focus.component.html
@@ -4,7 +4,8 @@
   <div class="form-group form-inline">
     <div class="input-group">
       <input class="form-control" name="dp" placeholder="yyyy-mm-dd" #d="ngbDatepicker" ngbDatepicker [(ngModel)]="model"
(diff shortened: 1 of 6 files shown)


fbasso:
Thanks for the PR @Smoggy 

Using the pointer-event seems to be the more natural and easy fix to do. But unfortunately it is not supported by IE10 : https://caniuse.com/#feat=pointer-events

I guess that this point is a problem.

maxokorokov:
@fbasso good catch, I wasn't thinking about it. However do you have any ideas for a better solution? 

I think others will involve adding JS handlers/code just to support this use case in IE10. This one costs nothing and improves for most browsers. I think I personally would prefer to keep this one, unless there is another simple one that works everywhere.

fbasso:
@maxokorokov, I agree. I think this css property must be kept anyway, it will have the priority on javascript behavior that we would (possibly, if someone complains about it) implement for ie10.

I don't know if ie10 is used a lot, as users usually have ie11 on their machine (automatically installed).

So I would go for this solution, and wait to see if someone complains for ie10.

maxokorokov:
I pushed an e2e test here, will merge when green

Issue #3627, opened 2020-02-26
Get all the open modal instances using the NgbModal


ahadjithoma:
#### Feature Request:
It would be nice if using the NgbModal we could get the instances of all the opened modals. 
Since the service has the `hasOpenModals` method, i expect that getting the instances of them won't be an difficult. 

#### What we are trying to do: 
Some modal components are extending a class for additional checks whether they can be closed or not. In our `app.component` we are subscribing to route events and we close all modals by calling the `dismissAll` method but we would like to check whether those modals can be closed or not before closing them. This can be done by calling a method inside the modal component. With that said, we need the instances of all the currently opened modals. 

Thanks

PR #3650, opened 2020-03-19
feat(modal): get all the open modal instances using the NgbModal (#3627)

gpolychronis:

### DIFF
--- a/demo/src/app/components/modal/demos/stacked/modal-stacked.html
+++ b/demo/src/app/components/modal/demos/stacked/modal-stacked.html
@@ -1,1 +1,3 @@
 <button class="btn btn-lg btn-outline-primary" (click)="open()">Launch demo modal</button>
+<hr/>
+<p>Number of modals: {{modalsNumber}}</p>
--- a/demo/src/app/components/modal/demos/stacked/modal-stacked.ts
+++ b/demo/src/app/components/modal/demos/stacked/modal-stacked.ts
@@ -53,7 +53,13 @@ export class NgbdModal2Content {
   templateUrl: './modal-stacked.html'
 })
 export class NgbdModalStacked {
-  constructor(private modalService: NgbModal) {}
+  modalsNumber = 0;
+
+  constructor(private modalService: NgbModal) {
+    this.modalService.activeInstances.subscribe((list) => {
+      this.modalsNumber = list.length;
+    });
+  }
 
   open() {
     this.modalService.open(NgbdModal1Content);
--- a/src/modal/modal-stack.ts
+++ b/src/modal/modal-stack.ts
@@ -3,6 +3,7 @@ import {
   ApplicationRef,
   ComponentFactoryResolver,
   ComponentRef,
+  EventEmitter,
   Inject,
   Injectable,
   Injector,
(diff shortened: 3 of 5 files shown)


pkozlowski-opensource:
Agree with @benouat concerns - can we discuss more the the public API we want to expose?

fbasso:
On my side, I agree with Benoit regarding the forEach API : it forces the developer to use a loop, so it's not flexible enough (what if we just want the number of opened modals, for example ?).

Regarding the solution, I would be be in favor of the first one, for practical reason (and considering the use case explained in #3627. With a subscribe, I guess that there would be a need to subscribe on each modal (to get the modal instance in the subscribe ?), and my feeling is it' can be painful.

On the other hand, the subscribe will allow to select modals where specific actions are required, letting the others apart. 

So I have no clear idea of what to do here :)

benouat:
> With a subscribe, I guess that there would be a need to subscribe on each modal

No, I was more thinking about the equivalent of a `QueryList` thing. You subscribe, you always get a refreshed copy of the activeModals in one shot, an array of modals

gpolychronis:
> 
> 
> > With a subscribe, I guess that there would be a need to subscribe on each modal
> 
> No, I was more thinking about the equivalent of a `QueryList` thing. You subscribe, you always get a refreshed copy of the activeModals in one shot, an array of modals

This seems a little counter-intuitive. The first time I read I though the observable would emit the ref any time a new child modal was opened.

Between the two options, I prefer the first and let the user iterate over the instances.

benouat:
> This seems a little counter-intuitive.

It might, arguably, feel counter-intuitive, but it's exactly how QueryList are working when using `ViewChidlren` and `ContentChildren`.
@gpolychronis maybe you're not familiar enough as well with Content Queries.

benouat:
I would like that we move on with this one.
I suggest we stick to the #&zwj;2 solution with an Observable return a live list of active modals.

@pkozlowski-opensource Could we use directly `QueryList` from Angular ? Sounds like a perfect fit actually, and it seems to be part of their @PublicAPI

> An unmodifiable list of items that Angular keeps up to date when the state of the application changes.

https://github.com/angular/angular/blob/master/packages/core/src/linker/query_list.ts

pkozlowski-opensource:
> Could we use directly QueryList from Angular ? Sounds like a perfect fit actually, 
@benouat hmmm, not sure which part of QueryList is a perfect fit here? Are you talking about notifications?

I guess I need more context as for now I'm not sure how we could take advantage of `QueryList` here...

benouat:
@pkozlowski-opensource We need to maintain a "live" list of item in a way that would be a Iterable, with helpers sur as `first` and `last` and the `length` and a way to be notified each time the list of items is changing via observables.

Looking at it, my description  is 100% matching what the class QueryList is providing. am I wrong ?

Regarding our feature, we would then maintain a `QueryList<ActiveModal>`
People could subscribe to the `changes` observable to be notified, access `first` and `last` if they do want to implement a FIFO/LIFO stack of modal...

pkozlowski-opensource:
@benouat I see. I would rather stay away from using `QueryList` given my understanding where things are heading.

 The main problem is that `QueryList` class is super-messy and mixes public and "private" APIs. To use it you would have to call methods that I consider "private" (ex. `reset(...)`) and those methods are likely to go away in the future. Even worst - we (Angular) are considering turning QueryList into an interface and make the implementation private - see  https://github.com/angular/angular/issues/29469
The other problem with the QueryList is that it emits itself on changes which might not be what you need here.

All in all I think it is safer to craft a minimal interface + impl and avoid dependency on sth that is likelly to change in the future.

benouat:
That's exactly for this kind of information that I pinged you! Thanks 🙏

gpolychronis:
UPDATE:

NgbModal has a new property: activeInstances
It's an observable that emits anytime a modal opens/closes. The emitted is an array of NgbModalRef.

@benouat Please review.

Issue #3575, opened 2020-02-05
The position of the popup is not updated for resizable textareas


5te1n:
### Bug description:
Situation:
We have the ngbPopover directive on a resizable textarea. The popover has the default configuration ("autoClose=true").

Current behaviour: 
The position of the popover is not updated when the textarea is resized. In addition, the resizing is interpreted as a click, so the popover closes or opens after the resize. 

Desired behavior:
The position is updated when the text area is resized and the resizing does not trigger the open/close mechanism.

### Link to minimally-working StackBlitz that reproduces the issue:
https://stackblitz.com/edit/angular-fr1ahy?embed=1&file=app/popover-basic.html

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 8.2.4

ng-bootstrap: 5.2.1

Bootstrap: 4.3.1

gpolychronis:
Hello @5te1n .
Resizing the text area is an "outside" pointer action. So closing the popover during resizing makes sense to me.
I propose to customize the triggers and autoclose to achieve what you need, i.e.:

```
<textarea style="resize: both;" placement="bottom" [autoClose]="'inside'" triggers="mouseup:mousedown"
          ngbPopover="Vivamus sagittis lacus vel augue laoreet rutrum faucibus." popoverTitle="Popover on bottom"
          #p="ngbPopover" (click)="p.open()">
  Popover on bottom
</textarea>
```

Issue #3579, opened 2020-02-10
Add Popper 2 as a dependency


maxokorokov:
Giving the number of issues we have about positioning, lack of resources and the fact that we're kind of rewriting popper on our own, we'll be adding `@popper/core` as a peer dependency in 12.0.0.

We'll be able to flip, reposition arrow (tooltip, popover), consider viewport dimensions, different parent containers.
Two major downsides are It adds around 14-17Kb (minified, NOT compressed) to the final app bundle in our tests and a 3rd dependency for this project

Affected components at the moment are datepicker, dropdown, popover, tooltip and typeahead.

**Positioning Issues we have**

- [ ] #4012
- [ ] #3981 
- [ ] #3927 
- [ ] #3836 
- [ ] #3834 
- [ ] #3743 
- [ ] #3575 
- [ ] #3525 
- [x] #3495 
- [ ] #3382 
- [ ] #3337 
- [ ] #3310 
- [ ] #3294 
- [ ] #3046 

**Related PRs**
#4129

bastienmoulia:
> check if/when bootstrap is going to use it

https://github.com/twbs/bootstrap/issues/29842

Not sure yet if Bootstrap 5 will:
- stay with popper 1
- use popper 2 with polyfill for ie11
- use popper 2 and drop support of ie11

maxokorokov:
Bootstrap 5 should be dropping IE support altogether, so no need in polyfills for popper 2

cc @fbasso

fbasso:
Close it, as ngBootstrap 13.1.0 has been released with feature

PR #3657, opened 2020-03-25
feat(popover): update popover position on mousemove (#3575)

gpolychronis:

### DIFF
--- a/src/popover/popover.spec.ts
+++ b/src/popover/popover.spec.ts
@@ -455,6 +455,21 @@ describe('ngb-popover', () => {
       expect(windowEl.textContent.trim()).toBe('Great tip!');
     });
 
+    it('should update on mousemove', () => {
+      const fixture = createTestComponent(`<div ngbPopover="Great tip!" placement="auto"></div>`);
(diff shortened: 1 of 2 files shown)


gpolychronis:
@benouat @maxokorokov 
Please review

fbasso:
I'm afraid about performance issues if the positioning is re-calculated on each mousemove.

Moreover, we already have performance issues opened because of the ngZone.onStable usage, and the direction (waiting for some decisions and code reviews for now) is more to remove onStable to use scroll or resize evens to re-positions the popups. So I'm not sure this is the fix to do.

Waiting for @benouat and @maxokorokov to see what they think about it.

benouat:
I am aligned with @fbasso I don't think it's a good idea.

gpolychronis:
The issue is that the popover should follow the resizing. Inevitably this means triggering on mouseover.
@fbasso @benouat How should I solve the problem, hmm? Any thoughts on that?

gpolychronis:
There already an issue on that.

benouat:
Have a look at #3579 to follow updates

Issue #2203, opened 2018-03-02
Should the bootstrap css be required in when using ng-bootstrap?


huntc:
Being new to this project, the following lines from the "getting started" threw me out completely:

> Should I add bootstrap.js or bootstrap.min.js to my project?
> No, the goal of ng-bootstrap is to completely replace JavaScript implementation for components. Nor should you include other dependencies like jQuery or popper.js. It is not necessary and might interfere with ng-bootstrap code.

I noticed that [the same problem was reported on Stackoverflow](https://stackoverflow.com/questions/48972235/angular-cli-angular-5-2-0-with-ng-boostrap-is-not-working).

Should the getting-started include some words on including the css file within `styles.css` upon having run an `npm install bootstrap`? For example:

```
@import "../node_modules/bootstrap/dist/css/bootstrap.css"
```

pkozlowski-opensource:
YES, you MUST include Boostrap CSS (only CSS, no JS). Could you suggest documentation update to our https://ng-bootstrap.github.io/#/getting-started page?

I would love to have a suggestion for the exact wording you would like to see to make things clear for you and outer users. Thnx!

huntc:
Done: https://github.com/ng-bootstrap/ng-bootstrap/pull/2204

huntc:
(haven't tested how the text renders though)

houfu:
I am thinking of this alternative that provides a bit more hand holding.

bastienmoulia:
Bootstrap should be in peerDependencies of the package.json

And for the import it's better to use the ~ instead of a relative path:
``` css
@import "~bootstrap/dist/css/bootstrap.css"
```

And maybe add an exemple of scss import:
``` scss
@import "~bootstrap/scss/bootstrap"
```

dotNetAthlete:
I think it is bizarre that the css isn't available through ng-bootstrap !

arturhungrig:
I am right now struggling getting the datepicker to be styled the way it is rendered on the Examples page. Absolutely not clear which stylesheet is required.

Not clear whether the default stylesheet should be enough `@import '~bootstrap/dist/css/bootstrap.css';` (in app.component.css for example)

or maybe one should go install `bootstrap-datepicker` and use:
`@import '~bootstrap-datepicker/dist/css/bootstrap-datepicker.css';`

or maybe 
`@import '~bootstrap-datepicker/dist/css/bootstrap-datepicker.standalone.css';`

None seems to work.
[...]

PR #3665, opened 2020-04-03
docs: update installation instructions

maxokorokov:
As more people are unclear what to do with CSS, here is another update to installation instructions. Based on  https://github.com/ng-bootstrap/ng-bootstrap/issues/2203#issuecomment-450644888

<details>
  <summary>Click to see the screenshot</summary>

Screen Shot 2020-04-03 at 14 59 47

</details>

But we'd better release updated schematics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Please check all components we have in action at https://ng-bootstrap.github.io
 ## Dependencies
 
 The only two dependencies are [Angular](https://angular.io) and [Bootstrap 4](https://getbootstrap.com) CSS.
-Here is the list of minimal required versions:
+The supported versions are:
 
 | ng-bootstrap | Angular | Bootstrap CSS |
 | ------------ | ------- | ------------- |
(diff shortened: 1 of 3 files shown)


maxokorokov:
@benouat, thanks, updated

Screen Shot 2020-04-06 at 11 01 46

4javier:
If you suggest to install bootstrap by npm, your examples should show the import from the correct initial path, that's node_modules, not home.

> /* update your 'styles.css' with */
@import 'node_modules/bootstrap/dist/css/bootstrap.css';
>
>/* or your 'styles.scss' with */
@import "node_modules/bootstrap/scss/bootstrap";

benouat:
It may be confusing, but it’s not referencing an home folder! It’s a SASS convention that points to `node_modules`.

4javier:
The most unintuitive convention I've seen in a long time...
Anyway, it seems that's a webpack fetaure, not a SASS one.
https://stackoverflow.com/a/37118406/6350583
Actually, I don't use webpack, and if I replace "node_module" with "~" inside my import statement this is what I get

> ERROR in ./src/styles.scss (./node_modules/@angular-devkit/build-angular/src/angular-cli-files/plugins/raw-css-loader.js!./node_modules/postcss-loader/src??embedded!./node_modules/sass-loader/dist/cjs.js??ref--15-3!./src/styles.scss)
> Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
> SassError: Can't find stylesheet to import.
>   ╷
> 7 │ @import "~/bootstrap/scss/bootstrap";
>   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   ╵
>   /mnt/dati/workspace/PetMenuAngularFrontend/src/styles.scss 7:9  root stylesheet

benouat:
> It’s a SASS convention.

My bad I was not clear... It's a `sass-loader` convention/feature. `sass-loader` being you're right, a `webpack` plugin, it can only be used via webpack.

Fortunately for us, Angular CLI relies on it under the hood.

It's true, though, that it might be confusing. Although, it is just `~` and not `~/` (that would point to the home directory)

> Actually, I don't use webpack, and if I replace "node_module" with "~"

So I guess in this case, you need to stick to the `node_modules` syntax

4javier:
Actually I manage my project with npm and ng. Maybe I was using ~/, because now it works.
Thanks for your clarifications.

PR #3669, opened 2020-04-07
Add schematics to `@ng-bootstrap/ng-bootstrap`

### LINKED ISSUE #2361: ng add with @ng-bootstrap/ng-bootstrap
When a new angular 6 project is created, why we have to use `ng add @ng-bootstrap/schematics` instead of `ng add @ng-bootstrap/ng-bootstrap`?

I see the change in #2289 but don't understand it.

Angular: 6

ng-bootstrap: 2

Bootstrap: 4

maxokorokov:
Based on the latest version of the `@ng-bootstrap/schematics` (it will be deprecated)

Mostly technical PR, **no new features/fixes**:
- copied code from `@ng-bootstrap/schematics`
- made it work with Angular 9
- made it work with our CI
- updated generated `@ng-bootstrap/ng-bootstrap` package with built schematics

Fixes #2361 

@benouat


schematics/ng-add/setup-project.ts
```diff
@@ -0,0 +1,117 @@
...
+  SchematicsException,
+  Tree,
+} from '@angular-devkit/schematics';
+import {addImportToModule} from '@schematics/angular/utility/ast-utils';
+import {InsertChange} from '@schematics/angular/utility/change';
+import {getWorkspace} from '@schematics/angular/utility/config';
+import {getAppModulePath} from '@schematics/angular/utility/ng-ast-utils';
+import {getProject} from '@schematics/angular/utility/project';
+import * as path from 'path';
+import * as ts from 'typescript';
```
devoto13, line 13: Recently fixed issue in another library, which was caused by the similar code. `@schematics/angular` bundles its own version of TS and as a result AST is created by the workspace version of TS (`ts.createSourceFile()` call), while it is modified by the version bundled by `@schematics/angular`. If these versions are different it crashes. The reproduction was to `ng new my-proj` with latest CLI, change workspace TS to 3.7.5 and `ng add <lib>`.

See https://github.com/FortAwesome/angular-fontawesome/pull/238 for more details.

maxokorokov: Thank you, will check this

maxokorokov:
Pushed the following changes:
- support for `@angular/localize` polyfill installation
- support for both `.sass` and `.scss`
- more unit tests

Full list of what is supported at the moment:
- `ng add @ng-bootstrap/ng-bootstrap` → will target a default CLI project
- `ng add @ng-bootstrap/ng-bootstrap --project xxx` → will target project `xxx`
- installing `bootstrap`, `@ng-bootstrap/ng-bootstrap` and `@angular/localize` dependencies
- adding `@angular/localize` polyfill
- adding `NgbModule` to the main application module
- adding styles for `.scss` and `.sass` in the style file directly
- warning for unsupported project styles for `less`, `stylus`
- adding `bootstrap.min.css` to `angular.json` in case of CSS styles / unsupported project styles

What is still missing:
- [ ] documentation updates
- [ ] versioning

@benouat, could you please take a look at this?

misc/copy-static-files.ts
```diff
@@ -1,7 +1,14 @@
-import {copyFileSync} from 'fs';
+import {copyFileSync, copySync} from 'fs-extra';
 
 /**
- * Copies missing required static assets after the ng-bootstap build
+ * Copies missing required static assets and Schematics after the ng-bootstrap build
```
benouat, line 4: 💄 ~~Schematics~~ → schematics

schematics/collection.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$schema": "../node_modules/@angular-devkit/schematics/collection-schema.json",
+  "schematics": {
+    "ng-add": {
+      "description": "Adds ng-bootstrap to the project and installs dependencies",
```
benouat, line 5: I am hesitating here, should we remplace `ng-bootstrap` with `@ng-bootstrap/ng-bootstrap`?

schematics/ng-add/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        throw new SchematicsException(messages.noProject(project));
+      }
+    }
+
+    // Installing dependencies
+    const angularCoreVersion = getPackageVersionFromPackageJson(tree, '@angular/core') !;
+    const angularLocalizeVersion = getPackageVersionFromPackageJson(tree, '@angular/localize');
+
+    addPackageToPackageJson(tree, '@ng-bootstrap/ng-bootstrap', `${NG_BOOTSTRAP_VERSION}`);
+    addPackageToPackageJson(tree, 'bootstrap', `${BOOTSTRAP_VERSION}`);
```
benouat, line 41: Why do we need `` `${ }` `` here ? It seems useless to me, both versions are already strings

schematics/ng-add/schema.ts
```diff
@@ -0,0 +1,6 @@
+export interface Schema {
+  /**
+   * Name of the project where ng-bootstrap should be installed
```
benouat, line 3: Same comment as a previous one (sounds like a lot of 💄)
I think we should be consistent and replace ~~`ng-bootstrap`~~ with `@ng-bootstrap/ng-bootstrap`

schematics/ng-add/steps/add-bootstrap.ts
```diff
@@ -0,0 +1,82 @@
...
+import {Schema} from '../schema';
+import * as messages from '../messages';
+import {getProjectStyleFile, getProjectTargetOptions} from '../../utils/project';
+
+
+const BOOTSTRAP_CSS_FILEPATH = 'node_modules/bootstrap/dist/css/bootstrap.min.css';
+const SUPPORTED_BOOTSTRAP_STYLE_IMPORTS = {
+  '.sass': `
+/* Importing Bootstrap SCSS file. */
+@import '~bootstrap/scss/bootstrap'
```
benouat, line 16: I assume bootstrap does not have a proper `sass` export here for that ?
Sounds weird to import `scss` into `sass`...

schematics/ng-add/index.ts
maxokorokov: yep, it was `^${BOOTSTRAP_VERSION}` initially

schematics/collection.json
maxokorokov: yeah, I was more thinking about the library name → `ng-bootsrtap`, maybe I should put "Adds ng-bootstrap **library** to the project"

schematics/ng-add/steps/add-bootstrap.ts
maxokorokov: Nope it does not. You would prefer to support only `scss` and in the case of `sass`, just install the `bootstrap.min.css` ?

benouat: No, let's keep it like that. I assume it's working ?

maxokorokov: It is working, the difference between the two snippets is actually  just the `;` symbol at the end of the import..

PR #3680, opened 2020-04-16
feat(datepicker): improve 'restoreFocus' input

r-ising:
Hello,

in my angular project I need to disable the restore focus feature on the date picker, so I created this pull request.

It's a very trivial fix, I changed the datatype from "true" to "boolean" and added one if statement

Thanks
Ramon

### DIFF
--- a/src/datepicker/datepicker-input-config.ts
+++ b/src/datepicker/datepicker-input-config.ts
@@ -17,5 +17,5 @@ export class NgbInputDatepickerConfig extends NgbDatepickerConfig {
   container: null | 'body';
   positionTarget: string | HTMLElement;
   placement: PlacementArray = ['bottom-left', 'bottom-right', 'top-left', 'top-right'];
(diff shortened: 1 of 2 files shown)


r-ising:
@maxokorokov

I created a StackBlitz, it's a problem of ie11, the restoreFocus isn't working very well there:

https://stackblitz.com/edit/angular-cn1jsd?file=src%2Fapp%2Fdatepicker-positiontarget.html

The user click in an input field and on focus the datepicker will open, if the user want to type the date by keyboard, the datepicker will close, if the user pick a date by datepicker, the date picker close too.. It's a very nice user experience, because the user has more options. It works without any problems in Chrome, Safari and Firefox but not in IE11. In IE11 the user can't enter a date by keyboard.

Maybe you can improve the ie11 support on the restore focus feature that it works like in Chrome, Firefox, etc..

Thanks,
Ramon

r-ising:
I create a issue of this

PR #3691, opened 2020-04-19
Update nav.ts

alexnoise79:
just an idea..

### DIFF
--- a/src/nav/nav.ts
+++ b/src/nav/nav.ts
@@ -353,16 +353,16 @@ export class NgbNavLink {
  *
  * @since 5.2.0
  */
-export interface NgbNavChangeEvent {
+export interface NgbNavChangeEvent<T> {
   /**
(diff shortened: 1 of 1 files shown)


src/nav/nav.ts
```diff
@@ -353,16 +353,16 @@ export class NgbNavLink {
  *
  * @since 5.2.0
  */
-export interface NgbNavChangeEvent {
+export interface NgbNavChangeEvent<T> {
```
peterblazejewicz, line 356:
```suggestion
export interface NgbNavChangeEvent<T = any> {
```
This would work, otherwise the change will break the existing stuff (see CI failure)
With the default value for generic, the change is no brainer and useful and people don't have to pay attention to generics

alexnoise79: right!

Issue #3678, opened 2020-04-15
Need aria-describedby on modal container


woehlck:
### Feature request:
This feature request is similar to https://github.com/ng-bootstrap/ng-bootstrap/issues/2472, but specific to modals. I would like to be able to set the aria-describedby property on a modal in addition to the aria-labelledby property.

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 8.1.0

ng-bootstrap: 5.1.0

Bootstrap: 4.3.1

pkozlowski-opensource:
Related: #2049

PR #3695, opened 2020-04-22
feat(modal): need aria-describedby on modal container (#3678)

gpolychronis:

### DIFF
--- a/src/modal/modal-config.ts
+++ b/src/modal/modal-config.ts
@@ -12,6 +12,12 @@ export interface NgbModalOptions {
   ariaLabelledBy?: string;
 
   /**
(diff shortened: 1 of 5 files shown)


src/modal/modal-config.ts
```diff
@@ -11,6 +11,14 @@ export interface NgbModalOptions {
    */
   ariaLabelledBy?: string;
 
+  /**
+   * `aria-describedby` attribute value to set on the modal window.
+   *
+   * @since 6.0.4
```
maxokorokov, line 17: You shouldn't put this, because it is not clear when this will be actually merged.
We're usually adding this only for features, so for `X.X.0` releases just before doing the release.

Issue #2927, opened 2018-12-13
Accordion accessibility


maxokorokov:
I've been looking into WAI-ARIA recommendation:
- docs → https://www.w3.org/TR/wai-aria-practices-1.1/#accordion
- accordion example → https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html

Markup they propose for accordion is:

```html
<dl id="accordionGroup" role="presentation">
  <dt role="heading" aria-level="3">
    <button aria-expanded="true" aria-controls="sect1" id="accordion1id" type="button" aria-disabled="true">
      <span>Title 1</span>
    </button>
  </dt>
  <dd id="sect1" role="region" aria-labelledby="accordion1id">
    <div>
      <!-- Variable content within section, may include any type of markup or interactive widgets. -->
    </div>
  </dd>
</dl>
```

That has nothing to do with what we have (roles, attributes):

```html
<ngb-accordion class="accordion" role="tablist" aria-multiselectable="true">
  <div class="card">
    <div role="tab" id="ngb-panel-0-header" class="card-header ">
      <button type="button" aria-expanded="true" aria-controls="ngb-panel-0"> Simple </button>
    </div>
    <div class="collapse show" role="tabpanel" id="ngb-panel-0" aria-labelledby="ngb-panel-0-header">
      <div class="card-body">
        <!-- Content goes here -->
      </div>
    </div>
  </div>
</ngb-accordion>
```

---

Differences:
- roles: `tablist / tab / tabpanel` vs `presentation / heading / region`
- they have `aria-disabled` on a button
- they have `aria-level` on a heading
- we have `aria-multiselectable` on the accordion

Fixed:
~~- we have an intermediate`<h5>` (should be addressed by #2847)~~

gpolychronisAmadeus:
@maxokorokov @benouat 
Please review.
Also: added ariaLevel input to ngbPanelHeader. It defaults to 3.

IAfanasovMob:
> they have aria-disabled on a button

we have `disabled` attribute 

According to [ARIA docs](https://www.w3.org/TR/html-aria/), we do it correct
> Only use the aria-disabled attribute for elements that are not allowed to have a disabled attribute in HTML.

IAfanasov:
I created the PR to resolve accessibility issues.
Some bullets from the @maxokorokov post outdated. I followed the current documentation and added references to ARIA documentation into ng-bootstrap docs

[...]

PR #3699, opened 2020-04-23
fix(accordion): increase accessibility

IAfanasov:
- align role attributes with ARIA practices
- removed aria-multiselectable attribute
- added option to set panel's content role attribute

Relates to #2927

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -155,20 +155,20 @@ export interface NgbPanelChangeEvent {
   selector: 'ngb-accordion',
   exportAs: 'ngbAccordion',
   encapsulation: ViewEncapsulation.None,
-  host: {'class': 'accordion', 'role': 'tablist', '[attr.aria-multiselectable]': '!closeOtherPanels'},
+  host: {'class': 'accordion'},
   template: `
-    <ng-template #t ngbPanelHeader let-panel>
+    <ng-template #defaultHeaderTemplate ngbPanelHeader let-panel>
       <button class="btn btn-link" [ngbPanelToggle]="panel">
         {{panel.title}}<ng-template [ngTemplateOutlet]="panel.titleTpl?.templateRef"></ng-template>
       </button>
     </ng-template>
     <ng-template ngFor let-panel [ngForOf]="panels">
       <div [class]="'card ' + (panel.cardClass || '')">
-        <div role="tab" id="{{panel.id}}-header" [class]="'card-header ' + (panel.type ? 'bg-'+panel.type: type ? 'bg-'+type : '')">
-          <ng-template [ngTemplateOutlet]="panel.headerTpl?.templateRef || t"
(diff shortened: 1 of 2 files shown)


IAfanasov:
I think it would be better to change the header template and instead and provide the developer chance to use a header tag (like h1, h2, etc) instead of `<div role='heading' aria-level='3'`. Can not imaging how to do it without breaking changes.

IAfanasov:
@benouat good point
I had a misunderstanding of [WAI-ARIA authoring practices for accordion](https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties). I will update the pr. Also I will change the custom header example to follow those practices

IAfanasov:
and thanks for review, @benouat !

IAfanasov:
I decided to split demo problem into a separate issue #3712
All the rest is pushed

src/accordion/accordion.ts
```diff
@@ -209,6 +209,13 @@ export class NgbAccordion implements AfterContentChecked {
    */
   @Input() type: string;
 
+  /**
+   * The panel's role attribute value. According to
+   * [WAI-ARIA practices](https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties)
+   * could be a '[region](https://www.w3.org/TR/wai-aria-1.1/#region)'
+   */
+  @Input() ariaPanelRole: 'region' | null;
```
benouat, line 217: I am still not convinced, as said previously, that we should introduce new Input for that.
Is there an actual need to change that? Should it always be a `panel` ? If yes, there's no need for the input. Let's change the previous `role="tab panel"` to the new one `role="region"`.

IAfanasov: Hi @benouat !
I totally understand your concerns and happy to discuss since I don't have 100% confidence here either.

the "region" role determines if the content should appear in the Landmarks navigation menu. It is one of menus opened by VO+U.

here is the extraction from [WAI-ARIA practices for accordion widget](https://www.w3.org/TR/wai-aria-practices-1.1/#wai-aria-roles-states-and-properties):

> Optionally, each element that serves as a container for panel content has role region and aria-labelledby with a value that refers to the button that controls display of the panel.
> - Avoid using the region role in circumstances that create landmark region proliferation, e.g., in an accordion that contains more than approximately 6 panels that can be expanded at the same time.
> - Role region is especially helpful to the perception of structure by screen reader users when panels contain heading elements or a nested accordion.

so, the need for this role depends on specific usage of the widget and only developer could determine when add role region and when not. 

we could consider avoid this to be part of accordion component and be part of the content template. `Optionally ... a container for panel content has role region and aria-labelledby ...`. It doesn't feel aligned with practices to me if the `aria-labelledby` would be in one HTML element and `role="region"` in the descendant. When it is one element, the region landmark  name would be calculated using `aria-labelledby` attribute.

IAfanasov: I would be thankful for the feedback (:

IAfanasov:
outdated

PR #3702, opened 2020-04-24
demo: better display of the right menu when the scrollbar is displayed

fbasso:
The issue came form the dimensions of the right menu, not well aligned with other elements. 
The main problem was the bottom of the scrollbar, which was out of the viewport.

### DIFF
--- a/demo/src/style/app.scss
+++ b/demo/src/style/app.scss
@@ -49,7 +49,8 @@ header.navbar {
 .sidebar {
   position: sticky;
   top: 1rem;
-  max-height: 100vh;
+  max-height: calc(100vh - 4rem);
+  margin-right: -1rem;
   overflow-x: hidden;
   overflow-y: auto;
 }
--- a/demo/src/app/shared/component-wrapper/component-wrapper.component.html
+++ b/demo/src/app/shared/component-wrapper/component-wrapper.component.html
@@ -20,7 +20,7 @@
       <ngbd-side-nav
         id="doc-nav"
         [ngbCollapse]="sidebarCollapsed"
-        class="d-lg-block my-3 collapse sidebar"
+        class="d-lg-block py-3 collapse sidebar"
       ></ngbd-side-nav>
     </div>
 


demo/src/style/app.scss
```diff
@@ -49,7 +49,8 @@ header.navbar {
 .sidebar {
   position: sticky;
   top: 1rem;
-  max-height: 100vh;
+  max-height: CALC(100vh - 4rem);
```
peterblazejewicz, line 52:
```suggestion
  max-height: calc(100vh - 4rem);
```
I know it works, but UPPERCASE can be seen as an error by reader.

fbasso: Thanks @peterblazejewicz, you're right, I don't know why I used uppercase here :)

benouat:
Can we have a screenshot of both the issue and the fix ? Because I don't understand the problem in the first place 😄

fbasso:
Here they are ! The issue is more visible on windows where you can see the scrollbars.

With the issue:

image

Improved version:

image

benouat:
😱
This is the kind of thing I never see on MacOS

maxokorokov:
@benouat but it also doesn't scroll right on Mac OS when you have a short screen, so you kinda see it actually...

benouat:
> @benouat but it also doesn't scroll right on Mac OS when you have a short screen, so you kinda see it actually...

For me it's working as it should based on the CSS I produced. Then, for sure we could improve it.

If you look carefully they have the exact same issue (if we want to call it issue) on getbootstrap.com!

fbasso:
Thanks @benouat, I will merge it !

By the way, yes, they have the same kind of issue on bootstrap, but it is much more visible when the end of scrollbar is not visible, at least on windows.

It's because, when you use the middle button of the mouse to scroll down, the system has some difficulties to go till the end of the sidebar, a condition required to continue the scrolling with the rest of the page.

It's even worse if you use the drag and drop on the scrollbar, you just can't reach the bottom of the sidebar :)

PR #3703, opened 2020-04-24
docs(modal): false positive A11y warning

### LINKED ISSUE #3319: ARIA hidden element must not contain focusable elements
This library is a great example of accessible software.  Thanks for the whole team for your attention to these things.  The following may not be resolvable, but I don't see it raised as a bug yet.  

### Bug description:
After opening a modal, axe-core reports [ARIA hidden element must not contain focusable elements](https://dequeuniversity.com/rules/axe/3.2/aria-hidden-focus).

When a modal is open, `aria-hidden="true"` is applied to the app-root.  All focusable elements in the page are nested and hidden, but still tabable.

As the focus trap _should_ prevent a user from activating any of these controls this may be a false positive.  If it is, a note in the documentation about this may be all that's needed / possible.

A programatic solution may be to look for all focusable elements in the DOM on modal open and store a reference to them.  Apply `tabindex="-1"` to these elements, and on modal close reset tabindex to its previous state per element.  Any changes to the hidden elements will not be tracked though, which is pretty error prone.  This solution feels janky.

Any of the StackBlitz's on the modal demo page work: https://ng-bootstrap.github.io/#/components/modal/examples

Steps:
1. Ensure you have a means to run an axe scan on an interacted website. I use the [axe-core chrome extension](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd).
2. Open any of the StackBlitz pages, then open the run in a new window
3. Open the modal
4. Run the scan

Angular: 8.1.3
ng-bootstrap:5.0.0
Bootstrap: 4.3.1

IAfanasov:
- describes false positive A11y warning related to `aria-hidden="true"` and focus trap

Closes #3319

### DIFF
--- a/src/modal/modal.ts
+++ b/src/modal/modal.ts
@@ -23,6 +23,10 @@ export class NgbModal {
    * then instances of those components can be injected with an instance of the `NgbActiveModal` class. You can then
    * use `NgbActiveModal` methods to close / dismiss modals from "inside" of your component.
    *
+   * The modal implementation adds `aria-hidden="true"` to parent DOM elements while keeping focus trapped inside
+   * the modal. This leads to false positive warnings from A11y tools, e.g. "aria-hidden elements do not contain
+   * focusable elements" from [axe](https://dequeuniversity.com/rules/axe/3.2/aria-hidden-focus).
+   *
    * Also see the [`NgbModalOptions`](#/components/modal/api#NgbModalOptions) for the list of supported options.
    */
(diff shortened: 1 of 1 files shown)


IAfanasov:
I would be thankful for the feedback (:

benouat:
Sorry for the late review!

Please have a look at my last comment in original issue #3319 but I don't think we are going to merge that one as main point is wrong

> The modal implementation adds `aria-hidden="true"` to parent DOM elements while keeping focus trapped inside

This is not valid

PR #3706, opened 2020-04-27
docs: homepage installation snippet

benouat:
image

I also changed the threshold for responsive switch from `md` to `lg`. Basically the change occurs earlier.

### DIFF
--- a/demo/src/app/default/default.component.ts
+++ b/demo/src/app/default/default.component.ts
@@ -1,5 +1,6 @@
 import {Component} from '@angular/core';
 import {environment} from '../../environments/environment';
+import { Snippet } from '../shared/code/snippet';
 
 @Component({
   selector: 'ngbd-default',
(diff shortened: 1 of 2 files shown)


demo/src/app/default/default.component.ts
```diff
@@ -1,10 +1,17 @@
...
 @Component({
   selector: 'ngbd-default',
   templateUrl: './default.component.html'
 })
 export class DefaultComponent {
   public version: string = environment.version;
+
+  installation = Snippet({
+    lang: 'shell',
+    code: `# recommanded installation: Angular CLI
```
maxokorokov, line 14: typo → recomm**a**nded

But more importantly, I'd:
- remove this comment about CLI altogether, leave just the snippet
- remove the code highlighting, so it would be all black text and `add` word would not be red
- change `See more installation details` to something shorter? `Get started` like bootstrap does?

Something like this, WDYT?

Screen Shot 2020-04-27 at 15 29 05

maxokorokov: Or even remove 'Install with Angular CLI' completely, those who know what it is don't need explanation, those who don't → will click on Get Started ?

maxokorokov: LGTM otherwise, might be nice to have it straight away

benouat:
I am going with

PR #3708, opened 2020-04-28
demo: npm link markup for better screen reader experience

IAfanasov:

### DIFF
--- a/demo/src/app/app.component.html
+++ b/demo/src/app/app.component.html
@@ -92,7 +92,8 @@
         [title]="downloadCount + ' downloads over the last 30 days on Npm'"
         class="d-flex flex-column align-items-center"
       >
-        {{ downloadCount }}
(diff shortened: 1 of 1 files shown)


IAfanasov:
without `aria-label` screen reader sees the link as a number, which is not descriptive

IAfanasov:
> I am confused because of several things I noticed on VoiceOver:
> 
> * VO is actually first reading just the number (which is disturbing), but then it reads the title.
> * both next links (GitHub and Twitter) that only contains svg icons, VO reads properly the title.
> 
> So instead of duplicating the title + aria-label, could we try to keep only the title and put the text inside the link as `aria-hidden="true"` ?
> 
> I'd rather prefer not duplicating both attributes with the exact same content.

Good point. Changed markup and force pushed.
That was about announcement and navigation which opens on pressing VO+U. I tested both cases. In both cases is a number plus the text.

Issue #1274, opened 2017-01-30
feature(typeahead): Support for grouped results


mijelin:
### Feature request:

Ability to display results by group/category.

![screen shot 2017-01-30 at 12 00 12 pm](https://cloud.githubusercontent.com/assets/7971973/22439257/c20098e2-e6e3-11e6-9c5d-5df7d02931e6.png)

#### Examples:
https://twitter.github.io/typeahead.js/examples#multiple-datasets (shown above)
https://valor-software.com/ng2-bootstrap/#/typeahead#grouping-results

plaetzchen79:
I hope this feature is coming soon

markmssd:
+1, really need this feature!

HAkbariMoghadam:
I think that it's very very useful feature

plaetzchen79:
What can we do to increase the priority? 

Good and supported search-as-you-type-Boxes for angular >=2 are still rare

pkozlowski-opensource:
@plaetzchen79 there are number of things you can do:
* vote on this feature to be implemented (thumbs up on the initial report), as this is what we take into account while prioritizing work. See [roadmap](https://github.com/ng-bootstrap/ng-bootstrap/wiki/Roadmap) for more details;
* you can contribute design and / or PR with code.

elisabetb:
Is this feature under development or not at all?

trickeyone:
Would love to see this feature added. I had accidentally been looking at the docs for ngx-bootstrap and got really excited when I saw that it had result grouping. But then my excitement disappeared when I realized I was looking at the wrong project. I really like ng-bootstrap's setup and components, not to mention the testing. This addition would be greatly appreciated!

dotNetAthlete:
How could I have each item in a type-ahead as a parent that when selected opened a second type-ahead to show the child items as based on the following JSON:

`{
        "id": 36,
        "logTypeName": "3rd Party Accident",
        "parentId": 2,
        "parent": {
            "id": 2,
            "classification": "Accident",
            "tblLogTypeLu": [
                {
                    "id": 34,
                    "logTypeName": "Personal Injury",
                    "parentId": 2
                },
                {
                    "id": 35,
                    "logTypeName": "Vehicle Accident",
                    "parentId": 2
                }
            ]
        }
    },`

dotNetAthlete:
just use the ng-select directive instead !!
https://stackblitz.com/edit/ng-select?file=app%2Fapp.component.ts

[...]

PR #3711, opened 2020-04-30
feat(typeahead): support for grouped results (#1274)

gpolychronis:

### DIFF
--- a/demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
+++ b/demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
@@ -1,0 +1,66 @@
+import {Component} from '@angular/core';
+import {Observable} from 'rxjs';
+import {debounceTime, distinctUntilChanged, map} from 'rxjs/operators';
+
+const states = ['Alabama', 'Alaska', 'American Samoa', 'Arizona', 'Arkansas', 'California', 'Colorado',
+  'Connecticut', 'Delaware', 'District Of Columbia', 'Federated States Of Micronesia', 'Florida', 'Georgia',
+  'Guam', 'Hawaii', 'Idaho', 'Illinois', 'Indiana', 'Iowa', 'Kansas', 'Kentucky', 'Louisiana', 'Maine',
+  'Marshall Islands', 'Maryland', 'Massachusetts', 'Michigan', 'Minnesota', 'Mississippi', 'Missouri', 'Montana',
+  'Nebraska', 'Nevada', 'New Hampshire', 'New Jersey', 'New Mexico', 'New York', 'North Carolina', 'North Dakota',
+  'Northern Mariana Islands', 'Ohio', 'Oklahoma', 'Oregon', 'Palau', 'Pennsylvania', 'Puerto Rico', 'Rhode Island',
+  'South Carolina', 'South Dakota', 'Tennessee', 'Texas', 'Utah', 'Vermont', 'Virgin Islands', 'Virginia',
+  'Washington', 'West Virginia', 'Wisconsin', 'Wyoming'].sort();
+
+@Component({
+  selector: 'ngbd-typeahead-grouping',
+  templateUrl: './typeahead-grouping.html',
+  styles: [`.form-control { width: 300px; }`]
+})
+export class NgbdTypeaheadGrouping {
+  public model: any;
+  categories: any[];
+
+  search = (text$: Observable<string>) =>
+    text$.pipe(
+      debounceTime(200),
+      distinctUntilChanged(),
+      map(term => term.length < 2 ? []
+        : this.filter(term))
+    )
+
+  private filter(term: any) {
+    const filtered = states.filter(v => v.toLowerCase().indexOf(term.toLowerCase()) > -1).slice(0, 10);
+    this.categories = this.group(filtered);
+    return filtered;
+  }
+
+  group(items: string[]) {
+    let idx = 0;
+    let categories = new Map();
+    for (const item of items) {
+      const key = this.key(item);
+      if (!categories.has(key)) {
+        categories.set(key, this.category(key));
+      }
+      categories.get(key).items.push(this.item(item, idx++));
+    }
+    return Array.from(categories.values());
+  }
+
+  private category(key: string) {
+    return {
+      label: key,
+      items: []
+    };
+  }
+
+  private item(label: string, idx: number) {
+    return {
+      label, idx
+    };
+  }
+
+  private key(item: string) {
+    return item.toUpperCase().charAt(0);
+  }
+}
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -71,7 +71,8 @@ export class NgbSlide {
           (click)="focus();select(slide.id, NgbSlideEventSource.INDICATOR);"></li>
     </ol>
     <div class="carousel-inner">
-      <div *ngFor="let slide of slides; index as i; count as c" class="carousel-item" [class.active]="slide.id === activeId" [id]="slide.id + '-slide'" role="tabpanel">
+      <div *ngFor="let slide of slides; index as i; count as c" class="carousel-item"
+      [class.active]="slide.id === activeId" [id]="slide.id + '-slide'" role="tabpanel">
         <span class="sr-only" i18n="Currently selected slide number read by screen reader@@ngb.carousel.slide-number">
           Slide {{i + 1}} of {{c}}
         </span>
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -1,5 +1,4 @@
 import {Component, EventEmitter, Input, OnInit, Output, TemplateRef, ViewEncapsulation} from '@angular/core';
-
 import {toString} from '../util/util';
 
 /**
@@ -17,25 +16,32 @@ export interface ResultTemplateContext {
   term: string;
 }
 
+/**
+ * The context for the typeahead content template in case you want to override the default one.
+ */
+export interface ContentTemplateContext<T> {
+  /**
+   * Your typeahead result items.
+   */
+  results: T[];
+}
+
 @Component({
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
   encapsulation: ViewEncapsulation.None,
   host: {'(mousedown)': '$event.preventDefault()', 'class': 'dropdown-menu show', 'role': 'listbox', '[id]': 'id'},
   template: `
-    <ng-template #rt let-result="result" let-term="term" let-formatter="formatter">
-      <ngb-highlight [result]="formatter(result)" [term]="term"></ngb-highlight>
+    <ng-template #defaultContentTemplate let-results="results">
+      <ng-template ngFor [ngForOf]="results" let-item let-index="index">
+        <ngb-typeahead-item
+          [item]="item"
+          [index]="index">
+        </ngb-typeahead-item>
+      </ng-template>
     </ng-template>
-    <ng-template ngFor [ngForOf]="results" let-result let-idx="index">
-      <button type="button" class="dropdown-item" role="option"
-        [id]="id + '-' + idx"
-        [class.active]="idx === activeIdx"
-        (mouseenter)="markActive(idx)"
-        (click)="select(result)">
-          <ng-template [ngTemplateOutlet]="resultTemplate || rt"
-          [ngTemplateOutletContext]="{result: result, term: term, formatter: formatter}"></ng-template>
-      </button>
-    </ng-template>
+    <ng-template [ngTemplateOutlet]="contentTemplate || defaultContentTemplate"
+                 [ngTemplateOutletContext]="{$implicit: results, results: results}"></ng-template>
   `
 })
 export class NgbTypeaheadWindow implements OnInit {
@@ -72,6 +78,11 @@ export class NgbTypeaheadWindow implements OnInit {
    * A template to override a matching result default display
    */
   @Input() resultTemplate: TemplateRef<ResultTemplateContext>;
+
+  /**
+   * A template to override a matching result default display
+   */
+  @Input() contentTemplate: TemplateRef<ContentTemplateContext<any>>;
 
   /**
    * Event raised when user selects a particular result row
--- a/src/typeahead/typeahead.ts
+++ b/src/typeahead/typeahead.ts
@@ -1,5 +1,7 @@
 import {
+  AfterViewInit,
   ChangeDetectorRef,
+  Component,
   ComponentFactoryResolver,
   ComponentRef,
   Directive,
@@ -15,7 +17,9 @@ import {
   Output,
   Renderer2,
   TemplateRef,
+  ViewChild,
   ViewContainerRef,
+  ViewEncapsulation,
   ApplicationRef
 } from '@angular/core';
 import {ControlValueAccessor, NG_VALUE_ACCESSOR} from '@angular/forms';
@@ -31,7 +35,7 @@ import {PlacementArray, positionElements} from '../util/positioning';
 import {isDefined, toString} from '../util/util';
 
 import {NgbTypeaheadConfig} from './typeahead-config';
-import {NgbTypeaheadWindow, ResultTemplateContext} from './typeahead-window';
+import {NgbTypeaheadWindow, ResultTemplateContext, ContentTemplateContext} from './typeahead-window';
 
 
 const NGB_TYPEAHEAD_VALUE_ACCESSOR = {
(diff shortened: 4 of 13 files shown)


gpolychronis:
@maxokorokov Please review

src/typeahead/typeahead-window.ts
```diff
@@ -116,6 +184,8 @@ export class NgbTypeaheadWindow implements OnInit {
 
   select(item) { this.selectEvent.emit(item); }
 
+  getContext() { return this; }
```
benouat, line 187: I am not keen of opening this kind of hole. Not for public API.

demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
```diff
@@ -0,0 +1,76 @@
...
+        (c as any).label = category;
+        categories.push(c);
+      }
+      categories[references[category]].push({
+        label: result, idx: idx++
+      });
+    }
+    this.categories = categories;
+  }
+}
```
benouat, line 76: I think this should be extracted as a standalone file, referenced in the demo. It's not clear for user that the demo contains this component definition. It would appear in the Code tab list

demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
```diff
@@ -0,0 +1,76 @@
...
+  'Northern Mariana Islands', 'Ohio', 'Oklahoma', 'Oregon', 'Palau', 'Pennsylvania', 'Puerto Rico', 'Rhode Island',
+  'South Carolina', 'South Dakota', 'Tennessee', 'Texas', 'Utah', 'Vermont', 'Virgin Islands', 'Virginia',
+  'Washington', 'West Virginia', 'Wisconsin', 'Wyoming'];
+
+@Component({
+  selector: 'ngbd-typeahead-grouping',
+  templateUrl: './typeahead-grouping.html',
+  styles: [`.form-control { width: 300px; }`]
+})
+export class NgbdTypeaheadGrouping {
```
benouat, line 19: I think our US list of states is not relevant for that kind of demo.
We could inspire ourselves from official Twitter Typeahead.js [multi-datasets example](https://twitter.github.io/typeahead.js/examples#multiple-datasets).

benouat (commented):
I left a few comments here.

Overall, I need to think more about this feature. We've done to many times wrong API design choices on the typeahead by the past that we need to put extra effort into that one.

src/typeahead/typeahead-window.ts
maxokorokov: It is not the "public" API, but it is required to pass the 'window' to the 'window template' context and then to the 'window item' as an input.
It is a public method, but not a public API (same as `.select()` method).

https://github.com/ng-bootstrap/ng-bootstrap/pull/3711/files#diff-2b05df9eab924379f039993fae7b433eR101

That is the biggest bottleneck in the PR, but I've no idea how to solve it...

demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
maxokorokov: I would add that results should not be grouped in the component, but in the search service directly. Component `NgbdTypeaheadGroupingInternal` should just do the display. Also the component should be renamed...

demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
maxokorokov: The demo is also broken if you type 2 letters, like "Co" (Colorado, Connecticut). They will be split in 2 groups

maxokorokov:
Also, @gpolychronis, please explain things in future when opening PRs in the description. It helps to understand what is done / not done, what was your thinking, etc. It should be clear what you're trying to do without looking into the code.

Ex. what is that you're introducing, what is the public API, etc.

cc @benouat

src/typeahead/typeahead-window.ts
benouat: My point here is to **not** return _this_ !! We return an object containing only what we need to access.

```typescript
getContext() {
return {
  markActive: ...
  select: ... 
  id: ...
  activeIdx: ...
  resultTemplate: ...
}
```

If we need them _live_ we could keep getters so they are executed when accessed, but we don't return _this_. To me, this is not safe/maintainable.

benouat: Other solution, we could also pass every single needed property as one on the `templateContext` itself, instead of forwarding _this_

Then people would write
```html
<.  ..... let-markActive="markActive" let-select="select"  .... >
```

maxokorokov: I don't think there was any intention to give this kind of API for people. What's the point, if we handle all this internally?

maxokorokov: On the other hand I see what you mean. If we have this API, we could use it and then people could use it. And it might be cleaner overall.

benouat: Let me rephrase my concern here:

👍 I am not against the surface API provided by an input `@Input() context` on the component.

👎 I am against providing `this` as a value to it. `this` will move over time, it's opaque, you don't know what it provides. Forcing us to declare what we export as context makes us design better API.

And finally, for pure documentation, we **should** use an interface to declare this context, to avoid such `any` effect as shown below

image

gpolychronis:
> 
> 
> Also, @gpolychronis, please explain things in future when opening PRs in the description. It helps to understand what is done / not done, what was your thinking, etc. It should be clear what you're trying to do without looking into the code.
> 
> Ex. what is that you're introducing, what is the public API, etc.
> 
> cc @benouat

I would like to enable user to write something like that:

<div *ngFor="let category of categories">
        <h3>{{category.label}}</h3>
        <div *ngFor="let item of category">
          <ngb-typeahead-window-item
            [result]="item.label"
            [index]="item.idx"
            [context]="context">
          </ngb-typeahead-window-item>
        </div>
      </div>

I don't like context either. I would like it not be visible to user at all. However as metnioned, the problem remains. How will typeahead instance be injected in items? Any ideas?

demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
gpolychronis: There is the issue with change detection. So a new component cannot be avoided. I will try to put in a separate file though.

src/typeahead/typeahead-window.ts
gpolychronis: > 
> 
> Let me rephrase my concern here:
> 
> 👍 I am not against the surface API provided by an input `@Input() context` on the component.
> 
> 👎 I am against providing `this` as a value to it. `this` will move over time, it's opaque, you don't know what it provides. Forcing us to declare what we export as context makes us design better API.
> 
> And finally, for pure documentation, we **should** use an interface to declare this context, to avoid such `any` effect as shown below
> 
> image

I would like to make context not visible by users at all. We would like to change what fields belong to context instead of exposing them to users. The only problem is how to inject context in items without users noticing. It isn't meant to be used by users, that's why it has the generic "any". It should not be used by users at all.

gpolychronis:
This is the blocking issue: https://github.com/angular/angular/issues/14935

benouat (changes requested):
@gpolychronis Have a look at TravisCI output, there is a circular dependency in your code. I don't know if the failing UT is due to that, most probably yes.
You'll have to use `forwardRef` I think to overcome this. Search for it on the project to find use case usage of it.

gpolychronis:
@benouat It seems there is an error on Travis side. Am I right, or there is an error on my side? Please chec.

gpolychronis:
UPDATED DESCRIPTION:

In order to enable result grouping, a new template has been added. The new template allows user to customize the layout of the typeahead results.

Technically, there is a problem. The NgbTypeaheadWindow instance must be injected into the items. Right now ngbTypeahead is bound to input element. Input element has no closing tag, so injection via html is not possible.

The solution provided is this: untie ngbTypeahead from input. This way ngbTypeahead can be used with a div element. The layout template can be included inside the this div as long as the input itself. 

At the same time the old way shall be supported too. So ngbTypeahead checks if there is any input child, else it assumes bound element is the input one. Also, if ngbTypeahed is bound to a div, then this div is supposed to include only the input, its label and the layout template.

@maxokorokov @benouat Please review

benouat:
@gpolychronis what is this last force push about ?

gpolychronis:
> 
> 
> @gpolychronis what is this last force push about ?

An attempt to rebuild and check again the tests. It worked.

gpolychronis:
1. Test if it fails gracefully if misused.
2. New PR: update all demo examples.
3. Replace  @ViewChild('input', {static: true}) childInput: ElementRef<HTMLInputElement>; with query
4a. "ngb-typeahead-window-item" -> "ngb-typeahead-item"
4b. "ngb-typeahead-window-item".result -> "ngb-typeahead-item".item

gpolychronis:
> 
> 
>     1. Test if it fails gracefully if misused.
> 
>     2. New PR: update all demo examples.
> 
>     3. Replace  @ViewChild('input', {static: true}) childInput: ElementRef; with query
>        4a. "ngb-typeahead-window-item" -> "ngb-typeahead-item"
>        4b. "ngb-typeahead-window-item".result -> "ngb-typeahead-item".item

Done. @maxokorokov , @benouat please review.

demo/src/app/components/typeahead/demos/grouping/typeahead-grouping.ts
benouat: My previous comment here is still valid to me, we should think about another problem domain for this demo. List of states is not good...

benouat (changes requested):
I have seen a lot of progress in the PR 👍
I have reviewed things again and do have some remarks and comments I left here and there.
Please @gpolychronis have a look a them

src/carousel/carousel.ts
```diff
@@ -71,7 +71,8 @@ export class NgbSlide {
           (click)="focus();select(slide.id, NgbSlideEventSource.INDICATOR);"></li>
     </ol>
     <div class="carousel-inner">
-      <div *ngFor="let slide of slides; index as i; count as c" class="carousel-item" [class.active]="slide.id === activeId" [id]="slide.id + '-slide'" role="tabpanel">
+      <div *ngFor="let slide of slides; index as i; count as c" class="carousel-item"
+      [class.active]="slide.id === activeId" [id]="slide.id + '-slide'" role="tabpanel">
```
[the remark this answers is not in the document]
gpolychronis: CI cannot work otherwise

gpolychronis:
Just a note: One typeahead template variable "result" is renamed to "item". This will probably break backwards compatibility.

src/typeahead/typeahead.ts
```diff
@@ -45,7 +48,7 @@ const NGB_TYPEAHEAD_VALUE_ACCESSOR = {
  */
 export interface NgbTypeaheadSelectItemEvent<T = any> {
   /**
-   * The item from the result list about to be selected.
+   * The item from the item list about to be selected.
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead.ts
```diff
@@ -97,7 +99,13 @@ export class NgbTypeahead implements ControlValueAccessor,
    *
    * @since 2.1.0
    */
-  @Input() autocomplete = 'off';
+  @Input()
+  set autocomplete(_autocomplete: string) {
+    this._autocomplete = _autocomplete;
+    if (this._inputRef) {
+      this._inputRef.nativeElement.setAttribute('autocomplete', this._autocomplete);
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead.ts
```diff
@@ -112,12 +120,12 @@ export class NgbTypeahead implements ControlValueAccessor,
...
 
   /**
-   * If `true`, the first item in the result list will always stay focused while typing.
+   * If `true`, the first item in the item list will always stay focused while typing.
    */
   @Input() focusFirst: boolean;
 
   /**
-   * The function that converts an item from the result list to a `string` to display in the `<input>` field.
+   * The function that converts an item from the item list to a `string` to display in the `<input>` field.
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead.ts
```diff
@@ -138,7 +146,7 @@ export class NgbTypeahead implements ControlValueAccessor,
   @Input() ngbTypeahead: (text: Observable<string>) => Observable<readonly any[]>;
 
   /**
-   * The function that converts an item from the result list to a `string` to display in the popup.
+   * The function that converts an item from the item list to a `string` to display in the popup.
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead.ts
```diff
@@ -210,13 +226,20 @@ export class NgbTypeahead implements ControlValueAccessor,
...
   }
 
   ngOnInit(): void {
+    this._inputRef = this.childInput || this._elementRef;
+    this._inputRef.nativeElement.setAttribute('autocomplete', this._autocomplete);
+    this._inputRef.nativeElement.setAttribute('autocapitalize', 'off');
+    this._inputRef.nativeElement.setAttribute('autocorrect', 'off');
+    this._valueChanges = fromEvent<Event>(this._inputRef.nativeElement, 'input')
+                             .pipe(map($event => ($event.target as HTMLInputElement).value));
+
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead.ts
```diff
@@ -210,13 +226,20 @@ export class NgbTypeahead implements ControlValueAccessor,
...
             this.container === 'body');
       }
     });
   }
 
   ngOnInit(): void {
+    this._inputRef = this.childInput || this._elementRef;
+    this._inputRef.nativeElement.setAttribute('autocomplete', this._autocomplete);
+    this._inputRef.nativeElement.setAttribute('autocapitalize', 'off');
+    this._inputRef.nativeElement.setAttribute('autocorrect', 'off');
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead.ts
```diff
@@ -405,3 +433,55 @@ export class NgbTypeahead implements ControlValueAccessor,
...
+      [id]="id + '-' + index"
+      [class.active]="index === activeIdx"
+      (mouseenter)="markActive()"
+      (click)="select()">
+        <ng-template [ngTemplateOutlet]="resultTemplate || rt"
+                     [ngTemplateOutletContext]="resultTemplateContext"></ng-template>
+    </button>
+  `
+})
+export class NgbTypeaheadItem {
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead-window.ts
```diff
@@ -17,25 +16,32 @@ export interface ResultTemplateContext {
   term: string;
 }
 
+/**
+ * The context for the typeahead content template in case you want to override the default one.
+ */
+export interface ContentTemplateContext {
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead-window.ts
```diff
@@ -17,25 +16,32 @@ export interface ResultTemplateContext {
...
-      <button type="button" class="dropdown-item" role="option"
-        [id]="id + '-' + idx"
-        [class.active]="idx === activeIdx"
-        (mouseenter)="markActive(idx)"
-        (click)="select(result)">
-          <ng-template [ngTemplateOutlet]="resultTemplate || rt"
-          [ngTemplateOutletContext]="{result: result, term: term, formatter: formatter}"></ng-template>
-      </button>
+    <ng-template #defaultContentTemplate let-r="results">
+      <ng-template ngFor [ngForOf]="r" let-item let-idx="index">
```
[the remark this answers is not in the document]
gpolychronis: done

src/typeahead/typeahead-window.ts
```diff
@@ -17,25 +16,32 @@ export interface ResultTemplateContext {
...
+        <ngb-typeahead-item
+          [item]="item"
+          [index]="idx">
+        </ngb-typeahead-item>
+      </ng-template>
     </ng-template>
+    <ng-template [ngTemplateOutlet]="contentTemplate || defaultContentTemplate"
+                 [ngTemplateOutletContext]="{results: results}"></ng-template>
```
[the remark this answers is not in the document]
gpolychronis: done

finau:
Any update on the progress team?  we need this feature please.

Nosfistis:
What happened to this PR? Any chance to progress it or any other typeahead group approach?

PR #3716, opened 2020-05-04
Alert animations

maxokorokov:
This is completely based on cherry-picked #2817. It contains extracted alert animations and initial version of the `ngbRunTransition`

What is does on top in 3 separate commits is:
- refactors `ngbRunTransition` leaving only what is necessary for alert
- adds unit tests for `ngbRunTransition` with full coverage
- adds unit tests for alert animations with full coverage

Alert animations API at the moment is:

```html
<ngb-alert #a #ngIf="closed" (close)="closed = true"></ngb-alert>
<button (click)="a.close()">Close</button>
```

Unclear points for me are:
- [ ] Is the public API ok?
- [x] ~~don't like this name → `running: 'continue' | 'restart';`~~
- [x] ~~unsure about the `runningTransitions` API, maybe too much generalisation? Maybe leave the boolean flag in the component itself. Also see next point.~~
- [x] ~~memory leaks with abrupt transition termination (ex. removed DOM before transition finishes)~~

cc @fbasso, @benouat 

P.S. Coverage:

Screen Shot 2020-05-04 at 11 01 32

### DIFF
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -12,6 +12,8 @@ import {
   ViewEncapsulation
 } from '@angular/core';
 
+import {Observable} from 'rxjs';
+
 import {NgbAlertConfig} from './alert-config';
 import {ngbRunTransition} from '../util/transition/ngbTransition';
 import {ngbAlertFadingTransition} from '../util/transition/ngbFadingTransition';
@@ -23,6 +25,7 @@ import {ngbAlertFadingTransition} from '../util/transition/ngbFadingTransition';
  */
 @Component({
   selector: 'ngb-alert',
+  exportAs: 'ngbAlert',
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.None,
   host: {'role': 'alert', 'class': 'alert show fade', '[class.alert-dismissible]': 'dismissible'},
@@ -64,7 +67,7 @@ export class NgbAlert implements OnInit,
   /**
    * An event emitted when the close button is clicked. It has no payload and only relevant for dismissible alerts.
    */
-  @Output('close') closeEvent = new EventEmitter<void>();
+  @Output() closed = new EventEmitter<void>();
 
 
   constructor(config: NgbAlertConfig, private _renderer: Renderer2, private _element: ElementRef) {
@@ -73,11 +76,20 @@ export class NgbAlert implements OnInit,
     this.animation = config.animation;
   }
 
-  close() {
-    ngbRunTransition(this._element.nativeElement, ngbAlertFadingTransition, {
-      animation: this.animation,
-      runningTransition: 'continue'
-    }).subscribe(() => this.closeEvent.emit());
+  /**
+   * Triggers alert closing programmatically (same as clicking on the close button (×)).
+   *
+   * The returned observable will emit and be completed once the closing transition has finished.
+   * If the animations are turned off this happens synchronously.
+   *
+   * Alternatively you could listen or subscribe to the `(closed)` output
+   */
+  close(): Observable<void> {
+    const transition = ngbRunTransition(
+        this._element.nativeElement, ngbAlertFadingTransition,
+        {animation: this.animation, runningTransition: 'continue'});
+    transition.subscribe(() => this.closed.emit());
+    return transition;
   }
 
   ngOnChanges(changes: SimpleChanges) {
--- a/src/alert/alert.ts
+++ b/src/alert/alert.ts
@@ -13,6 +13,8 @@ import {
 } from '@angular/core';
 
 import {NgbAlertConfig} from './alert-config';
+import {ngbRunTransition} from '../util/transition/ngbTransition';
+import {ngbAlertFadingTransition} from '../util/transition/ngbFadingTransition';
 
 /**
  * Alert is a component to provide contextual feedback messages for user.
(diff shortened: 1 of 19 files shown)


maxokorokov:
Also cosmetic changes I did here:
- renamed `NgbRunTransition` → `ngbRunTransition`, because it is a function
- renamed `NgbAlertFadeTransition` → `ngbAlertFadeTransition`, because it is a function
- ordered fields alphabetically

src/util/transition/ngbTransition.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const runningTransitions = new Map<HTMLElement, Subject<any>>();
+
+export const ngbRunTransition =
+    (element: HTMLElement, startFn: (element: HTMLElement) => void, options: NgbTransitionOptions):
+        Observable<undefined> => {
+
+          // checking if animations are enabled
+          if (!options.animation) {
+            return of(undefined);
```
maxokorokov, line 17: FYI, the difference between `of()` and `EMTPY` is that both are completed, but of also returns a value before completion. See unit tests for more details.

src/alert/alert.ts
```diff
@@ -23,43 +25,61 @@ import {NgbAlertConfig} from './alert-config';
...
    * Bootstrap provides styles for the following types: `'success'`, `'info'`, `'warning'`, `'danger'`, `'primary'`,
    * `'secondary'`, `'light'` and `'dark'`.
    */
   @Input() type: string;
+
   /**
    * An event emitted when the close button is clicked. It has no payload and only relevant for dismissible alerts.
    */
-  @Output() close = new EventEmitter<void>();
+  @Output('close') closeEvent = new EventEmitter<void>();
```
maxokorokov, line 68: Also I suggest we avoid doing this, because it is different in the template and in the code and it might be confusing:

```html
<ngb-alert (close)="...">
```
vs 
```ts
alert.closeEvent.subscribe(...)
```

As we're breaking API anyway, maybe we should do this:

```html
<ngb-alert #a #ngIf="closed" (closed)="closed = true"></ngb-alert> ←note 'closed', not 'close'
<button (click)="a.close()">Close</button>
```

It will be aligned with bootstrap's event name too

src/util/transition/ngbTransition.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    const errorSpy = createSpy();
+
+    ngbRunTransition(element, fadeFn, {animation: true, running: 'continue'}).subscribe(nextSpy, errorSpy, () => {
+      expect(nextSpy).toHaveBeenCalledWith(undefined);
+      expect(element.classList.contains('ngb-test-show')).toBe(false);
+      expect(window.getComputedStyle(element).opacity).toBe('0');
+      expect(errorSpy).not.toHaveBeenCalled();
+      done();
+    });
+  });
```
fbasso, line 36: Just an idea, and I don't know if it's necessary or not, but here you test that everything is fine after that the asynchronous action is done. But you don't test the 'asynchronous nature' of the action.

I mean, if in ngbTransition the observer is resolved synchronously for any (wrong) reason (in this case, the expected transition is not run), this test will not fail, right ?

Just adding `expect(window.getComputedStyle(element).opacity).not.toBe('0');` after ngbRunTransition should ensure it.

What do you think ?

maxokorokov: Done, thanks

maxokorokov:
Updated the PR with the `timer()` fallback that uses `getTransitionDurationMs()` in case of `transitionend` event is not fired for some reason (ex. browser becomes inactive, element invisible, etc)

src/util/transition/util.ts
```diff
@@ -0,0 +1,7 @@
+export function getTransitionDurationMs(element: HTMLElement) {
+  const {transitionDelay, transitionDuration} = window.getComputedStyle(element);
+  const transitionDelaySec = parseFloat(transitionDelay);
+  const transitionDurationSec = parseFloat(transitionDuration);
+
```
fbasso, line 5: We need to be careful with the returned value. At the moment, there is only one duration and delay with the Bootstrap transitions, but the value returned can be multiple, for example : "1s, 0.25s" (in case of of a custom theme for example).

fbasso:
LGTM !

src/alert/alert.ts
benouat: I am fine with that

src/test/common.ts
```diff
@@ -63,6 +63,14 @@ export function isBrowser(browsers: Browser | Browser[], ua = window.navigator.u
   }
 }
 
+export function isBrowserVisible(suiteName: string) {
+  if (document.hidden) {
+    console.warn(`${suiteName} tests were skipped because browser is not visible`);
```
benouat, line 68: we should explicitly  log that `browser tab` is not visible. The browser window might be in focus, but the tab in which tests are running not.

src/util/transition/ngbTransition.ts
```diff
@@ -0,0 +1,65 @@
...
+import {environment} from '../../environment';
+
+export type NgbTransitionStartFn = (element: HTMLElement) => void;
+
+export interface NgbTransitionOptions {
+  animation: boolean;
+  runningTransition: 'continue' | 'stop';
+}
+
+const TIMER_DELAY_MS = environment.production ? 5 : 500;
```
benouat, line 13: why can't we use 5 everytime ?

src/util/transition/util.ts
maxokorokov: For this one I just copied the bootstrap's behaviour, they pick the first one for some reason. I haven't investigated further, but added a unit test for this: https://github.com/ng-bootstrap/ng-bootstrap/pull/3716/files#diff-249b446faed753f89f2e46d4f08fef51R44-R47

src/alert/alert.ts
maxokorokov: Changing to `(closed)` then

src/util/transition/ngbTransition.ts
maxokorokov: Will add a comment, but basically some tests become unstable with timeout firing before transitionend

fbasso: Already discussed with @maxokorokov, but I would set the timer delay directly in environment, to avoid the test `environment.production`

maxokorokov:
Pushed API change → https://github.com/ng-bootstrap/ng-bootstrap/pull/3716/commits/e6a0eb496a003bb9e1bcf632180cc77203112c21

PR #3725, opened 2020-05-08
demo(modal): improve explanations on entryComponents

### LINKED ISSUE #3722: entryComponents + modals
Hey guys,

Sorry I have a question regarding the entryComponents and Angular 9. Is it still necessary to add modals to the entryComponents under your module(s) because everything seems to be working correctly when I take them out. 

I was looking through the examples on the main site; noticed they have been updated to utilize Angular 9 but are still using the entryComponents piece when dealing with modals.

Thanks!

jnizet:
entryComponents is not needed anymore since Angular 9.

fix #3722

### DIFF
--- a/demo/src/app/components/modal/demos/component/modal-component.module.ts
+++ b/demo/src/app/components/modal/demos/component/modal-component.module.ts
@@ -8,7 +8,7 @@ import { NgbdModalComponent, NgbdModalContent } from './modal-component';
   imports: [BrowserModule, NgbModule],
   declarations: [NgbdModalComponent, NgbdModalContent],
   exports: [NgbdModalComponent],
-  bootstrap: [NgbdModalComponent],
-  entryComponents: [NgbdModalContent]
+  bootstrap: [NgbdModalComponent]
+  // entryComponents: [NgbdModalContent] // this line would be needed in Angular 8 or older
 })
 export class NgbdModalComponentModule {}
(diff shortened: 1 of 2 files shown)


demo/src/app/components/modal/demos/component/modal-component.module.ts
```diff
@@ -8,7 +8,7 @@ import { NgbdModalComponent, NgbdModalContent } from './modal-component';
   imports: [BrowserModule, NgbModule],
   declarations: [NgbdModalComponent, NgbdModalContent],
   exports: [NgbdModalComponent],
-  bootstrap: [NgbdModalComponent],
-  entryComponents: [NgbdModalContent]
+  bootstrap: [NgbdModalComponent]
+  // entryComponents: [NgbdModalContent] // this line would be needed in Angular 8 or older
```
pkozlowski-opensource, line 12: It is more about ivy enabled or not - people can be on v9 and have ivy disabled. This situation will continue for a while (until we remove VE)

jnizet: Right. I didn't want to be too technical, and simply assume the defaults here. Not sure if all people know what Ivy is and what they're using without knowing it. But I can of course be more precise. How about this?

> this line is not needed if Ivy is used (i.e., by default, in Angular 9 or newer)

And the the text in the demo:

> In this case, if you're still using ViewEngine (which is the default in Angular 8 or before) remember to add the content component
in the <code>entryComponents</code> section of your <code>NgModule</code>. For Ivy (which is the default in Angular 9 or newer), it's not needed anymore.

benouat:
@jnizet @pkozlowski-opensource jumping into the conversation here

I realised several times in the past, and today this PR is no exception, that demos are definitely **not** the right place to document/describe a feature.

Thus, wouldn't it be the right time to introduce the Modal Overview page ?

WDYT ? cc @maxokorokov

demo/src/app/components/modal/demos/component/modal-component.module.ts
benouat: What about detailing the text in demo without even referring to ViewEngine or Ivy ?

We could state `up to Angular 8` and `Angular 9+` ? at least, no confusing, no technical detail...

PR #3727, opened 2020-05-13
Collapse animations

maxokorokov:
Again the collapsing implementation is cherry-picked and is based on #2817.

This contains several commits (**please review in the historical order with whitespaces/lines diff turned off for simplicity**):
- a bugfix for duration calculation
- add `endFn()` along with `startFn()` for transition
- support transition reverting with `runningTransition: stop`
- introduction of transition context to store parameters and data
- finally cherry-picked collapsed implementation and tests

Changes to the original PR are:
- end function has no arguments and simply using a closure (see these tests to get the idea of how it works → https://github.com/ng-bootstrap/ng-bootstrap/pull/3727/commits/5a8bebd0d3f2244b17da51f0fcd3b32a4d6b0723#diff-ec1addfdfcf53282cbf22a4f924fad6bR206-R264)
- using a transaction context for both data and initial arguments (so no need in storing things in DOM, because we already have the element-releated context)

cc @benouat @fbasso

### DIFF
--- a/src/collapse/collapse.ts
+++ b/src/collapse/collapse.ts
@@ -1,4 +1,7 @@
-import {Directive, Input} from '@angular/core';
+import {Directive, Input, ElementRef, Output, EventEmitter} from '@angular/core';
+import {ngbRunTransition} from '../util/transition/ngbTransition';
+import {ngbCollapsingTransition} from '../util/transition/ngbCollapseTransition';
+import {NgbConfig} from '../ngb-config';
 
 /**
  * A directive to provide a simple way of hiding and showing elements on the page.
@@ -10,7 +13,34 @@ import {Directive, Input} from '@angular/core';
 })
 export class NgbCollapse {
   /**
+   * If `true`, collapse will be animated.
+   *
+   * Animation is triggered only when clicked on triggering element
+   * or via the `.toggle()` function
+   */
+  @Input() animation = false;
+
+  /**
    * If `true`, will collapse the element or show it otherwise.
    */
   @Input('ngbCollapse') collapsed = false;
+
+  @Output() ngbCollapseChange = new EventEmitter<boolean>();
+
+  constructor(private _element: ElementRef, ngbConfig: NgbConfig) { this.animation = ngbConfig.animation; }
+
+  /**
+   * Triggers collapsing programmatically.
+   *
+   * If there is a collapsing transition running already, it will be reversed.
+   * If the animations are turned off this happens synchronously.
+   */
+  toggle(open: boolean = this.collapsed) {
+    this.collapsed = !open;
+    ngbRunTransition(this._element.nativeElement, ngbCollapsingTransition, {
+      animation: this.animation,
+      runningTransition: 'stop',
+      context: {direction: open ? 'show' : 'hide'}
+    }).subscribe(() => this.ngbCollapseChange.next(this.collapsed));
+  }
 }
--- a/src/util/transition/ngbCollapseTransition.ts
+++ b/src/util/transition/ngbCollapseTransition.ts
@@ -1,0 +1,63 @@
+import {NgbTransitionStartFn} from './ngbTransition';
+import {reflow} from '../util';
+
+export interface NgbCollapseCtx {
+  direction: 'show' | 'hide';
+  maxHeight?: string;
+}
+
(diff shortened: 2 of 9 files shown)


src/util/transition/ngbCollapseTransition.ts
```diff
@@ -0,0 +1,63 @@
...
+    classList.add('show');
+  }
+
+  element.style.height = '';
+  const height = element.getBoundingClientRect().height + 'px';
+
+  if (!hasShownClass) {
+    classList.remove('show');
+  }
+
```
fbasso, line 22: Just a detail here, but the 'show' string (for the classname) can be factorized in a constant.

benouat: it would not make the code more readable at all.

src/collapse/collapse.ts
```diff
@@ -9,8 +12,35 @@ import {Directive, Input} from '@angular/core';
...
+   *
+   * Animation is triggered only when clicked on triggering element
+   * or via the `.toggle()` function
+   */
+  @Input() animation = false;
+
   /**
    * If `true`, will collapse the element or show it otherwise.
    */
   @Input('ngbCollapse') collapsed = false;
```
benouat, line 26: I don't remember why we are _renaming_ this input here ?

src/util/transition/ngbTransition.spec.ts
```diff
@@ -2,14 +2,16 @@ import {ngbRunTransition} from './ngbTransition';
...
 /**
  * This is sometimes necessary only for IE when it fails to recalculate styles synchronously
  * after the 'transitionend' event was fired. To remove when not supporting IE anymore.
  */
-function getComputedStyleAsync(element: HTMLElement, style: keyof CSSStyleDeclaration) {
-  return new Promise<string>(resolve => setTimeout(() => resolve(window.getComputedStyle(element)[style])));
+function getComputedStyleAsync(element: HTMLElement, style: keyof CSSStyleDeclaration): Promise<string> {
+  const getStyle = () => window.getComputedStyle(element)[style];
+  return isBrowser('ie') ? new Promise<string>(resolve => setTimeout(() => resolve(getStyle()), 16)) :
+                           Promise.resolve(getStyle());
```
fbasso, line 14: Just another idea on how to manage it here. Instead of having a dedicated function to manage the 'get the computed style later for ie', I would have preferred to introduce a new generic utility, like `wait`. 

I think it would separate the the fact that the browser needs to wait, and the logic that is required after the wait. By the wait, this generic function can be useful if needed in other situation.

In this test file, it would be used like this:

```
import {isBrowser, isBrowserVisible, wait} from '../../test/common';
...
const DELAY = isBrowser('ie') ? 16 : 0;
...
expect(errorSpy).not.toHaveBeenCalled();
await wait(DELAY);
expect(window.getComputedStyle(element).opacity).toBe('0');
```
Instead of 
```
expect(await getComputedStyleAsync(element, 'opacity')).toBe('0');
```

What do you think ?

fbasso:
This PR is fine for me, I reviewed all the commits. My only remarks is about the getComputedStyleAsync, where I would have preferred an asynchronous wait, following by a usual expectation (I know it has been introduced in a PR before, but I didn't think about it at the time).

For the rest, it's fine for me ! 👍

src/collapse/collapse.ts
maxokorokov: My guess is just to match the directive name `ngbCollapse` AND avoid using expressions like `ngbCollapse = true` internally in favour of more readable `collapsed = true`

benouat: Ok, my point here was: I would be more comfortable with `@Input() collapsed = false`
So internaly still readable, externally **more** readable.

```
<div ngbCollapse [collapsed]="collapsed" ></div>
```

But I guess, here, I am falling in the trap of personal preference 🤔 Forget about it!

maxokorokov: That would be a breaking change, no? And not really related to this PR tbh :)

src/util/transition/ngbTransition.spec.ts
maxokorokov: I like your idea much more than mine. However I'll add another `refactor` commit for this instead of fixing all individual commits in this PR

Issue #3747, opened 2020-05-28
Animations progress


maxokorokov:
- [x] ~~Alert~~ (#3716)
- [x] ~~Toast~~ (#3749)
- [x] ~~Collapse~~ (#3727)
- [x] ~~Modal~~ (#3746)
  - [x] ~~Static backdrop~~ (#3768)
- [x] ~~Navs~~ (#3776)
- [x] ~~Popover~~ (#3792)
- [x] ~~Tooltip~~ (#3798)
- [x] ~~Carousel~~ (#3804)
- [x] ~~Accordion~~ (#3766)

Reference proof-of-concept PR: #2817 

Potentially to refactor:
- [x] ~~`[class.show]` bindings should be handled with transitions, not via host bindings to keep things the same with and without animations~~
- [x] ~~`runningTransition: 'stop' | 'continue'` could be an enum~~
- [x] ~~decide if we really need separate common transition functions like `ngbAlertFadeTransition`~~
- [x] ~~SSR is broken for ngbCollapse transition that does `.getClienBoundingRect` on init~~ (#3864)
- [x] ~~check all `shown/hidden/closed/opened` event names and imperative APIs like `.open()/.close()` for consistency across all components.~~
- [x] ~~some components set classes manually during initialisation (carousel), some just run a transition (collapse)~~
- [ ] review the way `reflow()` and `zone.onStable` are used
- [ ] Accordion might be broken if using numeric panel ids (don't have tests for this)

jnizet:
I was wondering if there was a plan to make the animation support (ngbRunTransition) public, i.e. usable from applications using ng-bootstrap.
I can see it useful for application-provided widgets or components.

maxokorokov:
@jnizet, yes, we had a couple of discussions to make the following public: `ngbAutoclose`, `ngbFocustrap`, `ngbRunTransition`, maybe something else, I'm definitely forgetting something. But no concrete steps were taken yet.

shivarajnaidu:
Any target for the stable v8 release date ?

PR #3749, opened 2020-05-28
Toast animations

fbasso:
Inspired by #2817, not cherry-picked because the toast animations are finally quite different compared to the alert.

This PR has 3 commits:
- toast animations implementation
- fix some test, due to a change on ngbTransition (the startFn function and its optional callback are now run even if animation = false)
- The changes in the demo examples

### Toast animations

The toast animations are defined as follow:

- The `fade` class exists only when the animations are defined,
- When visible, the toast contains the class `show`,
- When not visible, the toast contains the class `hide` (if, for any reason, the developer choose to keep the toast in the dom),
- When closing, the `show` class is removed to start the transition, then the `hide` class is added once finished,
- When opening, the `hide` class is removed, the `showing` class is added. Then the `showing` class is replaced the the `show` class,
- `show` (new api) and `hide` return the transition observer (to be able to implement code after the animations without the events),
- New events are available: `shown` and `hidden` (`hide`is deprecated, to stick with the official Bootstrap api).

### DIFF
--- a/demo/src/app/components/toast/demos/closeable/toast-closeable.ts
+++ b/demo/src/app/components/toast/demos/closeable/toast-closeable.ts
@@ -7,6 +7,6 @@ export class NgbdToastCloseable {
 
   close() {
     this.show = false;
-    setTimeout(() => this.show = true, 5000);
+    setTimeout(() => this.show = true, 3000);
   }
 }
--- a/src/toast/toast.spec.ts
+++ b/src/toast/toast.spec.ts
@@ -1,21 +1,21 @@
 import {Component} from '@angular/core';
-import {ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
+import {async, ComponentFixture, fakeAsync, TestBed, tick} from '@angular/core/testing';
 
-import {createGenericTestComponent} from '../test/common';
+import {createGenericTestComponent, isBrowserVisible} from '../test/common';
 import {NgbToastModule} from './toast.module';
+
+import {NgbConfig} from '../ngb-config';
+import {NgbConfigAnimation} from '../test/ngb-config-animation';
 
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
 
-const getElementWithSelector = (fixture: ComponentFixture<TestComponent>, className) =>
-    fixture.nativeElement.querySelector(className);
+const getElementWithSelector = (element: HTMLElement, className) => element.querySelector(className);
 
-const getToastElement = (fixture: ComponentFixture<TestComponent>): Element =>
-    getElementWithSelector(fixture, 'ngb-toast');
-const getToastHeaderElement = (fixture: ComponentFixture<TestComponent>): Element =>
-    getElementWithSelector(fixture, 'ngb-toast .toast-header');
-const getToastBodyElement = (fixture: ComponentFixture<TestComponent>): Element =>
-    getElementWithSelector(fixture, 'ngb-toast .toast-body');
+const getToastElement = (element: HTMLElement): Element => getElementWithSelector(element, 'ngb-toast');
+const getToastHeaderElement = (element: HTMLElement): Element =>
+    getElementWithSelector(element, 'ngb-toast .toast-header');
+const getToastBodyElement = (element: HTMLElement): Element => getElementWithSelector(element, 'ngb-toast .toast-body');
 
 describe('ngb-toast', () => {
 
@@ -30,12 +30,13 @@ describe('ngb-toast', () => {
     it('should have default classnames', () => {
       const fixture = createTestComponent(`<ngb-toast header="header">body</ngb-toast>`);
       // Below getter are using Bootstrap classnames
-      const toast = getToastElement(fixture);
-      const header = getToastHeaderElement(fixture);
-      const body = getToastBodyElement(fixture);
+      const toast = getToastElement(fixture.nativeElement);
+      const header = getToastHeaderElement(fixture.nativeElement);
+      const body = getToastBodyElement(fixture.nativeElement);
 
       expect(toast).toBeDefined();
       expect(toast).toHaveCssClass('toast');
+      expect(toast).not.toHaveCssClass('fade');
       expect(toast).toHaveCssClass('show');
       expect(header).toBeDefined();
       expect(body).toBeDefined();
@@ -43,13 +44,13 @@ describe('ngb-toast', () => {
 
     it('should not generate a header element when header input is not specified', () => {
       const fixture = createTestComponent(`<ngb-toast>body</ngb-toast>`);
-      const toastHeader = getToastHeaderElement(fixture);
+      const toastHeader = getToastHeaderElement(fixture.nativeElement);
       expect(toastHeader).toBeNull();
     });
 
     it('should contain a close button when header is specified', () => {
       const fixture = createTestComponent(`<ngb-toast header="header">body</ngb-toast>`);
-      const toastHeader = getToastHeaderElement(fixture);
+      const toastHeader = getToastHeaderElement(fixture.nativeElement);
       expect(toastHeader.querySelector('button.close')).toBeDefined();
     });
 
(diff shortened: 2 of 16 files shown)


src/toast/toast.ts
```diff
@@ -118,7 +144,19 @@ export class NgbToast implements AfterContentInit,
 
   hide() {
```
maxokorokov, line 145: Could we add documentation for both `.hide()` and `.show()` methods? At the moment there is no documentation for this on the demo site?

maxokorokov:
Also ideally the `BREAKING CHANGE: ` commit message should be for the `feat(toast): add animations` commit.

src/toast/toast.ts
```diff
@@ -97,17 +115,25 @@ export class NgbToast implements AfterContentInit,
...
       this.ariaLive = config.ariaLive;
     }
     this.delay = config.delay;
     this.autohide = config.autohide;
+    this.animation = config.animation;
   }
 
-  ngAfterContentInit() { this._init(); }
+  ngAfterContentInit() {
+    this._zone.onStable.asObservable().pipe(take(1)).subscribe(() => {
```
benouat, line 132: Why do we need zone here ?

src/util/transition/ngbFadingTransition.ts
```diff
@@ -3,3 +3,21 @@ import {NgbTransitionStartFn} from './ngbTransition';
...
+    classList.remove('showing');
+    classList.add('show');
+  };
+};
+
+export const ngbToastFadeOutTransition: NgbTransitionStartFn<NgbToastCtx> = ({classList}: HTMLElement) => {
+  classList.remove('show');
+
+  return () => { classList.add('hide'); };
+};
```
benouat, line 23: After 3 widgets conversion, I am still not convinced about the need for this abstraction.
These functions should be inlined in each component, even if used several times.

src/util/transition/ngbFadingTransition.ts
```diff
@@ -3,3 +3,21 @@ import {NgbTransitionStartFn} from './ngbTransition';
...
+  classList.remove('hide');
+  classList.add('showing');
+
+  return () => {
+    classList.remove('showing');
+    classList.add('show');
+  };
+};
+
+export const ngbToastFadeOutTransition: NgbTransitionStartFn<NgbToastCtx> = ({classList}: HTMLElement) => {
```
benouat, line 19: Don't forget to remove the context in the types also

src/util/transition/ngbFadingTransition.ts
```diff
@@ -3,3 +3,21 @@ import {NgbTransitionStartFn} from './ngbTransition';
 export const ngbAlertFadingTransition: NgbTransitionStartFn = ({classList}: HTMLElement) => {
   classList.remove('show');
 };
+
+export interface NgbToastCtx { direction: 'show' | 'hide'; }
+
+export const ngbToastFadeInTransition: NgbTransitionStartFn<NgbToastCtx> = ({classList}: HTMLElement) => {
```
benouat, line 9: Don't forget to remove the context in types also

src/util/transition/ngbTransition.ts
```diff
@@ -28,6 +28,7 @@ export const ngbRunTransition =
 
           // If animations are disabled, we have to emit a value and complete the observable
           if (!options.animation) {
+            (startFn(element, null !) || noopFn)();
```
benouat, line 31: Why is this needed? it's not clear to me. Is it because even without animations we need to remove the `hidden` class? If yes, let's not put it in case animation is false.

src/toast/toast.spec.ts
```diff
@@ -1,21 +1,21 @@
...
-const getToastElement = (fixture: ComponentFixture<TestComponent>): Element =>
-    getElementWithSelector(fixture, 'ngb-toast');
-const getToastHeaderElement = (fixture: ComponentFixture<TestComponent>): Element =>
-    getElementWithSelector(fixture, 'ngb-toast .toast-header');
-const getToastBodyElement = (fixture: ComponentFixture<TestComponent>): Element =>
-    getElementWithSelector(fixture, 'ngb-toast .toast-body');
+const getToastElement = (element: HTMLElement): Element => getElementWithSelector(element, 'ngb-toast');
+const getToastHeaderElement = (element: HTMLElement): Element =>
+    getElementWithSelector(element, 'ngb-toast .toast-header');
+const getToastBodyElement = (element: HTMLElement): Element => getElementWithSelector(element, 'ngb-toast .toast-body');
```
benouat, line 18: Why? Because of that change, you had to put `fixture.nativeElement` everywhere in the file...

src/toast/toast.ts
maxokorokov: I guess we need to add `.show` class when `.fade` is already in place, otherwise the browser won't triggger the fade in transition. And we need this to be synchronous.

src/util/transition/ngbTransition.ts
maxokorokov: I think that's the explanation → https://github.com/ng-bootstrap/ng-bootstrap/pull/3746#issuecomment-635186699

TLDR:
- `[class.fade]: 'animation'` always bindings
- `.show` always added/removed via the `ngbRunTransition`

Othwise you have to "duplicate" adding show class via binding and via `ngbRunTransition`

demo/src/app/components/toast/demos/closeable/toast-closeable.ts
```diff
@@ -6,7 +6,8 @@ export class NgbdToastCloseable {
   show = true;
 
   close() {
+    console.log('close called');
```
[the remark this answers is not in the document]
fbasso: Missed it ! 😕

src/toast/toast.ts
fbasso: Exactly. No animation is shown without it.

src/util/transition/ngbFadingTransition.ts
fbasso: It's a matter of coding preferences. Even if simple, I prefer to abstract this specific action done in the widget. 

`ngbToastFadeOutTransition` is a kind of documentation. When I read `
this._element.nativeElement, ngbToastFadeOutTransition, {animation: this.animation, runningTransition: 'stop'});` in the toast, I can easily understand what the code intend to do (a fade out), and separate the fade out code form the rest of the method code. In the fade out code, I can also easily know what are the Bootstrap specs on this transition.

But it's up to the main project decision and guidelines. No worries, I can change it if needed ;)

src/util/transition/ngbFadingTransition.ts
fbasso: Sure, thanks !

src/toast/toast.spec.ts
fbasso: It's because I have this error when I use it in the animation part:

```
Argument of type 'ComponentFixture<TestAnimationComponent>' is not assignable to parameter of type 'ComponentFixture<TestComponent>'
```

The possible fixes I found were:

- Remove the type in the signature of getToastElement:
`getToastElement(fixture) ...`, but we lost the type feature in this function.
- Extract TestAnimationComponent from its `it` context and move it to the top, then define getToastElement as follow : `const getToastElement = (fixture: ComponentFixture<TestComponent | TestAnimationComponent>)`

I ended up by follow what is done in alert.specs.ts.

What do you prefer ? Any other idea ?

src/util/transition/ngbFadingTransition.ts
maxokorokov: Added this to the list → https://github.com/ng-bootstrap/ng-bootstrap/issues/3747#issue-626345524

maxokorokov:
LGTM, will merge when CI is green!

@fbasso, I just also pushed this (https://github.com/ng-bootstrap/ng-bootstrap/pull/3749/commits/91c5356e16467e4034a589af78487c75e696feb8), because we forgot to fix documentation for the `(shown)` and `(hidden)`.
Also `.hide()` and `.show()` return types were missing in generated docs. You have to put them explicitly, so they would appear:

Screen Shot 2020-06-03 at 17 38 53

Issue #3753, opened 2020-05-28
Datepicker accessibility


PigeardSylvain:
Sorry for my bad english ...

1. Name of days are not well read by screen readers : 
For example screenreader read "mo" instead of "monday". This could be fixed adding an aria-label on column headers

![image](https://user-images.githubusercontent.com/13221271/83122595-2c011200-a0d4-11ea-90a8-cb9c94efed18.png)

2. Screen reader says nothing when using previous and next month buttons. It would be nice to ear the new selected date, for example : "april 2020".

This can be achived by :
- Adding a span at the end of the datepicker, [hidden on all devices except screen readers](https://getbootstrap.com/docs/4.5/utilities/screen-readers/).
- Adding aria-live attribute on this span
- Updating this span when a new day, month or year is selected by the user
Example : 
`<span class="sr-only" aria-live="polite" aria-atomic="true">april 2020</span>`

3. This span could be used has a label for the grid using aria-labelledby
Example : 
`<ngb-datepicker>`
`   ...`
`   <div role="grid" aria-labelledby="full-date">`
`      ...`
`   </div>`
`   <span id="full-date" class="sr-only" aria-live="polite" aria-atomic="true">monday april 6 2020</span>`
`</ngb-datepicker>`

4. Datepicker in a popup
![image](https://user-images.githubusercontent.com/13221271/83165833-5b347500-a10e-11ea-994f-d7334f8d62ca.png)
A label should be added on the button (for screen readers), example : 
`<button><span class="sr-only">choose date</span></button>`

5. Range selection
Some informations should be read to screen readers, using the span in sr-only (cf. 4).
Example : 
When user select the first date, the span could be updated with : "thuesday may 12 2020, select end date"
When user select the second date : "from thuesday 12 may 2020 from friday may 15 2020"

#3745

gpolychronis:
@PigeardSylvain  Hello.
About point 3, you talk about selected date, right?

PigeardSylvain:
I say that a span could be used to store the month and year, the same span could be used then to store the full date when the user select a day.

This span could also be used to label the grid using an aria attribute.

PR #3757, opened 2020-06-04
feat(datepicker): datepicker accessibility (#3753)

gpolychronis:

The following changes have been made to accommodate request 3753:
1) Screen Reader has been add on button of popup datepickers.
2) A screen reader about selected date has been added on datepicker.
3) A screen reader about current month/months has been added on datepicker navigation.


demo/src/app/components/datepicker/demos/adapter/datepicker-adapter.html
```diff
@@ -24,7 +24,9 @@
           <input class="form-control" placeholder="dd/mm/yyyy"
                  name="d2" #c2="ngModel" [(ngModel)]="model2" ngbDatepicker #d2="ngbDatepicker">
           <div class="input-group-append">
-            <button class="btn btn-outline-secondary calendar" (click)="d2.toggle()" type="button"></button>
+            <button class="btn btn-outline-secondary calendar" (click)="d2.toggle()" type="button">
+              <span class="sr-only" i18n="@@ngb.demo.datepicker.popup.choose">Choose date</span>
```
IAfanasovMob, line 28: I'm not an expert here and very curious (:
Why not `aria-label`?

gpolychronis:
@benouat @PigeardSylvain @BehnooshShiva 
Please review.

demo/src/app/components/datepicker/demos/adapter/datepicker-adapter.html
benouat, line 28: Why did you add i18n to the demo app ?? It seems useless as we don't build it with i18n.

src/datepicker/datepicker-i18n.ts
```diff
@@ -28,6 +28,13 @@ export abstract class NgbDatepickerI18n {
    */
   abstract getWeekdayShortName(weekday: number): string;
 
+  /**
+   * Returns the weekday name to listen from screen reader in the heading of the month view.
+   *
+   * With default calendar we use ISO 8601: 'weekday' is 1=Monday ... 7=Sunday.
+   */
+  getWeekdayName(weekday: number): string { return this.getWeekdayShortName(weekday); }
+
```
benouat, line 37: `NgbDatepickerI18n` is supposed to be an abstract class, with no simplementation.
Default implementation should go in `NgbDatepickerI18nDefault`

src/datepicker/datepicker-navigation.ts
```diff
@@ -38,6 +38,15 @@ import {NgbDatepickerI18n} from './datepicker-i18n';
         <span class="ngb-dp-navigation-chevron"></span>
       </button>
     </div>
+    <span class="sr-only" [attr.aria-live]="'polite'" [attr.aria-atomic]="true">
+      <span *ngIf="months.length > 1">
+        <span i18n="@@ngb.datepicker.screenreader.from">from</span>
+        {{ this.getScreenReaderFrom() }}
+        <span i18n="@@ngb.datepicker.screenreader.to">to</span>
+        {{ this.getScreenReaderTo() }}
```
benouat, line 46: I think I would name the `i18n` keys with `screenreader` inside. Just call it the way it should, then if we use it in SR context, it seems to be detail implementation.

src/datepicker/datepicker-navigation.ts
```diff
@@ -38,6 +38,15 @@ import {NgbDatepickerI18n} from './datepicker-i18n';
         <span class="ngb-dp-navigation-chevron"></span>
       </button>
     </div>
+    <span class="sr-only" [attr.aria-live]="'polite'" [attr.aria-atomic]="true">
+      <span *ngIf="months.length > 1">
+        <span i18n="@@ngb.datepicker.screenreader.from">from</span>
+        {{ this.getScreenReaderFrom() }}
```
benouat, line 44: Same remark here. Name the method according to what it does, ie `getLongDate`, `getShortMotnh`, anything, but not with ScreenReader inside.

src/datepicker/datepicker-navigation.ts
```diff
@@ -65,4 +74,19 @@ export class NgbDatepickerNavigation {
...
+
+  getScreenReaderFrom() {
+    const f = this.months[0].firstDate;
+    return this.i18n.getMonthFullName(f.month, f.year) + ' ' + this.i18n.getYearNumerals(f.year);
+  }
+
+  getScreenReaderTo() {
+    const t = this.months[this.months.length - 1].lastDate;
+    return this.i18n.getMonthFullName(t.month, t.year) + ' ' + this.i18n.getYearNumerals(t.year);
+  }
```
benouat, line 86: Aren't these things supposed to be localised (l10n) ? Meaning that depending on the lang, it might be `month + " " + year` or `year + ", " + month` ...

demo/src/app/components/datepicker/demos/adapter/datepicker-adapter.html
gpolychronis: Removed i18n from demo.

src/datepicker/datepicker-i18n.ts
gpolychronis: Removed implementation from NgbDatepickerI18n. Moved method implementation to class implementations.

src/datepicker/datepicker-navigation.ts
gpolychronis: Removed "screenreader" from i18n tag names.

src/datepicker/datepicker-navigation.ts
gpolychronis: Same

src/datepicker/datepicker-navigation.ts
gpolychronis: Moved implementation to NgbDatepickerI18n. Did it the same way as day aria label.

gpolychronis:
> 
> 
> Please have a look at my comments

Remarks have been answered.

gpolychronis:
Check if aria-live utility (live.ts) should be used.

gpolychronis:
Updated to use the Live in live.ts

gpolychronis:
> 
> 
> Updated to use the Live in live.ts

@maxokorokov , @benouat please review.

src/datepicker/datepicker.ts
```diff
@@ -158,7 +166,8 @@ export class NgbDatepickerContent {
       </ngb-datepicker-navigation>
     </div>
 
-    <div class="ngb-dp-content" [class.ngb-dp-months]="!contentTemplate" #content>
+    <div class="ngb-dp-content" [class.ngb-dp-months]="!contentTemplate" #content
+         role="grid" aria-labelledby="selected-date">
```
mattdistefano, line 170: The `aria-labelledby` here may have some issues.

1) I may just be missing it, but where is the corresponding element with id of `selected-date`?
2) Having a hard-coded ID isn't a great idea in any case - e.g. if there's multiple date pickers on the screen.
3) I'm not sure that the selected date is really an accurate label for the entire date picker.

demo/src/app/components/datepicker/demos/range-popup/datepicker-range-popup.html
```diff
@@ -43,7 +45,9 @@
              [value]="formatter.format(toDate)"
              (input)="toDate = validateInput(toDate, dpToDate.value)">
       <div class="input-group-append">
-        <button class="btn btn-outline-secondary calendar" (click)="datepicker.toggle()" type="button"></button>
+        <button class="btn btn-outline-secondary calendar" (click)="datepicker.toggle()" type="button">
+          <span class="sr-only">Choose date</span>
```
mattdistefano, line 49: Might be worth thinking about doing these as `aria-label`s instead so there's fewer instances of `sr-only` to rename when going to BS5.

achromial:
Hello maintainers & PR creator @mattdistefano @maxokorokov @gpolychronis 

Is there hope of merging these changes into the latest versions of NgBootstrap? What can I do to contribute to this effort? Over in issue  #4609 I've got an idea for a quicker fix but wanted to play nice with this more major feature IFF it's getting exhumed and put into master "officially".

To address 4609, I'm seeking to add (specifically for JAWS users) the `aria-selected` to the `role="gridcell"` div shown in this image.
image

MDN seems to indicate differently than a11ySupport on this specific combo.

MDN: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected#grid
a11ySupport: https://a11ysupport.io/tests/aria_gridcell%28aria-selected%29

PR #3766, opened 2020-06-11
feat(accordion): add animations

fbasso:
Cherry picked from #2817, with some changes.

The key points are:

- Animations are managed one panel at a time, in the original code, a loop was done on all panels to guess if an animation is required, for each panel. The code is simpler this way, and aligned with ngbTransition API,
- ngbCollapseChange event  have been added. It follows the collapse widget, to discuss if the name/API must be changed,
- A little change in ngbTransition has been required, to provide the context when animations are deactivated,
- Tests have been done the same way the collpase widget,
- Some `detectChanges` have been added in the previous tests, because of the `changeDetector.detectChanges()` required in `_runTransition`,
- Only one commit in this PR, as there are only a few impacted files.

### DIFF
--- a/src/accordion/accordion.spec.ts
+++ b/src/accordion/accordion.spec.ts
@@ -1,10 +1,12 @@
 import {TestBed, ComponentFixture, inject} from '@angular/core/testing';
 import {By} from '@angular/platform-browser';
-import {createGenericTestComponent} from '../test/common';
+import {createGenericTestComponent, isBrowserVisible} from '../test/common';
 
 import {Component} from '@angular/core';
 
 import {NgbAccordionModule, NgbPanelChangeEvent, NgbAccordionConfig, NgbAccordion} from './accordion.module';
+import {NgbConfig} from '../ngb-config';
+import {NgbConfigAnimation} from '../test/ngb-config-animation';
 
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
@@ -14,7 +16,7 @@ function getPanels(element: HTMLElement): HTMLDivElement[] {
 }
 
 function getPanelsContent(element: HTMLElement): HTMLDivElement[] {
-  return <HTMLDivElement[]>Array.from(element.querySelectorAll('.card > .collapse'));
+  return <HTMLDivElement[]>Array.from(element.querySelectorAll('.card > .collapse, .card > .collapsing'));
 }
 
 function getPanelsButton(element: HTMLElement): HTMLButtonElement[] {
@@ -51,8 +53,9 @@ function expectOpenPanels(nativeEl: HTMLElement, openPanelsDef: boolean[]) {
 describe('ngb-accordion', () => {
   let html = `
     <ngb-accordion #acc="ngbAccordion" [closeOthers]="closeOthers" [activeIds]="activeIds"
-      (panelChange)="changeCallback($event)" [type]="classType">
-      <ngb-panel *ngFor="let panel of panels" [id]="panel.id" [disabled]="panel.disabled" [type]="panel.type">
+      (panelChange)="changeCallback($event)" (shown)="shownCallback($event)" (hidden)="hiddenCallback($event)" [type]="classType">
+      <ngb-panel *ngFor="let panel of panels" [id]="panel.id" [disabled]="panel.disabled" [type]="panel.type"
+        (shown)="panelShownCallback($event)" (hidden)="panelHiddenCallback($event)">
         <ng-template ngbPanelTitle>{{panel.title}}</ng-template>
         <ng-template ngbPanelContent>{{panel.content}}</ng-template>
       </ngb-panel>
@@ -66,8 +69,8 @@ describe('ngb-accordion', () => {
   });
 
   it('should initialize inputs with default values', () => {
-    const defaultConfig = new NgbAccordionConfig();
-    const accordionCmp = new NgbAccordion(defaultConfig);
+    const defaultConfig = new NgbAccordionConfig(new NgbConfig());
+    const accordionCmp = TestBed.createComponent(NgbAccordion).componentInstance;
     expect(accordionCmp.type).toBe(defaultConfig.type);
     expect(accordionCmp.closeOtherPanels).toBe(defaultConfig.closeOthers);
   });
@@ -552,25 +555,31 @@ describe('ngb-accordion', () => {
     expect(panelContents.length).toBe(3);
   });
 
-  it('should emit panel change event when toggling panels', () => {
+  it('should emit panel events when toggling panels', () => {
     const fixture = TestBed.createComponent(TestComponent);
     fixture.detectChanges();
 
-    fixture.componentInstance.changeCallback = () => {};
-
     spyOn(fixture.componentInstance, 'changeCallback');
+    spyOn(fixture.componentInstance, 'shownCallback');
+    spyOn(fixture.componentInstance, 'hiddenCallback');
+    spyOn(fixture.componentInstance, 'panelShownCallback');
+    spyOn(fixture.componentInstance, 'panelHiddenCallback');
 
     // Select the first tab -> change event
     getButton(fixture.nativeElement, 0).click();
     fixture.detectChanges();
     expect(fixture.componentInstance.changeCallback)
         .toHaveBeenCalledWith(jasmine.objectContaining({panelId: 'one', nextState: true}));
+    expect(fixture.componentInstance.shownCallback).toHaveBeenCalledWith('one');
+    expect(fixture.componentInstance.panelShownCallback).toHaveBeenCalledWith(undefined);
 
     // Select the first tab again -> change event
     getButton(fixture.nativeElement, 0).click();
     fixture.detectChanges();
     expect(fixture.componentInstance.changeCallback)
         .toHaveBeenCalledWith(jasmine.objectContaining({panelId: 'one', nextState: false}));
+    expect(fixture.componentInstance.hiddenCallback).toHaveBeenCalledWith('one');
+    expect(fixture.componentInstance.panelHiddenCallback).toHaveBeenCalledWith(undefined);
   });
 
   it('should cancel panel toggle when preventDefault() is called', () => {
@@ -662,7 +671,7 @@ describe('ngb-accordion', () => {
   });
 
   describe('Custom config as provider', () => {
-    let config = new NgbAccordionConfig();
+    let config = new NgbAccordionConfig(new NgbConfig());
     config.closeOthers = true;
     config.type = 'success';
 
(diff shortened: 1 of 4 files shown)


src/accordion/accordion.spec.ts
```diff
@@ -852,11 +862,122 @@ describe('ngb-accordion', () => {
...
     });
   });
 });
 
+if (isBrowserVisible('ngb-accordion animations')) {
+  describe('ngb-accordion animations', () => {
+
+    @Component({
+      template: `
+      <ngb-accordion activeIds="first" (ngbCollapseChange)="onCollapseChange()">
```
maxokorokov, line 876: 1. Please add `$event` and check the values that are actually sent by the emitter
2. Please check that `(ngbCollapseChange)` is emitted with correct values in "normal" tests without animations

src/accordion/accordion.ts
```diff
@@ -181,6 +207,11 @@ export interface NgbPanelChangeEvent {
 export class NgbAccordion implements AfterContentChecked {
   @ContentChildren(NgbPanel) panels: QueryList<NgbPanel>;
 
+  /**
+   * If `true`, accordion will be animated.
+   */
+  @Input() animation = false;
```
maxokorokov, line 213: Please remove `=  false`

src/accordion/accordion.ts
```diff
@@ -216,7 +247,17 @@ export class NgbAccordion implements AfterContentChecked {
    */
   @Output() panelChange = new EventEmitter<NgbPanelChangeEvent>();
 
-  constructor(config: NgbAccordionConfig) {
+  /**
+   * Event emitted right after the panel toggle happens, after the animation has been completed.
+   *
+   * See [NgbPanelCollapseEvent](#/components/accordion/api#NgbPanelCollapseEvent) for payload details.
+   */
+  @Output() ngbCollapseChange = new EventEmitter<NgbPanelCollapseEvent>();
```
maxokorokov, line 255: I think The API should be a bit different. We have 2 events everywhere else. Ex. `(shown)/(hidden)`. Could we maybe do this ?:

```html
<ngb-accordion (collapsed)="..." (expanded)="...">.  ← $event is the id of the panel here
  <ngb-panel id="first" (collapsed)="..." (expanded)="..."></ngb-panel> ← $event is undefined here
</ngb-accordion>
```

We've discussed this for navs and to me it's kinda similar reasoning.

`ngbCollapse` itself is different, because the event is in 2-way binding: `<div [(ngbCollapse)]="someBoolean"></div>`

@benouat would be nice to have your input on this too.

src/accordion/accordion.ts
```diff
@@ -321,6 +364,26 @@ export class NgbAccordion implements AfterContentChecked {
   private _updateActiveIds() {
     this.activeIds = this.panels.filter(panel => panel.isOpen && !panel.disabled).map(panel => panel.id);
   }
+
+  private _runTransition(panel: NgbPanel) {
+    panel.transitionRunning = true;
+    this._changeDetector.detectChanges();
+    this._ngZone.onStable.pipe(take(1)).subscribe(() => {
```
maxokorokov, line 371: 1. I don't think we need this `onStable` here at all. Everything is fine in my tests if we don't do it + you can remove all double `fixture.detectChanges()` from the tests.

Adding those `fixture.detectChanges()` feels like something is wrong at the first place :)

2. Also we need to run `this._changeDetector.detectChanges()` only once, not for each panel in a loop !!! (in case of `_closeOthers`)

src/util/transition/ngbTransition.ts
```diff
@@ -26,15 +26,15 @@ export const ngbRunTransition =
     <T>(element: HTMLElement, startFn: NgbTransitionStartFn<T>, options: NgbTransitionOptions<T>):
         Observable<undefined> => {
 
+          // Getting initial context from options
```
maxokorokov, line 29: This is also fixed in my navs PR, but here you're missing tests... let's rollback this one.
My implementation is here → https://github.com/ng-bootstrap/ng-bootstrap/pull/3776/commits/c6b296b7d5e8e4118ff15a17212048775dba5ae6

src/accordion/accordion.ts
```diff
@@ -145,6 +154,23 @@ export interface NgbPanelChangeEvent {
   preventDefault: () => void;
 }
 
+/**
+ * An event emitted right before toggling an accordion panel visibility changed, once the animation has been completed.
+ */
```
benouat, line 159: Please, re-word this documentation. I can't manage to understand it as it is.

src/util/transition/ngbTransition.ts
fbasso: Yes sure. I'll intend to rebase the branch once this part will be delivered. It's just here because I need it to have this PR code working .

It's also required for the carousel. Maybe a separate PR would have been required for this part, to be available for all :)

src/accordion/accordion.ts
```diff
@@ -145,6 +154,23 @@ export interface NgbPanelChangeEvent {
...
+export interface NgbPanelCollapseEvent {
+  /**
+   * The id of the accordion panel that is being toggled.
```
benouat, line 162: The id of the accordion panel ~~that is~~ being toggled.

src/accordion/accordion.ts
```diff
@@ -145,6 +154,23 @@ export interface NgbPanelChangeEvent {
...
+   */
+  panelId: string;
+
+  /**
+   * The next state of the panel.
+   *
+   * `true` if it is collapsed, `false` if open.
```
benouat, line 169: > `true` if it is collapsed, `false` if open.

-> `true` when collapsed, `false` otherwise

src/accordion/accordion.ts
benouat: I am 💯% with you here. We should have `(collapsed)/(expanded)`

src/accordion/accordion.ts
benouat: > + you can remove all double fixture.detectChanges() from the tests

👍

fbasso: 1. It seems so. I remember I've added the onStable because of issues, but I think it was more due to `this._changeDetector.detectChanges()` I forgot to add.
2. Not sure how to handle it. The purpose of detectChanges is to ensure that `panel.transitionRunning = true;` is well reflected in the dom, just before the runTransition. So if I do it after the loop, it's too late. Need more time to think about it :)

src/accordion/accordion.ts
maxokorokov: Yeah +1 from me, I haven't left any comments yet on this because I think it will go away if we change the event API

src/util/transition/ngbTransition.ts
maxokorokov: I'll extract my commit into a a separate PR for this, no pb.

src/accordion/accordion.ts
maxokorokov: Run it before the loop IF there are any open tabs, no?

fbasso: :smile: I achieved almost the same code before reading your comment, the variable name is the only difference ! In this case, there are still 2 changeDetections if it's run from _changeOpenState (one for closeOthers, one for the active panel).

We can also improve that by:

- Changing the isOpen state of the panel to the desired state,
- Setting runningTranstion to true on the panel,
- run _runTransitions to perform the changeDetection (once) then run the pending transitions in a row.

This will be optimum, but I wonder if the additional complexity worth it.

What do you think ?

fbasso: ok, after a discussion with @maxokorokov, I'll try the optimization

src/accordion/accordion.ts
fbasso: Thanks, I just realized that I started to changed a pasted sentence, but without finishing it !
 Anyway, I will remove this event in favor of 'shown' and 'hidden' ones (as Bootstrap for the collapse)

src/accordion/accordion.spec.ts
```diff
@@ -857,6 +872,132 @@ describe('ngb-accordion', () => {
...
+        <ngb-panel id="second" (shown)="onSecondPanelShown()" (hidden)="onSecondPanelHidden()"></ngb-panel>
+      </ngb-accordion>
+    `,
+      host: {'[class.ngb-reduce-motion]': 'reduceMotion'}
+    })
+    class TestAnimationComponent {
+      reduceMotion = true;
+      onShown = (panelId) => panelId;
+      onHidden = (panelId) => panelId;
+      onFirstPanelShown = () => {};
```
maxokorokov, line 891: 1. you're not using this in ANY tests, please add checks that these callbacks are actually called
2. you could have 1 function and just check what id it is called with:
```ts
onPanelShown = () = {};
```

```html
(shown)="onPanelShown('one')"
```

src/accordion/accordion.spec.ts
```diff
@@ -552,25 +555,37 @@ describe('ngb-accordion', () => {
...
+    spyOn(fixture.componentInstance, 'panelShownCallback');
+    spyOn(fixture.componentInstance, 'panelHiddenCallback');
 
     // Select the first tab -> change event
     getButton(fixture.nativeElement, 0).click();
     fixture.detectChanges();
     expect(fixture.componentInstance.changeCallback)
         .toHaveBeenCalledWith(jasmine.objectContaining({panelId: 'one', nextState: true}));
+    expect(fixture.componentInstance.shownCallback).toHaveBeenCalledWith('one');
+    expect(fixture.componentInstance.panelShownCallback).toHaveBeenCalled();
```
maxokorokov, line 580: could we check with `.toHaveBeenCalledWith(undefined)` instead?

src/accordion/accordion.spec.ts
```diff
@@ -552,25 +555,37 @@ describe('ngb-accordion', () => {
...
 
     // Select the first tab again -> change event
     getButton(fixture.nativeElement, 0).click();
     fixture.detectChanges();
     expect(fixture.componentInstance.changeCallback)
         .toHaveBeenCalledWith(jasmine.objectContaining({panelId: 'one', nextState: false}));
+    expect(fixture.componentInstance.hiddenCallback).toHaveBeenCalledWith('one');
+    expect(fixture.componentInstance.panelHiddenCallback).toHaveBeenCalled();
```
maxokorokov, line 588: could we check with `.toHaveBeenCalledWith(undefined)` instead?

src/accordion/accordion.spec.ts
```diff
@@ -552,25 +555,37 @@ describe('ngb-accordion', () => {
     expect(panelContents.length).toBe(3);
   });
 
-  it('should emit panel change event when toggling panels', () => {
+  it('should emit panel events when toggling panels', () => {
     const fixture = TestBed.createComponent(TestComponent);
     fixture.detectChanges();
 
     fixture.componentInstance.changeCallback = () => {};
```
maxokorokov, line 562: please remove all these:

```ts
fuxture.componentInstance.xxx = () => {}
```

They're already declared in the `TestComponent`, no need to re-declare them

src/accordion/accordion.spec.ts
```diff
@@ -857,6 +872,132 @@ describe('ngb-accordion', () => {
...
+        fixture.componentInstance.reduceMotion = reduceMotion;
+        fixture.detectChanges();
+
+        const buttonEl = getPanelsButton(fixture.nativeElement)[0];
+        const panelEl = getPanelsContent(fixture.nativeElement)[0];
+        const onShownSpy = spyOn(fixture.componentInstance, 'onShown');
+        const onHiddenSpy = spyOn(fixture.componentInstance, 'onHidden');
+
+        // First we're going to collapse, then expand
+        let opened = true;
```
maxokorokov, line 918: why have this flag? it's not used anywhere

src/accordion/accordion.ts
maxokorokov: This is still not fixed

src/accordion/accordion.ts
```diff
@@ -216,7 +239,18 @@ export class NgbAccordion implements AfterContentChecked {
    */
   @Output() panelChange = new EventEmitter<NgbPanelChangeEvent>();
 
-  constructor(config: NgbAccordionConfig) {
+  /**
+   * An event emitted when a panel is shown, after the transition. The payload is the panel id.
```
maxokorokov, line 243: Could we be more specific, please ? meaning that "when the panel is shown" is not very clear...

Maybe `An event emitted when the expanding animation is finished on the panel` ?

src/accordion/accordion.ts
```diff
@@ -216,7 +239,18 @@ export class NgbAccordion implements AfterContentChecked {
...
+   */
+  @Output() shown = new EventEmitter<string>();
+
+  /**
+   * An event emitted when a panel is hidden, after the transition. The payload is the panel id.
```
maxokorokov, line 248: Same here, but `collapsing animation`. Would also be nice to precise, if DOM content is expected to be removed at this point or not.

src/accordion/accordion.ts
```diff
@@ -321,6 +360,34 @@ export class NgbAccordion implements AfterContentChecked {
   private _updateActiveIds() {
     this.activeIds = this.panels.filter(panel => panel.isOpen && !panel.disabled).map(panel => panel.id);
   }
+
+  private _runTransitions() {
+    this._changeDetector.detectChanges();
```
maxokorokov, line 365: Could we add a comment, explaining why we need CD here?

src/accordion/accordion.ts
```diff
@@ -321,6 +360,34 @@ export class NgbAccordion implements AfterContentChecked {
...
+    this.panels.forEach(panel => {
+      if (panel.transitionRunning) {
+        const panelElement = this._getPanelElement(panel.id);
+        ngbRunTransition(panelElement !, ngbCollapsingTransition, {
```
maxokorokov, line 369: Could we add a comment explaining why do we restart transitions, even though the transition might already be running?

src/accordion/accordion.spec.ts
fbasso: Finally, I need only the check on panel one, so I'll use it without parameter.

src/accordion/accordion.ts
fbasso: I push again with all the requested changes (I hope I didn't forgot anything !).

Unfortunately, I had to change the accordion code again. This is due to some failing tests on what classes are expected (after the ngbTransition refactor  with the startFn): classes managed by the transitions shouldn't be managed by bindings.

I'm not very happy with the code produced, but this is due to the initial code which would require a BIG refactor, probably in order to avoid ngAfterContentChecked usage (in favor of ngChanges for the accordion, and (change) usage in panels. There are also some inconsistency (in the base code), like:

- `activeIds` documentation : An array or comma separated strings of panel ids that should be opened **initially**,
- `should toggle panels based on "activeIds" values` testing just the opposite.

I did not want to include this refactor, because the review would be much harder.

fbasso:
I push again with all the requested changes (I hope I didn't forgot anything !).

Unfortunately, I had to change the accordion code again. This is due to some failing tests on what classes are expected (after the ngbTransition refactor  with the startFn): classes managed by the transitions shouldn't be managed by bindings.

I'm not very happy with the code produced, but this is due to the initial code which would require a BIG refactor, probably in order to avoid `ngAfterContentChecked` usage (in favor of ngChanges for the accordion, and (change) usage in panels. There are also some inconsistency (in the base code), like:

- `activeIds` documentation : An array or comma separated strings of panel ids that should be opened **initially**,
- `should toggle panels based on "activeIds" values` testing just the opposite.

I did not want to include this refactor, because the review would be much harder.

fbasso:
Last changes pushed ! 

I finally managed to eliminate the change detection change on `AfterContentChecked`, but the code required is (again) not satisfying in terms of simplicity and clarity ! The constraints given by the AfterContentChecked and the fact to not setup animation classes in the template (it produces other drawbacks and failures) is so hard that I failed to find something else 😞. 

@maxokorokov , @benouat, don't hesitate to share if you have other ideas !

Issue #3758, opened 2020-06-04
NgbDropdown attached to body tabindex issue


anton-gorbikov:
### Bug description:

NgbDropdown used in conjunction with `container="body"` breaks TAB keyboard navigation. Steps:

1. Get first toggle button focused using TAB button
1. Press ENTER to open the dropdown
1. Press TAB

Expected result: first item from the list should be focused
Actual result: the second dropdown toggle is focused

### Link to minimally-working StackBlitz that reproduces the issue:

[StackBlitz link](https://stackblitz.com/edit/angular-9prhp8) - this is fork of [basic dropdown demo](https://ng-bootstrap.github.io/stackblitzes/dropdown/basic/stackblitz.html) with only one attribute added

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 9.0.7

ng-bootstrap:

Bootstrap: 6.1.0

benouat:
Hey @anton-gorbikov 

Thank you to report this issue.
You're right, this is something we don't handle at all. I am afraid it will impact all widgets using the `container` @Input where we can set `body`.

We should think about a fix that would work whatever the widget using it.

benouat:
Hey @anton-gorbikov, sorry for the long time no response.

Would you mind testing again with a recent version ? We fixed some issues related to that in `v7`

anton-gorbikov:
Hi @benouat , it looks like it's working fine in the most recent version! Thanks a lot!

PR #3767, opened 2020-06-11
demo(dropdown): new container body example

benouat:
Working on a fix for #3758, I realised we don't have any demo for that.
As we had several time by the past issues where people did not know there is this option, I quickly crafted one.

Here is the visual aspect of it:

image


benouat:
Someone to quickly review this small demo ? cc @fbasso @maxokorokov

fbasso:
Reviewed, and fine for me!

PR #3771, opened 2020-06-14
feat(modal): add scale animation when static backdrop is clicked

### LINKED ISSUE #3768: Static backdrop modal should have a "shaky" animation
To see it live, open this demo, and click on the backdrop
https://getbootstrap.com/docs/4.4/components/modal/#static-backdrop

jnizet:
fix #3768

Note: this is my first PR on animations, and I don't really master everything. 
There are 3 points that I'm not sure about in the submitted code:

- should I pass 'continue' or 'stop'. I chose 'continue', and it seems to behave correctly, and the same way as the native bootstrap modal, but maybe I'm missing something
- in the unit tests, I don't know how to test that the `modal-static` class disappears at the end of the transition: nothing is emitted when that happens, and I didn't want to introduce an output only useful for tests. (see TODO in the test)
- I'm not sure exactly what the `reduceMotion` flag is supposed to do. I have noticed that the `modal-static` is not added when `reduceMotion` is true, and I guess it's a feature, but I'm not sure. (see TODO in the test)

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -32,6 +32,7 @@ import {ngbRunTransition, NgbTransitionOptions} from '../util/transition/ngbTran
     '[attr.aria-modal]': 'true',
     '[attr.aria-labelledby]': 'ariaLabelledBy',
     '[attr.aria-describedby]': 'ariaDescribedBy',
+    '(click)': 'bump($event)'
   },
   template: `
     <div #dialog [class]="'modal-dialog' + (size ? ' modal-' + size : '') + (centered ? ' modal-dialog-centered' : '') +
@@ -92,6 +93,19 @@ export class NgbModalWindow implements OnInit,
     this._restoreFocus();
 
     return transitions$;
+  }
+
+  bump(event: Event) {
+    const nativeElement = this._elRef.nativeElement;
+    // the animation must only happen if the backdrop is static
+    // and if the click is not inside the modal
+    if (event.target === nativeElement && this.backdrop === 'static') {
+      ngbRunTransition(nativeElement, ({classList}) => {
+        classList.add('modal-static');
+        return () => classList.remove('modal-static');
+      }, {animation: this.animation, runningTransition: 'continue'});
+      event.stopPropagation();
+    }
   }
 
   private _show() {
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -125,8 +125,9 @@ export class NgbModalWindow implements OnInit,
                   this._zone.run(() => this.dismiss(ModalDismissReasons.ESC));
                 }
               });
+            } else if (this.backdrop === 'static') {
+              this._bumpBackdrop();
             }
-            this._bumpIfStaticBackdrop();
           });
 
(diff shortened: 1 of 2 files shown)


src/modal/modal-window.ts
```diff
@@ -32,6 +32,7 @@ import {ngbRunTransition, NgbTransitionOptions} from '../util/transition/ngbTran
     '[attr.aria-modal]': 'true',
     '[attr.aria-labelledby]': 'ariaLabelledBy',
     '[attr.aria-describedby]': 'ariaDescribedBy',
+    '(click)': 'bump($event)'
```
maxokorokov, line 35: I also just realised while adding the comment about the <kbd>Esc</kbd> key, that it might be better to add the bump call in the `_enableEventHandling() → fromEvent<MouseEvent>(nativeElement, 'click')` ?

jnizet:
Thanks for the review @maxokorokov.

Since ngbRunTransition runs everything synchronously (i.e. the start and the end functions) when reduceMotion is true, my test already checks that the modal-static class disappears at the end of the transition, so I guess I'm fine there.

I'll do the remaining stuff, and add tests for the Escape key handling. 
I'll submit a new commit to help with the review, but if it suits you, feel free to squash the two commits before merging of course.

jnizet:
I'm fine doing that, but it will change the behevior. 
Currently, the if `keyboard` is true, pressing escape dismisses the modal whatever the backdrop value is.
With your suggested change, it will not dismiss the modal anymore if the backdrop is static.

I'm not sure what is the correct behavior, but my goal here was to introduce the animation, not to introduce a change of behavior (which could be seen as a breaking change).

Please tell me if this `else` should be added or not. To keep the current behavior, the code should be

```
// ESC handling
if (this.backdrop === 'static') {
  this._bumpBackdrop();
} // no else here!!!
if (this.keyboard) {
  requestAnimationFrame(() => {
    // ...
  });
}

// click handing
if (nativeElement === target) {
  if (this.backdrop === 'static') {
    this._bumpBackdrop();
  } else if (this.backdrop === true && !preventClose) {
    // ...
  }
}
```

@maxokorokov @benouat

benouat:
> Currently, the if `keyboard` is true, pressing escape dismisses the modal whatever the backdrop value is.
> With your suggested change, it will not dismiss the modal anymore if the backdrop is static.
> 
> I'm not sure what is the correct behavior, but my goal here was to introduce the animation, not to introduce a change of behavior (which could be seen as a breaking change).

Correct, we apparently did not implement it correctly in the first place!
I just checked on BS side, backdrop `static` is actually a real _modal dialog_ in a sense that only actions coming from the modal can close it. <kbd>Esc</kbd> does not work, neither does `click`.

So, return of the _never ending story_ ... 🤷‍♂️ what should we do ? Non backward compatible change ? or simply bug fixing a broken behavior that should never have been here in the first place ?

Back to reality here! I would say we do not change the behaviour in this animation PR introduction. We will then fix it/change it on master once animations are merged.

jnizet:
OK, thanks. I will then use the code I suggested above to keep the current buggy behavior.

PR #3773, opened 2020-06-15
feat(carousel): accessibility

### LINKED ISSUE #3755: Carousel accessibility
When the carousel advances automatically, it must pause when the focus is positioned inside.

More details : 
https://www.w3.org/WAI/tutorials/carousels/

#3745

divdavem:
This PR adds accessibility to the carousel:
 - the carousel is now automatically paused when the focus is inside it (this can be controlled with the `pauseOnFocus` property)
 - some roles and aria attributes are added to make it easier to use the carousel with a screen reader. This mostly relies on the carousel indicators that act like tabs.

In the future, it would be nice to integrate [assistive-webdriver](https://github.com/AmadeusITGroup/Assistive-Webdriver) to execute end-to-end tests with a screen reader.

Fixes #3755

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -6,8 +6,8 @@ import {
   Component,
   ContentChildren,
   Directive,
+  ElementRef,
   EventEmitter,
-  HostListener,
   Inject,
   Input,
   NgZone,
@@ -55,24 +55,34 @@ export class NgbSlide {
     'class': 'carousel slide',
     '[style.display]': '"block"',
     'tabIndex': '0',
-    '(keydown.arrowLeft)': 'keyboard && prev(NgbSlideEventSource.ARROW_LEFT)',
-    '(keydown.arrowRight)': 'keyboard && next(NgbSlideEventSource.ARROW_RIGHT)'
+    '(keydown.arrowLeft)': 'keyboard && arrowLeft()',
+    '(keydown.arrowRight)': 'keyboard && arrowRight()',
+    '(mouseenter)': 'mouseHover = true',
+    '(mouseleave)': 'mouseHover = false',
+    '(focusin)': 'focused = true',
+    '(focusout)': 'focused = false',
+    '[attr.aria-activedescendant]': 'activeId'
   },
   template: `
-    <ol class="carousel-indicators" *ngIf="showNavigationIndicators">
+    <ol class="carousel-indicators" [class.sr-only]="!showNavigationIndicators" role="tablist">
       <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId"
-          (click)="select(slide.id, NgbSlideEventSource.INDICATOR)"></li>
+          role="tab" [attr.aria-labelledby]="slide.id + '-slide'" [attr.aria-controls]="slide.id + '-slide'"
+          [attr.aria-selected]="slide.id === activeId"
+          (click)="focus();select(slide.id, NgbSlideEventSource.INDICATOR);"></li>
     </ol>
     <div class="carousel-inner">
-      <div *ngFor="let slide of slides" class="carousel-item" [class.active]="slide.id === activeId">
+      <div *ngFor="let slide of slides; index as i; count as c" class="carousel-item" [class.active]="slide.id === activeId" [id]="slide.id + '-slide'" role="tabpanel">
+        <span class="sr-only" i18n="Currently selected slide number read by screen reader@@ngb.carousel.slide-number">
+          Slide {{i + 1}} of {{c}}
+        </span>
         <ng-template [ngTemplateOutlet]="slide.tplRef"></ng-template>
       </div>
     </div>
-    <a class="carousel-control-prev" role="button" (click)="prev(NgbSlideEventSource.ARROW_LEFT)" *ngIf="showNavigationArrows">
+    <a class="carousel-control-prev" role="button" (click)="arrowLeft()" *ngIf="showNavigationArrows">
       <span class="carousel-control-prev-icon" aria-hidden="true"></span>
       <span class="sr-only" i18n="@@ngb.carousel.previous">Previous</span>
     </a>
-    <a class="carousel-control-next" role="button" (click)="next(NgbSlideEventSource.ARROW_RIGHT)" *ngIf="showNavigationArrows">
+    <a class="carousel-control-next" role="button" (click)="arrowRight()" *ngIf="showNavigationArrows">
       <span class="carousel-control-next-icon" aria-hidden="true"></span>
       <span class="sr-only" i18n="@@ngb.carousel.next">Next</span>
     </a>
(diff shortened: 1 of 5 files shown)


src/carousel/carousel.ts
```diff
@@ -163,10 +196,27 @@ export class NgbCarousel implements AfterContentChecked,
     this.wrap = config.wrap;
     this.keyboard = config.keyboard;
     this.pauseOnHover = config.pauseOnHover;
+    this.pauseOnFocus = config.pauseOnFocus;
     this.showNavigationArrows = config.showNavigationArrows;
     this.showNavigationIndicators = config.showNavigationIndicators;
   }
 
+  @HostListener('keydown.arrowLeft')
```
benouat, line 204: We don't use `@HostListener()` on the project. Would you mind moving all of them in the top `@Component()` declaration ?

benouat: I just realised we missed some other in the file on the previous PR. Please, take the opportunity to move _all_ of them

src/carousel/carousel.ts
```diff
@@ -53,26 +55,41 @@ export class NgbSlide {
   encapsulation: ViewEncapsulation.None,
   host: {
     'class': 'carousel slide',
-    '[style.display]': '"block"',
-    'tabIndex': '0',
-    '(keydown.arrowLeft)': 'keyboard && prev(NgbSlideEventSource.ARROW_LEFT)',
-    '(keydown.arrowRight)': 'keyboard && next(NgbSlideEventSource.ARROW_RIGHT)'
+    '[attr.tabindex]': 'keyboard && !showNavigationIndicators ? 0 : null',
```
benouat, line 58: Why `null` and not `-1` here?

src/carousel/carousel.ts
```diff
@@ -87,7 +104,9 @@ export class NgbCarousel implements AfterContentChecked,
   private _destroy$ = new Subject<void>();
   private _interval$ = new BehaviorSubject(0);
   private _mouseHover$ = new BehaviorSubject(false);
+  private _focused$ = new BehaviorSubject(false);
   private _pauseOnHover$ = new BehaviorSubject(false);
+  private _pauseOnFocus$ = new BehaviorSubject(false);
```
benouat, line 109: I think I would turn the default value to be `true`.

src/carousel/carousel.ts
divdavem: `null` means the attribute is not there.
`-1` makes the element still focusable when clicking on it, even though it is not reachable through tab.

src/carousel/carousel.ts
divdavem: Actually, the default value is not set here but in `carousel-config.ts`.
I just did like `_pauseOnHover$` which is also `true` by default.

src/carousel/carousel.ts
benouat: > `null` means the attribute is not there.

Oh yes! Stupid me 🤷 !
Because it's a binding, this is the default angular behaviour. `null` will remove the attribute. I always forget about this.

src/carousel/carousel.ts
benouat: But then, it means all our `BehaviorSubject` should be initialised in the constructor with the proper initial value coming from the config no ?

src/carousel/carousel.ts
divdavem: But I am thinking that it may actually be a good idea to allow focusing this element on click. Otherwise, depending on what is inside the carousel, it may happen that nothing is focused on click, and that could be confusing if the user then wants to interact with the carousel with the keyboard.

src/carousel/carousel.ts
divdavem: Thank you for your comment. I have added a commit to change this.

src/carousel/carousel.ts
divdavem: Actually, currently we do the following in the constructor:
```js
    this.interval = config.interval;
    this.wrap = config.wrap;
    this.keyboard = config.keyboard;
    this.pauseOnHover = config.pauseOnHover;
    this.pauseOnFocus = config.pauseOnFocus;
    this.showNavigationArrows = config.showNavigationArrows;
    this.showNavigationIndicators = config.showNavigationIndicators;
```
So those `BehaviorSubject` objects are created with an arbitrary value and then updated in the constructor (through the setters) with the correct default value, so the behavior is correct, even though we could improve a bit the performance by directly creating those objects in the constructor (I am not sure it would make a noticeable difference, though).

src/carousel/carousel.ts
divdavem: I have changed to use `-1` instead of `null` to allow focusing the `ng-carousel` element when clicking on it.

src/carousel/carousel.ts
```diff
@@ -163,20 +201,29 @@ export class NgbCarousel implements AfterContentChecked,
...
   }
 
+  set mouseHover(value: boolean) { this._mouseHover$.next(value); }
+
+  get mouseHover() { return this._mouseHover$.value; }
+
+  set focused(value: boolean) { this._focused$.next(value); }
+
+  get focused() { return this._focused$.value; }
+
```
benouat, line 226: It might sound nitpicking here, but could you please move these getters/setters above the constructor ?

src/carousel/carousel.ts
benouat: Ok, then let's keep it as it is today!

src/carousel/carousel.ts
divdavem: I have added a commit with this change.

benouat:
I still have one consideration regarding all these changes.

I was having a look a the first demo, if you click on previous/next arrow, I assume that focus will be somewhere caught inside the widget, but there is no visual hint such as an outline. Hence the carousel gets stuck in pause mode, and for the end user you don't know why...

It's not very intuitive that you have to click _elsewhere_ for the carousel to resume.

src/carousel/carousel.ts
```diff
@@ -55,24 +55,34 @@ export class NgbSlide {
...
+    <ol class="carousel-indicators" [class.sr-only]="!showNavigationIndicators" role="tablist">
       <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId"
-          (click)="select(slide.id, NgbSlideEventSource.INDICATOR)"></li>
+          role="tab" [attr.aria-labelledby]="slide.id + '-slide'" [attr.aria-controls]="slide.id + '-slide'"
+          [attr.aria-selected]="slide.id === activeId"
+          (click)="focus();select(slide.id, NgbSlideEventSource.INDICATOR);"></li>
     </ol>
     <div class="carousel-inner">
-      <div *ngFor="let slide of slides" class="carousel-item" [class.active]="slide.id === activeId">
+      <div *ngFor="let slide of slides; let i=index;" class="carousel-item" [class.active]="slide.id === activeId" [id]="slide.id + '-slide'" role="tabpanel">
```
benouat, line 74: You could actually write the ngFor with single `let`
```html
<div *ngFor="let slide of slides; index as i">
```
and length of the collection can also be accessed by a local variable: `count`
```html
<div *ngFor="let slide of slides; index as i; count as c">
```

src/carousel/carousel.ts
```diff
@@ -55,24 +55,34 @@ export class NgbSlide {
...
+        <span class="sr-only" i18n="Currently selected slide number read by screen reader@@ngb.carousel.slide">
+          Slide {{i + 1}} of {{slides.length}}
```
benouat, line 76: `@@ngb.carousel.slide` sounds too generic to me here. what about `@@ngb.carousel.slide-number`

divdavem:
@benouat Thank you for your review and your comments. I have updated the code accordingly.

PR #3776, opened 2020-06-17
Nav animations

maxokorokov:
Again this is inspired by #2817, not cherry-picked though, because there were several issues with the implementation.

Contains 2 main changes:
- a bugfix in `ngbRunTransition` so it would still pass context even if animations are disabled
- nav animations implementation

## Nav animations
- `NgbNav` notifies `NgbNavOutlet` when the active item was changed via `(click)`, `[activeId]` or `.select()`
- `NgbNavOutlet` manages the stream of these `navItemChange` events and handles fade in/out transitions on `NgbNavPanes`
- new additional `(shown)` and `(hidden)` outputs both at `NgbNavItem` and at `NgbNav` level:

```html
<ul ngbNav #n="ngbNav" class="nav-tabs" (shown)="..." (hidden)="...">
  <li ngbNavItem (shown)="..." (hidden)="...">
    <a ngbNavLink>link</a>
    <ng-template ngbNavContent>content</ng-template>
  </li>
</ul>
<div [ngbNavOutlet]="n"></div>
```

- tests were pretty challenging to write here...

### DIFF
--- a/src/nav/nav-outlet.ts
+++ b/src/nav/nav-outlet.ts
@@ -1,5 +1,42 @@
-import {Component, Input, ViewEncapsulation} from '@angular/core';
-import {NgbNav} from './nav';
+import {
+  AfterViewInit,
+  ChangeDetectorRef,
+  Component,
+  Directive,
+  ElementRef,
+  Input,
+  QueryList,
+  ViewChildren,
+  ViewEncapsulation
+} from '@angular/core';
+
+import {distinctUntilChanged, skip, startWith} from 'rxjs/operators';
+
+import {
+  ngbNavFadeInNoReflowTransition,
+  ngbNavFadeInTransition,
+  ngbNavFadeOutTransition
+} from '../util/transition/ngbFadingTransition';
+import {ngbRunTransition, NgbTransitionOptions} from '../util/transition/ngbTransition';
+import {NgbNav, NgbNavItem} from './nav';
+
+@Directive({
+  selector: '[ngbNavPane]',
+  host: {
+    '[id]': 'item.panelDomId',
+    'class': 'tab-pane',
+    '[class.fade]': 'nav.animation',
+    '[attr.role]': 'role ? role : nav.roles ? "tabpanel" : undefined',
+    '[attr.aria-labelledby]': 'item.domId'
+  }
+})
+export class NgbNavPane {
+  @Input() item: NgbNavItem;
+  @Input() nav: NgbNav;
+  @Input() role: string;
+
+  constructor(public elRef: ElementRef<HTMLElement>) {}
+}
 
 /**
  * The outlet where currently active nav content will be displayed.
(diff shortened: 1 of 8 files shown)


maxokorokov:
@fbasso I removed the `ngbRunTransition` commit from here and moved it into the #3781

src/nav/nav-outlet.ts
```diff
@@ -34,4 +73,80 @@ export class NgbNavOutlet {
...
+    this._activePane ?.elRef.nativeElement.classList.add('show');
+    this._activePane ?.elRef.nativeElement.classList.add('active');
+
+    // this will be emitted for all 3 types of nav changes: .select(), [activeId] or (click)
+    this.nav.navItemChange$
+      .pipe(filter(nextItem => this._activePane?.item.id !== nextItem?.id))
+      .subscribe(nextItem => {
+      const options: NgbTransitionOptions<undefined> = {animation: this.nav.animation, runningTransition: 'stop'};
+
+      // Case 1: "fade out" is not running and it doesn't matter if "fade in" is currently running or not.
```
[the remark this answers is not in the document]
maxokorokov: I'll look again with fresh eyes, but couldn't do simpler than it is at the moment...

src/util/transition/ngbFadingTransition.ts
```diff
@@ -19,3 +20,19 @@ export const ngbToastFadeOutTransition: NgbTransitionStartFn = ({classList}: HTM
...
 };
+
+export const ngbNavFadeOutTransition: NgbTransitionStartFn = ({classList}) => {
+  classList.remove('show');
+  return () => classList.remove('active');
+};
+
+export const ngbNavFadeInTransition: NgbTransitionStartFn = (element: HTMLElement) => {
+  element.classList.add('active');
+  reflow(element);
```
maxokorokov, line 31: @fbasso I think we might be able to remove this `reflow` after recent changes to the `ngbRunTransition`. Will test this theory after reviews are done

maxokorokov: nope, we can't

maxokorokov:
re-pushed, because `nav-outlet` was simplified/refactored based on @fbasso's ideas

src/nav/nav.spec.ts
```diff
@@ -1094,6 +1150,253 @@ describe('nav', () => {
...
+          // 8. adding 3rd content, 2nd still active (only starting fading in)
+          expectContents(fixture, ['content 2', 'content 3'], 0);
+          [first, second] = getContents(fixture);
+          expectContentState(first, ['fade'], ['show'], '0');
+          expectContentState(second, ['fade'], ['show'], '0');
+        }
+      });
+
+      onNavShownSpy.and.callFake(shownId => {
+        // 9. (shown) is fired 2 times, for 1 -> 2, then for 2 -> 3
```
fbasso, line 1367: Are you sure ?

> // TEST: 1 is active, switching 1 -> 2, switching 2 -> 3 while fading in 2, catching nav (shown)

I'd say if you click on 3 when 2 is fade in, `(shown)` shouldn't be called as it fades out before being completely shown. So I'd expect to have two `(hidden)` (for 1 and 2) and only one `(shown)` (for 3) called at the end of the test.

src/nav/nav.spec.ts
```diff
@@ -1094,6 +1150,253 @@ describe('nav', () => {
...
+          fixture.detectChanges();
+          expect(shownId).toBe(3);
+          expect(onNavHiddenSpy).toHaveBeenCalledWith(1);
+          expect(onNavHiddenSpy).toHaveBeenCalledWith(2);
+          expect(onItemHiddenSpy).toHaveBeenCalledWith(1);
+          expect(onItemHiddenSpy).toHaveBeenCalledWith(2);
+          expect(onItemShownSpy).toHaveBeenCalledWith(3);
+          expectContents(fixture, ['content 3']);
+          expectContentState(getContent(fixture), ['fade', 'show'], [], '1');
+          done();
```
fbasso, line 1379: Following my previous comment, maybe you should add (in each test) checks like:

```
expect(onNavHiddenSpy).toHaveBeenCalledTimes(x);
expect(onNavShownSpy).toHaveBeenCalledTimes(x);
```
just before the `done()`, to be sure that there is no extra call/work done before the test end.

What do you think ?

src/nav/nav.spec.ts
maxokorokov: you're right, the comment is wrong. it is called 1 time in reality, i'll remove the `if`

maxokorokov: And by the way the test is coorect as well, as I'm expecting only:
```ts
expect(onItemShownSpy).toHaveBeenCalledWith(3);
```

src/nav/nav.spec.ts
maxokorokov: you mean just to make sure that is called more than two times?

this kind of imlies that is is called two times:

```ts
        expect(onItemHiddenSpy).toHaveBeenCalledWith(1);
        expect(onItemHiddenSpy).toHaveBeenCalledWith(2);
```

fbasso:
Thanks @maxokorokov, I finished the review, I like the nav-outlet code, much clearer (at least for me !) than the first version.

The code looks good to me. Just a note about the tests (see my comments). In short, the idea is to add some `haveBeenCalledTimes` at the end of each animation, to be more deterministic on the events called.

maxokorokov:
@fbasso thanks for the review! 
Will add `toHaveBeenCalledTimes(x)` to tests and push again, good idea.

maxokorokov:
Updated based on @fbasso's comments. Now using
- `expect(onItemHiddenSpy.calls.allArgs()).toEqual([[1], [2]]);` checks instead of `.toHaveBeenCalledWith(x)` for stricter checks

PR #3781, opened 2020-06-22
fix(animations): pass transition context with animation disabled

maxokorokov:
Fixes an issue when the context was not passed across two transitions when `animation = false` (see test for the example)


maxokorokov:
@fbasso thanks, merging!

Issue #3783, opened 2020-06-23
Dropdown does not autoclose, throwing: Object doesn't support property or method 'closest'


Istanful:
### Bug description:
When I open a dropdown with autoclose and then click on the scrollbar in edge, an error is thrown in the console. It throws the error `Object doesn't support property or method 'closest'`. It is thrown because the element in `event.target` is `HTMLDocument` which does not support the `closest` method.

### Link to minimally-working StackBlitz that reproduces the issue:
StackBlitz: https://angular-dpdswh-r6tj5y.stackblitz.io

1. Open the StackBlitz in `Edge 44.18362.449.0`.
2. Open the dropdown by clicking on it. Click on the scrollbar.
3. The dropdown does not close. An error is thrown in the console.

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 8.2.13

ng-bootstrap: 4.2.2

Bootstrap: 4.3.1

Browser: Edge 44.18362.449.0, EdgeHTML 18.18362

benouat:
Hey @Istanful 
thanks to report this issue!

I we refer to [browser support](https://angular.io/guide/browser-support) page on angular.io for v8, only the 2 most recent major versions of Edge are supported.

With the release of Chromium based Edge back in January, I would say that we don't support EdgeHTML anymore...

Istanful:
@benouat 

Thank you for the swift answer! Makes perfect sense. 

About 60% of the application I'm working on uses those old Edge versions, which of course is a shame. Because of that I will have to do some kind of workaround. Do you know of any specific settings I can use that will circumvent this issue?

The current solution would be to write a custom auto close strategy and disable yours entirely. :/

benouat:
I would suggest to use a polyfill 🙂 Way much easier.

You can have a look at this [MDN](https://developer.mozilla.org/fr/docs/Web/API/Element/closest) page

benouat:
oohhh my bad!
I did not read carefully your issue description!!

EdgeHTML actually do support closest 😄 It's only when you click on the scrollbar.
Right, then you might have spotted a bug in a sense, that we should discard doing the call to closest if `event.target` is `document`.

Sorry for that.
We'll have a look

Istanful:
@benouat Cool! I will make an attempt on a PR if you don't get to it first. :)

Istanful:
[...]

PR #3786, opened 2020-06-25
fix(dropdown): support closest for HTMLDocument (#3783)

Istanful:
## Support closest for HTMLDocument

When autoclosing a dropdown by clicking the scrollbar, an error would be
thrown in Edge 44.18362.449.0. This was due to the HTMLDocument not
supporting `Element.prototype.closest`.

By emulating `Element.prototype.closest` for this specific node, this
commit circumvents the issue.

### DIFF
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -66,8 +66,21 @@ if (typeof Element !== 'undefined' && !Element.prototype.closest) {
   };
 }
 
-export function closest(element: HTMLElement, selector): HTMLElement | null {
+export function closest(element: HTMLElement, selector?: string): HTMLElement | null {
   if (!selector) {
+    return null;
+  }
+
+  /*
+   * In certain browsers (e.g. Edge 44.18362.449.0) HTMLDocument does
+   * not support `Element.prototype.closest`. To emulate the correct behaviour
+   * we return null when the method is missing.
+   *
+   * Note that in evergreen browsers `closest(document.documentElement, 'html')`
+   * will return the document element whilst in Edge null will be returned. This
+   * compromise was deemed good enough.
+   */
+  if (typeof element.closest === 'undefined') {
     return null;
   }
 
--- a/src/util/util.spec.ts
+++ b/src/util/util.spec.ts
@@ -1,6 +1,38 @@
-import {toInteger, toString, getValueInRange, isInteger, isString, hasClassName} from './util';
+import {toInteger, toString, getValueInRange, isInteger, isString, hasClassName, closest} from './util';
 
 describe('util', () => {
+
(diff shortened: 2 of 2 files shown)


src/util/util.ts
```diff
@@ -66,10 +66,19 @@ if (typeof Element !== 'undefined' && !Element.prototype.closest) {
   };
 }
 
-export function closest(element: HTMLElement, selector): HTMLElement | null {
+export function closest(element: HTMLElement, selector?: string): HTMLElement | null {
   if (!selector) {
     return null;
   }
 
+  /*
```
benouat, line 74: I think we should do thing a little bit differently here because we can't just access `document` like that.

Any angular app could rendered server-side, where basically `document` won't be defined. So in order to properly use it, we usually have to get it by injection so that DI can provide a _special_ one in environment where there is no `document`. (Have a [look](https://github.com/ng-bootstrap/ng-bootstrap/search?q=%40Inject%28DOCUMENT&unscoped_q=%40Inject%28DOCUMENT) at how we do it)

But here, as we are in an utility function, it would mean we need to pass it as an argument, to have access to DI from the caller standpoint.

I found this a bit heavy, for such a small use case. (It apparently only fails in EdgeHTML)

What about we simply
```typescript
try {
  return element.closest(selector)
} catch(err) { 
  return null;
}
```

WDYT ?

benouat:
Could you please also rename the commit message when you'll push again with something starting with `fix:`
It's not `build:` related.

Istanful:
@benouat Sure! I'll rename the commit. I wasn't sure what category it would fit in but figured `:build` from the description. :)

> build: ... → any change for the **utility** scripts, configurations, dependencies, etc.

src/util/util.ts
Istanful: Nice catch. It slipped my mind. 😛 

I'm a bit hesitant to use try-catch since any error would be caught. How about using a more exact check? 

```typescript
if (typeof element.closest === 'undefined') {
  return null;
}
```

Beware though; in modern browsers `document.documentElement.closest('html')` would return `document`. By returning null always, we get a differing behaviour in Edge. How would you go about that problem?

Istanful: Ping @benouat. Would you be ok with that behaviour and check? :)

benouat: Yes, go ahead with that!

Don't forget to comment out things (this is needed for old edgeHTML generation) because I am pretty sure in 3 months from now we'll have forgotten why we do have this test!

Istanful: @benouat Cool. I've changed the commit and code according to discussion. :)

PR #3790, opened 2020-06-29
build: updated Angular to v10

santam85:
Ran update schematics to v10 and fixed tests


santam85:
Had to push last commit again to retrigger the build, due to this morning's GitHub instability it wasn't added automatically to this PR after several hours of waiting.

src/util/accessibility/live.ts
```diff
@@ -33,6 +33,10 @@ function getLiveElement(document: any, lazyCreate = false): HTMLElement | null {
 
 
 
+// TODO: Add Angular decorator.
+// TODO: Add Angular decorator.
+// TODO: Add Angular decorator.
+// TODO: Add Angular decorator.
```
pkozlowski-opensource, line 39: This is due to https://github.com/angular/angular/issues/37726 I believe so in this sense this is Angular migrations "fault". Still, we should remove those TODOs from this PR.

santam85: Ooops, missed those

e2e-app/src/app/datepicker/datepicker.po.ts
```diff
@@ -1,7 +1,7 @@
 import {$} from 'protractor';
 
 export class DatepickerPage {
-  getDatepicker(selector = 'ngb-datepicker') { return $(selector); }
+  getDatepicker(selector = 'ngb-datepicker') { return $(selector) as any; }
```
santam85, line 4: As for the other `ElementFinder as any`, this is needed due to additional type checks added for the updated Typescript version, where either an actual promise or an object without a `then` method are required when using async/await.

schematics/ng-add/setup-project.spec.ts
```diff
@@ -16,7 +16,7 @@ import {createTestApp} from '../utils/testing';
 
     async function createAppWithOptions(appOptions = {}): Promise<{ tree: UnitTestTree, project: WorkspaceProject }> {
       // 'app' is the default application, so we're not passing '--project' option
-      const options: Schema = projectName === 'app' ? {} : {project: projectName};
+      const options: Schema = {project: projectName};
```
santam85, line 19: Without `projectName` @angular/localise schematics would fail without a value for it.

e2e-app/src/app/datepicker/datepicker.po.ts
maxokorokov: Hmm, I think the fix for this  and other cases should be not using `await` when calling this function instead of adding `as any`, because everything is actually synchronous.

To me the error is here:

```ts
const blah = await page.getDatepicker();
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ng-bootstrap/ng-bootstrap",
-  "version": "6.1.0",
+  "version": "7.0.0",
```
maxokorokov, line 3: Oh, could you please also not modify the version for now? Leave at `6.1.0`

santam85: Reverted to 6.1.0

e2e-app/src/app/datepicker/datepicker.po.ts
santam85: Yes you are right, made sure there are no cases where it's resolved asynchronously in the e2e tests.

PR #3792, opened 2020-06-29
Popover animations

maxokorokov:
Again this is cherry-picked and fixed/adapted from #2817.

The popover API is the same as before:

```html
<div ngbPopover="Look at me!" (shown)="..." (hidden)="..."></div>
```

`(shown)` and `(hidden)` emit synchronously as before in case of reduced motion or no animations

Will open the same thing for the tooltip once this is merged

### DIFF
--- a/src/util/popup.ts
+++ b/src/util/popup.ts
@@ -1,13 +1,19 @@
 import {
-  Injector,
-  TemplateRef,
-  ViewRef,
-  ViewContainerRef,
-  Renderer2,
-  ComponentRef,
+  ApplicationRef,
   ComponentFactoryResolver,
-  ApplicationRef
+  ComponentRef,
+  Injector,
+  NgZone,
+  Renderer2,
+  TemplateRef,
+  ViewContainerRef,
+  ViewRef
 } from '@angular/core';
+
+import {Observable, of} from 'rxjs';
+import {mergeMap, take, tap} from 'rxjs/operators';
+
+import {ngbRunTransition} from './transition/ngbTransition';
 
 export class ContentRef {
   constructor(public nodes: any[], public viewRef?: ViewRef, public componentRef?: ComponentRef<any>) {}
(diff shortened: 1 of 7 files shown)


src/util/popup.ts
```diff
@@ -19,31 +25,47 @@ export class PopupService<T> {
...
+    if (!this._windowRef) {
+      return of(undefined);
     }
+
+    return ngbRunTransition(
+               this._windowRef.location.nativeElement, ({classList}) => classList.remove('show'),
+               {animation, runningTransition: 'continue'})
+        .pipe(tap(() => {
+          if (this._windowRef) {
+            this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._windowRef.hostView));
```
benouat, line 60: Can't we take the opportunity here to upgrade to `this._windowRef.hostView.destroy()` instead of `this._viewContainerRef.remove(this._viewContainerRef.indexOf(this._windowRef.hostView))` ?

maxokorokov: We could, sure. I wonder why it was not done before... it is already done for `this._contentRef.viewRef.destroy();` 5 lines below

maxokorokov:
> I assume that because of the change inside the popupService, this is why you also had to update tooltip and typeahead inside this PR ?

@benouat sure. As you can see, the popup service API has changed from:

```ts
open(...): ComponentRef<T>;
close(...): void;
```

to:

```ts
open(...): {windowRef: ComponentRef<T>, transition$: Observable<void>}
close(...): Observable<any>
```

src/util/popup.ts
```diff
@@ -19,31 +25,47 @@ export class PopupService<T> {
...
     }
 
-    return this._windowRef;
-  }
+    const {nativeElement} = this._windowRef.location;
+    const onStable$ = this._ngZone.onStable.asObservable().pipe(take(1));
+    const transition$ = onStable$.pipe(
+        mergeMap(
+            () => ngbRunTransition(
+                nativeElement, ({classList}) => classList.add('show'), {animation, runningTransition: 'continue'})));
```
fbasso, line 45: To be discussed, but I'm not sure but here we can use `stop` instead of `continue`. I don't know if there is a reason to not be able to revert it

The question is : do we want to behave like the alert (not reversible) or like the dialog, toast or collapse (reversible) ?

maxokorokov: I would say reversible, as it works in two ways - opening/closing. I'll change that, thanks.

fbasso:
Thanks @maxokorokov, the code is fine for me ! 

Just two little points (in addition to @benouat comments):

- [ ] Rebase the branch on the latest commit of `animations` branch. It shouldn't be a problem, it's just to be sure that there is no problem with the #3793 changes,
- [ ] If it's now reversible, don't forget to add the corresponding tests.

maxokorokov:
Pushed again, integrated both @benouat's  and @fbasso's comments:
- rebased
- fixed comment typo
- using `this._windowRef.destroy()` now
- popup opening can now be reverted (@fbasso now works in the same way as modal)
- added test for opening revert → https://github.com/ng-bootstrap/ng-bootstrap/pull/3792/commits/3b2055d0e117e623fcc55f7c504b58c71c0cf5d9#diff-df50abf4356a61091d4c11859f693908R851-R880

PR #3793, opened 2020-06-29
feat(animations): refactor ngbTransition

fbasso:
This is a refactor of ngbTransition to fix an issue around the transition detection.

The main problem comes from the order startFn is run and the transition computed style.

It's easier to explain the issue with the collapse widget:

What Bootstrap do :

- The transition is set on the `collapsing` class,
- When reduce motion is activated, Bootstrap set this transition to none.

Before this PR :

- We get the transition value for mthe element computed style,
- if this transition is `none`, we consider that there is no animation and we run the startFn  an endFn synchronously.

This is wrong, as it's the startFn purpose is to change collapse to collapsing, we can't compute the `transition` property of `collapsing` before running it.

What this PR do:

- Always run startFn 
- If no animation (via the options or transition none), run endFn synchronously

At the same time, for the collpase widget:

- The shown and hidden were missing to well managed the transitions: there are emitted at the transition end, unlike the collapseChange.
- The tests have been refactor to well reflect the asynchronous/synchronous stuff depending on the reduce motion activation

The other transitions tests need to be refactor as well (not done yet).

### DIFF
--- a/src/collapse/collapse.ts
+++ b/src/collapse/collapse.ts
@@ -1,4 +1,13 @@
-import {Directive, Input, ElementRef, Output, EventEmitter} from '@angular/core';
+import {
+  Directive,
+  ElementRef,
+  EventEmitter,
+  Input,
+  OnChanges,
+  OnInit,
+  Output,
+  SimpleChanges,
+} from '@angular/core';
 import {ngbRunTransition} from '../util/transition/ngbTransition';
 import {ngbCollapsingTransition} from '../util/transition/ngbCollapseTransition';
 import {NgbCollapseConfig} from './collapse-config';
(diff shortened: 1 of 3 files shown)


src/collapse/collapse.ts
```diff
@@ -27,8 +32,31 @@ export class NgbCollapse {
...
+   * An event emitted when the collapse element is hidden, after the transition. It has no payload.
+   */
+  @Output() hidden = new EventEmitter<void>();
+
+
   constructor(private _element: ElementRef, config: NgbCollapseConfig) { this.animation = config.animation; }
 
+  ngOnInit() {
+    const {classList} = this._element.nativeElement;
+    classList.add('collapse');
```
maxokorokov, line 50: I think this one we should add via the `'class': 'collapse'` binding as before, what's the point doing it imperatively, if we could do this declaratively with Angular...

src/util/transition/ngbTransition.ts
```diff
@@ -47,18 +48,15 @@ export const ngbRunTransition =
             }
           }
 
-          // If animations are disabled, we have to emit a value and complete the observable
-          if (!options.animation) {
-            (startFn(element, context) || noopFn)();
-            return of(undefined);
-          }
+          reflow(element);
```
maxokorokov, line 51: Could you please explain why in a comment? because we'll never remember this later.

src/collapse/collapse.ts
fbasso: I would prefer not merged what is done by binding and what is done by the transition engine. 

As far as remember, the collapse class was added back just after starting the transition (direction: show), because `collapsed` is true. The result is there is `collapsing`and `collapse`at the same time in the dom.

I will check, but I thing the tests in this PR prevent this, so it should fail.

fbasso:
Thanks !

Issue #2169, opened 2018-02-14
Popover sticky on hover


jimmymajed:
Hello,

Would it be possible for the popover triggered on hover to remain displayed even if the mouse is out of the trigger area but inside the popover itself?

I found this solution on stackoverflow in jQuery:
https://stackoverflow.com/questions/15989591/how-can-i-keep-bootstrap-popover-alive-while-the-popover-is-being-hovered

Best,
Jimmy

Safiyya:
For anyone interested, here is a sticky popover that extends the functionnalities of `NgbPopover`

```
import {
    ElementRef,
    Directive, Input, TemplateRef,
    EventEmitter,
    Renderer2,
    Injector,
    ComponentFactoryResolver,
    ViewContainerRef,
    NgZone
} from "@angular/core";
import { NgbPopover, NgbPopoverConfig } from "@ng-bootstrap/ng-bootstrap";
import { ContentRef } from "@ng-bootstrap/ng-bootstrap/util/popup";

@Directive({
    selector: "[stickyPopover]",
    exportAs: "stickyPopover"
})
export class StickyPopoverDirective extends NgbPopover {
    @Input() stickyPopover: TemplateRef<any>;

    popoverTitle: string;

    placement: "auto" | "top" | "bottom" | "left" | "right" | "top-left" | "top-right" | "bottom-left" | "bottom-right" | "left-top" | "left-bottom" | "right-top" | "right-bottom" | ("auto" | "top" | "bottom" | "left" | "right" | "top-left" | "top-right" | "bottom-left" | "bottom-right" | "left-top" | "left-bottom" | "right-top" | "right-bottom")[];

    triggers: string;

    container: string;

    shown: EventEmitter<{}>;

    hidden: EventEmitter<{}>;

    toggle(): void {
        super.toggle()
    }

    isOpen(): boolean {
        return super.isOpen()
    }

    ngpPopover: TemplateRef<any>;

    canClosePopover: boolean;

    constructor(private _elRef: ElementRef, private _render: Renderer2, injector: Injector,
        componentFactoryResolver: ComponentFactoryResolver, private viewContainerRef: ViewContainerRef, config: NgbPopoverConfig,
        ngZone: NgZone) {
        super(_elRef, _render, injector, componentFactoryResolver, viewContainerRef, config, ngZone);
        this.triggers = "manual"
        this.popoverTitle = "Permissions";
        this.container = "body"
    }
    ngOnInit(): void {
[...]

PR #3794, opened 2020-06-29
Sticky popovers and tooltips

earshinov:

---

This PR attempts to solve the problem described in https://github.com/ng-bootstrap/ng-bootstrap/issues/2169,

# Motivation

In our project we have popovers with information that users might like to copy, and also popovers with links for users to click. I consider these scenarios fairly basic and common, but ngbPopover doesn't really support them because it closes as soon as the user moves the mouse out of the element the popover is bound to (I will call it _the target element_ from here on).

Here are some "solutions" we considered:
  - Force our users to make a click to open the popup
  - Confugure `[closeDelay]` large enough for users to be able to click a link before the popup closes
  - Use code snippets from https://github.com/ng-bootstrap/ng-bootstrap/issues/2169.

None of these are acceptable.

# Proposal

The _basic idea_ is to listen to trigger events both on the target element AND the popover element. To illustrate, in case of `triggers="mouseenter:mouseleave"` the popover will close on `mouseleave` UNLESS `mouseenter` was triggered when the user moved the cursor from the target element to the popover.

Of course, the cursor isn't moved instantaneously, so some delay is required. You can find it in the new `NgbPopoverConfig.mouseleaveCloseDelay` field and `NgbPopover.mouseleaveCloseDelay` input. I chose 100 ms as the default, but you are welcome to suggest any other positive number. I will tell you more, there is a similar field and a similar input for `focusout` as well.

You might argue that this approach is wrong and these events shouldn't be treated as special citizens. Maybe they shouldn't, but although I am open to suggestions regarding design and refactoring I would prefer to keep the code dumb and not over-complicate things until the basic idea is approved and my approach reviewed and tested. (Also I must note here that events will need to be handled differently anyway because they require different delays: mouse movement does not happen instantly while clicks and focus changes do.)

Among the enhancements themselves this PR includes:
1. A couple of new unit tests which appropriately fail without the enhancements
2. A couple of new buttons on the popover demo page so that the updated behavior can be seen in action.
3. Similar enhancements, unit tests and demos for ngbTooltip

There are some technical details I would like to point out:
1. In order to detect focus events, a popover needs to be focusable (have `tabIndex`)
2. "Click" triggers are excluded from the general rule: in contrast to other trigger events, clicks are not listened to on the popover element. This is because clicks outside the target element are already nicely handled by `autoClose`.

_Update_: fixed some typos.

### DIFF
--- a/src/popover/popover.ts
+++ b/src/popover/popover.ts
@@ -20,11 +20,14 @@ import {
   SimpleChanges,
   ViewEncapsulation,
   ChangeDetectorRef,
-  ApplicationRef
+  ApplicationRef,
+  HostBinding
 } from '@angular/core';
 import {DOCUMENT} from '@angular/common';
+import {Subscription, Subject} from 'rxjs';
+import {switchMap} from 'rxjs/operators';
 
-import {listenToTriggers} from '../util/triggers';
+import {parseTriggers, observeTriggers, triggerDelay} from '../util/triggers';
 import {ngbAutoClose} from '../util/autoclose';
 import {positionElements, PlacementArray} from '../util/positioning';
 import {PopupService} from '../util/popup';
@@ -47,13 +50,23 @@ let nextId = 0;
     <div class="popover-body"><ng-content></ng-content></div>`,
   styleUrls: ['./popover.scss']
 })
-export class NgbPopoverWindow {
+export class NgbPopoverWindow implements OnInit {
   @Input() title: undefined | string | TemplateRef<any>;
   @Input() id: string;
   @Input() popoverClass: string;
   @Input() context: any;
+  @HostBinding('attr.tabIndex') tabIndex = -1;
+  @Output() inited = new EventEmitter<HTMLElement>();
+
+  set focusable(value: boolean) { this.tabIndex = value ? 0 : -1; }
+
+  get focusable() { return this.tabIndex !== -1; }
+
+  constructor(private _elementRef: ElementRef) {}
 
   isTitleTemplate() { return this.title instanceof TemplateRef; }
+
+  ngOnInit() { this.inited.next(this._elementRef.nativeElement); }
 }
 
 /**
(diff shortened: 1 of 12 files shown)


src/popover/popover.ts
```diff
@@ -148,6 +161,21 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
    */
   @Input() closeDelay: number;
 
+  /**
+   * The delay before the popover is closed on mouseleave.  Only takes effect if "mouseleave" is among the close
+   * triggers.
+   *
+   * @since *Unreleased*
+   */
+  @Input() mouseleaveCloseDelay: number;
```
earshinov, line 170: Do I need to prepare API documentation other than these comments?

src/popover/popover.ts
```diff
@@ -148,6 +161,21 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
    */
   @Input() closeDelay: number;
 
+  /**
+   * The delay before the popover is closed on mouseleave.  Only takes effect if "mouseleave" is among the close
+   * triggers.
+   *
+   * @since *Unreleased*
```
earshinov, line 168: `*Unreleased*` to be replaced with an appropriate version at the time of merge.

benouat:
Hey @earshinov! Sorry to comment out this late on your PR.

Allow me to challenge you that what you try to fix in this PR is not needed in the first place. 😉 
Please have a look a my https://github.com/ng-bootstrap/ng-bootstrap/issues/2169#issuecomment-661965023 in which I try to describe and explain the confusion around both popovers and tooltips.

My main concern was to try to give some answers somehow based on your description
> In our project we have popovers with information that users might like to copy, and also popovers with links for users to click. I consider these scenarios fairly basic and common, but ngbPopover doesn't really support them because it closes as soon as the user moves the mouse out of the element the popover is bound to (I will call it the target element from here on).

I have gathered my solution in a StackBlitz available [here](https://stackblitz.com/edit/angular-ivy-mvpupx)
Could you please have a look at it, and see if it could suit your needs ? Thanks! 🙏

PR #3798, opened 2020-07-02
Tooltip animations

maxokorokov:
Same as popover in #3792, but for the tooltip

Again this is cherry-picked and fixed/adapted from #2817.

The popover API is the same as before:

<div ngbPopover="Look at me!" (shown)="..." (hidden)="..."></div>
(shown) and (hidden) emit synchronously as before in case of reduced motion or no animations

Will open the same thing for the tooltip once this is merged

### DIFF
--- a/src/test/test-styles.css
+++ b/src/test/test-styles.css
@@ -9,6 +9,10 @@
   transition: opacity 0.01s linear !important;
 }
 
+.tooltip.show {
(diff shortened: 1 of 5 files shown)


maxokorokov:
Pushed changes to tests to fix Safari rounding issues when reading opacity

PR #3803, opened 2020-07-03
feat(datepicker): datepicker week day format via injection token (#2516)

### LINKED ISSUE #2516: Defaults for days of week too short - should be 3 letters, Mon, Tue,

### Bug description:
By default, the i18n choice for the days of the month from the datepicker is: Mo, Tu, We

It should be Mon, Tue, Wed etc like the old one:

https://angular-ui.github.io/bootstrap/ (search for datepicker)

IT would be an easy fix on your end, the enum Const that stores the i18n length just needs to be Abbreviation not Short.

Overridding it is hard,  https://stackoverflow.com/questions/50900233/internationalization-of-datepickers

You can see that in order to change it we have to override that provider with our own.

Angular: All

ng-bootstrap: latest

Bootstrap: no matter

AceVentura:
This PR fixes #2516. 
It's not mandatory for the developer to provide the token in his app, by default it will just inject the Short format.

In terms of API, getWeekdayShortName is no longer the correct name, but it's up to you if you want a breaking change.


AceVentura:
@benouat I'll get on to in. Just one question. Should the tests (e.g. datepicker-i18n.spec.ts) test the new method only or both of them?

benouat:
Just test the new one. As the old one will be just an alias to the new one, there's no point in testing both of them

AceVentura:
@benouat it's done. Tell me if you need anything else!

demo/src/app/components/datepicker/demos/i18n/datepicker-i18n.ts
```diff
@@ -27,6 +27,11 @@ export class CustomDatepickerI18n extends NgbDatepickerI18n {
   getWeekdayShortName(weekday: number): string {
     return I18N_VALUES[this._i18n.language].weekdays[weekday - 1];
   }
+
+  getWeekdayName(weekday: number): string {
+    return I18N_VALUES[this._i18n.language].weekdays[weekday - 1];
+  }
+
```
benouat, line 34: I would have actually call the new method from all the old ones instead of just duplicating.

```typescript
getWeekdayShortName(weekday: number): string {
    return this.getWeekdayName(weekday);
}
```

AceVentura:
yeah, that made no sense from my part. fixed now

AceVentura:
@benouat is there something missing for this pull request?

maxokorokov:
Sorry, had to revert it after merging into master for two reasons:
- it contained breakinf changes
- the whole number 4 from this list → https://github.com/ng-bootstrap/ng-bootstrap/issues/2516#issuecomment-404462241 doesn't really fit APIs we have in ng-bootstrap (usually we configure things via the `NgbXXXConfig` historically)

So I opened the #4028 instead

PR #3804, opened 2020-07-03
feat(carousel): add animations

fbasso:
Cherry-picked form #2817, then simplified and adapted to be aligned with recent widget development.

To be noticed:

- The `slid` event has been added (when the transition finish),
- The revert can be done only if the previous slide is selected.

### DIFF
--- a/src/carousel/carousel.spec.ts
+++ b/src/carousel/carousel.spec.ts
@@ -1,12 +1,15 @@
 import {fakeAsync, discardPeriodicTasks, tick, TestBed, ComponentFixture, inject} from '@angular/core/testing';
-import {createGenericTestComponent} from '../test/common';
+import {createGenericTestComponent, isBrowserVisible} from '../test/common';
 
 import {By} from '@angular/platform-browser';
 import {ChangeDetectionStrategy, Component} from '@angular/core';
 
 import {NgbCarouselModule} from './carousel.module';
-import {NgbCarousel, NgbSlideEvent, NgbSlideEventDirection, NgbSlideEventSource} from './carousel';
+import {NgbCarousel, NgbSlideEvent, NgbSlideEventSource, NgbSingleSlideEvent} from './carousel';
 import {NgbCarouselConfig} from './carousel-config';
+import {NgbConfig} from '../ngb-config';
+import {NgbConfigAnimation} from '../test/ngb-config-animation';
+import {NgbSlideEventDirection} from './carousel-transition';
 
 const createTestComponent = (html: string) =>
     createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
@@ -35,8 +38,8 @@ describe('ngb-carousel', () => {
   });
 
   it('should initialize inputs with default values', () => {
-    const defaultConfig = new NgbCarouselConfig();
-    const carousel = new NgbCarousel(new NgbCarouselConfig(), null, <any>null, <any>null, <any>null);
+    const defaultConfig = new NgbCarouselConfig(new NgbConfig());
+    const carousel = new NgbCarousel(new NgbCarouselConfig(new NgbConfig()), null, <any>null, <any>null, <any>null);
 
     expect(carousel.interval).toBe(defaultConfig.interval);
     expect(carousel.wrap).toBe(defaultConfig.wrap);
@@ -108,7 +111,7 @@ describe('ngb-carousel', () => {
 
        const fixture = createTestComponent(html);
 
-       fixture.componentInstance.activeSlideId = 1;
+       fixture.componentInstance.activeSlideId = '1';
        fixture.detectChanges();
        expectActiveSlides(fixture.nativeElement, [true, false]);
 
@@ -380,7 +383,7 @@ describe('ngb-carousel', () => {
        fixture.detectChanges();
        expect(fixture.componentInstance.carouselSlideCallBack).toHaveBeenCalledWith(jasmine.objectContaining({
          direction: NgbSlideEventDirection.RIGHT,
-         source: NgbSlideEventSource.INDICATOR
+         source: NgbSlideEventSource.INDICATOR,
        }));
 
        spyCallBack.calls.reset();
@@ -423,8 +426,8 @@ describe('ngb-carousel', () => {
   it('should fire a slide event with correct direction and source on carousel control click', fakeAsync(() => {
        const html = `
       <ngb-carousel (slide)="carouselSlideCallBack($event)">
-        <ng-template ngbSlide>foo</ng-template>
-        <ng-template ngbSlide>bar</ng-template>
+        <ng-template ngbSlide (slid)="carouselSingleSlideCallBack($event, 'foo')">foo</ng-template>
+        <ng-template ngbSlide (slid)="carouselSingleSlideCallBack($event, 'bar')">bar</ng-template>
       </ngb-carousel>
     `;
 
@@ -432,6 +435,7 @@ describe('ngb-carousel', () => {
        const prevControlElm = fixture.nativeElement.querySelector('.carousel-control-prev');
        const nextControlElm = fixture.nativeElement.querySelector('.carousel-control-next');
        const spyCallBack = spyOn(fixture.componentInstance, 'carouselSlideCallBack');
+       const spySingleCallBack = spyOn(fixture.componentInstance, 'carouselSingleSlideCallBack');
 
        prevControlElm.click();
        fixture.detectChanges();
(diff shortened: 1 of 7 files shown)


src/util/transition/ngbCarouselTransition.ts
```diff
@@ -0,0 +1,62 @@
+import {NgbTransitionStartFn} from './ngbTransition';
+import {reflow} from '../util';
+
+/**
+ * Defines the carousel slide transition direction.
+ */
+export enum NgbSlideEventDirection {
+  LEFT = 'left',
+  RIGHT = 'right'
+}
```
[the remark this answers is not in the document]
fbasso: If I let the definition in carousel.ts, I have a circular dependency:
```
WARNING in Circular dependency detected:
src\carousel\carousel.ts -> src\util\transition\ngbCarouselTransition.ts -> src\carousel\carousel.ts
```

So it has to be moved in another file. 

@benouat, @maxokorokov, what do you think about the best pattern in this case ? A separate file for the definitions, in the carousel folder ?

src/util/transition/ngbCarouselTransition.ts
```diff
@@ -0,0 +1,62 @@
...
+const removeDirectionClasses = (classList) => {
+  classList.remove('carousel-item-left');
+  classList.remove('carousel-item-right');
+};
+
+const removeClasses = (classList) => {
+  removeDirectionClasses(classList);
+  classList.remove('carousel-item-prev');
+  classList.remove('carousel-item-next');
+};
```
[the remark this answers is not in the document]
fbasso: You're right, and the destructuring notation bring the same readability. I'm going to change it, thanks !

src/util/transition/ngbCarouselTransition.ts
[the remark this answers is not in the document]
maxokorokov: My take is still the same that all custom transitions should be placed with components → `ngbCarouselTransition.ts` should become either a part of the `carousel.ts` or something like `carousel-transition.ts` inside the carousel folder. I see why common transitions should be placed in the `util/trainsition` folder, I don't see why component-specific should end there.

src/carousel/carousel.ts
```diff
@@ -60,11 +68,11 @@ export class NgbSlide {
   },
   template: `
     <ol class="carousel-indicators" *ngIf="showNavigationIndicators">
-      <li *ngFor="let slide of slides" [id]="slide.id" [class.active]="slide.id === activeId"
+      <li *ngFor="let slide of slides" [id]="'ngb-slide-indicator-' + slide.id" [class.active]="slide.id === activeId"
           (click)="select(slide.id, NgbSlideEventSource.INDICATOR)"></li>
     </ol>
     <div class="carousel-inner">
-      <div *ngFor="let slide of slides" class="carousel-item" [class.active]="slide.id === activeId">
+      <div *ngFor="let slide of slides" class="carousel-item" [id]="'ngb-slide-' + slide.id">
```
maxokorokov, line 75: This generates the following ids: ex. `ngb-slide-ngb-slide-13`, why is the `ngb-slide-` prefix necessary?
Can't we just use `slide.id`?

src/carousel/carousel.ts
```diff
@@ -91,6 +99,11 @@ export class NgbCarousel implements AfterContentChecked,
   private _pause$ = new BehaviorSubject(false);
   private _wrap$ = new BehaviorSubject(false);
 
+  /**
+   * A flag to enable/disable the animation when closing.
```
maxokorokov, line 103: When closing ?

src/carousel/carousel.ts
```diff
@@ -180,6 +207,15 @@ export class NgbCarousel implements AfterContentChecked,
   ngAfterContentInit() {
     // setInterval() doesn't play well with SSR and protractor,
     // so we should run it in the browser and outside Angular
+
+    // Initialize the 'active' class (not managed by the template)
+    this._ngZone.onStable.pipe(take(1)).subscribe(() => {
```
maxokorokov, line 212: it seems strange that we need the `onStable` here. Not necessary for the navs, for example

src/carousel/carousel.ts
```diff
@@ -246,11 +282,39 @@ export class NgbCarousel implements AfterContentChecked,
...
+      const activeSlide = this._getSlideById(this.activeId);
+      if (activeSlide) {
+        observers.push(ngbRunTransition(this._getSlideElement(activeSlide.id), ngbCarouselTransitionOut, option));
+      }
+
+      const previousId = this.activeId;
       this.activeId = selectedSlide.id;
+      observers.push(ngbRunTransition(this._getSlideElement(selectedSlide.id), ngbCarouselTransitionIn, option));
+
+      zip(...observers).subscribe(() => {
```
maxokorokov, line 313: This should be unsubscribed, I think. To verfiy.

src/carousel/carousel.ts
```diff
@@ -246,11 +282,39 @@ export class NgbCarousel implements AfterContentChecked,
...
+      return;
+    }
+
     let selectedSlide = this._getSlideById(slideIdx);
     if (selectedSlide && selectedSlide.id !== this.activeId) {
+      this._transitionIds = [this.activeId, slideIdx];
       this.slide.emit(
           {prev: this.activeId, current: selectedSlide.id, direction: direction, paused: this._pause$.value, source});
+
+      const option: NgbTransitionOptions<NgbCarouselCtx> = {
```
maxokorokov, line 297: `option` → `options`?

src/carousel/carousel.ts
```diff
@@ -246,11 +282,39 @@ export class NgbCarousel implements AfterContentChecked,
...
+        animation: this.animation,
+        runningTransition: 'stop',
+        context: {direction},
+      };
+
+      const observers: Array<Observable<any>> = [];
```
maxokorokov, line 303: maybe `transitions`? this is definitely not `observers`

src/carousel/carousel.ts
```diff
@@ -290,6 +354,10 @@ export class NgbCarousel implements AfterContentChecked,
     return isFirstSlide ? (this.wrap ? slideArr[slideArr.length - 1].id : slideArr[0].id) :
                           slideArr[currentSlideIdx - 1].id;
   }
+
+  private _getSlideElement(tabId: string): HTMLElement {
```
maxokorokov, line 358: why `tabId`? maybe `slideId`?

src/carousel/carousel.spec.ts
```diff
@@ -882,6 +885,186 @@ describe('ngb-carousel', () => {
 
 });
 
+if (isBrowserVisible('ngb-carousel animations')) {
+  describe('ngb-carousel animations', () => {
+
+    @Component({
+      template: `
+      <ngb-carousel (slid)="onSlid($event)" [interval]="-1">
+        <ng-template ngbSlide id="one">One</ng-template>
```
maxokorokov, line 894: Just an idea, but don't we want to have `(slid)` at this level as well? as in other components?

src/carousel/carousel.ts
fbasso: If I'm not wrong, having dom ids starting with numbers is a problem. It's just to avoid this case.

maxokorokov: it is already prefixed with `ngb-slide-`, you're prefixing it twice

fbasso: Yes, you're right, I answered too quickly 😟. I think I prefixed it the same way than the the indicators, but it's not needed. I'll remove it.

src/carousel/carousel.ts
fbasso: The elements (the slides) are not here yet, but I think I can use the `AfterViewInit` hook instead.

fbasso:
I pushed back most of the required changes. Still working on the `slid`event.

src/carousel/carousel.ts
fbasso: Finally, I need to prefix it. We prefixed the id only when it's not provided, this way:

```
@Input() id = `ngb-slide-${nextId++}`;
```

See the test `should mark the requested slide as active` for example.

benouat: What about not prefixing it when provided, but making sure they **do not** provide a number as ID ?

fbasso: IMO, I think the simpler (what have done in the latest version) is to prefix it in the template, not in the generated id. It keeps the freedom to provide what you want, without having to check anything.

fbasso:
Latest changes pushed.  The tests fails at the moment because of a mismatch between chrome driver (support only 85) and Chrome (officially 84 at the moment).

I integrated all the required changes, but maybe still things to discuss:

- I had to change a unit test in order to have slides id starting with letters, and not numbers. The issue comes that I need to query elements in order to initialize the `active` class. But:

  - Html 5/Javascript is weird on this point : numbers are valid for id attributes, but `querySelector('#1')` leads to `#1 is not a valid selector` error
  - I can't prefix ids coming from the user without breaking the compatibility (even if we already breaks the compatibility here : numbers can't be used. But there is maybe less chances to met this use case (?)). It's to be discussed, as we can break the compatibility in order to prefix the id all the time (generated or given by the user): in this case, there will be no more javascript errors, but users would not be able to query the dom with the ids they gave.
- (slid) has been added at slide level. NgbSingleSlideEvent is used for the payload. I'm not sure about the naming used.

fbasso:
After some discussion with @maxokorokov, it seems that the id are not used the way we thought. At the end, on this last changes, I:

  - Removed the ids for the indicators (not used)
  - Change the slides ids to prefix them by 'slide-' instead of a suffix (initially done by #3773, to avoid the id issue (querySelector not possible with id starting with a number)
  - Revert the 'should mark the requested slide as active' test (panel id with numbers)

PR #3805, opened 2020-07-03
Removed tabindex=-1 so that opening the modal does not unfocus already focused inputs.

danfergo:
Removed tabindex=-1 so that opening the modal does not unfocus already focused inputs.

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -25,7 +25,7 @@ import {ModalDismissReasons} from './modal-dismiss-reasons';
   host: {
     '[class]': '"modal fade show d-block" + (windowClass ? " " + windowClass : "")',
     'role': 'dialog',
-    'tabindex': '-1',
+    // 'tabindex': '-1',
     '[attr.aria-modal]': 'true',
     '[attr.aria-labelledby]': 'ariaLabelledBy',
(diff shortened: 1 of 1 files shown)


benouat:
Hey @danfergo 

Would you mind detailing a bit more what is your primary use case for that ?

danfergo:
Hi @benaut,
Sorry about that.

I was trying to prevent an input (outside the modal) from loosing the focus when the modal opens. Its a small modal, and in the desktop it's not a problem to loose the focus of the input, but on mobile it means the soft keyboard opening and closing.

Initially I thought it was due to the tabindex that my input was loosing focus, but then I realized that the modal controller is in fact taking over the focus. I didn't manage to find any parameter so I ended up cloning the modal module source code and implementing it on my one. Basically an if/else. If in the future you would like to add such parameter, that would be great.

benouat:
From what you describe, it sounds more like you are trying to use a modal for something that does not sound a modal.

In term of a11y, it's very very bad to implement a modal/dialog without a focus trap. I mean, having the modal to _steal_ the focus when opened is made on purpose! It's by design.

So maybe you should try experimenting with a popover, or something completely custom, but not a modal.

I hope it clarifies things on your side, meanwhile I'll close this PR because  as you could understand we won't merge it.

PR #3807, opened 2020-07-08
Animations carousel debug

fbasso:
This PR is not designed to be merged, but it is used to investigate the failing tests on travis for the  carousel animations (works locally).


fbasso:
Investigation finished.

PR #3826, opened 2020-08-05
feat(modal): add modalDialogClass to modal-dialog div

MisuleT:


jnizet:
Hi @MisuleT 

Thanks for the PR, but could you please explain why you think we should add such a configuration option? Can't you use the already supported `windowClass` option?

For example with such a selector in your stylesheet:

```
.myCustomWindowClass .modal-dialog {
  ...
}
```

MisuleT:
Hi @jnizet 

We depend on another library in our project, that is based on Bootstrap. We already use windowClass, but we have to add another class to modal-dialog div for the correct library function.

We created workaround where we add class 'g-modal' to size option. But it is not clear solution and if you change this option it won't work for us. 

` options.size += ' g-modal';`

MisuleT:
Hi @jnizet 

I fixed the test and this pull request is ready to merge :)

PR #3829, opened 2020-08-07
demos: improved UX for sidebar

benouat:
Instead of a long description, I'll put a GIF

Kapture 2020-08-07 at 10 27 25

### DIFF
--- a/demo/src/app/app.routing.ts
+++ b/demo/src/app/app.routing.ts
@@ -23,15 +23,24 @@ import {ROUTES as TYPEAHEAD_ROUTES} from './components/typeahead/typeahead.modul
 import {DefaultComponent} from './default';
 import {GettingStartedPage} from './pages/getting-started/getting-started.component';
 import {AnimationsPage} from './pages/animations/animations.component';
+import {I18nPage} from './pages/i18n/i18n.component';
 import {PositioningPage} from './pages/positioning/positioning.component';
 
 const routes: Routes = [
   {path: '', pathMatch: 'full', redirectTo: 'home'},
   {path: 'home', component: DefaultComponent},
   {path: 'getting-started', component: GettingStartedPage},
-  {path: 'animations', component: AnimationsPage},
-  {path: 'positioning', component: PositioningPage},
-  {path: 'components', pathMatch: 'full', redirectTo: 'components/alert'},
+  {path: 'animations', redirectTo: 'guides/animations'},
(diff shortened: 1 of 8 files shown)


benouat:
I did the change from `*ngIf` to `[ngbCollapse]` as suggested by @jnizet 

Maybe we should also introduce **Guides** as a top level navigation item in the blue header ?

@pkozlowski-opensource @maxokorokov are you guys ok with the update behind the PR ? 🤔 visually speaking I mean

pkozlowski-opensource:
@benouat is it about grouping guides into one block / section? If so, I think this is a good move 👍

benouat:
> is it about grouping guides into one block / section? If so, I think this is a good move 👍
 
yes, part of it is that, other part is to collapse the non-active section, or expand the active one 😄

pkozlowski-opensource:
> the non-active section, or expand the active one

Hmm, not sure about collapsing the components section - maybe we want to see a list of components all the time? Just a random thought, not feeling strongly either way.

benouat:
it's actually mimicking bootstrap's website. You see the list of components while you are on a component's page. Same for guides, you see the list while you are on a guide...

demo/src/app/app.routing.ts
```diff
@@ -31,7 +31,7 @@ const routes: Routes = [
   {path: 'getting-started', component: GettingStartedPage},
   {path: 'animations', component: AnimationsPage},
   {path: 'positioning', component: PositioningPage},
-  {path: 'components', pathMatch: 'full', redirectTo: 'components/alert'},
+  {path: 'components', pathMatch: 'full', redirectTo: 'components/accordion'},
```
maxokorokov, line 34: 🎉

maxokorokov (approved):
@benouat, I like the idea

- I think I would extract i18n from the getting started to the "Guides"
- Also "Guides" is not easily discoverable now, I would either just put them in the header or put the "Getting Started" under the guides menu:

```
Guides
  Getting Started
  Animations
  I18n
  Positioning

Components
```

benouat:
I kept `Getting Started` as stand alone section, and introduced `Guides` in the top level navigation.

image

I also moved `I18n` out of `Getting Started` to move it to `Guides`
I changed  a bit the spacing and added a background color when hovered on each nav item.

Here is the new demo
Kapture 2020-08-25 at 15 54 41

benouat:
I also moved guides to a new `#/guides/` route but kept redirection for both old `/animations` and `/positioning` routes

PR #3841, opened 2020-08-24
feat(progressbar): simplify template

zVolodymyr:
Reduced number of elements by moving "progress" class to host element. As result height can be changed via style.

### DIFF
--- a/demo/src/app/components/progressbar/demos/basic/progressbar-basic.ts
+++ b/demo/src/app/components/progressbar/demos/basic/progressbar-basic.ts
@@ -2,12 +2,7 @@ import {Component} from '@angular/core';
 
 @Component({
   selector: 'ngbd-progressbar-basic',
-  templateUrl: './progressbar-basic.html',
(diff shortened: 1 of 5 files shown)


benouat:
Hey, the build is failing, you have unformatted files
Please run `yarn check-format` locally and have a look at what's reported there.

maxokorokov (approved):
Thanks for the PR!

mradcliffe:
A couple of workarounds in case anyone else had extended NgbProgressBar and relied on a wrapper class with bootstrap's `progress` class to make fancy progress bars.

- Use increase specificity in CSS e.g. `my-custom-progressbar.progress { }`
- Set the new `height` property to `'auto'`, use `HostBinding` on other styles added by .progress on the component selector, use a wrapper class to reset the height to the default value because the height property applies to all elements that use .progress.
- Don't extend NgbProgressBar, re-write everything from scratch using bootstrap classes directly.

PR #3854, opened 2020-09-16
feat(demo): enable purgeCss

fbasso:
A small imporvement, but this PR enable purgeCss (https://purgecss.com/) for the demo site. After this change, the size of style.css decrease from 152kB (27kB gzipped) to 61kB (13kB gzipped).


maxokorokov (commented):
Hey, @fbasso!

I have 2 big comments to make about this:

1. The website is broken visually in some places after a quick look: for example datepicker doesn't look the same and tooltip bottom positioning is broken in demos

2. I would argue that it doesn't improve anything for the end-user, but adds more configuration/pain for us:

If you look at the timeline, we're in no way blocked by CSS, but by the main JS (highlighted with the red box).
Screenshot 2020-10-12 at 15 09 39

What the PR in question is doing is:

1. making CSS parsing 7ms faster (with 2x CPU slowdown):

Before purge:
Screenshot 2020-10-12 at 15 14 55

After purge:
Screenshot 2020-10-12 at 15 15 25


2. Improving the total website compressed transfer size by 13Kb from 435Kb → 422Kb, which is 3%

On the other hand it doesn't improve page load time or first paint at all.
So to me at the moment changes/gain ratio is not worth it at all.

I suggest we keep the PR open as a reference, but I don't see any improvement it will bring in our use case at the moment. Maybe later, if we implement lazy-loading.

cc @benouat

benouat:
I kinda agree, it's not worth it to me.

I'd like to see it integrated into CLI so we don't have to set it up ourselves, and thus it would make sense, as it would be purely transparent, and no burden at all

fbasso:
Thanks for the review @maxokorokov,

Yes, I didn't expect too much from this optimization, as the bandwidth used for the js is too important compared to the styles. The purpose was to get a state about the feasibility and to know what it brings (what you've done with your last comment).

Just a note about the configuration and the CLI : integrating it with webpack and Angular was quite hard, that's why I did it in an isolated manner. Even if it adds a config, I find it not so hard to configure (only the source directory and some white list keywords. Sure, we would need to carefully set it the first time, to not break anything), I think it would have been more tricky if done with the CLI.

But anyway, I'm fine to keep this PR open as a reference as you said, just in case...

maxokorokov:
Actually I'm going to close it, so it just wouldn't be popping in the list. We can always find it if necessary.

PR #3876, opened 2020-10-19
refactor(rating): simplify _getFillValue implementation

pkozlowski-opensource:
A tiny simplification to the existing `_getFillValue` in the rating:
* we've got already a utility to trim numbers to a given range so let's use it
* we can round numbers without converting them to a string :-)


maxokorokov (approved):
Well... yes :)
LGTM

PR #3909, opened 2020-11-23
fix(popover,tooltip): run transaction inside angular zone

### LINKED ISSUE #3896: Popover: Does not update parent view on shown event
### Bug description:
This is a regression bug of 8.0.0 release.

I have a subject that fires on each `shown`/`hidden` event, and is merged to an observable to update a count:

```Typescript
  count$: Observable<number> = merge(
    interval(3000),
    this.isShownSubject.pipe(mapTo(0))
  );
```

However, the view does not update on `shown`. Even though the event is fired, and the observable emits, the entire containing view does not update at all. The `hidden` event works as intended.

I have checked the current code and realized that the `NgbPopover.close()` marks for change detection after the `hidden` event is fired, but the `NgbPopover.open()` does not. Code:
https://github.com/ng-bootstrap/ng-bootstrap/blob/1f466edaa222812a15d233ab6a604ece68e3a58b/src/popover/popover.ts#L258

https://github.com/ng-bootstrap/ng-bootstrap/blob/1f466edaa222812a15d233ab6a604ece68e3a58b/src/popover/popover.ts#L267-L276

 I have manually tried adding the `this._changeDetector.markForCheck()`, but the view still does not update. Only explicitly adding `this._changeDetector.detectChanges()` fixes the issue.

It is even stranger that not only the directive content, but also the entire parent component view does not update (in the StackBlitz, the description bottom text does not change either). This occurs in OnPush and Default change detection strategies, even with animations disabled.

https://stackblitz.com/edit/angular-vsfpjy

Angular: 10.2.3

ng-bootstrap: 8.0.0

Bootstrap: 4.5.3

zVolodymyr:
Forcing to run ngbRunTransition inside angular zone, because NgZone.onStable is called outside angular zone.

closes #3896

### DIFF
--- a/src/util/popup.ts
+++ b/src/util/popup.ts
@@ -39,10 +39,10 @@ export class PopupService<T> {
 
     const {nativeElement} = this._windowRef.location;
     const onStable$ = this._ngZone.onStable.asObservable().pipe(take(1));
-    const transition$ = onStable$.pipe(
-        mergeMap(
-            () => ngbRunTransition(
-                nativeElement, ({classList}) => classList.add('show'), {animation, runningTransition: 'continue'})));
+    const transition$ = onStable$.pipe(mergeMap(() => this._ngZone.run(() => {
+      return ngbRunTransition(
+          nativeElement, ({classList}) => classList.add('show'), {animation, runningTransition: 'continue'});
+    })));
 
     return {windowRef: this._windowRef, transition$};
   }


maxokorokov (approved):
LGTM, thanks!
I was trying to come up with tests for this for some time, but couldn't.
If you have any ideas - will be helpful.

PR #3913, opened 2020-11-26
fix(datepicker): set 00:00 instead of 12:00 for the native adapter time

terencehonles:
I believe this is a regression and I'm not sure why it was added in 83930cd84190a077f0305b4f450b1485546bd874

Specifically: https://github.com/ng-bootstrap/ng-bootstrap/commit/83930cd84190a077f0305b4f450b1485546bd874#diff-3330a3121736b2b8be485023ca85083e3a69337d8ebc28edaa0a7029284ee007R13

### DIFF
--- a/src/datepicker/adapters/ngb-date-native-adapter.spec.ts
+++ b/src/datepicker/adapters/ngb-date-native-adapter.spec.ts
@@ -35,12 +35,12 @@ describe('ngb-date-native model adapter', () => {
     });
 
     it('should convert a valid date',
-       () => { expect(adapter.toModel({year: 2016, month: 10, day: 15})).toEqual(new Date(2016, 9, 15, 12)); });
+       () => { expect(adapter.toModel({year: 2016, month: 10, day: 15})).toEqual(new Date(2016, 9, 15)); });
 
     it('should convert years between 0 and 99 correctly', () => {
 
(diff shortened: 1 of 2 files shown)


peterblazejewicz:
Probably a duplicate of topic discussed here: #3378 
This is how NgBS computes day diffs, defaulting to high-noon, not midnight, see details.
I"m not if that is a best approach, but it's supposed to help with timezone related problems (I'm convinced it won't work always, as for Kiribati timezone case).
It has no relation to TypeScript related changes you've referenced IMO

terencehonles:
> Probably a duplicate of topic discussed here: #3378
> This is how NgBS computes day diffs, defaulting to high-noon, not midnight, see details.
> I"m not if that is a best approach, but it's supposed to help with timezone related problems (I'm convinced it won't work always, as for Kiribati timezone case).
> It has no relation to TypeScript related changes you've referenced IMO

Yes it's a duplicate since it's the same diff (and closing in favor of #3378), but I also linked to the same commit you point to so I'm not sure how you can say it's not related to those changes as the more specific link is the same you give.

Anyways, it looks like there's a bug somewhere else that they are working around and they're not super keen on changing this. It might need someone looking at where the adapter is used and only have the adapter being used for the end user but have any date related logic happen regardless of the adapter so they don't need the date to be in the middle of the day. This may have already changed since the original bug #1615 and may not be needed anymore.

PR #3932, opened 2021-01-07
fix(carousel): update active class when the content changes

### LINKED ISSUE #3911: Odd behaviour with carousel items using ngFor directive
### Bug description:
When using `*ngFor` with an observable subscription that determines the content, sometimes the newly added content (images) will not load properly even though it's there (network tab). I'm not sure of the process behind the scenes however I would expect the carousel to always load properly and even re-render itself if the slides were to change. What I think is happening, is that the carousel is being changed mid-transition, and then it's waiting until the next slide to fix itself, which isn't always the best when I just want to change the contents of the carousel and start it again.

Production use case is runtime localisation that provides data from an observable. Same concept as the stackblitz just not on a timer.

I have this issue currently on another site I'm working on which may cause differences, but to save complexity I have been able to repro it in a stackblitz and provide a demonstration of the behaviour as well.

https://stackblitz.com/edit/angular-49qw2x?file=src/app/carousel-basic.ts

https://www.youtube.com/watch?v=2cX3Q9dLumo

The reproduction is rather random, as it seems to happen when I click the force button, even though that's literally the same function being called on the 5 second interval.

Further to that I'm not sure of how the rendering happens at all, but should it be slide 400+? Should it not reset/recalculate based on the contents.

image

Angular: Stackblitz is 10.0.14, also present in v11

ng-bootstrap: 8.0.0

Bootstrap: 4 in stackblitz, 5-alpha in production (I don't think there's much difference with the carousel)

fbasso:
fix #3911 

This fix force the (missing) active class on the active slide, when the content has changed. Any running transition is stopped, but trying to keep all the current classes (to try to continue the transition with the new content) produces some strange behaviors.

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -25,7 +25,7 @@ import {NgbCarouselConfig} from './carousel-config';
 
 import {BehaviorSubject, combineLatest, NEVER, Observable, Subject, timer, zip} from 'rxjs';
 import {distinctUntilChanged, map, startWith, switchMap, take, takeUntil} from 'rxjs/operators';
-import {ngbRunTransition, NgbTransitionOptions} from '../util/transition/ngbTransition';
+import {ngbCompleteTransition, ngbRunTransition, NgbTransitionOptions} from '../util/transition/ngbTransition';
 import {
   ngbCarouselTransitionIn,
   ngbCarouselTransitionOut,
@@ -280,7 +280,25 @@ export class NgbCarousel implements AfterContentChecked,
       });
     }
 
-    this.slides.changes.pipe(takeUntil(this._destroy$)).subscribe(() => this._cd.markForCheck());
+    this.slides.changes.pipe(takeUntil(this._destroy$)).subscribe(() => {
+      this._transitionIds ?.forEach(id => ngbCompleteTransition(this._getSlideElement(id)));
+      this._transitionIds = null;
+
+      this._cd.markForCheck();
+
+      // The following code need to be done asynchronously, after the dom becomes stable,
+      // otherwise all changes will be undone.
+      this._ngZone.onStable.pipe(take(1)).subscribe(() => {
+        for (const { id } of this.slides) {
+          const element = this._getSlideElement(id);
+          if (id === this.activeId) {
+            element.classList.add('active');
+          } else {
+            element.classList.remove('active');
+          }
+        }
+      });
+    });
   }
 
   ngAfterContentChecked() {
--- a/src/util/transition/ngbTransition.spec.ts
+++ b/src/util/transition/ngbTransition.spec.ts
@@ -233,7 +233,7 @@ if (isBrowserVisible('ngbRunTransition')) {
       expect(completeSpy1).toHaveBeenCalled();
       expect(element.classList.contains('ngb-test-during')).toBe(false);
       expect(element.classList.contains('ngb-test-after')).toBe(true);
-      expect(window.getComputedStyle(element).opacity).toBe('0');
+      expectOpacity(element, '0');
     });
 
     it(`should create and allow modifying context when running a new transition`, (done) => {
--- a/src/util/transition/ngbTransition.spec.ts
+++ b/src/util/transition/ngbTransition.spec.ts
@@ -1,4 +1,4 @@
-import {ngbRunTransition, NgbTransitionStartFn} from './ngbTransition';
+import {ngbCompleteTransition, ngbRunTransition, NgbTransitionStartFn} from './ngbTransition';
 import createSpy = jasmine.createSpy;
 import {Component, ElementRef, ViewChild} from '@angular/core';
 import {ComponentFixture, TestBed} from '@angular/core/testing';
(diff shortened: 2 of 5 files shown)


src/carousel/carousel.ts
```diff
@@ -274,7 +274,15 @@ export class NgbCarousel implements AfterContentChecked,
       });
     }
 
-    this.slides.changes.pipe(takeUntil(this._destroy$)).subscribe(() => this._cd.markForCheck());
+    this.slides.changes.pipe(takeUntil(this._destroy$)).subscribe(() => {
+      this._cd.detectChanges();
+      if (this.activeId) {
```
maxokorokov, line 279: What's with `'0'` id ?
Also do we care about this `if (this.activeId)` at all ?

maxokorokov (commented):
Hmm, actually we already have this issue here → https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/carousel/carousel.ts#L293

Also it would be nice to come up with a test for this

fbasso:
Updated the last version, with a more complete and  better approach I think:

  - Any running transitions are completed directly, running the endFn in order to set back the right classes on the elements,
  - On the next tick, the active classes are toggled depending on the activeId, to ensure that the active class is on the element, and only one active class,
  - trackBy has been added on the ngFor, as it's better to rely on this id instead of the content.

The code needs to be discussed of course (not sure if it's the best code to finish the transition for example), and a test still need to be added at the moment.

fbasso:
Last push has been done with two separated commits:

  - one for the new ngbCompleteTransition to force the transition end without delay,
  - one for the fix

I finally removed the trackBy (compared to my previous push), as it causes more issue: object comparison is better here.

It's ready for a review.

src/carousel/carousel.ts
```diff
@@ -280,7 +280,30 @@ export class NgbCarousel implements AfterContentChecked,
       });
     }
 
-    this.slides.changes.pipe(takeUntil(this._destroy$)).subscribe(() => this._cd.markForCheck());
+    this.slides.changes.pipe(takeUntil(this._destroy$)).subscribe(() => {
+      const transitionIds = this._transitionIds;
```
maxokorokov, line 284: nitpick: why have `const transitionIds` ?

Shorter version:
```ts
this._transitionIds?.forEach(id => ngbCompleteTransition(this._getSlideElement(id)));
this._transitionIds = null;
```

src/carousel/carousel.ts
```diff
@@ -280,7 +280,30 @@ export class NgbCarousel implements AfterContentChecked,
...
+        for (const id of transitionIds) {
+          ngbCompleteTransition(this._getSlideElement(id));
+        }
+        this._transitionIds = null;
+      }
+
+      this._cd.markForCheck();
+
+      this._ngZone.onStable.pipe(take(1)).subscribe(() => {
+        const activeId = this.activeId;
```
maxokorokov, line 295: nitpick: what's the point of having `const activeId` and `const id`?

Shorter version:
```ts
for (const {id} of this.slides) {
  const element = this._getSlideElement(id);
  if (id === this.activeId) {
    element.classList.add('active');
  } else {
    element.classList.remove('active');
  }
}
```

src/util/transition/ngbTransition.ts
```diff
@@ -15,6 +15,7 @@ export interface NgbTransitionOptions<T> {
 
 export interface NgbTransitionCtx<T> {
   transition$: Subject<any>;
+  stop: () => void;
```
maxokorokov, line 18: I think I would prefer this function to be called `complete: () => {}`, because it properly completes the transition...

WDYT?

src/carousel/carousel.ts
```diff
@@ -280,7 +280,30 @@ export class NgbCarousel implements AfterContentChecked,
...
+      if (transitionIds) {
+        for (const id of transitionIds) {
+          ngbCompleteTransition(this._getSlideElement(id));
+        }
+        this._transitionIds = null;
+      }
+
+      this._cd.markForCheck();
+
+      this._ngZone.onStable.pipe(take(1)).subscribe(() => {
```
maxokorokov, line 294: A comment about why we need `onStable` would be helpful. I don't remember already :)

src/util/transition/ngbTransition.spec.ts
```diff
@@ -203,6 +203,39 @@ if (isBrowserVisible('ngbRunTransition')) {
...
+      expect(element.classList.contains('ngb-test-after')).toBe(false);
+      expect(window.getComputedStyle(element).opacity).toBe('1');
+
+      ngbCompleteTransition(element);
+
+      expect(nextSpy1).toHaveBeenCalled();
+      expect(completeSpy1).toHaveBeenCalled();
+      expect(element.classList.contains('ngb-test-during')).toBe(false);
+      expect(element.classList.contains('ngb-test-after')).toBe(true);
+      expectOpacity(element, '0');
```
maxokorokov, line 236: We should change:

```css
.ngb-test-after {
  opacity: 0;
  transition: none;
}
```

For this to stop failing in FF

maxokorokov (changes requested):
Hey, @fbasso !

LGTM, apart from failing test and a couple of nitpicks :)

src/carousel/carousel.ts
fbasso: Yes, better version indeed ! I'm not used enough with the optional chaining operator.

src/util/transition/ngbTransition.ts
fbasso: Yes, I prefer it too.

maxokorokov (approved):
LGTM!
Please merge if CI is green.

PR #3936, opened 2021-01-11
fix(nav): avoid doing double reflow for fade in transition

### LINKED ISSUE #3900: switching nav scrolls to the top of the page
In a page where there is some text at the top, not filing the whole viewport height, and a nav at the bottom, large enough for the page to scroll, if you scroll down, then switch to another tab, the browser scrolls back to the top of the page.
This happens with or without animation, and with or without destroyOnHide. 

bug-nav

https://stackblitz.com/edit/angular-hyx6po?file=src%2Findex.html

Angular: 10 / 11

ng-bootstrap: 8.0.0

Bootstrap: 4.5.3

maxokorokov:
Double reflow issue caused styles + layout calculation with tab pane hidden (without `.active`).
This caused window scroll positition adjustment as if tab pane wasn't there.

```
Before:
- reflow() in ngbRunTransition
- add('.active') in ngbNavFadeInTransition
- reflow() in ngbNavFadeInTransition
- add('.show') in ngbNavFadeInTransition

After:
- add('.active') in NgbNavOutlet
- reflow() in ngbRunTransition
- add('.show') in ngbNavFadeInTransition

```

P.S. No idea how to add tests for this...

Fixes #3900

### DIFF
--- a/src/nav/nav-outlet.ts
+++ b/src/nav/nav-outlet.ts
@@ -12,7 +12,7 @@ import {
 } from '@angular/core';
 import {distinctUntilChanged, skip, startWith, takeUntil} from 'rxjs/operators';
 
-import {ngbNavFadeInNoReflowTransition, ngbNavFadeInTransition, ngbNavFadeOutTransition} from './nav-transition';
+import {ngbNavFadeInTransition, ngbNavFadeOutTransition} from './nav-transition';
 import {ngbRunTransition, NgbTransitionOptions} from '../util/transition/ngbTransition';
 import {NgbNav, NgbNavItem} from './nav';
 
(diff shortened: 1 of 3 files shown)


fbasso:
Thanks for the fix, @maxokorokov

The code is fine for me. 

Regarding the tests, I don't know if it's feasible easily, but an idea would be to reproduce the issue scenario:

  - Include the navs in a container with an overflow container with a scrollbar,
  - Force programmatically the scroll,
  - Change the nav
  - Check the scroll position

Another idea (but I never used this feature before, so I don't know if it will work or not for the test), would be to use the [MutationObserver ](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) on the navs and check that the active class is maintained all each reflow (See for example : https://javascriptinfo.com/view/156199/listening-browser-reflow-event).

maxokorokov:
Hey, @fbasso!

Thanks for the review. 
Actually added a pretty simple test → https://github.com/ng-bootstrap/ng-bootstrap/pull/3936/files#diff-b3fabf31e70b94a2c4a491cb8d1f0c82263a1e5f4615405b34a01e7199241828R688

fbasso (approved):
Thanks for the test @maxokorokov. 

LGTM !

PR #3939, opened 2021-01-13
fix(animations): delegate 'animation' flag to 'NgbConfig'

### LINKED ISSUE #3893: animation delegation in configs is fragile
I'm not absolutely sure if anything deserves to be changed, but anyway...

### Bug description:

In my tests, I import a TestingModule which intends to disable animations globally, the following way:

```typescript
class TestingModule {
  constructor(config: NgbConfig) {
    config.animation = false;
  }
}
```

And I have an actual module configuring the modals using something like

```typescript
class SharedModule {
  constructor(modalConfig: NgbModalConfig) {
    modalConfig.backdrop = 'static';
  }
}
```

The problem is that if Angular decides to instantiate the shared module first, then the NgbModalConfig is created *before* the NgbConfig has been customized, and thus *copies* the `animation` flag when it's still true. Disabling the animations globally in TestingModule thus doesn't have any effect for modals.

The workaround is simple enough: also inject NgbModalConfig in the testing module, and disable its animation flag too. But it's still a bit unintuitive to do that, and finding out why tests fail and why animations are not disabled is not trivial.

I was wondering if a better solution would not be to actually delegate to NgbConfig instead of copying its animation flag value at construction. Something like:

```typescript
export class NgbModalConfig implements Required<NgbModalOptions> {
  private _animation: boolean | undefined;

  constructor(private ngbConfig: NgbConfig) {}

  get animation(): boolean {
    return (this._animation === undefined) ? this.ngbConfig.animation : this._animation;
  }
  set animation(animation: boolean) {
    this._animation = animation;
  }

  // ...
}
```

Angular: 10.2

ng-bootstrap: 8.0.0

maxokorokov:
Component configs now delegate the `animation` flag to `NgbConfig` if it is not set explicitly.

Fixes #3893

### DIFF
--- a/demo/src/app/app.routing.ts
+++ b/demo/src/app/app.routing.ts
@@ -64,4 +64,4 @@ const routes: Routes = [
 ];
 
 export const routing: ModuleWithProviders<RouterModule> =
-    RouterModule.forRoot(routes, {enableTracing: false, useHash: true, scrollPositionRestoration: 'enabled'});
+    RouterModule.forRoot(routes, { enableTracing: false, useHash: true, scrollPositionRestoration: 'enabled' });
--- a/src/accordion/accordion-config.ts
+++ b/src/accordion/accordion-config.ts
@@ -9,9 +9,13 @@ import {NgbConfig} from '../ngb-config';
  */
 @Injectable({providedIn: 'root'})
 export class NgbAccordionConfig {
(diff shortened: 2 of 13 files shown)


fbasso (approved):
LGTM, as I have no other idea on how to resolve this.animation differently !

Issue #3778, opened 2020-06-18
Typeahead z-index problem when used inside Modal


hughanderson4:
### Bug description:

When using `container="body"` on Typeahead, and placing Typeahead inside Modal, I am seeing the Typeahead's dropdpown appear behind the Modal in the z-index order. (Typeahead dropdown z-index is 1000 while Modal is 1050).

Please see the attached Stackblitz.

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-7dfepa

1. Open the modal by clicking "Launch Demo Modal"
2. Interact with the first Typeahead (type "te") - notice it works fine
3. Interact with the second Typeahead (type "te") - noticed there is no dropdown visible on the screen.  Using Chrome Element Inspector we can see that the `<ngb-typeahead-window>` appended to body has a class of `.dropdown-menu` setting the `z-index` of 1000 (below the modal of 1050).
4. Push down arrow key of keyboard, then enter key of keyboard, to select invisible 1st option of Typeahead.
5. Interact with the Dropdown, notice it works correctly, this also is using `container="body"`, I see a dynamically added z-index on the element style, presumably by ngb code.

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 9.1.0

ng-bootstrap: 6.1.0

Bootstrap: 4.3.0

### Thanks

Thank you team, we ❤️ ng-bootstrap lib!

hughanderson4:
As a temporary workaround, I am applying the following CSS at the global level
```css
ngb-modal-window + ngb-typeahead-window.dropdown-menu {
    z-index: 1051;
}
```

benouat:
Thanks to report this issue.
Nonetheless, I am afraid we can't afford a fully dynamic `z-index` engine whenever we are using `container: body`. It's a kind of pandora box, with a never ending story outcome...

Your workaround seems fine to me for now.
Thus, I'll close the issue as we probably won't fix anything here. Sorry

Jacobra93:
This workaround is not working for me, is there a different one or are y'all going to address this issue in the future? It seems pretty important to be able to use this inside a modal.

dexluce:
> This workaround is not working for me, is there a different one or are y'all going to address this issue in the future? It seems pretty important to be able to use this inside a modal.

Hello @Jacobra93 
[...]

PR #3947, opened 2021-01-15
feat(typeahead): allow custom class on window for modal issue

eliprand:
This PR enables providing one or more custom classes to the typeahead window. It's intended to help solve #3778 as per [my comment](https://github.com/ng-bootstrap/ng-bootstrap/issues/3778#issuecomment-759833406).

I modeled the behavior and implementation after `windowClass` from `NgbModal`.The main intent here is allowing users to provide a custom `z-index` value along with the `container="body"` option for their specific needs. But ultimately, it can be used to override any attributes on the `ngb-typeahead-window`.

I did not feel we needed to update the demo, but I would be happy to do so if needed.

### DIFF
--- a/src/typeahead/typeahead-window.ts
+++ b/src/typeahead/typeahead-window.ts
@@ -21,7 +21,12 @@ export interface ResultTemplateContext {
   selector: 'ngb-typeahead-window',
   exportAs: 'ngbTypeaheadWindow',
   encapsulation: ViewEncapsulation.None,
-  host: {'(mousedown)': '$event.preventDefault()', 'class': 'dropdown-menu show', 'role': 'listbox', '[id]': 'id'},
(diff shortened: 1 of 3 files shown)


maxokorokov (approved):
LGTM, thanks!

eliprand:
@maxokorokov thanks for merging. FYI, can I/we update the inline docs for the `typeahead.ts` changes I made? I was optimistic and say `@since 8.1.0` but it looks like we're going out with `9.1.0`. Can I update this code here? or is it too late?

maxokorokov:
@eliprand we're usually doing it before the feature release, don't worry
I've also renamed `windowClass` to `popupClass` later, because publicly (in docs) everywhere we're referring to it as typeahead popup

PR #3959, opened 2021-01-20
fix(modal): missing entry animation in prod mode

### LINKED ISSUE #3952: Modal's entry animations are missing in Angular production build
### Bug description:

I noticed that the modals are missing their entry animations (the downwards movement) when running the production build of an Angular 11 application:

animation-missing

The animation works correctly when running the development server:

animation-works

Also, the exit animation (upwards movement) works in both cases.

I can only reproduce the problem with Angular's production build, which StackBlitz doesn't support as far as I know. But I created a minimal example repository that contains ng-bootstrap's modal demo page, and where the problem can be reproduced: https://github.com/peruukki/ng-bootstrap-modal-example.

Angular: 11.0.9

ng-bootstrap: 9.0.0

Bootstrap: 4.5.3

fbasso:
fix #3952 

I can't explain why there is this difference between the prod and non prod mode, but adding a onStable during the ngAfterViewInit life cycle fixes the issue.

### DIFF
--- a/src/modal/modal-window.ts
+++ b/src/modal/modal-window.ts
@@ -71,7 +71,9 @@ export class NgbModalWindow implements OnInit,
 
   ngOnInit() { this._elWithFocus = this._document.activeElement; }
 
-  ngAfterViewInit() { this._show(); }
+  ngAfterViewInit() {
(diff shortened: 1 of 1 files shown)


maxokorokov:
This should be fixed by #3962

PR #3962, opened 2021-01-20
fix(animations): don't reflow in 'ngbRunTransition'

### LINKED ISSUE #3952: Modal's entry animations are missing in Angular production build

### LINKED ISSUE #3954: Performance issues with ngbRunTransition within ngbCollapse
### Bug description:
Actually we are experiencing significant performance impacts by your collapse module.
We have on site with many cards in it which are collapseable by your directive:

```html
<div class="card {{bgClass}} {{borderClass}} {{ accentClass }} {{cardClass}} mb-0 overflow-hidden w-100"
  *ngIf="showBody || showList" [style.height]="height">
  <!-- Header -->
  <div
    class="card-header validation d-flex flex-row justify-content-between {{bgClass}} {{borderClass}} {{headerBorderClass}} {{ headerClickToCollapse ? 'cursor-pointer' : '' }}"
    *ngIf="showHeader" (click)="headerClickToCollapse && toggleCollapseState($event)"
    [class.border-bottom-0]="isCollapsed">
    {{ headerText }}
    <div #contentWrapperHeader class="w-100">
      <ng-content select="[card-header]"></ng-content>
    </div>
    <ul class="nav nav-{{navType}} card-header-{{navType}}" *ngIf="navigationItems">
      <li class="nav-item" *ngFor="let item of listGroupItems">
        <a class="nav-link {{item.state}}" href="{{item.link}}">{{item.text}}</a>
      </li>
    </ul>
    <div class="card-header-actions">
      <ng-content select="[card-header-actions]"></ng-content>
      <a class="card-header-action" *ngIf="collapsable" (click)="toggleCollapseState($event)" aria-expanded="true">
        <vng-controls-icon *ngIf="isCollapsed" icon="chevron-down"></vng-controls-icon>
        <vng-controls-icon *ngIf="!isCollapsed" icon="chevron-up"></vng-controls-icon>
      </a>

    </div>
  </div>

  <!-- Body -->
  <div class="card-body animated {{customClassBody}} p-3" id="collapseBasic" [class.border-top-0]="isCollapsed"
    [(ngbCollapse)]="isCollapsed && collapsable" *ngIf="showBody" [animation]="false">
    <span #contentWrapperBody>
      <ng-content select="[card-body]"></ng-content>
    </span>
  </div>
</div>
```

Because we've noticed some serious issues on performance we've investigated the overall performance via chrome performance monitor.

We can clearly see the transitions are significantly slower than other things.
Since version 8.0.0 there is an animation for the components. These animations are already turned off but the problem persists:

image

Why are these transitions executed when we've deactivated them? How can we tunr them completely off?

I've already tried to reproduce that issue within stackblitz but I didn't managed to reprocude this issue.

Please note that we can _not_ act on bug reports without a _minimal_ reproduction scenario in a StackBlitz. Here is why:

Angular: 11.0.5

ng-bootstrap: 8.0.1

Bootstrap: 4.5.3

We have attached the per profile

[ngb-collapse-perf-issues.zip](https://github.com/ng-bootstrap/ng-bootstrap/files/5828774/ngb-collapse-perf-issues.zip)

maxokorokov:
Replace a generic `reflow()` call in `ngbRunTransition`:
- introduce granular per-component reflows only when necessary
- take into account `animation` flag for reflows

Fixes #3954
Fixes #3952

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -348,11 +348,9 @@ export class NgbAccordion implements AfterContentChecked {
         if (panelElement) {
           if (!panel.initClassDone) {
             panel.initClassDone = true;
-            const {classList} = panelElement;
-            classList.add('collapse');
-            if (panel.isOpen) {
-              classList.add('show');
-            }
+            ngbRunTransition(
+                this._ngZone, panelElement, ngbCollapsingTransition,
+                {animation: false, runningTransition: 'stop', context: {direction: panel.isOpen ? 'show' : 'hide'}});
           }
         } else {
           // Classes must be initialized next time it will be in the dom
@@ -397,7 +395,7 @@ export class NgbAccordion implements AfterContentChecked {
     this.activeIds = this.panels.filter(panel => panel.isOpen && !panel.disabled).map(panel => panel.id);
   }
 
-  private _runTransitions(animation: boolean, emitEvent = true) {
+  private _runTransitions(animation: boolean) {
     // detectChanges is performed to ensure that all panels are in the dom (via transitionRunning = true)
     // before starting the animation
     this._changeDetector.detectChanges();
@@ -413,15 +411,13 @@ export class NgbAccordion implements AfterContentChecked {
           context: {direction: panel.isOpen ? 'show' : 'hide'}
         }).subscribe(() => {
           panel.transitionRunning = false;
-          if (emitEvent) {
-            const {id} = panel;
-            if (panel.isOpen) {
-              panel.shown.emit();
-              this.shown.emit(id);
-            } else {
-              panel.hidden.emit();
-              this.hidden.emit(id);
-            }
+          const {id} = panel;
+          if (panel.isOpen) {
+            panel.shown.emit();
+            this.shown.emit(id);
+          } else {
+            panel.hidden.emit();
+            this.hidden.emit(id);
           }
         });
       }
--- a/src/collapse/collapse.ts
+++ b/src/collapse/collapse.ts
@@ -54,14 +54,11 @@ export class NgbCollapse implements OnInit, OnChanges {
     this.animation = config.animation;
   }
 
-  ngOnInit() {
-    this._element.nativeElement.classList.add('collapse');
-    this._runTransition(this.collapsed, false, false);
-  }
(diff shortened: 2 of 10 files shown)


src/accordion/accordion.ts
```diff
@@ -397,7 +395,7 @@ export class NgbAccordion implements AfterContentChecked {
     this.activeIds = this.panels.filter(panel => panel.isOpen && !panel.disabled).map(panel => panel.id);
   }
 
-  private _runTransitions(animation: boolean, emitEvent = true) {
+  private _runTransitions(animation: boolean) {
```
maxokorokov, line 398: note: removing this, because it is never used

src/collapse/collapse.ts
```diff
@@ -54,14 +54,11 @@ export class NgbCollapse implements OnInit, OnChanges {
     this.animation = config.animation;
   }
 
-  ngOnInit() {
-    this._element.nativeElement.classList.add('collapse');
-    this._runTransition(this.collapsed, false, false);
-  }
+  ngOnInit() { this._runTransition(this.collapsed, false).subscribe(); }
```
fbasso, line 57: I don't understand why we need the subscribe() here.

To be removed if there is no reason, or add a comment if there is one,.

maxokorokov: Have just removed it, thanks!

fbasso (approved):
Thanks for the PR @maxokorokov, LGTM !

Besides fising the issues, I feel the code much better this way, with a more fine grain control on waht to do for each animation, with or without animation enabled.

PR #3968, opened 2021-01-22
test: stabilize e2e navigation with Playwright

fbasso:
Added some waitForSelector to stabilize the navigation

### DIFF
--- a/e2e-app/src/app/tools.pw-po.ts
+++ b/e2e-app/src/app/tools.pw-po.ts
@@ -94,12 +94,16 @@ export const expectFocused = async(selector, message) => {
  */
 let hasBeenLoaded = false;
 export const openUrl = async(url: string) => {
+  const currentPage = page();
   if (hasBeenLoaded) {
-    await page().click(`#navigate-home`);
-    await page().click(`#navigate-${url.replace('/', '-')}`);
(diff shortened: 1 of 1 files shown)


e2e-app/src/app/tools.pw-po.ts
```diff
@@ -94,12 +94,16 @@ export const expectFocused = async(selector, message) => {
  */
 let hasBeenLoaded = false;
 export const openUrl = async(url: string) => {
+  const p = page();
```
maxokorokov, line 97: Oh, I actually don't like this at all :D

maxokorokov (commented):
I don't understand the fix, to be honest...

Specifically why we're waiting for `ng-component` in detached state...

fbasso:
It is not a fix of an open issue, but I think it's better to be sure that the navigation produce the right effect before continuing more actions.

  - waitForSelector for 'app-root' after the goto : ensure that the page is well loaded,
  - waitForSelector for 'ng-component', with the detached : be sure that the previous test is unloaded (the menu page is ready then)
  - waitForSelector for 'ng-component' : be sure that the component page is ready.

maxokorokov (approved):
LGTM, please merge when green!

PR #3975, opened 2021-01-26
fix(rating): avoid inconsistent disabling

FFKL:
Hello! I've noticed that rating component has inconsistent disabling feature. It wasn't fully disabled. You can use tab for focusing readonly element. Also I reworked `disabled` property.

### DIFF
--- a/src/rating/rating.spec.ts
+++ b/src/rating/rating.spec.ts
@@ -439,31 +439,84 @@ describe('ngb-rating', () => {
 
     it('updates aria-disabled when readonly', () => {
       const fixture = createTestComponent('<ngb-rating></ngb-rating>');
-      let ratingEl = fixture.debugElement.query(By.directive(NgbRating));
+      const ratingEl = fixture.debugElement.query(By.directive(NgbRating));
+      const ratingComp = <NgbRating>ratingEl.componentInstance;
+
       fixture.detectChanges();
       expect(ratingEl.attributes['aria-disabled'] == null).toBeTruthy();
 
-      let ratingComp = <NgbRating>ratingEl.componentInstance;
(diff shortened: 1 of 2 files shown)


FFKL:
@maxokorokov Hello! Can you review my PR?

maxokorokov (commented):
Hey, @FFKL!

Thank you for the PR, however I don't quite understand what you're trying to achieve.

`Readonly` and `Disabled` are two distinct states:
1. `Disabled` → actually means disabled from the form point of view. Not focusable, not accesible to screen readers, not modifiable. Ex. like `<input disabled>`
2. `Readonly` → can de focused, the value can be read by screen readers, but the value can't be changed. Ex. like `<input aria-disabled="true">`

So I don't think what you're suggesting here is correct behavior.

P.S. We should probably update the `readonly` docs to be more explicit about it.

WDYT?

Cheers,
Max

FFKL:
@maxokorokov Hi! Thanks for answering;) I get your point, but
1. `[attr.aria-disabled]': 'readonly ? true : null`. I think, It's strange behaviour. What about `aria-readonly` attribute? As I understand - `aria-disabled` is used here to replace the native `disabled` for custom input.
2. I can disable element focus only if I use `FormControl` but It can be necessary to disable element without `FormControl`.
I tried to fix these problems in my PR. Maybe I should improve it. What do you think?

maxokorokov:
Will just drop a stackblitz with current behavior here → https://stackblitz.com/edit/angular-uhnyui-7lmv6p?file=src%2Fapp%2Frating-form.html

FFKL:
@maxokorokov I don't understand you;)

maxokorokov:
Fixed in #4603

PR #3993, opened 2021-02-02
test(datepicker): migrate e2e tests to Playwright

maxokorokov:


e2e-app/src/app/datepicker/focus/datepicker-focus.pw-spec.ts
```diff
@@ -0,0 +1,356 @@
...
+    await preSelectDate();  // 10 AUG 2018
+    await openDatepicker();
+
+    await expectFocused(SELECTOR_TOGGLE, `Toggling element should stay focused`);
+
+    const dayElement = SELECTOR_DAY(new Date(2018, 7, 10));
+    await test.page.waitForSelector(dayElement);
+    let message = '';
+    try {
+      await test.page.click(dayElement, {timeout: 100});
```
fbasso, line 227: On window, webkit is much slower and is not enough to get the message `checking that element receives pointer events` (it fails with element still moving).

I tested it successfully on my laptop with 200, but I wonder it can introduce a flaky test, depending on the machine where it runs.

fbasso: I tried to use [waitForElementState](https://playwright.dev/docs/api/class-elementhandle?_highlight=waitforelementstate#elementhandlewaitforelementstatestate-options) before the click (to be sure that the element is not moving), but it doesn't help.

I guess it's something 'internally' which is slower, not giving the right message, so setting the timeout to 200 must be a good solution.

maxokorokov: ok

PR #4000, opened 2021-02-09
Playwright waitForSelector refactoring

maxokorokov:
- removed redundant checks around `waitForSelector`
- added waiting for explicit conditions before each test, ex `await test.page.waitForSelector('h3:text("Datepicker autoclose")');` to ensure the page is actually loaded


fbasso (commented):
LGTM !

PR #4005, opened 2021-02-12
test: waitForFocus instead of expectFocused

fbasso:
The purpose is to render the tests more stable with all the browsers

### DIFF
--- a/e2e-app/src/app/typeahead/focus/typeahead-focus.pw-spec.ts
+++ b/e2e-app/src/app/typeahead/focus/typeahead-focus.pw-spec.ts
@@ -1,14 +1,14 @@
-import {expectFocused, Key, openUrl, sendKey} from '../../tools.pw-po';
+import {waitForFocus, Key, openUrl, sendKey} from '../../tools.pw-po';
 import {test} from '../../../../playwright.conf';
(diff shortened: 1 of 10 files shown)


maxokorokov (approved):
LGTM!

e2e-app/src/app/typeahead/focus/typeahead-focus.pw-spec.ts
```diff
@@ -69,6 +74,7 @@ describe('Typeahead', () => {
     if (process.env.BROWSER !== 'webkit') {
       it(`should select element on tab`, async() => {
         await test.page.focus(SELECTOR_TYPEAHEAD);
+        await waitForFocus(SELECTOR_TYPEAHEAD);
```
divdavem, line 77: As far as I could test (and as discussed on Wednesday), I think this extra line does not only make this test more reliable on Chromium and Firefox but also makes it pass on `webkit` (which is currently excluded).

fbasso: Yes, I did a helper function in a previous try. Maybe I should re-integrate it, as if the focus('something') is not reliable, it can leads to flaky tests.

PR #4016, opened 2021-02-22
test: enable console log checks

fbasso:


e2e-app/setup.pw-spec.ts
```diff
@@ -11,6 +11,23 @@ beforeAll(async() => {
   } catch (e) {
     console.error('Unable to setup a new page with playwright', e);
   }
+
+  test.page.on('console', async msg => {
+    let hesUnexpectedMessage = true;
+    const consoleMsgs: Array<string> = ['Unexpected console messages found'];
+    for (let i = 0; i < msg.args().length; ++i) {
+      const message: string = await msg.args()[i].jsonValue();
+      if (message.includes('Angular is running') || message.includes('Live Reloading')) {
```
maxokorokov, line 20: We should simply turn off live reloading for tests, like it is done for the demo for example:
- run demo with a specific config for playwright `yarn e2e-app:serve -c playwright`
- similar to https://github.com/ng-bootstrap/ng-bootstrap/blob/master/angular.json#L155-L158

I guess it should be something like:

```
"playwright": {
  "browserTarget": "demo:build:production",
  "liveReload": false,
  "watch": false,
  "extractLicenses": false
}
```

maxokorokov (changes requested):
Hey, @fbasso,

- I think we should turn off all reporting that causes issues instead of filtering it out in the code
- Also maybe we should look only for warnings and errors

e2e-app/setup.pw-spec.ts
```diff
@@ -11,6 +11,23 @@ beforeAll(async() => {
   } catch (e) {
     console.error('Unable to setup a new page with playwright', e);
   }
+
+  test.page.on('console', async msg => {
+    let hesUnexpectedMessage = true;
+    const consoleMsgs: Array<string> = ['Unexpected console messages found'];
```
maxokorokov, line 17: I think `Array<string>` is redundant here

package.json
```diff
@@ -32,7 +32,7 @@
     "ngb:lint": "ng lint ng-bootstrap && yarn schematics:lint",
     "ngb:test": "ng test ng-bootstrap --code-coverage --source-map true --progress false --watch false",
     "ngb:tdd": "ng test ng-bootstrap --source-map false",
-    "ngb:e2e": "concurrently --success first --kill-others \"yarn e2e-app:serve --prod --no-watch\" \"yarn ngb:e2e-playwright BROWSER=chromium && yarn ngb:e2e-playwright BROWSER=firefox && yarn ngb:e2e-playwright BROWSER=webkit\"",
+    "ngb:e2e": "concurrently --success first --kill-others \"yarn e2e-app:serve --prod --no-watch -c playwright\" \"yarn ngb:e2e-playwright BROWSER=chromium && yarn ngb:e2e-playwright BROWSER=firefox && yarn ngb:e2e-playwright BROWSER=webkit\"",
```
maxokorokov, line 35: I think you can remove `--prod` and `--no-watch` and add 
`"browserTarget": "e2e-app:build:production"` in the `angular.json` configuration

fbasso:
Thanks for the review @maxokorokov !

I did the changes. I also listen for the 'pageerror' event (uncaught exceptions), which normally don't happen (every thing seems to be caught at the higher level) , but just in case, it good to report too.

e2e-app/setup.e2e-spec.ts
```diff
@@ -11,6 +11,23 @@ beforeAll(async() => {
   } catch (e) {
     console.error('Unable to setup a new page with playwright', e);
   }
+
+  // Listen for all console events and handle errors
+  test.page.on('console', async msg => {
```
maxokorokov, line 16: type would be helpfull→ `(msg: ConsoleMessage)`

e2e-app/setup.e2e-spec.ts
```diff
@@ -11,6 +11,23 @@ beforeAll(async() => {
...
+    const type = msg.type();
+    if (type === 'error' || type === 'warning') {
+      const output = ['Unexpected console error:'];
+      const length = msg.args().length;
```
maxokorokov, line 20: maybe 

```ts
for (let m of msg.args()) {
  output.push(await m.jsonValue());
}
```

angular.json
```diff
@@ -225,6 +225,10 @@
           "configurations": {
             "production": {
               "browserTarget": "e2e-app:build:production"
+            },
+            "playwright": {
+              "liveReload": false,
```
maxokorokov, line 230: what about the flag about extracting licenses?

maxokorokov (commented):
Hey, @fbasso !

Left some comments

maxokorokov (approved):
LGTM!

PR #4018, opened 2021-02-23
test: remove Protractor from the project

maxokorokov:
I've also renamed targets to match `demo`, `ssr` and the way it was done before:

```
// now
yarn ngb:e2e → for all ng-bootstrap e2e tests
yarn ngb:e2e-playwright BROWSER=XXX → to run playwright in a browser XXX

// before
yarn e2e:playwright
yarn playwright
yarn playwright:firefox
yarn playwright:webkit
```


e2e-app/tsconfig.json
```diff
@@ -6,23 +6,13 @@
...
     "src/main.ts",
     "src/polyfills.ts"
   ],
-  "include": [
-    "src/**/*.d.ts"
-  ],
-  "exclude": [
-    "src/**/*.po.ts",
-    "src/**/*.e2e-spec.ts"
-  ],
```
fbasso, line 25: These changes are fine, but there other tsconfig where po.ts and e2e-spec.ts are referenced. Should we remove them too ?

fbasso (commented):
The 'po' has been not used (compared to protractor).  `e2e-app\src\app\tools.pw-po.ts` is the only file usig this extension. Is it the right PR to change its name (tools.ts ? tools.pw.ts ?)

maxokorokov:
- renamed test files to `*.e2e-spec.ts` and `*.po.ts`
- reverted to `tools.po.ts`

PR #4021, opened 2021-02-25
test: script for launching e2e tests

fbasso:
The purpose is to have a more flexible way to launch e2e test suites


playwright.js
```diff
@@ -0,0 +1,60 @@
...
+    .option('debug', {description: 'enable debugging environment', type: 'boolean'})
+    .option('slow', {alias: 's', description: 'enable slow motion', type: 'boolean'})
+    .example('yarn ngb:e2e-playwright -d ssr-app', "Run the ssr-app tests suite")
+    .example('yarn ngb:e2e-playwright -b firefox --headless', "Run playwright with firefox in headless mode")
+    .example('yarn ngb:e2e-playwright --debug', "Run playwright in debug mode")
+    .example('yarn ngb:e2e-playwright --slow', "Run playwright in slow motion mode")
+    .wrap(null);
+var parameters = yargs.parse(process.argv);
+
+async function startSuite() {
```
maxokorokov, line 22: nitpick: just `start()` maybe, why have `suite` here?

maxokorokov (commented):
Hey, @fbasso!

LGTM, just two questions:
- do you think we can remove `wait-on` package now?
- could we put `playwright.js` into the `playwright` folder and make it `*.ts`, not `*.js`?

For the last point, I'm not sure about creating Jasmine instance from TS. I couldn't do it on my side for some reason.

playwright.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-console */
```
maxokorokov, line 1: I don't think we use `eslint`

fbasso:
Thanks for the comments, @maxokorokov.

I did the all the possible changes:

  - Move playwright.js to misc/playwright.js
  - Remove eslint comments

Regarding the wait-on package, it depends what it means. 
  - If the solution is to wait with Playwright, it seems it's not possible, as for Playwright, a wrong response is a response, so it fails.
  - We can do the same kind of mechanism as wait-on before launching the suite, would you prefer to implement it ourselves ?

Regarding the ts instead of js, it was my first intention, but I face strange errors I was not able to solve:

node-ts misc/playwright.ts leads to the following:

```
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
e2e-app/setup.e2e-spec.ts:5:43 - error TS2345: Argument of type 'undefined' is not assignable to parameter of type 'Function'.

5 beforeAll(() => __awaiter(void 0, void 0, void 0, function* () {
                                            ~~~~~~
e2e-app/setup.e2e-spec.ts:16:42 - error TS2345: Argument of type 'undefined' is not assignable to parameter of type 'Function'.

16 afterAll(() => __awaiter(void 0, void 0, void 0, function* () {
```

It seems it's a kind of conflict between the ts-node used first and the one used in the script (I can't remove it neither).

maxokorokov:
@fbasso, could you please rebase/merge this?

PR #4032, opened 2021-03-12
fix(toast): default delay and toasts container removal

fbasso:
Changes related to https://getbootstrap.com/docs/5.0/migration/#toasts-1

- Default delay changed from 500 to 5000
- ngb-toasts container removed, as Bootstrap has one now.

The removal of ngb-toasts is a breaking change, as there is no style for this tag anymore.


src/toast/toast.spec.ts
```diff
@@ -74,12 +74,12 @@ describe('ngb-toast', () => {
       expect(fixture.componentInstance.hide).toHaveBeenCalled();
     });
 
-    it('should emit hide output after default delay (500ms)', fakeAsync(() => {
+    it('should emit hide output after default delay (5000ms)', fakeAsync(() => {
          const fixture = createTestComponent(`<ngb-toast header="header" (hidden)="hide()">body</ngb-toast>`);
-         tick(499);
+         tick(4099);
```
maxokorokov, line 79: 4999? :)

maxokorokov (commented):
LGTM,

However technically removing `.ngb-toasts` is a breaking change.
I think it was meant as a helpful utility if you want to build your own toast service.

But it was documented nowhere.

You don't think it might be helpful?

fbasso:
Yes, about `.ngb-toasts`, as Bootstrap 5 has one, I think this one is redundant and not necessary any more. 
I think we should remove it, even if it's a breaking change.

PR #4076, opened 2021-04-24
fix(tooltip/popover): enable pointer events only for tooltip/popover content

### LINKED ISSUE #3997: Tooltip closes immediately when hovered from tooltip direction
### Bug description:

In our application, tooltips triggered on hover don't open if the mouse enters the triggering element from the direction where the tooltip is placed. This doesn't happen always but most of the time. Hovering from any other direction works fine.

Looking at the DOM, the tooltip is shown but then immediately hidden.

I'm able to reproduce the problem quite well on the [demo site](https://ng-bootstrap.github.io/#/components/tooltip/examples) when using Chrome, but not with Firefox or Safari. In our application it seems to occur in all browsers. Also, a Linux-using colleague reported that they can reproduce the problem more frequently when zooming the browser at e.g. 125%. However, I'm on a Mac and didn't notice any difference related to the zooming level.

It also reproduces in [StackBlitz](https://sk4r8k--run.stackblitz.io/) (I can't get a working editor URL, only an application URL):

https://user-images.githubusercontent.com/1781172/106900971-d586b180-66ff-11eb-879d-2fa78209f9c5.mp4

Angular: 11.0.9

ng-bootstrap: 9.0.2

Bootstrap: 4.5.3

Chrome: 88.0.4324.96

macOS: 11.2

peruukki:
This fixes #3997. I was able to reproduce the problem in the issue manually with tooltips and popovers on the demo site in Chrome, but writing automated tests for it would not be straightforward. It could be possible with a test that simulates moving the mouse slowly on top of a tooltip/popover, but I'm not sure if that's worth the effort.

I did verify that if I completely disable pointer events from these elements (i.e. omit the `"pointer-events: auto;"` definitions), several "Tooltip Autoclose" and "Popover Autoclose" e2e tests fail. So at least the existing tests verify that these changes don't cause regressions in those scenarios.

### DIFF
--- a/src/popover/popover.scss
+++ b/src/popover/popover.scss
@@ -2,6 +2,13 @@
 $arrow-size: 1rem;
 
 ngb-popover-window {
+  pointer-events: none;
+
+  .popover-header,
+  .popover-body {
+    pointer-events: auto;
+  }
+
   &.bs-popover-top > .arrow,
   &.bs-popover-bottom > .arrow {
     left: 50%;
--- a/src/tooltip/tooltip.scss
+++ b/src/tooltip/tooltip.scss
@@ -2,6 +2,12 @@
 $arrow-size: 0.8rem;
 
 ngb-tooltip-window {
+  pointer-events: none;
+
+  .tooltip-inner {
+    pointer-events: auto;
+  }
+
   &.bs-tooltip-top .arrow,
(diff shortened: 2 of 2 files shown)


maxokorokov (approved):
LGTM, thanks!

PR #4106, opened 2021-06-01
fix(tabset): Add role to li

ashutosh1608:
Added role="none" because otherwise, it causes issues in fast pass for Accessibility testing.

Screenshot (236)

image

Screenshot (237)


maxokorokov:
Thanks for the PR, however Tabset was deprecated and is not supported anymore.
It's not a part of the library deliverables since `8.0.0`.

Its replacement `NgbNav` has role generation strategy that might be of use for you (see "customization" and `[roles]` in the API) → https://ng-bootstrap.github.io/#/components/nav/overview#customization

Issue #1107, opened 2016-12-02
NgbAccordion - there is no way to change the default style.


bbehzadi:
It is not possible to change the style of NgbAccordion, NgbPanel, NgbPanelTitle and NgbPanelContent.  e.g. If using 'type' is not desirable, there is no way to change the background color or remove the borders without studying the internals of the implementation and hacking up a solution.  The API should explicitly support changing all the styles for these elements.

pkozlowski-opensource:
@bbehzadi what would be your API proposal for this? We don't want to explode API surface either but if you can propose a reasonable solution we could look into implementing it. A PR with a super-nice implementation would be even better :-)

pkozlowski-opensource:
One thing we should probably do is to change classes used from "solid" to "outline" as using a custom type makes the accordion looks ... well, not very pretty....

bbehzadi:
Using Bootstrap .card, .card-header, .card-block and all other classes, can be extended and even overridden.  But they are all hidden here.

The problem is beyond this particular component. But to answer your question @pkozlowski-opensource, it is desirable to have everything about the component that can be styled or tweaked  exposed by the API - without any fear of 'API surface explosion'.

The Polymer project has done an excellent job on this front.  Every element in their [catalog](https://elements.polymer-project.org/) is documented in well structured subsections which include Styling, Properties, Events, Examples and comprehensive API Reference. And ALL that can be customized is exposed through the API properties and styling with ample examples and animation hints, etc.

pkozlowski-opensource:
@bbehzadi I'm sorry, but you would have to be more specific (that is - exact proposal not just a link to another impl) if you want to push it forward.... 

I've casually looked at https://elements.polymer-project.org/elements/paper-header-panel and honestly I'm not sure what you are proposing. I don't see any specific properties that would control styling. You have styles but so Bootstrap does so you can override Bootstrap classes if you don't like the default theme.

I've got the impression that we don't understand each other that why I'm insisting on _concrete_ examples. Try to prepare a plunk with what you are trying to do and what you believe is "hidden".

bbehzadi:
As an example, to customize ngb-accordion background color, I had to **dig in its source code** and find out that in order to achieve that I had to change the background-color and the border-color of the .card, .card-header and .card-block.  These 3 classes are internal to this component and if in some future version of ngb-accordion some other classes are used or simply these are renamed, I am screwed.

[...]

Issue #3898, opened 2020-11-12
New Accordion in Bootstrap 5


maxokorokov:
There is a new accordion component in the Bootstrap 5.

We should take the opportunity to re-implement the one we have with a set of directives (same as we did with `nav`) for more flexbility and direct DOM access.

Here is the markup from Bootstrap 5:

```html
<div class="accordion">
  <div class="accordion-item">
    <h2 class="accordion-header" id="headingOne">
      <button class="accordion-button" type="button" aria-expanded="true" aria-controls="collapseOne">...</button>
    </h2>
    <div id="collapseOne" class="accordion-collapse collapse show" aria-labelledby="headingOne">
      <div class="accordion-body">...</div>
    </div>
  </div>
</div>
```

That **might** translate into something like:

```html
<div ngbAccordion>
  <div [ngbAccordionItem]="itemOne">
    <h2 ngbAccordionHeader>
      <button ngbAccordionToggle>...</button>
    </h2>
    <div ngbAccordionBody>
      <ng-template>...</ng-template>
    </div>
  </div>
</div>
```

maxokorokov:
Fixed by #4437

PR #4110, opened 2021-06-12
feat(accordion): added inputs to NgbPanel to allow local custom styling

J-Guenther:
Hi,
I recently learned that changing the default style of `NgbAccordion `on a component level is hard to achieve. According to issue #1107 it only seems to be possible when using a global stylesheet or when setting `ViewEncapsulation.None`. I also do not consider `::ng-deep` a desirable solution since it is deprecated according to the Angular [documentation](https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep)

The responses for the issues #1107 and #2564 show that multiple people try to achieve a styling on component level. For example, I personally would like to be able to adjust the padding of the div with the `accordion-body` class. I only want to do this for some of my components that use `NgbAccordion`. I do not want to apply a global theme or set `ViewEncapsulation.None`.

In my opinion there should be a way to achieve this. Therefore I would like to propose a change of the API of the accordion component. The `NgbPanel ` already has an optional input, called `cardClass`, to set a class for styling the card. In Bootstrap 5 terms this would style the “accordion-item” (see [bootstrap accordion example](https://getbootstrap.com/docs/5.0/components/accordion/#example) ). I suggest to add more inputs that provide similar “access” to the `accordion-header`, `accordion-button` and the `accordion-body. ` Those inputs could be of type `{ [klass: string]: any; }` which is the type of the [ngStyle directive](https://angular.io/api/common/NgStyle#properties) and would allow to overwrite properties of those Bootstrap css classes and to add a custom style on component level.

I think it could add to the flexibility of the styling of the Accordion component. It would also change the Accordion component to be closer to how an Accordion would behave when using plain Bootstrap. When using plain Bootstrap it is possible to extend or overwrite the Bootstrap style locally on every single div of an Accordion.

This proposal should not cause breaking changes since it only adds optional functionality. Also this way of allowing to style specific component parts could easily be transferred to other NgBootstrap components to offer this feature for all components.

This is the first time I am trying to contribute to this project. I am not sure if it is a good practice to start with a pull request rather than with opening an issue and discussing this feature proposal first. But this pull request should make clear what and how I was trying to achieve things. Also I am not sure if choosing to work on the bootstrap5 branch is considered good for this purpose.

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -117,6 +117,27 @@ export class NgbPanel implements AfterContentChecked {
   @Input() cardClass: string;
 
   /**
+   * An optional style applied to the accordion button element that represents the button in the accordion header.
+   *
+   *
+   */
+  @Input() buttonStyle: {[klass: string]: any;};
+
+  /**
+   * An optional style applied to the accordion header element that wraps the title.
+   *
+   *
+   */
+  @Input() headerStyle: {[klass: string]: any;};
+
+  /**
+   * An optional style applied to the accordion body element that wraps the content.
+   *
+   *
+   */
+  @Input() bodyStyle: {[klass: string]: any;};
+
+  /**
    * An event emitted when the panel is shown, after the transition. It has no payload.
    *
    * @since 8.0.0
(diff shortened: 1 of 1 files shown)


maxokorokov:
Hey,

thanks for the detailed explanation, however I don't agree that making things more flexible by making API bigger / adding more inputs is a good way to go.

The idea we had for the accordion is to migrate it to implementation based on directives. It seems to have worked well for the `Tabset` (ex. [basic tabset](https://ng-bootstrap.github.io/#/components/tabset/examples#basic)) → `Nav` (ex. [basic nav](https://ng-bootstrap.github.io/#/components/nav/examples#basic)) migration. 

Here is the issue in question → #3898

maxokorokov:
As we're deprecating component-based accordion, I'm closing this. Please see #3898.

PR #4116, opened 2021-07-01
fix: replace the division by math.div for sass

fbasso:
The division is now deprecated with sass-dart, and must be replaced by math.div

### DIFF
--- a/src/popover/popover.scss
+++ b/src/popover/popover.scss
@@ -1,3 +1,4 @@
+@use "sass:math";
 
 $arrow-size: 1rem;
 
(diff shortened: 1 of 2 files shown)


maxokorokov:
It is failing, isn't it?
Screenshot 2021-07-01 at 11 17 53

fbasso:
> It is failing, isn't it?

Hum... Strange, I haven't seen that locally. It's probably the same thing of the warning displayed only once after the first build, even after several consecutive builds.

I'll check it.

lionel-meunier (approved):

lionel-meunier:
the sass@1.32.12 package is installed while the ^1.33.0 package is needed

need update angular to ^12.1

fbasso:
Thanks @lionel-meunier ! I'll try it.

maxokorokov:
not relevant anymore, closing

PR #4129, opened 2021-07-20
feat(positioning): popper 2

fbasso:


src/datepicker/datepicker-input.spec.ts
```diff
@@ -945,57 +944,6 @@ describe('NgbInputDatepicker', () => {
 
   describe('positionTarget', () => {
 
-    let positionElementsSpy: jasmine.Spy;
```
maxokorokov, line 948: Removing this decreases coverage for `positionTarget` tests

src/dropdown/dropdown.spec.ts
```diff
@@ -131,49 +131,52 @@ describe('ngb-dropdown', () => {
...
+       expect(dropdownEl()).toHaveCssClass('my-class');
+       expect(dropdownEl()).toHaveCssClass('dropdown');
 
-    // open
-    dropdown.open();
-    fixture.detectChanges();
+       // open
+       dropdown.open();
+       fixture.detectChanges();
+       tick(25);
```
maxokorokov, line 156: Seems like popper uses promises for scheduling, so 25 is not necessary, just `tick()`

src/dropdown/dropdown.spec.ts
```diff
@@ -131,49 +131,52 @@ describe('ngb-dropdown', () => {
     expect(dropdownEl).toHaveCssClass('dropdown');
   });
 
-  it('should allow setting a custom dropdown class (container="body")', () => {
-    const html = `
+  it('should allow setting a custom dropdown class (container="body")', fakeAsync(() => {
```
maxokorokov, line 134: If I'm not mistaken this specific test actually works fine without any changes

src/dropdown/dropdown.ts
```diff
@@ -63,6 +63,7 @@ export class NgbDropdownItem {
     '[class.dropdown-menu]': 'true',
     '[class.show]': 'dropdown.isOpen()',
     '[attr.x-placement]': 'placement',
```
maxokorokov, line 65: should remove `x-placement` I guess?

src/popover/popover.ts
```diff
@@ -251,6 +244,12 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
       // mark the parent component to be checked.
       this._windowRef.changeDetectorRef.markForCheck();
 
+      this._ngZone.runOutsideAngular(() => {
```
maxokorokov, line 247: I think we should do (as a generic algoritm for all components, not only popover):
- create a new popper on opening. outside angular
- when zone is stable AND popup is opened do `popper.update()`
- when popup is closed release popper resources

We should probably rename `positionELements` to someting like `createPopper` and call it only initially and then use the returned popper instance inside the component to call `.update()` Just an idea.

src/popover/popover.ts
```diff
@@ -269,6 +268,7 @@ export class NgbPopover implements OnInit, OnDestroy, OnChanges {
       this._renderer.removeAttribute(this._elementRef.nativeElement, 'aria-describedby');
       this._popupService.close(this.animation).subscribe(() => {
         this._windowRef = null;
+        this._popperDestroy();
```
maxokorokov, line 271: Yeah, popper is not destroyed for datepicker, dropdown and typeahead

src/util/positioning.ts
```diff
@@ -1,167 +1,49 @@
...
+import {
+  popperGenerator,
+  defaultModifiers,
+} from '@popperjs/core/lib/popper-lite';
+import arrow from '@popperjs/core/lib/modifiers/arrow';
+import flip from '@popperjs/core/lib/modifiers/flip';
+import offset from '@popperjs/core/lib/modifiers/offset';
+import preventOverflow from '@popperjs/core/lib/modifiers/preventOverflow';
+
+const createPopper = popperGenerator({
```
maxokorokov, line 12: I don't think we need the generator here, what I would do is:

```ts
import {createPopperLite, offset, flip, preventOverflow, arrow} from '@popperjs/core';

const popperInstance = createPopperLite(hostElement, targetElement, {
  modifiers: [
    bsModifier,
    offset, flip, preventOverflow, arrow,
etc ...
```

maxokorokov: And we should have imports only from `@popper/core`

src/util/positioning.ts
```diff
@@ -1,167 +1,49 @@
...
-      elOffset.left = Math.round(elOffset.left);
-      elOffset.right = Math.round(elOffset.right);
-    }
-
-    return elOffset;
+const popperStartPrimaryPlacement = /^left/;
+const popperEndPrimaryPlacement = /^right/;
+function convertPlacementToBsClasses(baseClass: string, placement: PopperPlacement): string {
+  const newPlacement =
+      placement.replace(popperStartPrimaryPlacement, 'start').replace(popperEndPrimaryPlacement, 'end') as Placement;
```
maxokorokov, line 35: I would extract this as utiliy function to have 2:

```
convertPlacementToPopper();
convertPlacementToBootstrap();
```

src/util/positioning.ts
```diff
@@ -1,167 +1,49 @@
...
-      elOffset.top = Math.round(elOffset.top);
-      elOffset.bottom = Math.round(elOffset.bottom);
-      elOffset.left = Math.round(elOffset.left);
-      elOffset.right = Math.round(elOffset.right);
-    }
-
-    return elOffset;
+const popperStartPrimaryPlacement = /^left/;
+const popperEndPrimaryPlacement = /^right/;
+function convertPlacementToBsClasses(baseClass: string, placement: PopperPlacement): string {
```
maxokorokov, line 33: maybe something  like `getBootstrapBaseClassPlacement`. I was struggling to understand what is does.

Also unit tests for these kind of utility functions would be nice...

src/util/positioning.ts
```diff
@@ -1,167 +1,49 @@
...
-
-    // Check if the targetElement is inside the viewport
-    const targetElBCR = targetElement.getBoundingClientRect();
-    const html = document.documentElement;
-    const windowHeight = window.innerHeight || html.clientHeight;
-    const windowWidth = window.innerWidth || html.clientWidth;
-
-    return targetElBCR.left >= 0 && targetElBCR.top >= 0 && targetElBCR.right <= windowWidth &&
-        targetElBCR.bottom <= windowHeight;
+  if (baseClass) {
```
maxokorokov, line 41: Like can we have `baseClass` that is not set really? Once again unit tests would answer all these questions...

src/util/positioning.ts
```diff
@@ -214,42 +77,68 @@ export function positionElements(
...
-    // If nothing match, the first placement is the default one
-    testPlacement = placementVals[0];
-    addClassesToTarget(testPlacement);
-    positionService.positionElements(hostElement, targetElement, testPlacement, appendToBody);
-  }
+  let mainPlacement = popperPlacements.shift();
+
+  const bsModifier: Partial<Modifier<any, any>> = {
+    name: 'bootstrapClasses',
+    enabled: true,
```
maxokorokov, line 86: I think it should be `enabled: !!baseClass` to avoid executing it for components without the base class

src/util/positioning.ts
```diff
@@ -214,42 +77,68 @@ export function positionElements(
...
 
-  if (!isInViewport) {
-    // If nothing match, the first placement is the default one
-    testPlacement = placementVals[0];
-    addClassesToTarget(testPlacement);
-    positionService.positionElements(hostElement, targetElement, testPlacement, appendToBody);
-  }
+  let mainPlacement = popperPlacements.shift();
+
+  const bsModifier: Partial<Modifier<any, any>> = {
```
maxokorokov, line 84: maybe `bootstrapBaseClassModifier` ?

src/util/positioning.ts
```diff
@@ -214,42 +77,68 @@ export function positionElements(
...
+  const popperInstance = createPopper(hostElement, targetElement, {
+    placement: mainPlacement,
+    modifiers: [
+      bsModifier,
+      {
+        name: 'offset',
+        enabled: true,
+        options: {
+          offset: () => {
+            return baseClass === 'bs-popover' ? [0, 12] : [];
```
maxokorokov, line 121: This should be passed as a partial config only from calling `positionElements` from the popover...
We should not have component-specific code here. Also 8px, not 12 → https://github.com/twbs/bootstrap/blob/main/js/src/popover.js#L25

src/util/positioning.ts
```diff
@@ -214,42 +77,68 @@ export function positionElements(
...
+        },
+      },
+      {
+        name: 'flip',
+        options: {
+          fallbackPlacements: popperPlacements,
+        },
+      },
+      {
+        name: 'preventOverflow',
```
maxokorokov, line 132: don't think this is necessary

maxokorokov (changes requested):
Hey @fbasso,

here is what I've found so far:

**Big issues**
- [ ] Calling `positionElements` recreates popper every time is called. This is an issue for example in dropdown when we do it every time zone is stable. We should redesign the API for repositioning I guess and either create a new popper or do `popper.update()` when necessary.
- [ ] I think we should change `posisitonElements` to accept popper config as an argument to avoid component-specific code in `positioning.ts` (ex. for popover offset 8px should be hardcoded) and allowing end-users to provide custom popper configs if they need (separate PR for the last one I guess).
- [ ] Popper is destroyed correctly only in tooltip and popover

**Minor issues**
- [ ] tooltip/popover arrow is centered correctly only AFTER the next CD, but not initially (might be related to the 2nd major issue)
- [ ] dropdown/datepicker start flipping randomly on CD (definitely related to the 2nd major issue)
- [ ] we should not use deep imports, but import everything form `@popper/core`
- [ ] dropdown → we should check that keyboard works as expected and dropup/dropdown works right when popper flips menu positioning automatically
- [ ] tests broken with latest popper `2.10` (not sure which ones though, but you asked me to add this)

Everything else is inline in code!

Cheers,
Max

Issue #4100, opened 2021-05-23
navbar dropdown with .dropdown-menu-end is not aligned right


jnizet:
### Bug description:

In ng-bootstrap 11.0.0-beta.1, a dropdown with the class .dropdown-menu-end is not aligned right, but is aligned left. This can be reproduced by simply clicking the link "Static right" in the "Dynamic positioning in a navbar" demo of the dropdown component.

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-1w9gq9?file=src%2Fapp%2Fdropdown-navbar.html

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 12.0.1

ng-bootstrap: 11.0.0.beta-1

Bootstrap: 5.0.1

### Temporary workaround:

Add the folowing CSS rules in the component CSS:

```css
.dropdown-menu-end {
  right: 0;
  left: auto;
}
```

beginor:
container="body" doesn't work too.

PR #4136, opened 2021-07-31
fix(dropdown): fix for menu alignment issue mentioned in #4100

donalfenwick:
bootstrap v5 no longer sets `right: 0` as part of the `dropdown-menu-end` style rule. Instead it does the positioning using javascript by checking the elements computed styles and looking for `--bs-position: end;` in the CSS to figure out if the element should be positioned on the right.
Source :https://github.com/twbs/bootstrap/blob/23fd488c380c347a3d0121a49b6285bdd65418a2/js/src/dropdown.js#L285

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -513,6 +513,16 @@ export class NgbDropdown implements AfterContentInit, OnDestroy {
         renderer.removeClass(bodyContainer, 'dropdown');
         renderer.addClass(bodyContainer, dropdownClass);
       }
+
+      // We need to trim the value because custom properties can also include spaces
+      const isEnd = getComputedStyle(this._menu.nativeElement).getPropertyValue('--bs-position').trim() === 'end';
+      if (isEnd) {
+        renderer.setStyle(this._menu.nativeElement, 'left', 'auto');
+        renderer.setStyle(this._menu.nativeElement, 'right', '0');
+      } else {
+        renderer.removeStyle(this._menu.nativeElement, 'left');
+        renderer.removeStyle(this._menu.nativeElement, 'right');
+      }
     }
   }
 }


lionel-meunier (approved):

src/dropdown/dropdown.ts
```diff
@@ -513,6 +513,16 @@ export class NgbDropdown implements AfterContentInit, OnDestroy {
         renderer.removeClass(bodyContainer, 'dropdown');
         renderer.addClass(bodyContainer, dropdownClass);
       }
+
+      // We need to trim the value because custom properties can also include spaces
+      const isEnd = getComputedStyle(this._menu.nativeElement).getPropertyValue('--bs-position').trim() === 'end';
+      if (isEnd) {
+        renderer.setStyle(this._menu.nativeElement, 'left', 'auto');
```
bbarry, line 520: Instead of using setStyle, the appropriate css class should be added, right?

setting style attributes directly makes it harder to comply with certain security policies.

donalfenwick: @bbarry that's a good point. 

In bootstrap v4 the left and right properties were being set as part of the `dropdown-menu-right` class.
image

However in V5 this class (renamed to `dropdown-menu-end`) has those styles [removed](https://getbootstrap.com/docs/5.0/components/dropdowns/#menu-alignment) 
image

Instead of using the class to position the dropdown menu bootstrap v5 now relies on popper to do its positioning in this situation (see code [here](https://github.com/twbs/bootstrap/blob/23fd488c380c347a3d0121a49b6285bdd65418a2/js/src/dropdown.js#L285)).

This would be fine except that ng-bootstrap handles this a little differently when the dropdowns are placed within a navbar. The original issue #4100 was raised relating to the navbar dropdown in the docs. (see following image)
image

In the case where the dropdown is placed inside a navbar, the display property gets set to a value of "static" which [ignores the placement configuration](https://github.com/ng-bootstrap/ng-bootstrap/blob/9598506dfcbb0112afe21a1d6157c4645fd47522/src/dropdown/dropdown.ts#L501).

Maybe we could add the `end-0` class to the element instead of the call to `setStyle`, but at this point I'm wondering if this would be better included as part of the popper2 migration work done on PR #4129. It might be  more inline with the way bootstrap handles this itself.

@fbasso dis you come across this issue with the static nav menu items when looking into integrating popper v2?

fbasso:
Hi @donalfenwick, thanks for your (deep) investigation , it helped !

I close this PR in favor of #4240. The problem here is that it's not completely responsive (for example if you change the screen size when the dropdown is open, with for example 'dropdown-menu-lg-end', and it differs from the Bootstrap example : https://getbootstrap.com/docs/5.1/components/dropdowns/#responsive-alignment.

Issue #4178, opened 2021-11-04
Angular 13 support


maxokorokov:
Latest releases:
- ng-bootstrap 10 with Angular 12 and BS 4.5 (Stable)
- ng-bootstrap 11 with Angular 13 and BS 4.6 → [RC.0](https://github.com/ng-bootstrap/ng-bootstrap/releases/tag/11.0.0-rc.0)
- ng-bootstrap 12 with Angular 13 and BS 5 → [Beta.4](https://github.com/ng-bootstrap/ng-bootstrap/releases/tag/12.0.0-beta.4)

---
Internal issues:
- [x] Looks like StackBlitz doesn't work with Angular 13 yet → https://github.com/stackblitz/core/issues/1657, so no demo site update
- [x] `eslint-plugin-deprecation` doesn't support `eslint 8`→ using `angular-eslint 12` version at the moment
- [x] e2e coverage is not working. Looks like issues with `coverage-istanbul-reporter` → temporarily disabling custom webpack config
- [x] BS warnings in e2e tests in console → temporarily disabling warning checks
- [x] ~~unit tests for schematics are broken due to `externalSchematic` calling `@angular/localize` ESM issues → temporarily disabling~~ (Fixed in Angular 13.0.2 with #4183)
- [x] ~~no support for Angular 13 in `ngx-build-plus` → no e2e coverage at the moment~~ → moved to @angular-builders with #4180

inexuscore:
Any idea when Angular 13 will be supported for the Bootstrap 4.x version? We're trying to update to Angular 13 but a few libraries have conflicting peer dependencies. It would be great to bump the versions and target ng 13 if it's not difficult. Thank you very much.

terencehonles:
@maxokorokov any thoughts on when the move to rxjs 7.4 might happen? This is possible in Angular 13, but rxjs is declared as using `^6.5.5` so I'm not sure if that would be something that would happen in ng-bootstrap 11 or ng-bootstrap 12. I can see what breaks when migrating, but not sure what branch to target and when/if the changes would be merged.

maxokorokov:
@terencehonles we'll do the same as all Angular packages and have `"rxjs": "^6.5.3 || ^7.4.0"` in peer deps the next 11 release

maxokorokov:
@terencehonles several issues we're currently facing:
- no `ngx-build-plus` support for Angular 13
- no `angular-eslint` support for Angular 13

[...]

PR #4180, opened 2021-11-10
build: replace ngx-build-plus with @angular-builders/custom-webpack

terencehonles:
This is to help unblock Angular 13 support as mentioned in https://github.com/ng-bootstrap/ng-bootstrap/issues/4178#issuecomment-965018042. Opening a PR to see what the CI says since I believe this config is used by the CI.

Before submitting a pull request, please make sure you have at least performed the following:

 - [x] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) and [DEVELOPER.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/DEVELOPER.md) guide.
 - [ ] built and tested the changes locally.
 - [x] added/updated any applicable tests.
 - [x] added/updated any applicable API documentation.
 - [x] added/updated any applicable demos.

PR #4182, opened 2021-11-10
Angular 13

maxokorokov:
See #4178  for details


maxokorokov:
Rebased on #4180

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.angular/cache
```
peterblazejewicz, line 1:
```suggestion
.angular/
```
should be probably just `.angular`
https://github.com/angular/components/blob/master/.gitignore#L14

maxokorokov: Thanks, will check. That's what cli13 is generating

PR #4196, opened 2021-12-07
fix(accordion): ids that begin with number were not accepted, #4195

### LINKED ISSUE #4195: multiple Accordion's build with ngFor ['Element': '#3.1' is not a valid selector.]
### Bug description:
- in template(html) I want multiple accordions buld with for lop, how to define them in typescript 
- in typescript I want for specific accordion to toggel specific panel

is this even possible?

- HTML:

```html
<div class="panel-group"
         *ngFor="let fqa of faqs, let i = index">

        <h5 [id]="fqa.id">{{fqa.section}}</h5>

        <ngb-accordion #acc="ngbAccordion"
                       [id]="fqa.id">
            <ngb-panel *ngFor="let q of fqa.qa"
            [id]="q.id"
                       title="V: {{q.question}}">

                <ng-template ngbPanelContent>
                    <span class="badge badge-primary">O:</span>&nbsp;
                    <span style="text-align: justify"
                          [innerHtml]="q.answer"></span>
                </ng-template>
            </ngb-panel>
        </ngb-accordion>
</div>
```
- typescript:
```typescript
@ViewChildren('acc')
  accordions: QueryList<NgbAccordion>;
```

- error:
```
Failed to execute 'querySelector' on 'Element': '#1.1' is not a valid selector.
    at NgbAccordion._getPanelElement (http://localhost:4200/vendor.js:1874:44)
```

Angular: 13

ng-bootstrap:  10.0.0

Bootstrap:  4.6.0

ExFlo:
fix #4195

### DIFF
--- a/src/accordion/accordion.ts
+++ b/src/accordion/accordion.ts
@@ -14,6 +14,8 @@ import {
   TemplateRef,
   ViewEncapsulation,
   NgZone,
+  OnInit,
+  OnDestroy,
 } from '@angular/core';
 
 import {isString} from '../util/util';
@@ -134,6 +136,7 @@ export class NgbPanel implements AfterContentChecked {
   titleTpl: NgbPanelTitle;
   headerTpl: NgbPanelHeader;
   contentTpl: NgbPanelContent;
+  panelDiv: HTMLElement | null;
 
   @ContentChildren(NgbPanelTitle, {descendants: false}) titleTpls: QueryList<NgbPanelTitle>;
   @ContentChildren(NgbPanelHeader, {descendants: false}) headerTpls: QueryList<NgbPanelHeader>;
@@ -172,6 +175,16 @@ export interface NgbPanelChangeEvent {
   preventDefault: () => void;
 }
 
+@Directive({selector: '[ngbRef]'})
+export class NgbRefDirective implements OnInit, OnDestroy {
+  @Output() ngbRef = new EventEmitter<HTMLElement | null>();
+  constructor(private _El: ElementRef) {}
+
+  ngOnInit() { this.ngbRef.emit(this._El.nativeElement); }
+
+  ngOnDestroy() { this.ngbRef.emit(null); }
+}
+
 /**
  * Accordion is a collection of collapsible panels (bootstrap cards).
  *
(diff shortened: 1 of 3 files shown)


src/accordion/accordion.ts
```diff
@@ -425,10 +436,6 @@ export class NgbAccordion implements AfterContentChecked {
       }
     });
   }
-
-  private _getPanelElement(panelId: string): HTMLElement | null {
-    return this._element.nativeElement.querySelector('#' + panelId);
```
maxokorokov, line 430: Can't we just `querySelector('[id="{panelId}"]')` ?

src/accordion/accordion.ts
```diff
@@ -195,7 +208,7 @@ export interface NgbPanelChangeEvent {
           <ng-template [ngTemplateOutlet]="panel.headerTpl?.templateRef || t"
                        [ngTemplateOutletContext]="{$implicit: panel, opened: panel.isOpen}"></ng-template>
         </div>
-        <div id="{{panel.id}}" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'"
+        <div id="{{panel.id}}" (ngbRef)="panel.panelDiv = $event" role="tabpanel" [attr.aria-labelledby]="panel.id + '-header'"
```
maxokorokov, line 211: Oh, Im so not convinced by passing id as the event on init :)

maxokorokov (changes requested):
Hey,

is there actually an issue for this?
The one that is referenced is about something else.

I'm not convinced by passing the id as the event on init and having a special directive for this on my side :)

Can't we just change the query? WDYT?

src/accordion/accordion.ts
divdavem: Note that this is not an id, it is the reference to the element itself, thus removing the need of an id.

maxokorokov: Sure. But on the other hand I just thought that it also avoids doing `querySelector` for each panel on every `ngAfterContentChecked`...

divdavem: `ngbRef` seems to be inspired from [callback refs in react](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs).
We could instead use [refs object as in react](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) which would perhaps convince you more than the callback form.

maxokorokov (approved):
Hmm, on reflection:

I don't like the idea of firing events in lifecycle hooks, it sometimes leads to nasty side-effects, doesn't it? However in this case ids never change and it's pretty simple, so it looks ok to me. I can't find any simpler ways to do it once on my side.

Simpler version would be to just change selector to use the attribute, but this PR also removes doing `querySelector` on each `ngAfterContentChecked which is a good thing.

I think it is LGTM.

src/accordion/accordion.ts
maxokorokov: I see, and on my side I can't really find a simpler way to pass the element ref once during the initialization

maxokorokov: maybe I would prefer passing a function instead of using event emitter

Issue #4211, opened 2022-01-05
NgbDatepicker breaks on server side rendering when accessing `window`


ppicom:
Please remember, the issues forum is __NOT__ for support requests. It is for bugs and feature requests only.
Please read https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md and search
existing issues (both open and closed) prior to opening any new issue and ensure you follow the instructions therein.

### Bug description:
NgInputDatePicker tries to access the `window` object on the server side, hence throwing a TypeError.

### Link to minimally-working StackBlitz that reproduces the issue:
As the issue involves server side rendering I'm not sure how to reproduce it in a StackBlitz, but the logs have the detailed location on where in the code this is happening.

```
ReferenceError: window is not defined
    at Object.getAllStyles (/app/dist/gh-angular/server/main.js:1:428526)
    at Object.getStyle (/app/dist/gh-angular/server/main.js:1:428601)
    at Object.position (/app/dist/gh-angular/server/main.js:1:429126)
    at Object.positionElements (/app/dist/gh-angular/server/main.js:1:430900)
    at positionElements (/app/dist/gh-angular/server/main.js:1:433769)
    at NgbInputDatepicker._updatePopupPosition (/app/dist/gh-angular/server/main.js:1:442657)
    at /app/dist/gh-angular/server/main.js:1:437642
    at SafeSubscriber.schedulerFn [as _next] (/app/dist/gh-angular/server/main.js:1:4474855)
    at SafeSubscriber.__tryOrUnsub (/app/dist/gh-angular/server/main.js:1:1686779)
    at SafeSubscriber.next (/app/dist/gh-angular/server/main.js:1:1685478)
ReferenceError: window is not defined
    at Object.getAllStyles (/app/dist/gh-angular/server/main.js:1:428526)
    at Object.getStyle (/app/dist/gh-angular/server/main.js:1:428601)
    at Object.position (/app/dist/gh-angular/server/main.js:1:429126)
    at Object.positionElements (/app/dist/gh-angular/server/main.js:1:430900)
    at positionElements (/app/dist/gh-angular/server/main.js:1:433769)
    at NgbInputDatepicker._updatePopupPosition (/app/dist/gh-angular/server/main.js:1:442657)
    at /app/dist/gh-angular/server/main.js:1:437642
    at SafeSubscriber.schedulerFn [as _next] (/app/dist/gh-angular/server/main.js:1:4474855)
[...]

PR #4212, opened 2022-01-05
fix(datepicker): prevent the datepicker from calling `window` server side

ppicom:
Fixes ng-bootstrap/ng-bootstrap#4211

### DIFF
--- a/src/util/positioning.ts
+++ b/src/util/positioning.ts
@@ -11,7 +11,9 @@ interface ClientRect {
 }
 
 export class Positioning {
-  private getAllStyles(element: HTMLElement) { return window.getComputedStyle(element); }
+  private getAllStyles(element: HTMLElement) { 
+    return !!window ? window.getComputedStyle(element) : new CSSStyleDeclaration();
+  }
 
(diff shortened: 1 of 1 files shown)


maxokorokov (changes requested):
@ppicom I'm sorry, but I don't think we'll be able to patch `ng-bootstrap: 6.2.0` at this point.

We could do a release for the `11.X.X` (it seems to still have this problem). Please open the PR against the `11_0_x` branch → https://github.com/ng-bootstrap/ng-bootstrap/branches

However it would be helpful to understand the use-case that led to this, so an SSR test would be necessary. Test should be added here → https://github.com/ng-bootstrap/ng-bootstrap/blob/master/ssr-app/src/app/components/datepicker.component.ts

We can discuss it beforehand if help is needed.

```
yarn ssr → ssr test suite
yarn ssr-app:serve → run ssr app
```

Since `12.X.X` this doesn't seem to be relevant anymore as we use popper for positioning.

maxokorokov:
It would also be preferrable to fix it using something like `isPlatformBrowser` and `PLATFORM_ID`

maxokorokov:
Closing as no response, please reopen if necessary.

PR #4244, opened 2022-02-11
doc(carousel): adding since/before badges regarding left/right or start/end

divdavem:


fbasso:
Fine, thanks @divdavem !

divdavem:
@fbasso Thank you for reviewing and merging my PR!

PR #4254, opened 2022-02-16
fix(carousel): indicators not visible

### LINKED ISSUE #4253: ngbCarousel - showNavigationIndicators not working
### Bug description:

After updating to ng-bootstrap 12 navigation indicators not showing any more in ngbCarousel.

Go to https://ng-bootstrap.github.io/#/components/carousel/examples and click "Toggle Navigation Indicators" and nothing happens.

Angular: 13.2.2

ng-bootstrap: 12.0.0

Bootstrap: 5.1.3

fbasso:
fix #4253

### DIFF
--- a/src/carousel/carousel.ts
+++ b/src/carousel/carousel.ts
@@ -68,6 +68,7 @@ export class NgbSlide {
   changeDetection: ChangeDetectionStrategy.OnPush,
   encapsulation: ViewEncapsulation.None,
   host: {
+    '[id]': 'id',
     'class': 'carousel slide',
     '[style.display]': '"block"',
     'tabIndex': '0',
@@ -83,7 +84,10 @@ export class NgbSlide {
     <ol class="carousel-indicators" [class.visually-hidden]="!showNavigationIndicators" role="tablist">
       <li *ngFor="let slide of slides" [class.active]="slide.id === activeId"
           role="tab" [attr.aria-labelledby]="'slide-' + slide.id" [attr.aria-controls]="'slide-' + slide.id"
+          [attr.data-bs-target]="id"
+          [attr.data-bs-to]="'slide-' + slide.id"
           [attr.aria-selected]="slide.id === activeId"
+          [attr.aria-curent]="slide.id === activeId"
           (click)="focus();select(slide.id, NgbSlideEventSource.INDICATOR);"></li>
     </ol>
     <div class="carousel-inner">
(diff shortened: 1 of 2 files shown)


maxokorokov:
this should be marked as a feature, not fix, so we could add `@since` decorator when releasing

src/carousel/carousel.ts
```diff
@@ -83,7 +84,10 @@ export class NgbSlide {
     <ol class="carousel-indicators" [class.visually-hidden]="!showNavigationIndicators" role="tablist">
       <li *ngFor="let slide of slides" [class.active]="slide.id === activeId"
           role="tab" [attr.aria-labelledby]="'slide-' + slide.id" [attr.aria-controls]="'slide-' + slide.id"
+          [attr.data-bs-target]="id"
```
divdavem, line 87: If you want to do as bootstrap, you should add a `#` before the id.
cf https://getbootstrap.com/docs/5.1/components/carousel/#with-indicators

But actually, I don't understand the point of adding an id at all. You could simply add the `data-bs-target` attribute with any value (or even without any value) for the CSS to be properly applied Do we actually need an id?

src/carousel/carousel.ts
```diff
@@ -83,7 +84,10 @@ export class NgbSlide {
...
+          [attr.data-bs-to]="'slide-' + slide.id"
```
divdavem, line 88: What is the point of adding this attribute? If you want to do as bootstrap, there is a `data-bs-slide-to` attribute and it contains only the slide number, but I don't see any `data-bs-to` attribute.
cf (again) https://getbootstrap.com/docs/5.1/components/carousel/#with-indicators

src/carousel/carousel.ts
```diff
@@ -83,7 +84,10 @@ export class NgbSlide {
...
           [attr.aria-selected]="slide.id === activeId"
+          [attr.aria-curent]="slide.id === activeId"
```
divdavem, line 90: According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current :

> When something is selected rather than current, such as a [tab](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role) in a [tablist](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role), use [aria-selected](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) to indicate the currently-displayed tabpanel.

Unlike bootstrap, we are using `role="tablist"` and `role="tab"` for carousel indicators, because they are somehow similar to tabs in a tablist. So is it correct to use both `aria-selected` and `aria-current`?

divdavem:
Also, if we want to do like bootstrap, shouldn't we use `<button>` inside `<div class="carousel-indicators">` instead of `<li>` inside `<ol class="carousel-indicators">`?

maxokorokov:
As discussed closing in favor of #4259

PR #4258, opened 2022-02-22
fix: dropdown and datepicker respond to autoClose input changes

### LINKED ISSUE #4219: Bug: NgbDropdown does not respond to autoClose changes while open
### Bug description:
Changing the NgbDropdown autoClose input while it is open has no effect, only its initial state is taken into account.

https://stackblitz.com/edit/angular-prenvw?file=src/app/dropdown-basic.ts

Angular: 13.0.3
ng-bootstrap: 11.0.0
Bootstrap: 4.6.1

### Possible solution

Maybe this is one of the motivations for #3014 (Re-implement AutoClose as a service) but there is also a simpler solution:
- inside _setCloseHandlers add
```ts 
this._close$.next() // to unsubscribe the previous closeHandler, may be a good idea to rename this _clearCloseHandler$
```
- inside ngOnChanges add the following:
```ts
if (changes.autoClose && this._open) {
  this._setCloseHandlers();
}
```
I would be happy to submit a pull request along those lines.

BistroStu:
Closes #4219 

I added the feature to datepicker as well as dropdown because it's equally applicable there. popover, tooltip and typeahead also use autoClose but I can't imagine any use case for this feature in a non-interactive component.

### DIFF
--- a/e2e-app/src/app/datepicker/autoclose/datepicker-autoclose.e2e-spec.ts
+++ b/e2e-app/src/app/datepicker/autoclose/datepicker-autoclose.e2e-spec.ts
@@ -147,6 +147,21 @@ describe('Datepicker Autoclose', () => {
         await clickOnDay(DATE_OUTSIDE_AFTER);
         await expectDatepickerToBeClosed(`Datepicker should be closed on outside day click`);
       });
+      it(`should change autoClose setting dynamically`, async() => {
+        // initially set autoClose to 'inside' because selectAutoClose is an outside click
+        await selectAutoClose('inside');
+        await openDatepicker(`Open datepicker with autoclose 'inside'`);
+
+        // change autoClose to false whilst open
+        await selectAutoClose('false');
+        await clickOnDay(DATE_SELECT);
+        await expectDatepickerToBeOpen(`Datepicker should not close after autoClose is changed to false whilst open`);
+
+        // change autoClose to 'inside' whilst open
+        await selectAutoClose('inside');
+        await clickOnDay(DATE_SELECT);
+        await expectDatepickerToBeClosed(`Datepicker should close after autoClose is changed to 'inside' whilst open`);
+      });
     });
   }
 });
--- a/e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.e2e-spec.ts
+++ b/e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.e2e-spec.ts
@@ -180,5 +180,20 @@ containers.forEach((container) => {
       await sendKey(' ');
       await expectDropdownToBeHidden(`Dropdown should be closed on Space`);
     });
+    it(`should change autoClose setting dynamically`, async() => {
+      // initially set autoClose to 'inside' because selectAutoClose is an outside click
+      await selectAutoClose('inside');
+      await openDropdown(`Open dropdown with autoclose 'inside'`);
+
+      // change autoClose to false whilst open
+      await selectAutoClose('false');
+      await clickDropdownItem();
+      await expectDropdownToBeVisible(`Dropdown should not close after autoClose is changed to false whilst open`);
+
+      // change autoClose to 'inside' whilst open
+      await selectAutoClose('inside');
+      await clickDropdownItem();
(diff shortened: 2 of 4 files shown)


e2e-app/src/app/datepicker/autoclose/datepicker-autoclose.e2e-spec.ts
```diff
@@ -147,6 +147,23 @@ describe('Datepicker Autoclose', () => {
         await clickOnDay(DATE_OUTSIDE_AFTER);
         await expectDatepickerToBeClosed(`Datepicker should be closed on outside day click`);
       });
+      it(`should change autoClose setting dynamically`, async() => {
+        // initially set autoClose to true
+        await selectAutoClose('true');
```
maxokorokov, line 152: Can't really see the point in setting `autoClose` to `true` initially, closing and resetting it to `false`

src/datepicker/datepicker-input.ts
```diff
@@ -278,6 +279,7 @@ export class NgbInputDatepicker implements OnChanges,
   private _onChange = (_: any) => {};
   private _onTouched = () => {};
   private _validatorChange = () => {};
+  private _destroyCloseHandlers$ = new Subject<void>();
```
maxokorokov, line 282: Could you please move this one ±100 lines higher to other private field group, so it won't get lost here accidentally

src/datepicker/datepicker-input.ts
```diff
@@ -288,6 +290,7 @@ export class NgbInputDatepicker implements OnChanges,
       config: NgbInputDatepickerConfig) {
     ['autoClose', 'container', 'positionTarget', 'placement'].forEach(input => this[input] = config[input]);
     this._zoneSubscription = _ngZone.onStable.subscribe(() => this._positioning.update());
+    this.closed.subscribe(this._destroyCloseHandlers$);
```
maxokorokov, line 293: Could you do the same as in dropdown and just `this._destrotCloseHandlers$.next()` inside the `close()` method where we actually firing `this.close.next()` to avoid having this subscription?

Also like this we never unsubscribe

e2e-app/src/app/dropdown/autoclose/dropdown-autoclose.e2e-spec.ts
```diff
@@ -180,5 +180,22 @@ containers.forEach((container) => {
       await sendKey(' ');
       await expectDropdownToBeHidden(`Dropdown should be closed on Space`);
     });
+    it(`should change autoClose setting dynamically`, async() => {
+      // initially set autoClose to true
+      await selectAutoClose('true');
```
maxokorokov, line 185: Same comment as or the datepicker test

maxokorokov (changes requested):
Hey @BistroStu,

thank you for your first contribution!

It looks good to me apart from some minor comments I left.

Could we also please either split it into 2 separate commits:
`fix(datepicker): ....`
`fix(dropdown): ...` 
to keep history cleaner and to make it fixes, not features, so we could plan it for `12.0.1` ?

Cheers,
Max

e2e-app/src/app/datepicker/autoclose/datepicker-autoclose.e2e-spec.ts
BistroStu: Calling selectAutoClose constitutes an outside click, which interferes with the test. I will change to testing with 'inside' setting.

src/datepicker/datepicker-input.ts
BistroStu: Yes no problem - I didn't notice that group

src/datepicker/datepicker-input.ts
BistroStu: Fair enough , there's no other reason to call this.close.next anyway.

BistroStu:
Hi @maxokorokov, 

I reset my branch to clean up the commits as you suggested and the pull request closed. Oops.

I have incorporated all your feedback now. 

Thanks for your time,
Stu.

maxokorokov (approved):
LGTM, thanks!

PR #4260, opened 2022-02-22
refactor(test): use @playwright/test instead of jasmine

divdavem:


demo/src/app/components/components.e2e-spec.ts
```diff
@@ -10,60 +9,57 @@ const COMPONENTS = [
...
+  test.afterEach(async() => {
+    if (messages.length > 0) {
+      console.log(messages);
+      test.fail(true, 'Unexpected console messages found');
     }
   });
 
-  COMPONENTS.forEach(component => {
-    describe(`${component} page`, () => {
+  test('should work', async({page}) => {
```
maxokorokov, line 31: should work? :)

package.json
```diff
@@ -23,29 +23,25 @@
...
-    "playwright": "node misc/playwright.js",
     "e2e-app:serve": "ng serve e2e-app",
     "e2e-app:lint": "ng lint e2e-app",
     "ngb:static": "ts-node --project misc/tsconfig.json misc/copy-static-files.ts",
     "ngb:lint": "ng lint ng-bootstrap",
     "ngb:test": "ng test ng-bootstrap --code-coverage --source-map true --progress false --watch false",
     "ngb:tdd": "ng test ng-bootstrap --source-map false",
-    "ngb:e2e": "concurrently --success first --kill-others \"yarn e2e-app:serve -c playwright\" \"yarn ngb:e2e-playwright && yarn ngb:e2e-playwright -b firefox && yarn ngb:e2e-playwright -b webkit\"",
-    "ngb:e2e-playwright": "yarn playwright",
+    "ngb:e2e": "playwright test -c e2e-app",
```
maxokorokov, line 32: I think we can remove `concurrently` as well from dependencies

maxokorokov (approved):
LGTM, left a couple of comments.
Maybe @fbasso should look too?

package.json
divdavem: Indeed, thank you!

demo/src/app/components/components.e2e-spec.ts
divdavem: I have replaced it with "should load successfully". Is it now precise enough?

e2e-app/src/app/tools.po.ts
```diff
@@ -18,7 +18,7 @@ export const sendKey = async(key: string) => {
  * @param message to display in case of error
  */
 export const waitForFocus = async(selector: string, message = `Unable to focus '${selector}'`) => {
-  const {page} = test;
+  const page = getPage();
   const el = await page.$(selector);
   await timeoutMessage(page.waitForFunction(function(_el) { return _el === document.activeElement; }, el), message);
```
fbasso, line 23: As we have all the playwright test runner capabilities now, I would use the [toBeFocused ](https://playwright.dev/docs/test-assertions#expectlocatortobefocusedoptions)feature:
```typescript
await expect(selector, message).toBeFocused()
```

Maybe this utility function is not needed anymore, as the code is simple enough to be used inline.

e2e-app/src/app/tools.po.ts
```diff
@@ -76,7 +46,7 @@ const roundBoundingBox = (rect: {x: number, y: number, width: number, height: nu
  * Returns the element bounding box
  */
 export const getBoundingBox = async(selector: string) => {
-  const element = await test.page.$(selector);
+  const element = await getPage().$(selector);
   const boundingBox = element ? await element.boundingBox() : {x: 0, y: 0, width: 0, height: 0};
   return roundBoundingBox(boundingBox !);
```
fbasso, line 51: Can we simplify this code by using locators ? (not directly linked to the test runner, but at the time locators didn't exist).

```typescript
   const boundingBox = (await getPage().locator('selector').boundingBox()) || {x: 0, y: 0, width: 0, height: 0};
```

What do you think ?

e2e-app/src/app/modal/autoclose/modal-autoclose.po.ts
```diff
@@ -1,26 +1,26 @@
...
+    await getPage().click(`#option-${option}`);
   }
 
-  await test.page.click('#open-modal');
+  await getPage().click('#open-modal');
   await waitForModalCount(1);
 };
 
 export const waitDismissReason = async(expected, error) => {
   await timeoutMessage(
```
fbasso, line 24: Probably timeoutMessage is not needed anymore (everywhere), as we can use the expect features for that purpose.

For example here:

```typescript
await expect(getPage().locator('#dismiss-reason'), error).toHaveText(expected);
```

e2e-app/src/app/tools.po.ts
```diff
@@ -129,5 +99,5 @@ export const mouseMove = async(selector: string) => {
   const rect = await getBoundingBox(selector);
```
fbasso, line 99: This function is not needed anymore, it's now better to use [ locator.hover](https://playwright.dev/docs/api/class-locator#locator-hover)

PR #4262, opened 2022-02-23
build(schematics): clean from rollup config instead of using rimraf

divdavem:


maxokorokov (commented):
Actually both with `rimraf` and without are fine with me, so yo decide :)

But, there is no lockfile? is this ok?

Merge if you want, as you'll probably have conflicts with the playwright PR

divdavem:
> Actually both with `rimraf` and without are fine with me, so you decide :)

I prefer without actually.

> But, there is no lockfile? is this ok?

It looks like `rimraf` is included as a dependency of some other dependency so it does not change `node_modules` to remove it from direct dependencies.

> Merge if you want, as you'll probably have conflicts with the playwright PR

It looks like git is able to automatically resolve conflicts in this case. I will merge it now.

PR #4268, opened 2022-02-28
fix(test): increase timeout in datepicker-focus.e2e-spec (especially for webkit)

divdavem:
It looks like this test sometimes fails on webkit because of a timeout that is too small.

(cf for example https://github.com/ng-bootstrap/ng-bootstrap/actions/runs/1910949270)


fbasso:
LGTM

divdavem:

PR #4269, opened 2022-03-01
Use test.describe.parallel

divdavem:


fbasso:
Finally, I don't know if it's worth doing this now, because of https://github.com/microsoft/playwright/pull/12446 (merged, so I guess available in the next release of Playwright).

What do you think ?

divdavem:
> Finally, I don't know if it's worth doing this now, because of [microsoft/playwright#12446](https://github.com/microsoft/playwright/pull/12446) (merged, so I guess it will be available in the next release of Playwright).
> 
> What do you think ?

Indeed, I will close this PR. Let's use the `fullyParallel` option when it is available in playwright.

Issue #4272, opened 2022-03-01
Documentation for latest Bootstrap 4 branch


jnizet:
Unless I missed something, the only available documentation now is for the v12 version , which is based on Bootstrap 5. But since there still is a branch for Bootstrap 4, and since I expect many projects are still on Bootstrap 4, it would be nice to have the documentation online for this branch too. 
Even if the ng-bootstrap API itself hasn't changed much, the HTML markup in the demos has changed.

maxokorokov:
Having said that, I'll take a look if we can publish a static version under `https://ng-bootstrap.github.io/11.0.0/...`. My fears might be unfounded :)

PR #4279, opened 2022-03-09
Older versions documentation

maxokorokov:
Adds the documentation selector that would look like this:

Screenshot 2022-03-09 at 17 26 58

And on mobile:

Screenshot 2022-03-09 at 17 27 10

---

Several implementation notes:
- older website versions will be available under `ng-bootstrap.github.io/releases/11.x/`-like URLs
- if I'm at `ng-bootstrap.github.io/#/components/datepicker/overview` and I switch to version `11`, I'll be redirected to `ng-bootstrap.github.io/releases/11.x/#/components/datepicker/overview`
- It will be possible to navigate from older versions to newer versions (available version list is published in `/versions.json`)
- these `/releases/x.x` versions will not be indexed
- they will be just committed into `demo/src/app/public/releases/11.x`
- a script `demo:release-snapshot` is a helper to create the above directory
- I'll open a separate PR to add older versions

Part of #4272


demo/src/public/versions.json
```diff
@@ -0,0 +1 @@
+["12.x"]
```
maxokorokov, line 1: Hmm, I think I'd like to put all data like text and URLs here as well. So in case things change, we won't need to rebuild older versions.

demo/src/app/demo-versions.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class NgbdDemoVersionsComponent {
+  current = environment.version;
+  routerUrl = '';
+  versions$: Promise<Version[]> = (window as any).NGB_DEMO_VERSIONS;
+
+  constructor(router: Router) {
+    router.events.pipe(
+      filter(event => event instanceof NavigationEnd)
+    ).subscribe((event: NavigationEnd) => {
```
divdavem, line 34: I know the component is supposed to be always present, but in case it changes in the future, wouldn't be better to have a correct clean-up mechanism to unsubscribe if the component is destroyed?

maxokorokov: Yep, I haven't add it explicitly because it is always present. I really doubt it will change, as the idea for this component is to be always visible. I still can add the cleanup code, sure.

maxokorokov: fixed

PR #4282, opened 2022-03-12
fix(datepicker): ensure popup is shown above modal

### LINKED ISSUE #4278: DatePicker popup attached to body is shown behind modal window
### Bug description:

Configuring a `NgbDatepicker` with `container="body"` and placing it within a `NgbActiveModal ` dialog causes the datepicker's popup to be shown underneath the modal's backdrop.

image

https://stackblitz.com/edit/angular-ukonh5?file=src/app/modal-basic.html

This Stackblitz is forked from the "
Modal with default options" [example here](https://ng-bootstrap.github.io/#/components/modal/examples), and the only change is adding `container="body"` to the input.

Angular: 13.0.3

ng-bootstrap: 12.0.0

Bootstrap: 5 (whatever ng-bootstrap 12.0.0 depends on, locally I use bootstrap 5.1.3)

HexagonSun:
When using container="body", the datepicker's `z-index` must be equal
or above the modal window's one (as defined in `modal.scss`).

Fixes #4278

 - [n/a] added/updated any applicable API documentation.
 - [n/a] added/updated any applicable demos.

### DIFF
--- a/e2e-app/src/app/app.module.ts
+++ b/e2e-app/src/app/app.module.ts
@@ -9,6 +9,7 @@ import {AppComponent} from './app.component';
 import {NavigationComponent} from './navigation.component';
 
 import {DatepickerAutoCloseComponent} from './datepicker/autoclose/datepicker-autoclose.component';
+import {DatepickerContainerComponent} from './datepicker/container/datepicker-container.component';
 import {DatepickerFocusComponent} from './datepicker/focus/datepicker-focus.component';
 import {DatepickerMultipleComponent} from './datepicker/multiple/datepicker-multiple.component';
 import {DropdownAutoCloseComponent} from './dropdown/autoclose/dropdown-autoclose.component';
@@ -36,6 +37,7 @@ import {TypeaheadValidationComponent} from './typeahead/validation/typeahead-val
     AppComponent,
     NavigationComponent,
     DatepickerAutoCloseComponent,
+    DatepickerContainerComponent,
     DatepickerFocusComponent,
     DatepickerMultipleComponent,
     DropdownAutoCloseComponent,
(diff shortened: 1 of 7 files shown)


HexagonSun:
I just realized that the same issue occurs with `NgbDropdown` as well, see [dropdown.ts#L486](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/dropdown/dropdown.ts#L486).

shyamal890:
Actually instead of moving datepicker, popover, tooltip to a higher z-index it would make sense to lower down modal's z-index as it is also interferring with external libraries like ng-select.

maxokorokov (approved):
LGTM, thanks!

However it would be nice to think about the @shyamal890's comment before merging this

HexagonSun:
> it would make sense to lower down modal's z-index as it is also interferring with external libraries like ng-select.
> …
> it would be nice to think about the @shyamal890's comment before merging this

I was starting with the premise that there was a reason for the modal's z-index to be raised. Of course it would work, too, but we'd have to find out first why it was changed in [fc84cb1d](https://github.com/ng-bootstrap/ng-bootstrap/commit/fc84cb1d021dc8ff60cb17a908a203f2bba8579f).


This PR adds an E2E test that is designed to work regardless of what the modal's and datepicker's z-indices absolute z-indices are, so I feel that it's a good thing to have.


That said: seeing that there's an issue with `NgbDropdown` too, from a design standpoint I feel that there should be broader changes to how ng-boostrap determines and assignes these z-indices. Having three separate literal values `1050`/`1055` in three separate files (and applied in three different ways, once in `.scss`, once as a host directive and once through `Renderer2`) will always lead to inconsistencies.  
I really don't know enough of this code base and how these things are usually handled here. I feel that there should be at least a small utility function that provides the modal-related z-index value in a central location. Or a common service, or decorator or somthing like that. 
A very nice solution from a user's standpoint would be for ng-bootstrap to accept SCSS-variables so that one can use whatever value one likes. I realize this is a fundamental change, though, and _way_ out of scope :).

PR #4283, opened 2022-03-12
feat(offcanvas): implement offcanvas components and service

jnizet:

This PR implements the Offcanvas component of Bootstrap 5, in a way that is very similar to the Modal component.

It has a demo showing most of the options in action, but more demos should be added.

The design is very similar to the design of the Modal service and components. There are some notable differences, though:
- Bootstrap doesn't use a window component covering the whole screen as it does for the modal. So I did the same thing as Bootstrap. As a result, the clicks on the backdrop which closes the offcanvas are handled by the backdrop itself
- Only one offcanvas must be opened at once. The code itself doesn't ensure that: it's up the developer to ensure that it's the case. But since most of the time, there should be a backdrop, clicking on two different buttons that both open an offcanvas should be impossible. 
- The scrollbar is restored only after the offcanvas is hidden (i.e. after the animation rather than before). This is especially important when the offcanvas is on the end (right), because otherwise, the scrollbar restoration makes the offcanvas shift to the left before the animation starts. This should also be done on the modal, BTW, which would avoid the shifting bug that we see now when closing a modal.

Pros of the design:
- very flexible
- familiar, because it's similar to the modal design, so developers already using modals should be able to use the offcanvas very easily
- since the code is close to the code of the modal, maintenance should be easy

Cons:
- retains the same historic baggage as the modal
  - a result promise
  - the markup for the offcanvas panel is hidden in the ng-bootstrap component, forcing us to pass various options like aria attributes and css classes

### DIFF
--- a/e2e-app/src/app/offcanvas/autoclose/offcanvas-autoclose.component.html
+++ b/e2e-app/src/app/offcanvas/autoclose/offcanvas-autoclose.component.html
@@ -23,7 +23,7 @@
     <div class="col-12">
       <button class="btn btn-outline-secondary ms-3" id="reset-button" (click)="reset()">Reset</button>
       <button class="btn btn-outline-secondary ms-1" id="option-backdrop-false" (click)="options['backdrop'] = false">backdrop = false</button>
-      <button class="btn btn-outline-secondary ms-1" id="option-no-keyboard" (click)="options['keyboard'] = false">keyboard = 'false'</button>
+      <button class="btn btn-outline-secondary ms-1" id="option-no-keyboard" (click)="options['keyboard'] = false">keyboard = false</button>
     </div>
   </div>
 </form>
--- a/e2e-app/src/app/offcanvas/autoclose/offcanvas-autoclose.component.html
+++ b/e2e-app/src/app/offcanvas/autoclose/offcanvas-autoclose.component.html
@@ -1,0 +1,33 @@
+<h3>
+  Offcanvas autoclose tests
+  <span class="ms-1 badge bg-info text-dark" id="dismiss-reason">{{ reason }}</span>
+</h3>
+
+<!-- Offcanvas content -->
+<ng-template #content>
+  <div class="offcanvas-body">
+    <h2>Hello from offcanvas</h2>
+    <button id="offcanvas-close-button" class="btn btn-primary" (click)="closeOffcanvas()">Close offcanvas</button>
+  </div>
+</ng-template>
+
+<!-- Controls -->
+<form id="default" (contextmenu)="$event.preventDefault()">
+  <div class="mb-3 row row-cols-lg-auto">
+    <div class="col-12">
+      <div class="input-group">
+        <button class="btn btn-outline-secondary" type="button" id="open-offcanvas" (click)="openOffcanvas(content)">Open offcanvas</button>
+      </div>
+    </div>
+
+    <div class="col-12">
+      <button class="btn btn-outline-secondary ms-3" id="reset-button" (click)="reset()">Reset</button>
+      <button class="btn btn-outline-secondary ms-1" id="option-backdrop-false" (click)="options['backdrop'] = false">backdrop = false</button>
+      <button class="btn btn-outline-secondary ms-1" id="option-no-keyboard" (click)="options['keyboard'] = false">keyboard = 'false'</button>
+    </div>
+  </div>
+</form>
+
+<!-- Options -->
+<h4>Options</h4>
+<pre>{{ options | json }}</pre>
--- a/src/offcanvas/offcanvas-config.ts
+++ b/src/offcanvas/offcanvas-config.ts
@@ -1,0 +1,108 @@
+import {Injectable, Injector} from '@angular/core';
+import {NgbConfig} from '../ngb-config';
+
+/**
+ * Options available when opening new offcanvas windows with `NgbOffcanvas.open()` method.
+ */
+export interface NgbOffcanvasOptions {
+  /**
+   * If `true`, opening and closing will be animated.
+   */
+  animation?: boolean;
+
+  /**
+   * `aria-describedby` attribute value to set on the offcanvas panel.
+   */
+  ariaDescribedBy?: string;
+
+  /**
+   * `aria-labelledby` attribute value to set on the offcanvas panel.
+   */
+  ariaLabelledBy?: string;
+
+  /**
+   * If `true`, the backdrop element will be created for a given offcanvas.
+   *
+   * Default value is `true`.
+   */
+  backdrop?: boolean;
+
+  /**
+   * A custom class to append to the offcanvas backdrop.
+   */
+  backdropClass?: string;
+
+  /**
+   * Callback right before the modal will be dismissed.
+   *
+   * If this function returns:
+   * * `false`
+   * * a promise resolved with `false`
+   * * a promise that is rejected
+   *
+   * then the offcanvas won't be dismissed.
+   */
+  beforeDismiss?: () => boolean | Promise<boolean>;
+
+  /**
+   * A selector specifying the element all new offcanvas panels and backdrops should be appended to.
+   *
+   * If not specified, will be `body`.
+   */
+  container?: string | HTMLElement;
+
+  /**
+   * The `Injector` to use for offcanvas content.
+   */
+  injector?: Injector;
+
+  /**
+   * If `true`, the offcanvas will be closed when `Escape` key is pressed
+   *
+   * Default value is `true`.
+   */
+  keyboard?: boolean;
+
+  /**
+   * A custom class to append to the offcanvas panel.
+   */
+  panelClass?: string;
+
+  /**
+   * The position of the offcanvas
+   */
+  position?: 'start' | 'end' | 'top' | 'bottom';
+
+  /**
+   * Scroll content while offcanvas is open (false by default).
+   */
+  scroll?: boolean;
+}
+
+/**
+ * A configuration service for the [`NgbOffcanvas`](#/components/offcanvas/api#NgbOffcanvas) service.
+ *
+ * You can inject this service, typically in your root component, and customize the values of its properties in
+ * order to provide default values for all offcanvases used in the application.
+ */
+@Injectable({providedIn: 'root'})
+export class NgbOffcanvasConfig implements Required<NgbOffcanvasOptions> {
+  ariaDescribedBy: string;
+  ariaLabelledBy: string;
+  backdrop: boolean = true;
+  backdropClass: string;
+  beforeDismiss: () => boolean | Promise<boolean>;
+  container: string | HTMLElement;
+  injector: Injector;
+  keyboard = true;
+  panelClass: string;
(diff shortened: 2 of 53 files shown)


src/offcanvas/offcanvas-panel.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  ngOnDestroy() { this._disableEventHandling(); }
+
+  hide(): Observable<any> {
+    const {nativeElement} = this._elRef;
+    const context: NgbTransitionOptions<any> = {animation: this.animation, runningTransition: 'stop'};
+
+    const offcanvasTransition$ = ngbRunTransition(this._zone, this._elRef.nativeElement, (element) => {
+      nativeElement.classList.remove('show');
+      return () => element.style.visibility = 'hidden';
```
peterblazejewicz, line 71: why this one is needed (and L.94)? removing/adding 'show' class sets element visibility:
https://github.com/twbs/bootstrap/blob/main/scss/_offcanvas.scss#L9
https://github.com/twbs/bootstrap/blob/main/scss/_offcanvas.scss#L24

src/offcanvas/offcanvas-config.ts
```diff
@@ -0,0 +1,108 @@
...
+@Injectable({providedIn: 'root'})
+export class NgbOffcanvasConfig implements Required<NgbOffcanvasOptions> {
+  ariaDescribedBy: string;
+  ariaLabelledBy: string;
+  backdrop: boolean = true;
+  backdropClass: string;
+  beforeDismiss: () => boolean | Promise<boolean>;
+  container: string | HTMLElement;
+  injector: Injector;
+  keyboard = true;
```
peterblazejewicz, line 97:
```suggestion
  keyboard: boolean = true;
```
let's stick to one style of defining type, explicit or implicit

src/offcanvas/offcanvas-config.ts
```diff
@@ -0,0 +1,108 @@
...
+  panelClass: string;
+  position: 'start' | 'end' | 'top' | 'bottom';
+  scroll: boolean;
```
peterblazejewicz, line 100:
```suggestion
  scroll: boolean = false;
```
let's stick to one style of defining details, explicit vs implicit

src/offcanvas/offcanvas-panel.ts
jnizet: I'm not sure if it's a workaround for a bug in scss or in bootstrap, or something else, but the css looks like this: https://github.com/twbs/bootstrap/blob/main/dist/css/bootstrap.css#L6341. 
And if you look at the Bootstrap documentation and inspect the style of the offcanvas in the demos, they do set the visible style explicitly, too. And I confirm that only adding and removing the `show` class isn't sufficient.

src/offcanvas/offcanvas-panel.ts
```diff
@@ -0,0 +1,154 @@
...
+export class NgbOffcanvasPanel implements OnInit,
+    OnDestroy {
+  private _closed$ = new Subject<void>();
+  private _elWithFocus: Element | null = null;  // element that is focused prior to offcanvas opening
+
+  @Input() animation: boolean;
+  @Input() ariaLabelledBy?: string;
+  @Input() keyboard = true;
+  @Input() panelClass: string;
+  @Input() position: 'start' | 'end' | 'top' | 'bottom' = 'start';
```
peterblazejewicz, line 46:
```suggestion
  @Input() position: 'start' | 'end' | 'top' | 'bottom';
```
there is no default placement for offcanvas in native implementation. Is the different behavior in NG bootstrap desired?
https://getbootstrap.com/docs/5.1/components/offcanvas/#placement

jnizet: I think it's a good choice. It's a reasonable default, that can very easily be changed, either by passing an explicit option to open(), or by setting the default in the NgbOffcanvasConfig. 
And it makes the usage consistent with modals, where there is no need to pass any option to `open()`.
The alternative would be to allow no option to be passed (because NgbOffcanvasConfig might have set a default), bit to fail at runtime if the option isn't specified in the options nor in the config. I find it more developer-friendly to just use a reasonable default value.

src/offcanvas/offcanvas-stack.ts
```diff
@@ -0,0 +1,201 @@
...
+      this._scrollBarRestoreFn = this._scrollBar.hide();
+    }
+  }
+
+  open(moduleCFR: ComponentFactoryResolver, contentInjector: Injector, content: any, options: NgbOffcanvasOptions):
+      NgbOffcanvasRef {
+    const containerEl = options.container instanceof HTMLElement ? options.container : isDefined(options.container) ?
+                                                                   this._document.querySelector(options.container) :
+                                                                   this._document.body;
+    const renderer = this._rendererFactory.createRenderer(null, null);
```
peterblazejewicz, line 65:
```suggestion
```
it's not used

src/offcanvas/offcanvas-panel.ts
jnizet: Mistery solved: the CSS fix was added to Bootstrap recently: https://github.com/twbs/bootstrap/commit/0d054bb0f1484dbc0c778d3643e96e37ff46d708. Before that fix, they did indeed set the visibility explicitly.

src/offcanvas/offcanvas-stack.ts
```diff
@@ -0,0 +1,201 @@
...
+    containerEl.appendChild(panelCmptRef.location.nativeElement);
+    return panelCmptRef;
+  }
+
+  private _applyPanelOptions(windowInstance: NgbOffcanvasPanel, options: NgbOffcanvasOptions): void {
+    this._panelAttributes.forEach((optionName: string) => {
+      if (isDefined(options[optionName])) {
+        windowInstance[optionName] = options[optionName];
+      }
+    });
```
peterblazejewicz, line 127:
```suggestion
    this._panelAttributes.forEach((optionName) => windowInstance[optionName] ??= options[optionName]);
```
this should work same (skipping `isDefined`)

src/offcanvas/offcanvas-stack.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  private _applyBackdropOptions(backdropInstance: NgbOffcanvasBackdrop, options: NgbOffcanvasOptions): void {
+    this._backdropAttributes.forEach((optionName: string) => {
+      if (isDefined(options[optionName])) {
+        backdropInstance[optionName] = options[optionName];
+      }
+    });
```
peterblazejewicz, line 135:
```suggestion
    this._backdropAttributes.forEach(optionName => backdropInstance[optionName] ??= options[optionName]);
```
this should work exactly same way, skipping `isDefined` details

src/offcanvas/offcanvas-stack.ts
```diff
@@ -0,0 +1,201 @@
...
+  NgZone,
+  RendererFactory2,
+  TemplateRef
+} from '@angular/core';
+import {finalize, Subject} from 'rxjs';
+
+import {ngbFocusTrap} from '../util/focus-trap';
+import {ContentRef} from '../util/popup';
+import {ScrollBar} from '../util/scrollbar';
+import {isDefined, isString} from '../util/util';
```
peterblazejewicz, line 19:
```suggestion
import {isString} from '../util/util';
```
if `??=` change assignments are accepted, this has to be removed to prevent linting failure

src/offcanvas/offcanvas-stack.ts
jnizet: I don't think so. The existing code modifies the `windowInstance` option only if the corresponding `options` option is defined. Your code would modify the `windowInstance` option only if the `windowInstance option` is nullish, and whatever the value of the `options` option is. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment)

src/offcanvas/offcanvas-panel.ts
peterblazejewicz: thx!

src/offcanvas/offcanvas-stack.ts
peterblazejewicz: Oh, you're correct. I misread the purpose here

src/offcanvas/offcanvas-stack.ts
peterblazejewicz: disregard, I misread the purpose

src/offcanvas/offcanvas-stack.ts
peterblazejewicz: disregards, misread intent in other place

jnizet:
Thanks for the review @peterblazejewicz !

maxokorokov (changes requested):
Thanks a lot @jnizet for this PR, LGTM overall!

I would also prefer to keep the API close to modal one at the moment.
Just left a couple of minor comments.

Also I've removed `tabset` completely from the demo site, so you'll have to rebase.
And I guess we should update demos as well separately.

src/offcanvas/offcanvas-config.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * A custom class to append to the offcanvas panel.
+   */
+  panelClass?: string;
+
+  /**
+   * The position of the offcanvas
+   */
+  position?: 'start' | 'end' | 'top' | 'bottom';
```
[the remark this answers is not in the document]
jnizet: I don't think it would be a good idea: Bootstrap only supports those four positions, and the position is used to generate a Bootstrap CSS class. So if one passes a string that is not one of those four values, it would use an unexisting class, and the offcanvas wouldn't work.

jnizet:
The goal of the demo was mainly so that you can quickly see the result of my work, and most of the options. If you think this is going to be merged, I can find some time to work on more demos (and once again, get inspiration from the modal demos).

jnizet:
I added several other demos for the offcanvas (basic, options, focus, config).

e2e-app/src/app/offcanvas/autoclose/offcanvas-autoclose.component.html
```diff
@@ -0,0 +1,33 @@
...
+    <div class="col-12">
+      <div class="input-group">
+        <button class="btn btn-outline-secondary" type="button" id="open-offcanvas" (click)="openOffcanvas(content)">Open offcanvas</button>
+      </div>
+    </div>
+
+    <div class="col-12">
+      <button class="btn btn-outline-secondary ms-3" id="reset-button" (click)="reset()">Reset</button>
+      <button class="btn btn-outline-secondary ms-1" id="option-backdrop-false" (click)="options['backdrop'] = false">backdrop = false</button>
+      <button class="btn btn-outline-secondary ms-1" id="option-no-keyboard" (click)="options['keyboard'] = false">keyboard = 'false'</button>
```
cexbrayat, line 26: ni: any reason to use `backdrop = false` and `keyboard = 'false'`?

maxokorokov (approved):
LGTM, thanks @jnizet!
Will merge after releasing `12.0.1`!

PR #4294, opened 2022-03-23
fix(datepicker): do not scroll when opening with container='body'

### LINKED ISSUE #4290: ngbDatepicker - container body causes page to scroll to bottom
### Bug description:

When using ngbDatepicker with the container set as "body", opening the datepicker popup causes the screen to scroll to the bottom of the page.

https://angular-ivy-q3pj7e.stackblitz.io

Angular: 13.3.0
ng-bootstrap: 12.0.0
Bootstrap: 5.1.3

maxokorokov:
Delaying setting focus until after the popup is positioned with popper.

Fixes #4290


johannesnormannjensen:
Bump

PR #4297, opened 2022-03-28
fix(positionning): add offset to match Bootstrap

### LINKED ISSUE #4261: Check BS5 `data-***` styles
In some cases with BS 5 they start to use `data-bs-*` and `data-popper-*` attributes in the style rules, that we didn't have previously (ex. it was the case with broken carousel indicators and missing `data-bs-target` → #4259 

I think we should double-check this these rules.

Looking at the CSS, I see the following:
```css
/* dropdown */
.dropdown-menu[data-bs-popper]

.dropdown-menu-start[data-bs-popper]
.dropdown-menu-end[data-bs-popper]

.dropdown-menu-sm-start[data-bs-popper]
.dropdown-menu-sm-end[data-bs-popper]
.dropdown-menu-md-start[data-bs-popper]
.dropdown-menu-md-end[data-bs-popper]
.dropdown-menu-lg-start[data-bs-popper]
.dropdown-menu-lg-end[data-bs-popper]
.dropdown-menu-xl-start[data-bs-popper]
.dropdown-menu-xl-end[data-bs-popper]
.dropdown-menu-xxl-start[data-bs-popper]
.dropdown-menu-xxl-end[data-bs-popper]

/* tooltip */
.bs-tooltip-auto[data-popper-placement^=top] // with various placement values

/* popover */
.bs-popover-auto[data-popper-placement^=top] // with various placements

/* carousel */
.carousel-indicators [data-bs-target] // this is fixed by #4259 
```

---

1. There several places in the dropdown where we mess with `data-`:

I'm not sure we need this at all:
https://github.com/ng-bootstrap/ng-bootstrap/blob/1425962d40aff923192abc5e203ef6a10f72309a/src/dropdown/dropdown.ts#L66

This is for the `static` case:
https://github.com/ng-bootstrap/ng-bootstrap/blob/1425962d40aff923192abc5e203ef6a10f72309a/src/dropdown/dropdown.ts#L514-L520

And we never set `data-bs-popper` anywhere.

Also, it looks like dropdown menu should be lower, no?

ng-bootstrap:
Screenshot 2022-02-23 at 10 35 08

bootstrap:
Screenshot 2022-02-23 at 10 35 48

2. For the tooltip/popover I can't find the case where we ever set `.bs-popover-auto` classes

@fbasso, could you take a look at it, as you've been working on dropdown most ?

fbasso:
Fix #4261

### DIFF
--- a/src/datepicker/datepicker-input.ts
+++ b/src/datepicker/datepicker-input.ts
@@ -42,6 +42,7 @@ import {NgbDatepickerConfig} from './datepicker-config';
 import {isString} from '../util/util';
 import {take} from 'rxjs/operators';
 import {Subject} from 'rxjs';
+import {addPopperOffset} from '../util/positioning-util';
 
 /**
  * A directive that allows to stick a datepicker popup to an input field.
@@ -394,6 +395,7 @@ export class NgbInputDatepicker implements OnChanges,
             targetElement: this._cRef.location.nativeElement,
             placement: this.placement,
             appendToBody: this.container === 'body',
+            updatePopperOptions: addPopperOffset([0, 2])
           });
         }
       });
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -175,46 +175,6 @@ describe('ngb-dropdown', () => {
     dropdown.close();
   });
 
-  it('should have data-popper attribute reflecting placement', fakeAsync(() => {
-       const html = `
-      <div ngbDropdown [open]="true" placement="bottom-right">
-          <button ngbDropdownAnchor></button>
-          <div ngbDropdownMenu>
-            <a class="dropdown-item">dropDown item</a>
-            <a class="dropdown-item">dropDown item</a>
-          </div>
-      </div>`;
-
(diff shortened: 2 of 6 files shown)


src/util/positioning.ts
```diff
@@ -112,6 +113,8 @@ export function getPopperOptions({placement, baseClass}: PositioningOptions): Pa
       bsModifier,
       flip,
       preventOverflow,
+      offset,
+      {name: 'offset', options: {offset: [0, 2]}},
```
maxokorokov, line 117: but you've added it for all components, not only to the popover...
tooltips are now not positioned correctly, for example

maxokorokov (changes requested):

src/util/positioning.ts
fbasso: Oh yes, you're right @maxokorokov, sorry, I should have known it, as it's already done for the popover. 
New pushed done.

maxokorokov (approved):
LGTM, thanks!

maxokorokov:
As discussed:
- added offsets for all dropdown-based components
- fixed popover offset 12 → 8
to match Bootstrap

PR #4303, opened 2022-04-04
fix(dropdown): disable host button of disabled dropdown item

### LINKED ISSUE #4301: NgbDropdownItem [disabled] not working properly
When I use `ngbDropdownItem` directive on a button element as suggested for keyboard navigation, the `disabled` option is not working properly. It just adds a `disabled` CSS class, but does not add the `disabled` button property. So the disabled option is not clickable or selectable via keyboard navigation, but is still accessible via `Tab` and selectable via `Space` or `Enter`. 
And while the directive is "catching" the disabled property I can't add them to the button by myself.

https://stackblitz.com/edit/angular-vwghsr?file=src/app/dropdown-basic.html

jnizet:
Note that a host binding would work fine on buttons but wouldn't on links,
which don't have a disabled property.

fix #4301

### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -49,11 +49,12 @@ export class NgbDropdownItem {
   @Input()
   set disabled(value: boolean) {
     this._disabled = <any>value === '' || value === true;  // accept an empty attribute as true
+    this._renderer.setProperty(this.elementRef.nativeElement, 'disabled', this._disabled);
   }
 
   get disabled(): boolean { return this._disabled; }
 
-  constructor(public elementRef: ElementRef<HTMLElement>) {}
+  constructor(public elementRef: ElementRef<HTMLElement>, private _renderer: Renderer2) {}
 }
 
 /**
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -40,7 +40,10 @@ export class NgbNavbar {
  *
  * @since 4.1.0
  */
-@Directive({selector: '[ngbDropdownItem]', host: {'class': 'dropdown-item', '[class.disabled]': 'disabled'}})
+@Directive({
+  selector: '[ngbDropdownItem]',
+  host: {'class': 'dropdown-item', '[class.disabled]': 'disabled', '[tabIndex]': 'disabled ? -1 : 0'}
+})
 export class NgbDropdownItem {
   static ngAcceptInputType_disabled: boolean | '';
 
@@ -49,6 +52,9 @@ export class NgbDropdownItem {
   @Input()
   set disabled(value: boolean) {
     this._disabled = <any>value === '' || value === true;  // accept an empty attribute as true
+    // note: we don't use a host binding for disabled because when used on links, it fails because links don't have a
+    // disabled property
+    // setting the property using the renderer, OTOH, works fine in both cases.
     this._renderer.setProperty(this.elementRef.nativeElement, 'disabled', this._disabled);
   }
 
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -298,6 +298,25 @@ describe('ngb-dropdown', () => {
     expect(fixture.componentInstance.isOpen).toBe(false);
     expect(fixture.componentInstance.stateChanges).toEqual([true, false]);
   });
+
+  it('should disable a button dropdown item', () => {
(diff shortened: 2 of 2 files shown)


maxokorokov (commented):
Hey, @jnizet!

Thanks for the PR, I was thinking about this a bit yesterday. The thing is that `ngbDropdownItem` is not limited to buttons, but works on both `<button>` and `<a>` elements. As well as bootstrap styles, and moreover they use mostly `<a>`s in their documentation.

So I have a couple of questions here:
1. should we add `disabled` property on a button → I'd say yes, it is confusing now
2. should we add `disabled` property on a link → probably no, but no harm done I guess?
4. should we "disable" a link with say `tablindex` → not sure

This LGTM for buttons, but I'm not sure about link. WDYT?

jnizet:
Hi @maxokorokov 

My first idea was to test whether the host element was a button before setting the property. But I wasn't sure if that would work fine with SSR. So I tried just setting the property without testing the type of the host element. And apparently, `Renderer2.setProperty()` doesn't do anything if the property doesn't exist. BTW, the PR adds a test with buttons and another test with links, and both pass without emitting any error or warning.

I also tried with a host binding BTW, and that fails with links (but works with buttons).

Regarding tabindex, that's a good point: I don't think anyone will ever want to customize the tab index on dropdown items, so setting it to -1 when disabled and unsetting it when enabled is probably a good idea. 

WDYT?

maxokorokov:
@jnizet, thanks for explanations, didn't know about the renderer.

The more I think about it, the more I confused I get :)

A couple more things I've learned:
- bootstrap emphasise in docs they only make an item "appear" disabled with `.disabled` (visually disabled). And don't care about anything else like functionally or semantically disabling items.
- there is `aria-disabled="true"` we forgot about (semantically disabled)
- to disable (functionally) a link the common recommentdation is to remove the `href` completely, but I don't think we want to mess with `href` (ex. can have routerlinks, can be dynamic, etc)
- bootstrap has confusing keyboard navigation (arrowup/down between non-disabled dropdown items vs Tabs) :(

It seems to me we have two options:

1. Truly disabling an item (straightforward one)
```html
<!-- adding 'disabled' to the button, adding tabindex too for simplicity won't hurt -->
<button class="dropdown-item disabled" disabled>

<!-- removing href from the link is better, but this would do I guess -->
<a href="#" class="dropdown-item disabled" tabindex="-1">
```

bad: not accessible at all for screen readers, neither with Tab nor with arrows

2. Marking an item as disabled and documenting the rest (a11y-focused)

```html
<!-- just adding aria-label and leaving  -->
<button class="dropdown-item disabled" aria-disabled="true">
<a href="#" class="dropdown-item disabled" aria-disabled="true">
```

bad: not fully functionally disabled, can select with keyboard still; confusion with `<button [disabled]="true">` not adding the disabled property.

---

I think at the moment I'd prefer the option 1 and just adding the `tabindex="-1"` in the PR.

---

Just one use-case that won't work is that if somebody wants to have a11y AND arrowup/down keyboard navigation:

```html
<!-- This won't work, because arrowup/down works based on 
@ContentChildren(NgbDropdownItem) that have `disabled` field === false  -->
<a href="#" role="button" ngbDropdownItem [class.disabled]="disabled" [attr.aria-disabled]="disabled">
```
https://github.com/ng-bootstrap/ng-bootstrap/blob/108fa52e727f8539b9b82a20d51d6193b00119cc/src/dropdown/dropdown.ts#L444

jnizet:
@maxokorokov I also prefer option 1. 

But why not combine it with option 2 and also add aria-disabled="true" on the button or link when it's disabled? TBH, I'm still baffled that using disabled binding binds the input of the directive but not the property of the button. I guess it's a feature and not a bug of Angular, but it's very surprising.

maxokorokov:
I think `aria-disabled` won't do much, if you can't focus the element?

jnizet:
Alright, I see. I will add a commit which also sets tabindex.

src/dropdown/dropdown.ts
```diff
@@ -49,11 +52,15 @@ export class NgbDropdownItem {
   @Input()
   set disabled(value: boolean) {
     this._disabled = <any>value === '' || value === true;  // accept an empty attribute as true
+    // note: we don't use a host binding for disabled because when used on links, it fails because links don't have a
```
maxokorokov, line 55: 👍

maxokorokov (approved):
LGTM!

PR #4309, opened 2022-04-11
feat(rating): Allow to set tabindex using [tabIndex].

AMehlem:
Hi,
I am using the rating control and was unable to set the tabindex on the control. Have I missed something or is this pull request required to be able to set the tabindex of the html element?
The change was tested using the demo-site.

Best regards,
Ashwani Mehlem


maxokorokov (changes requested):
Hey, thanks for the PR!

I think you're on the right track for this one.
Just please: 
- [ ] add unit tests
- [ ] add tabindex 0 `NgbRatingConfig` and initialize `NgbRating`'s tabindex with it (as for other inputs)

PR #4324, opened 2022-05-04
Fix popper setup and updates

maxokorokov:
Popper is now created outside Angular and updates are scheduled only after popups are shown

cc @fbasso


src/datepicker/datepicker-input.ts
```diff
@@ -494,7 +496,9 @@ export class NgbInputDatepicker implements OnChanges,
 
   ngOnDestroy() {
     this.close();
-    this._zoneSubscription.unsubscribe();
+    if (this._zoneSubscription) {
```
fbasso, line 499: I would use `this._zoneSubscription?.unsubscribe();` instead of these three lines (same for ngOnDestroy)

maxokorokov: Fixed

PR #4330, opened 2022-05-12
fix(popover): not all classnames are removed after forceupdate

### LINKED ISSUE #4327: Popover classlist gets fludded with extra classes related to placement on document scroll
### Bug description:
Popover (ngb-popover-window) classlist gets fludded with extra classes during document scrolling if one of the placement options contains '-start', '-left', '-end' or '-right' prefix.

https://stackblitz.com/edit/angular-qmskoz?file=src/app/popover-basic.html

Angular: 13.0.3

ng-bootstrap: 12.1.1

Bootstrap: 5

image_2022_05_06T10_42_22_519Z

volvachev:
Fixed fluding popover classlist during updating with extra classes on document scroll

Fixes #4327.


maxokorokov (approved):
Thanks @volvachev !

I've just pushed a commit with a test case for correct class updates that fails without your fix

PR #4336, opened 2022-05-30
fix(positionning): fix start/end placements for popper

fbasso:
The purpose is to have the right placement for widgets using popper when the dir="rtl" is set on the html tag

### DIFF
--- a/demo/src/app/app.module.ts
+++ b/demo/src/app/app.module.ts
@@ -29,6 +29,7 @@ import {PositioningPage} from './pages/positioning/positioning.component';
 import {NgbdSharedModule} from './shared';
 import {NgbdDemoVersionsComponent} from './demo-versions.component';
 import {NgbdOffcanvasModule} from './components/offcanvas/offcanvas.module';
+import {NgbdComponentsSharedModule} from './components/shared';
 
 
 const DEMOS = [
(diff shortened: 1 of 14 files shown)


maxokorokov (commented):
I left 2 comments, could you please rebase and I'll review again :)

src/test/ngb-rtl.ts
```diff
@@ -0,0 +1,20 @@
...
+export class NgbTestLTR {
+  constructor(@Inject(DOCUMENT) private _document: any) {}
+
+  isRTL() { return false; }
+
+  getLang() { return 'en'; }
+}
+
+@Injectable({providedIn: 'root'})
+export class NgbTestRTL {
```
maxokorokov, line 14: This is used nowhere

src/test/ngb-rtl.ts
```diff
@@ -0,0 +1,20 @@
+import {Injectable, Inject} from '@angular/core';
+import {DOCUMENT} from '@angular/common';
+
+@Injectable({providedIn: 'root'})
+export class NgbTestLTR {
```
maxokorokov, line 5: If you remove all `provides` from tests, this will not be used anywhere either

maxokorokov (changes requested):
Hey,

left another review, please see inline comments.

Big things:
1. you shouldn't provide anything special in any of the tests, so `NgbTestRTL` and `NgbTestLTR` are not needed
2. integration tests are missing, meaning we don't have tests that say that "if we have a correct RTL attribute on <html> element, placement classes are correctly generated". Well, at least one test :)
3. We don't need `getLang()` at the moment, so let's remove it

Cheers,
Max

fbasso:
Thanks for your review @maxokorokov, I'm going to see the changes to be done.

PR #4349, opened 2022-07-10
fix(toast): ensure `onStable` always emits outside of the Angular zone

arturovt:


maxokorokov (commented):
I'm sorry, but we'll need more information to continue with this.

- I don't think ng-bootstrap was ever tested with zone-less scenarios
- An issue with reproduction scenario would be helpful (ex. a minimal GitHub project if StackBlitz is not enough)
- This is only a small toast use-case. We have around 15 places where we use `onStable`. Maybe a cetralized way of fixing this would be better?
- What is the testing strategy for these use-cases?
- is this only a performance issue, or a functional issue?

I'm defeintely open for discussion, just need more details.

arturovt:
Hey @maxokorokov 

This change is related to the internal stuff so I don't think that reproducible example will bring some helpful context. Lemme try to describe what's happening.

The `onStable` always emits outside of the Angular zone by default, thus the execution context resolves to the root zone within the subscribe callback:
```ts
ngZone.onStable.subscribe(() => {
  console.log(Zone.current); // <root>
  this._init();
  this.show();
})
```

Which means the code is executed within the `<root>` zone. The `_init` method installs a DOM timer (through `setTimeout`) which would've triggered change detection if it was installed within the Angular zone (currently, it doesn't).

When using `zone-patch-rxjs`, it captures the zone where `subscribe` was called in.
> Note: the patch may be included by doing this
```ts
// polyfills.ts
import 'zone.js';
import 'zone.js/plugins/zone-patch-rxjs';
```

Basically, no matter if the `onStable` emits outside of the Angular (since Angular does `runOutsideAngular(() => onStable.emit(null))` internally). The callback will be always executed within the Angular zone:

```ts
import 'zone.js/plugins/zone-patch-rxjs';

ngZone.onStable.subscribe(() => {
  console.log(Zone.current); // Angular
  this._init();
  this.show();
})
```

This means that toast is doing all of the stuff within the Angular zone (e.g., installing a timer, etc.); this causes unnecessary change detection cycles.

The change in my PR ensures that `onStable` is always called outside of the Angular zone, thus the behavior is always consistent with and without RxJS patch.

---

I didn't touch other places because I'm making this PR for the first time; it's easier to discuss the change in the scope of this single change.

PR #4370, opened 2022-08-08
feat(collapse): add horizontal collapse

ValentinNelu:
I have added the horizontal option to the collapse widget.
It has the same requirement as in [Bootstrap 5](https://getbootstrap.com/docs/5.2/components/collapse/#horizontal): the immediate child elements must have a width.

### DIFF
--- a/demo/src/app/components/collapse/collapse.module.ts
+++ b/demo/src/app/components/collapse/collapse.module.ts
@@ -1,16 +1,22 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
-import { NgModule } from '@angular/core';
+import {NgModule} from '@angular/core';
 
-import { NgbdSharedModule } from '../../shared';
-import { ComponentWrapper } from '../../shared/component-wrapper/component-wrapper.component';
-import { NgbdComponentsSharedModule, NgbdDemoList } from '../shared';
-import { NgbdApiPage } from '../shared/api-page/api.component';
-import { NgbdExamplesPage } from '../shared/examples-page/examples.component';
-import { NgbdCollapseBasic } from './demos/basic/collapse-basic';
-import { NgbdCollapseBasicModule } from './demos/basic/collapse-basic.module';
-import { NgbdCollapseNavbar } from './demos/navbar/collapse-navbar';
-import { NgbdCollapseNavbarModule } from './demos/navbar/collapse-navbar.module';
-import { Routes } from '@angular/router';
+import {NgbdSharedModule} from '../../shared';
+import {
+  ComponentWrapper
+} from '../../shared/component-wrapper/component-wrapper.component';
+import {NgbdComponentsSharedModule, NgbdDemoList} from '../shared';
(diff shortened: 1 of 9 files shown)


src/util/transition/ngbCollapseTransition.ts
```diff
@@ -18,19 +19,22 @@ function measureCollapsingElementHeightPx(element: HTMLElement): string {
...
 
   if (!hasShownClass) {
     classList.remove('show');
   }
 
-  return height;
+  return dimensionSize;
 }
 
 export const ngbCollapsingTransition: NgbTransitionStartFn<NgbCollapseCtx> =
```
maxokorokov, line 32: You're not cleaning up the `width` in the return of this function, only `height`.

maxokorokov (changes requested):
Hey, LGTM overall, thanks for adding the demo.

There is a couple of small issues:
1. you're not clearing up the width, it stays in the DOM after the 
2. transition "feels" different when I compare it with the one on Bootstrap's site demo. As if the container width is wrong. It's much faster on ng-bootstrap demo side, especially when showing the content.

I've also left a couple of minor comments on the code.

Cheers,
Max

demo/src/app/components/collapse/collapse.module.ts
```diff
@@ -19,6 +25,14 @@ const DEMOS = {
     code: require('!!raw-loader!./demos/basic/collapse-basic').default,
     markup: require('!!raw-loader!./demos/basic/collapse-basic.html').default
   },
+  vertical: {
```
jnizet, line 28: nit: rename this to 'horizontal'

src/util/transition/ngbCollapseTransition.ts
```diff
@@ -48,30 +49,40 @@ export const ngbCollapsingTransition: NgbTransitionStartFn<NgbCollapseCtx> =
...
-
-        reflow(element);
 
         // Start the animation
         classList.add('collapsing');
-      }
 
-      // Start or revert the animation
-      element.style.height = direction === 'show' ? maxHeight : '0px';
+        const scrollDimension = `scroll${dimension[0].toUpperCase()}${dimension.slice(1)}`;
```
maxokorokov, line 60: Fell free to ignore this. Personally I would find it simpler/more readable to do something like this:

```ts
const SCROLL_DIMENSIONS = {
  'height': 'scrollHeight',
  'width': 'scrollWidth'
};

element.style[dimension] = element[SCROLL_DIMENSIONS[dimension]] + 'px';
```

maxokorokov (approved):
LGTM, thanks!

I've just pushed a commit with unit test for `collapse-horizontal` class and added a small remark.

P.S. I'll merge unless you want to change something, but be careful as I've rebased and force-pushed a commit to your remote.

src/util/transition/ngbCollapseTransition.ts
ValentinNelu: I would agree with you, but I wanted to remain consistent with the Bootstrap code as much as possible, that's why I went for this approach

PR #4374, opened 2022-08-10
fix(collapse): solved running transition twice

ValentinNelu:
I have removed the OnChange in the ngbCollapse directive to avoid starting the animation twice.
I have used a setter and a flag. The setter executes the "_runTransitionWithEvents" while the flag is used to avoid running the animation during initialisation.


maxokorokov (approved):
LGTM, thanks!

I've rebased, there were conflicts.
Will merge when CI is green

PR #4379, opened 2022-08-17
demo: migrate nav and sidenav to offcanvas

Dafnik:
Hello!

I've updated the ng-bootstrap demo page to a more modern look and to be more similar to the new bootstrap docs design.

The pr is mostly restructuring work and implementing bootstrap docs scss.
I hope this isn't frowned upon.

Would be glad for feedback or suggestions! Also if this isn't needed or wanted, no worries.

Sreenshots:
image

image

image

image

image

image

image

image

image

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -77,7 +77,7 @@ The most useful commands are:
 
 #### `yarn demo`
 
-Serves the demo site locally in dev mode at [http://localhost:4200/](http://localhost:4200/). You can optionally add `--prod` argument to serve demo in production mode or `--aot` to serve demo in dev mode, but with AOT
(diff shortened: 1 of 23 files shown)


maxokorokov (commented):
Hey,

thanks for taking the time to tinker with the demo site.

I agree that overall this should improve experience, especially on mobile.

However, there is a couple of points I can make:
- there are too many changes in 1 PR to review
- offcanvas looks unfinished (paddings, margins, the fact that you can't see subpages, icon colors that don't really fit to exising site)
- I'm not a fan of copying styles as-is with copyrights from the Boostrap's demo site

Don't get me wrong, I like most of the changes overall. It makes things lighter, removes borders, etc. But it needs more polishing and a more step-by-step approach for me.

I'm going to close this one, but we can continue the discussion, if you want to.

Cheers,
Max

PR #4388, opened 2022-09-15
fix(collapse): restore previous transition logic

divdavem:
This PR restores the previous collapse transition logic that was changed with #4370.

Without this PR, the show and hide phases are inconsistent: the hide phase is from the full size of the container to `0px`, and the show phase is from `0px` to the `scrollWidth`/`scrollHeight` of the container and then there is a sudden change when the classes are changed at the end of the transition.

Using `scrollWidth` / `scrollHeight` does not feel right for me. We should use whatever size the container will have when changing the classes at the end of the transition, which is what `measureCollapsingElementDimensionPx` measures correctly. I have added `style="max-width:300px;"` on the container in the demo so that the animation matches the visible content.

### DIFF
--- a/demo/src/app/components/collapse/demos/horizontal/collapse-horizontal.html
+++ b/demo/src/app/components/collapse/demos/horizontal/collapse-horizontal.html
@@ -5,7 +5,7 @@
   </button>
 </p>
 <div style="min-height: 100px;">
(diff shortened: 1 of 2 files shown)


maxokorokov (approved):
LGTM, thanks. It is better like this indeed

divdavem:
@maxokorokov Thank you for reviewing and merging my PR!

PR #4408, opened 2022-10-24
demo: use `tab-size: 2` for demo examples after Prettier migration

maxokorokov:


maxokorokov:
Screenshot 2022-10-24 at 16 10 01

VS

Screenshot 2022-10-24 at 16 09 37

PR #4409, opened 2022-10-24
fix(typeahead): remove aria-multiline attribute

### LINKED ISSUE #4181: The aria-multiline attribute is invalid on combobox role
### Bug description:
When using the the typeahead component `aria-multiline="false"` is added to the input but is not allowed since is valid only for the `textbox` role while by using the typeahead component the role is `combobox`.
This is an accesibility issue because elements must only use allowed ARIA attributes.
  

https://stackblitz.com/run?file=app/typeahead-basic.ts

Angular: 10.2.3

ng-bootstrap: 7.0.0

Bootstrap: 4.3.1

stephanieleary:
Removes `aria-multiline="false"`, which is invalid on comboboxes.

Fixes #4181

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -796,7 +796,6 @@ describe('ngb-typeahead', () => {
 			fixture.detectChanges();
 
 			expect(input.getAttribute('role')).toBe('combobox');
-			expect(input.getAttribute('aria-multiline')).toBe('false');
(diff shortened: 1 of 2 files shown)


maxokorokov (changes requested):
LGTM, could you just update the test as well?

https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/typeahead/typeahead.spec.ts#L799

stephanieleary:
Thanks for pointing that out!

maxokorokov (approved):

PR #4411, opened 2022-10-25
refactor(toast): update toast transitions to Bootstrap 5.2

erikbarke:
Toast animations have been broken since the update to Bootstrap v5.2, which changed it's internal api for transitions. This PR mimics the behavior of the logic in bootstrap.js and removes the use of the deprecated class `hide`.


maxokorokov (approved):
LGTM, thanks! 
A bit strange that demos are fading in on page open, but it matches BS. We missed this one indeed...

PR #4425, opened 2022-11-20
feat: upgrade to Angular 15

jnizet:
This is a draft PR that upgrades the project to Angular 15. 

Here's what has been done:

- upgrade using the `ng update` tool
- remove the calls to `enableProdMode()` which is now useless
- delete and regenerate the yarn.lock file because some dependencies are obsolete
- remove the polyfills.ts files (except one, see below) and replace them by polyfills entries directly in the angular.json file. That's how it's done for new projects created by the CLI
- add "@angular/localize" to the types of the tsconfig.json. That's what `ng add @angular/localize` does now
- fix the ComponentWrapper component in the demo, which caused an NG0100 error (due to CD now running before the component is added next to the outlet)
- remove the `types` overwrite in the ssr-app/tsconfig.json file, because the parent tsconfig.json file adds the localize type which is needed.
- fix the parse-demo.ts file because decorators aren't handled the same way in TS 4.8
- fix the setup-project.spec.ts in the schematics because calling the `ng add @angular/localize` doesn't have the same effect as before: it now updates the tsconfig.json instead of udating the polyfills.ts file (which doesn't exist anymore)
- add types for jasmine and node in the schematics tsconfig, otherwise
It's a draft because of the following issues:

- the custom-webpack builder does not support Angular 15 yet. So I had to add resolutions in package.json that should be removed once it supports angular 15
- due to this lack of support, I had to leave the polyfills.ts file in the e2e-app/src directory. Once the custom-webpack builder supports Angular 15, we should remove this file, remove its reference in e2e-app/tsconfig.json, and replace its reference in angular.json by `["zone.js"]`

I'm not sure if the CI will pass because locally, on macOS, when I run `yarn ci`, the ssr part fails because the demo still runs and playwright conects to the demo rather than the ssr app. 


demo/src/app/shared/component-wrapper/component-wrapper.component.ts
```diff
@@ -84,37 +84,39 @@ export class ComponentWrapper implements OnDestroy {
...
-					title: demo.title,
-				};
-			});
-		} else if (component instanceof NgbdApiPage) {
-			let toc = getLinks(component.components);
-
-			if (component.classes.length > 0) {
-				const klasses = getLinks(component.classes);
-				toc = toc.concat(toc.length > 0 ? [<any>{}, ...klasses] : klasses);
+		setTimeout(() => {
```
samvloeberghs, line 87: Just curious, but why is the setTimeout needed here?

jnizet: I have not digged too much, but I'm pretty sure it's because of this change: https://angular.io/guide/update-to-version-15#v15-bc-09

This causes a NG0100 error, because this method is called when the router is activated and causes new changes.

maxokorokov: I did a `queueMicrotask()` in #4428, but haven't checked the profile on what is better to do

maxokorokov (approved):
Hey, @jnizet!

LGTM, I haven's seen this one and opened #4428, but this one is better than mine as a base for Angular 15 migration.

I missed changes to polyfills and localized on my side, but have a couple of more minor fixes.

Could you maybe:
- rollback to playwright `1.27`
- add eslint `15` packages that are available now

We'll merge this one and deal with `custom-webpack` later when it is avalable

I'll rebase  #4428 with standalone components and will do an RC release. WDYT?

Also if you have any opinion on https://github.com/ng-bootstrap/ng-bootstrap/pull/4428/commits/0c8de6492d74505c141abc04b65be68d59bdbbc2, that would be helpful!

maxokorokov:
Also https://github.com/ng-bootstrap/ng-bootstrap/blob/master/.github/workflows/ci.yml#L28 `scripts:test` will be failing, as API doc extraction is broken. I'll be fixing that separately I think.

Now:
Screenshot 2022-11-23 at 14 07 40

After TS 4.8 migration:
Screenshot 2022-11-23 at 14 08 02

jnizet:
Hi @maxokorokov 
I've just pushed the changes you asked for. 
I have no opinion on setTimeout vs queueMicrotask, so feel free to change that if you know what's best.
I'll review your commit when I find some time.

PR #4428, opened 2022-11-23
Angular 15 and standalone components

maxokorokov:

 - bumping to Angular 15, Bootstrap 5.3.2 and Popper 2.11.6
 - migrating to standalone components
 - temporarily disabling schematics and e2e coverage


demo/src/app/shared/api-docs/api-docs-config.component.ts
```diff
@@ -19,7 +19,7 @@ const CONFIG_SUFFIX_LENGTH = 'Config'.length;
 @Component({
 	selector: 'ngbd-api-docs-config',
 	standalone: true,
-	imports: [RouterLinkWithHref, NgbdApiDocsBadge, NgForOf, NgIf, NgbdFragment],
+	imports: [RouterLink, NgbdApiDocsBadge, NgForOf, NgIf, NgbdFragment],
```
jnizet, line 22: NgForOf should be replaced with NgFor (everywhere)

maxokorokov: Fixed with #4433

PR #4430, opened 2022-11-24
Angular 15 fixes

maxokorokov:
More minor fixes for Angular 15


demo/src/main.ts
```diff
@@ -13,6 +12,6 @@ bootstrapApplication(AppComponent, {
 			provide: LocationStrategy,
 			useClass: HashLocationStrategy,
 		},
-		importProvidersFrom([HttpClientModule]),
+		provideHttpClient(),
```
jnizet, line 15: nit: we could use `provideHttpClient(withNoXsrfProtection())`

misc/api-doc.ts
```diff
@@ -63,6 +83,32 @@ function isInternalMember(member) {
...
+		isIdentifier(decorator.expression.expression) &&
+		types.includes(decorator.expression.expression.text)
+	);
+}
+
+function getDecoratorOfType(node: Node, decoratorType: string): Decorator | null {
+	const decorators = canHaveDecorators(node) ? getDecorators(node) || [] : [];
+
+	for (const decorator of decorators) {
+		if (isCallExpression(decorator.expression) && isIdentifier(decorator.expression.expression)) {
```
jnizet, line 98: Couldn't we use `isDecoratorOfType` here rather than duplicating the logic?

misc/api-doc.ts
```diff
@@ -130,22 +150,21 @@ class APIDocVisitor {
...
+			} else if (isInterfaceDeclaration(statement)) {
 				return directivesSoFar.concat(this.visitInterfaceDeclaration(fileName, statement));
 			}
 
 			return directivesSoFar;
 		}, []);
 	}
 
-	visitInterfaceDeclaration(fileName, interfaceDeclaration) {
+	visitInterfaceDeclaration(fileName, interfaceDeclaration: InterfaceDeclaration) {
```
jnizet, line 163: fileName: string

misc/api-doc.ts
```diff
@@ -314,80 +362,43 @@ class APIDocVisitor {
...
-			defaultValue: property.initializer ? this.stringifyDefaultValue(property.initializer) : undefined,
-			type: this.visitType(property),
-			description: displayPartsToHtml(property.symbol.getDocumentationComment(this.typeChecker)),
+			...this.visitNamedDeclaration(property),
+			defaultValue: property.initializer?.getText(),
 		};
 	}
 
-	visitType(node) {
+	visitType(node: Node & { type? }) {
```
jnizet, line 395: missing type for type?

jnizet:
Also, in my Angular 15 upgrade PR, I commented:

> I had to leave the polyfills.ts file in the e2e-app/src directory. Once the custom-webpack builder supports Angular 15, we should remove this file, remove its reference in e2e-app/tsconfig.json, and replace its reference in angular.json by ["zone.js"]

Since the custom webpack builder now supports Angular 15, I guess we could now do the above (but I haven't tested it)

misc/api-doc.ts
maxokorokov: Will check that, thanks

PR #4437, opened 2022-12-01
[WIP] feat(accordion): wip accordion directive

ValentinNelu:

### DIFF
--- a/src/accordion/accordion.directive.ts
+++ b/src/accordion/accordion.directive.ts
@@ -1,0 +1,322 @@
+import {
+	AfterContentInit,
+	ChangeDetectorRef,
+	Component,
+	ContentChild,
+	ContentChildren,
+	Directive,
+	EventEmitter,
+	forwardRef,
+	Inject,
+	Input,
+	OnDestroy,
+	Output,
+	QueryList,
+	TemplateRef,
+} from '@angular/core';
+import { Subscription } from 'rxjs';
+import { NgbAccordionConfig } from './accordion-config';
+import { NgTemplateOutlet } from '@angular/common';
+import { NgbCollapse } from '../collapse/collapse';
+
+let nextId = 0;
+
+@Component({
+	selector: '[ngbAccordionBody]',
+	standalone: true,
+	imports: [NgTemplateOutlet],
+	host: { '[class.accordion-body]': 'true' },
+	template: ` <ng-template [ngTemplateOutlet]="template()" ]></ng-template> `,
+})
+export class NgbAccordionBody {
+	constructor(@Inject(forwardRef(() => NgbAccordionItem)) private _item: NgbAccordionItem) {}
+
+	@ContentChild(TemplateRef, { static: true }) bodyTpl: TemplateRef<any>;
+
+	template() {
+		return this._item.destroyOnHide === false || this._item.animatingBodyCollapse ? this.bodyTpl : null;
+	}
+}
+
+@Directive({
+	exportAs: 'ngbAccordionCollapse',
+	standalone: true,
+	selector: '[ngbAccordionCollapse]',
+	host: {
+		role: 'region',
+		'[class.accordion-collapse]': 'true',
+		'[id]': 'item.collapseId',
+		'[attr.aria-labelledby]': 'item.toggleId',
+	},
+	hostDirectives: [
+		{
+			directive: NgbCollapse,
+		},
+	],
+})
+export class NgbAccordionCollapse {
+	constructor(
+		@Inject(forwardRef(() => NgbAccordionItem)) public item: NgbAccordionItem,
+		public ngbCollapse: NgbCollapse,
+	) {}
+}
+
+@Directive({
+	selector: '[ngbAccordionToggle]',
+	standalone: true,
+	host: {
+		'[id]': 'item.toggleId',
+		'[disabled]': 'item.disabled',
+		'[class.accordion-button]': 'true',
+		'[class.collapsed]': 'item.collapsed',
+		'[attr.aria-controls]': 'item.collapseId',
+		'[attr.aria-expanded]': '!item.collapsed',
+		type: 'button',
+		'(click)': '!item.disable ? accordion.toggle(item.id): undefined',
+	},
+})
+export class NgbAccordionToggle {
+	constructor(
+		@Inject(forwardRef(() => NgbAccordionItem)) public item: NgbAccordionItem,
+		@Inject(forwardRef(() => NgbAccordionDirective)) public accordion: NgbAccordionDirective,
+	) {}
+}
+
+@Directive({
+	selector: '[ngbAccordionHeader]',
+	standalone: true,
+	host: {
+		role: 'heading',
+		'[class.accordion-header]': 'true',
+	},
+})
+export class NgbAccordionHeader {
+	constructor() {}
+}
+
+@Directive({
+	selector: '[ngbAccordionItem]',
+	standalone: true,
+	host: {
+		'[class.accordion-item]': 'true',
+		'[id]': 'id',
+	},
+})
+export class NgbAccordionItem implements AfterContentInit, OnDestroy {
+	constructor(
+		@Inject(forwardRef(() => NgbAccordionDirective)) private _accordion: NgbAccordionDirective,
+		private _cd: ChangeDetectorRef,
+	) {}
+
+	private _subscriptions: Subscription[] = [];
+	private _collapsed = true;
+	private _id = `ngb-accordion-item-${nextId++}`;
+	private _disabled = false;
+
+	animatingBodyCollapse = false;
+
+	@ContentChild(NgbAccordionCollapse, { static: true }) private _collapse: NgbAccordionCollapse;
+
+	@Input('ngbAccordionItem') set id(inputId: string) {
+		if (inputId) {
+			this._id = inputId;
+		}
+	}
+	@Input() destroyOnHide = this._accordion.destroyOnHide;
+	@Input() set disabled(isDisabled: boolean) {
+		if (isDisabled != this.disabled) {
+			this._disabled = isDisabled;
+		}
+	}
+	@Input() set collapsed(collapsed: boolean) {
+		if (this.collapsed !== collapsed) {
+			this._collapsed = collapsed;
+			// we need force CD to get template into DOM before starting animation to calculate its height correctly
+			if (!this._collapsed) {
+				this.animatingBodyCollapse = true;
+				this._cd.detectChanges();
+			}
+			// we also need to make sure 'animation' flag is up-to- date
+			this._collapse.ngbCollapse.animation = this._accordion.animation;
+			this._collapse.ngbCollapse.collapsed = collapsed;
+		}
+	}
+
+	@Output() shown = new EventEmitter<void>();
+	@Output() hidden = new EventEmitter<void>();
+
+	get collapsed(): boolean {
+		return this._collapsed;
+	}
+	get disabled() {
+		return this._disabled;
+	}
+	get id() {
+		return `${this._id}`;
+	}
+
+	get toggleId() {
+		return `${this.id}-toggle`;
+	}
+	get collapseId() {
+		return `${this.id}-collapse`;
+	}
+
+	ngAfterContentInit(): void {
+		const { ngbCollapse } = this._collapse;
+		// we need to disable the animation for the first init
+		ngbCollapse.animation = false;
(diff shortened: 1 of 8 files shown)


maxokorokov:
```html
<div
	ngbAccordion
	[animation]="animation"
	[closeOthers]="closeOthers"
	[destroyOnHide]="true"
	(shown)="log('shown ' + $event)"
	(hidden)="log('hidden ' + $event)"
>
<!--	ngb-accordion-header-XXX-->
<!--	ngb-accordion-collapse-XXX-->
	<div ngbAccordionItem [destroyOnHide]="true" [disabled] (shown) (hidden) [collapse]="true">
		<h2 ngbAccordionHeader>
			<button ngbAccordionToggle>Accordion Item #1</button>
		</h2>
		<div ngbAccordionCollapse>
			<div ngbAccordionBody>
				<ng-template>
					<strong>This is the first item's accordion body.</strong> It is shown by default, until the collapse plugin
					adds the appropriate classes that we use to style each element. These classes control the overall appearance,
					as well as the showing and hiding via CSS transitions. You can modify any of this with custom CSS or
					overriding our default variables. It's also worth noting that just about any HTML can go within the
					<code>.accordion-body</code>, though the transition does limit overflow.
				</ng-template>
			</div>
		</div>
	</div>
</div>
```

ValentinNelu:
Need to add proper documentation, and do the demo page.
Should we split it in a different module or keep all in the ngb-accordion module?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+import { NgbCollapse } from '../collapse/collapse';
+
+let nextId = 0;
+
+@Component({
+	selector: '[ngbAccordionBody]',
+	standalone: true,
+	imports: [NgTemplateOutlet],
+	host: { '[class.accordion-body]': 'true' },
+	template: ` <ng-template [ngTemplateOutlet]="addToDOM() ? bodyTpl : null" ]></ng-template> `,
```
maxokorokov, line 28: idea: maybe it is even better to extract it into something like 

```html
<ng-template [ngTemplateOutlet]="template()"></ng-template>
```

WDYT?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+})
+export class NgbAccordionBody {
+	constructor(@Inject(forwardRef(() => NgbAccordionItem)) private item: NgbAccordionItem) {}
+
+	@ContentChild(TemplateRef) bodyTpl: TemplateRef<any>;
+
+	addToDOM() {
+		return this.item.destroyOnHide === false || this.item.animatingBodyCollapse;
```
maxokorokov, line 36: `this.item.destroyOnHide` is only initialized once, so it might be wrong I guess.

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+	selector: '[ngbAccordionBody]',
+	standalone: true,
+	imports: [NgTemplateOutlet],
+	host: { '[class.accordion-body]': 'true' },
+	template: ` <ng-template [ngTemplateOutlet]="addToDOM() ? bodyTpl : null" ]></ng-template> `,
+})
+export class NgbAccordionBody {
+	constructor(@Inject(forwardRef(() => NgbAccordionItem)) private item: NgbAccordionItem) {}
+
+	@ContentChild(TemplateRef) bodyTpl: TemplateRef<any>;
```
maxokorokov, line 33: Probably should have `{static: true}` option

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+		return this.item.destroyOnHide === false || this.item.animatingBodyCollapse;
+	}
+}
+
+@Directive({
+	exportAs: 'ngbAccordionCollapse',
+	standalone: true,
+	selector: '[ngbAccordionCollapse]',
+	host: {
+		role: 'region',
```
maxokorokov, line 45: Will a user be able to override a `role` is necessary?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+	selector: '[ngbAccordionToggle]',
+	standalone: true,
+	host: {
+		'[id]': 'item.toggleId',
+		'[disabled]': 'item.disabled',
+		'[class.accordion-button]': 'true',
+		'[class.collapsed]': 'item.collapsed',
+		'[attr.aria-controls]': 'item.collapseId',
+		'[attr.aria-expanded]': '!item.collapsed',
+		'[type]': '"button"',
```
maxokorokov, line 73: just `'type': 'button'` ?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+
+@Directive({
+	selector: '[ngbAccordionItem]',
+	standalone: true,
+	host: {
+		'[class.accordion-item]': 'true',
+		'[id]': 'id',
+	},
+})
+export class NgbAccordionItem implements AfterContentInit {
```
maxokorokov, line 104: Must implement `OnDestroy` and unsubscribe, as you're leaking subscriptions (never unsubscribing)

Something like:
```ts
for (let s of this._subscriptions) {
  s?.unsubscribe();
}
```

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+	private _collapsed = true;
+	private _id = `${nextId++}`;
+	private _disabled = false;
+
+	animatingBodyCollapse = false;
+
+	@ContentChild(NgbAccordionCollapse, { static: true }) private collapse: NgbAccordionCollapse;
+
+	@Input('ngbAccordionItem') set id(inputId: string) {
+		if (inputId) {
```
maxokorokov, line 120: Will it work for `id` `'0'` ?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+			this._id = inputId;
+		}
+	}
+	@Input() destroyOnHide = this.accordion.destroyOnHide;
```
maxokorokov, line 124: Now, what if it changes at the accordion level via binding at runtime? The expression in `NgbAccorionBody` will be wrong I think.

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+		if (isDisabled != this.disabled) {
+			this._disabled = isDisabled;
+			if (this._disabled) {
+				this._collapsed = true;
+				this.collapse.ngbCollapse.collapsed = true;
+			}
+		}
+	}
+	@Input() set collapsed(collapsed: boolean) {
+		if (this.disabled) {
```
maxokorokov, line 135: This seems strange to me, seems redundant... is this `if` and the nested one because it might change via the `@Input()` binding?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+	@Output() shown = new EventEmitter<void>();
+	@Output() hidden = new EventEmitter<void>();
+
+	get collapsed(): boolean {
+		return this._collapsed;
+	}
+	get disabled() {
+		return this._disabled;
+	}
+	public get id() {
```
maxokorokov, line 161: `public` is redundant

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+		return `ngb-accordion-item-${this._id}`;
+	}
+
+	get toggleId() {
```
maxokorokov, line 165: Maybe just `{this.id}-toggle` (note the `id`, not `_id`) ?

It will give:

```
ngb-accordion-item-1
ngb-accordion-item-1-toggle
ngb-accordion-item-1-collapse
```

And I think it is better, because toggle and collapse are nested in the item

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+	ngAfterContentInit(): void {
+		const { ngbCollapse } = this.collapse;
+		// we need to disable the animation for the first init
+		ngbCollapse.animation = false;
+		ngbCollapse.collapsed = this.collapsed;
+		// we set the animation to the default of the accordion
+		ngbCollapse.animation = this.accordion.animation;
+		// event forwarding from 'ngbCollapse' to 'ngbAccordion'
+		this.subscriptions.push(
+			ngbCollapse.hidden.subscribe(() => {
```
maxokorokov, line 181: no unsubscribe for both

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+			}),
+			ngbCollapse.shown.subscribe(() => {
+				this.shown.emit();
+				this.accordion.shown.emit(this.id);
+			}),
+		);
+	}
+
+	toggle() {
+		this.collapsed = !this.collapsed;
```
maxokorokov, line 195: Hmm, what about the fact is can be disabled? You're doing quite a lot of checks in `set collapse()`, but not here?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+		this.closeOthers = config.closeOthers;
+	}
+
+	/**
+	 * Toggles a panel with the given id.
+	 *
+	 * Has no effect if the panel is disabled.
+	 */
+	toggle(itemId: string) {
+		if (this.canBeToggled(itemId)) {
```
maxokorokov, line 243: doing `this.canBeToggled` seemed like a lot of redundant work, but indeed I don't see how to do it better...

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+					item.collapsed = false;
+				} else if (this.closeOthers) {
+					item.collapsed = true;
+				}
+			});
+		}
+	}
+
+	expandAll() {
+		if (this.closeOthers) {
```
maxokorokov, line 272: Seems complex, isn't this doing the same?

```ts
for (const item of this._items) {
	if (!item.disabled) {
		item.collapsed = false;
		if (this.closeOthers) {
			break;
		}
	}
```

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+			const available = this.items.filter((item) => !item.disabled);
+			//we check if there are any item not disabled and none of them is expanded
+			if (available.length > 0 && !available.find((item) => !item.collapsed)) {
+				available[0].collapsed = false;
+			}
+		} else {
+			this.items.forEach((item) => {
+				if (!item.disabled) {
```
maxokorokov, line 280: I'm not sure about `disabled` handling again... why can't you work with `disabled` items programmatically? I understand that the user can't do anything with them, but why do APIs not allow it?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+					item.collapsed = false;
+				}
+			});
+		}
+	}
+
+	collapse(itemId: string) {
+		this.items.find((item) => item.id === itemId)!.collapsed = true;
```
maxokorokov, line 288: It should fail when `find` finds no matching item. Tests missing?

src/accordion/accordion.directive.ts
```diff
@@ -0,0 +1,305 @@
...
+	/**
+	 * Checks if a panel with a given id is expanded.
+	 */
+	isExpanded(panelId: string): boolean {
+		const item = this.items.find((item) => item.id === panelId);
+		return item ? !item.collapsed : false;
+	}
+
+	private canBeToggled(itemId: string) {
+		return this.items.find((item) => item.id === itemId && !item.disabled) ? true : false;
```
maxokorokov, line 303: nit: might be simpler with `!!this._items.find...`

maxokorokov (changes requested):
Hey, @ValentinNelu, looks good overall!

Left some comments (details in code comments):
1. Please use underscore for all private vars `private _varName`, not `private varName`. Sorry, it's the repository code style.
2. I'm not convinces with how `disabled` is handled, also the demo doesn't seem to work to me? You should be able to open/ close disabled panels via API, but user shouldn't. Some old tests checking this are missing: ex. we've had `it('should not open disabled panels from click')` test and `it('should open/collapse disabled panels')`. Now you seem to have different behavior.(let's discuss it)
3. Subscriptions are leaking
4. Seems like `destroyOnHide` might not work properly at runtime

I haven't checked test code coverage and API coverage yet.
Will do after your fixes!

Cheers,
Max

PR #4453, opened 2023-01-03
feat(rating): add customization to aria-valuetext

ValentinNelu:
Small PR to add the possibility to change the aria-valuetext in the rating component.

### DIFF
--- a/demo/src/app/components/rating/demos/decimal/rating-decimal.ts
+++ b/demo/src/app/components/rating/demos/decimal/rating-decimal.ts
@@ -29,4 +29,8 @@ import { NgbRatingModule } from '@ng-bootstrap/ng-bootstrap';
 })
 export class NgbdRatingDecimal {
 	currentRate = 3.14;
+
+	ariaValueText(current: number, max: number) {
+		return `${current} out of ${max} hearts`;
+	}
 }
--- a/src/rating/rating.ts
+++ b/src/rating/rating.ts
@@ -50,7 +50,7 @@ export interface StarTemplateContext {
 		'aria-valuemin': '0',
 		'[attr.aria-valuemax]': 'max',
 		'[attr.aria-valuenow]': 'nextRate',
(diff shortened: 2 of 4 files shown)


src/rating/rating.ts
```diff
@@ -100,6 +100,13 @@ export class NgbRating implements ControlValueAccessor, OnInit, OnChanges {
 	 */
 	@Input() resettable: boolean;
 
+	/**
+	 * Allows setting a custom function to provide aria text
+	 */
+	@Input() set ariaText(f: (rating: NgbRating) => string) {
```
maxokorokov, line 106: why not simply this? why setters and private values?

`@Input() ariaText: (current: number, max: number) => string;`

demo/src/app/components/rating/demos/decimal/rating-decimal.ts
```diff
@@ -29,4 +29,8 @@ import { NgbRatingModule } from '@ng-bootstrap/ng-bootstrap';
 })
 export class NgbdRatingDecimal {
 	currentRate = 3.14;
+
+	ariaText(r: NgbRating) {
+		return `${r.nextRate} out of ${r.max} hearts`;
```
maxokorokov, line 34: `nextRate` is not a public API, so we shouldn't show it in demos. That's also why I', not a fan of passing `NgbRating`

src/rating/rating.ts
maxokorokov: It also should stay `ariaValueText` otherwise it's not clear which text... I would just modify existing function, as it is not technically a public API anyway.

maxokorokov (changes requested):
Hey, @ValentinNelu, thanks!

Added a couple of comments about the API

maxokorokov (approved):
LGTM, thanks!
Looks much simpler now!

Will merge when CI is green

PR #4471, opened 2023-01-20
feat(dropdown): replace NgbNavbar usage by DOM call to closest()

### LINKED ISSUE #4462: NgbNavbar is not documented
### Bug description:

For dropdowns to look correctly inside a collapsed navbar, the `NgbNavbar` directive is necessary. It's automatically applied when importing `NgbDropdownModule` (because its selector is simply `.navbar`), but when using standalone components, it's far from obvious that this directive needs to be imported, since it's not even documented.

jnizet:
The NgbNavbar directive is not removed, but simply deprecated, in order to avoid a breaking change for people who have imported it in their modules or standalone components.

fix #4462

### DIFF
--- a/src/dropdown/dropdown.module.ts
+++ b/src/dropdown/dropdown.module.ts
@@ -1,12 +1,5 @@
 import { NgModule } from '@angular/core';
-import {
-	NgbDropdown,
-	NgbDropdownAnchor,
-	NgbDropdownToggle,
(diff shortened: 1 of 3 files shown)


jnizet:
Note: I wondered if using a call to closest() would be an issue in a server-side environment, which I don't have any experience with, but Angular Universal now apparently uses domino, which should thus support the call.

Issue #4498, opened 2023-05-04
Add support for Angular 16


saerx:
### Bug description:

I'm sure it is in the works, but thought I would add the request for a pre-release version of ng-bootstrap that supports Angular 16 now that the latter has been released.

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 16 

Bootstrap: any

abrahampe:
There is a PR on the road [feat: upgrade to Angular 16](https://github.com/ng-bootstrap/ng-bootstrap/pull/4501)

ramonleenders:
> There is a PR on the road [feat: upgrade to Angular 16](https://github.com/ng-bootstrap/ng-bootstrap/pull/4501)

It hasn't been updated since creating the PR not replied upon. Is this still in active development? Would have thought it would be a bit quicker to release, since it is a pretty straight forward upgrade Angular wise? Got a project which I'd want to update to Angular 16, but this is a bit in its way due to CI/CD.

GabrielMJin:
i updated my project, and it stopped picking up the lib

abrahampe:
> i updated my project, and it stopped picking up the lib

So I did! I had to rollback to NG 15 again :-(

md13381:
The problem is probably angular-builders hasnt been updated yet, which is probably delaying things

klexas:
Manually installing via npm for example and not the ng builder works with some manual importing. But it's really frustrating as out of the box, its not compatible.

klexas:
> Thanks for this! So is 14.2 compatible with Angular 16?

It is compatible, but you need to -force the install. 

I didn't see a `-force` option on the `ng add` so I had to use the `npm install` option - same packages, just a bit finer grain control. 

Forcing the packages (just ignores the version <16 warnings) I can confirm works perfectly.

detobel36:
The new version on the master branch
Release version (14.2.0) is not compatible yet

ClaasWilke:
I think it is a bit confusing that the issue was close althoug ng-bootstrap 14.2.0 is still not fully compatible with Angular 16. Building Angular projects with --force or --legacy-peer-dep is no real option for your CI/CD when handling multiple dependencies or working with dependency management tools like Renovate.

pbarranis:
[...]

PR #4501, opened 2023-05-06
feat: upgrade to Angular 16

jnizet:
- upgrade dependencies versions
- removed a re-generate yarn.lock
- fix formatting due to prettier upgrade
- fix deprecations in schematics
- remove usage of getFileContent (which doesn't exist anymore) in schematics spec
- revert @types/node dependency to 16.x (because the 17 version, which I guess is unstable caused a typing error)
- fix schematics test to apply the version 16 of the project generation and check in the right file that localize is being added to the project

The build currently fails when running the playwright tests, with the following error message:

```
[WebServer] This version of CLI is only compatible with Angular versions ^15.0.0,
but Angular version 16.0.0 was found instead.
```

I guess I get this error because of the custom webpack builder, which hasn't been upgraded to Angular 16 yet.

Remaining work:

- make the complete build pass
- change documentation 
- mark some inputs as required (but I don't thing there are many)


md13381:
Custom webpack just added the beta version for v16 so might now work with updated dependency

gnegrini:
Hey guys, I appreciate your work on this! At the moment, this is the only package holding my team to migrate to Angular 16, hope to have it soon!

jnizet:
@gnegrini I'm using ng-bootstrap with Angular 16 since Angular 16 came out. All you need to do is to tell npm to not care about the version mismatch. Or to use yarn, which just prints a warning in case of version mismatch.
Angular is backward-compatible, and even though the current version has not been formally tested with Angular 16, it works just fine.

That said, I sure hope that this PR is merged soon and that a new release comes out.

gnegrini:
> @gnegrini I'm using ng-bootstrap with Angular 16 since Angular 16 came out. All you need to do is to tell npm to not care about the version mismatch. Or to use yarn, which just prints a warning in case of version mismatch. Angular is backward-compatible, and even though the current version has not been formally tested with Angular 16, it works just fine.
> 
> That said, I sure hope that this PR is merged soon and that a new release comes out.

Awesome! We have started migrating one project using --legacy-peer-deps on npm install, and so far so good. But we are afraid components from other projects might break.  Also, the #4498 is tagged with "breaking change", so there I wonder... Did you experience any component break in your projects?

jnizet:
> Did you experience any component break in your projects?

No. As I said, it works just fine.

Using Angular 16 in ng-bootstrap is a breaking change because you won't be able to use it with a version of Angular older than 16, but the reverse isn't true: Libraries built for Angular 15 should normally be compatible out of the bow with  Angular 16.

md13381:
Can confirm as well i have had no issues with using the current version in angular 16

maxokorokov (approved):
LGTM, thanks, @jnizet.
Will do RC release ASAP.

PR #4521, opened 2023-06-01
fix(dropdown): set display when inside a navbar and guarded by ngIf

### LINKED ISSUE #4520: Dropdown in navbar doesn't have static display if guarded by ngIf
### Bug description:

A dropdown in a navbar is supposed to automatically have its display set to `static`. It's not the case if the dropdown or an ancestor is guarded by an `*ngIf`

https://stackblitz.com/edit/angular-ljqpmb?file=src%2Fapp%2Fdropdown-navbar.html,src%2Fapp%2Fdropdown-navbar.ts

Make sure to test the two dropdowns with a narrow street, where the navbar is initially collapsed, to see the difference.

Angular: 16.0.0

ng-bootstrap: 15.0.0

Bootstrap: 5.2.3

jnizet:
fix #4520

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -518,6 +518,35 @@ describe('ngb-dropdown-toggle', () => {
 			);
 		});
 
+		it(`shouldn't position the menu even if ngIf is applied`, () => {
+			const html = `
+      <nav class="navbar">
+        <div class="collapse navbar-collapse" id="navbarNav">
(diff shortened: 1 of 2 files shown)


maxokorokov (approved):
Oh, this is a good one...

`@Optional() NgbNavbar` directive worked in constructor with injection. But now we have to move it to `ngOnInit()` as `.closest()` doesn't work, because `ngIf` attaches the view when handling the `@Input() ngIf`, right?

LGTM!

PR #4522, opened 2023-06-02
demo: use `provideHttpClient` in stackblitz demos

terencehonles:
This change provides an HttpClient to all the stackblitz demos to ensure they function like they do on the demo site. Currently the HttpClient is the only provider needed, but the stackblitz template `misc/stackblitzes-templates/main.ts.ejs` should be kept in sync with `demo/src/main.ts` to ensure the individual examples have all their dependencies provided.

The needed providers can also be provided to the demo modules themselves or as context to the template renderer, but that may be more effort than its worth and while providing the dependencies to all the examples is overkill it will ensure they all work correctly.

### DIFF
--- a/misc/stackblitzes-templates/main.ts.ejs
+++ b/misc/stackblitzes-templates/main.ts.ejs
@@ -1,8 +1,13 @@
 import './polyfills';
+import {provideHttpClient, withNoXsrfProtection} from '@angular/common/http';
 import {bootstrapApplication} from '@angular/platform-browser';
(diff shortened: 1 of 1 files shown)


terencehonles:
this fixes the typeahead demo https://ng-bootstrap.github.io/stackblitzes/typeahead/http/stackblitz.html and it doesn't look like any others are currently using an `HttpClient`

maxokorokov (commented):
Thanks!

PR #4532, opened 2023-06-26
fix(dropdown): add `show` class to dropdown toggle when dropdown is open

### LINKED ISSUE #4531: Dropdown toggle doesn't have "show" class when dropdown is open
### Bug description:

When dropdown is open, the toggle button should have `.show` class indicating that the button is active. Picture below shows DOM of [Bootstrap's dropdown example](https://getbootstrap.com/docs/5.0/components/dropdowns/#overview) when the dropdown is open.

b5-dropdown-open

The `.show` class is applied to the toggle button and to the menu element, but not to the wrapper `.dropdown` element. In Bootstrap 4 the `.show` class was applied to wrapper and menu element, but not to the toggle button.

Picture below shows DOM of [NG Bootstrap's dropdown example](https://ng-bootstrap.github.io/#/components/dropdown/examples#basic) where `.show` class applied as it where in Bootstrap 4.

ngb-dropdown-open

This small change between Bootstrap v4 and v5 has gone unnoticed until now.

https://ng-bootstrap.github.io/#/components/dropdown/examples#basic

Angular: 16

ng-bootstrap: 15.0.1

Bootstrap: 5.0.0 and up

jsarem:
The `show` class on dropdown toggle will keep the button element in active state, inditicating that dropdown is open. This small change was made between Bootstrap v4 and v5.

Fixes #4531

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -13,35 +13,47 @@ import { Options } from '@popperjs/core';
 const createTestComponent = (html: string) =>
 	createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
 
-function getDropdownEl(tc) {
-	return tc.querySelector(`[ngbDropdown]`);
+function getDropdownEl(tc: HTMLElement) {
+	return tc.querySelector(`[ngbDropdown]`)!;
 }
 
-function getMenuEl(tc) {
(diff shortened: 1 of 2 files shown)


jsarem:
I hesitate to remove `.show` class from `NgbDropdown` directive because it would be breaking change for all who have custom styles relying on it. It doesn't serve any other purpose on Bootstrap 5. Let me know what you think.

Issue #870, opened 2016-10-11
feat(popover,tooltip): Support something akin to 'selector' in BS4 popover


spongessuck:
ref: http://v4-alpha.getbootstrap.com/components/popovers/#options

Use case: allow popover to be delegated to 3rd-party components. It could support selectors or (more importantly) element refs.

For instance, I'm using a component from a module and I want to have a popover point to it, but its not positioned exactly where its parent element is, so putting the popover on the parent doesn't look right. The only way I can get it to work properly is to add `ngbPopover` directly to the component's template.

pkozlowski-opensource:
@spongessuck I'm trying to determine how useful this would be in a general case so could you please share a plunker with your real-life scenario? Thnx.

spongessuck:
It's something like this:

https://plnkr.co/edit/8q9Fgp3eZw1oDr1cIdz3?p=preview

Consider the DemoComponent template to be something I don't want to modify. I want to keep it generic so that it doesn't depend on Ngb. But I want to expose the inner element so that a popover can be targeted on that instead of the host element.

MickL:
Real-life example faced:

We are using a third-party non-angular library called vis.js to display a timeline. Vis.js is giving me events like click, hover which i can bind/subscribe to in Angular. Now when i want to display a tooltip i can't because ng-bootstrap does not have the option to add a popover or tooltip to a DOM selector like original Bootstrap has: `$('#example').tooltip(options)`

keljoundi:
Similar real-life example:

I am using Fullcalendar in an application and want to be able to bind popovers to the calendar events to show details at a glance.

I am able to work around the problem and get popovers to work in a dirty way. I defined a single popover element in the HTML and in the TS component I reference the popover via ViewChild and set title and content. To correct the location I set the **private** _elementRef variable and it seems to work, and just ignore the ts errors with @ts-ignore.

This is a quick and dirty workaround and something as simple as exposing the _elementRef or better yet adding a function to change the _elementRef could do the trick. My first thought was to extend the NgbPopover class or add a prototype function but wasn't successful and have read on other issues that extending features of NG-Bootstrap isn't really feasible.

cribbstechnologies:
@keljoundi What version of fullcalendar were you using? I'm running into the exact problem you ran into and I can't get your workaround to work. Angular complains that this.popover.ngbPopover is undefined. How did you define your popover in the HTML?

keljoundi:
[...]

PR #4551, opened 2023-08-09
feat(tooltip/popover/dropdown): add custom container support

### LINKED ISSUE #1884: Use any selector for popover container
Now we append popover in the element or body. It will be good to give possibility append it in any element
```html
<button type="button" ngbPopover="text..." container=".my-block">
    Popover appended to element with class .my-block
</button>
```
Furthermore it removes hardcode

```js
if (this.container === 'body') { //can change to `if (this.container) {...}`
   window.document.querySelector(this.container).appendChild(this._windowRef.location.nativeElement);
}
```

Angular: 4.4.4

ng-bootstrap: 1.0.0-beta.5

Bootstrap: 4.0.0-beta

steve3d:

Currently in ng-bootstrap, there is no way of setting the container of tooltip/popover/dropdown to other values, the only accepted values are null and `body`, which is not enough. For example, I have something like 
```
<body>
  <app-root>
      <div id="screen"></div>
  </app-root>
</body>
```

or somthing like above, then if I need to make `#screen` div fullscreen, then all tooltip/popover/dropdown with `container=body` will not work, without the `container=body` the actual container must not be `overflow: hidden`, this is huge restriction. And my special case has much more problem, I have a something in table, the outer component also need to be fullscreen, in the cells of this table, I have dropdown as actions menu, have tooltip on some elements, and also the popover on some elements, without this custom container support, my use case will only work without the component fullscreen.

So I've changed the `container` input, which accept an valid css selector or a `HTMLElement`, then tooltip/popover/dropdown will use this container element as the real container. this will solve the many problems.

and I don't think there will be any new test, because the `body` string value in `container` is a valid CSS selector. so the old tests works fine.

And I believe this will also fix #1884 and #870 .

### DIFF
--- a/demo/src/app/components/dropdown/demos/container/dropdown-container.html
+++ b/demo/src/app/components/dropdown/demos/container/dropdown-container.html
@@ -3,50 +3,87 @@
 	it belongs. This option is useful if the dropdown is defined inside an element that clips its contents (i.e.
 	<code>overflow: hidden</code>).
 </p>
-<table class="table table-striped">
-	<thead>
-		<tr>
-			<th scope="col">#</th>
-			<th scope="col">Items</th>
-			<th scope="col">Actions</th>
-			<th scope="col">Actions</th>
-		</tr>
-	</thead>
-	<tbody>
-		<tr>
-			<td>1</td>
-			<td>
-				Item
-				<div>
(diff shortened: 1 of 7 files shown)


SantoshN7:
This seems unnecessary, could be resolve via [custom target feature](https://ng-bootstrap.github.io/#/components/popover/examples#custom-target)

PR #4555, opened 2023-08-27
Fix NgbAlert dismissable mentions 'close' output instead of 'closed'

### LINKED ISSUE #4553: [Docs] NgbAlert dismissable mentions 'close' output instead of 'closed'
Documention for NgbAlert's dismissable input mentions 'close' output but the output is actually named 'closed'

https://ng-bootstrap.github.io/#/components/alert/api

> The close button (×) will be displayed and you can be notified of the event with the (close) output.

should be 

> The close button (×) will be displayed and you can be notified of the event with the **(closed)** output.

bastienmoulia:
Fix #4553


maxokorokov (approved):
Thanks!

PR #4557, opened 2023-08-31
Update modal-backdrop.ts

bailzbrooke16:
This is a universal solve for the issue of modal layering and backdrop indexing in the right way


karsa-mistmere:
@maxokorokov: Can we get this merged? [1050 is definitely the right value here](https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L1138).

Although I'm not even sure why the z-index value needs to be inserted as an inline style to begin with, when the `modal-backdrop` class is also set which in turn should set the [correct backdrop z-index](https://github.com/twbs/bootstrap/blob/main/scss/_modal.scss#L115) anyways.

PR #4588, opened 2023-10-25
feat: accessibility fixes + e2e axe

ValentinNelu:

### DIFF
--- a/demo/src/app/components/components.e2e-spec.ts
+++ b/demo/src/app/components/components.e2e-spec.ts
@@ -1,4 +1,5 @@
-import { ConsoleMessage, test, expect } from '@playwright/test';
+import { ConsoleMessage, test, expect, type Page } from '@playwright/test';
+import AxeBuilder from '@axe-core/playwright';
 
 const SELECTOR_TAB_LINKS = 'header.title a.nav-link';
 const SELECTOR_SIDE_NAV_COMPONENT_LINKS = 'ngbd-side-nav >> a[href^="#/components/"]';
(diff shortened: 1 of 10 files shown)


src/carousel/carousel.ts
```diff
@@ -111,13 +111,25 @@ export class NgbSlide {
...
+		<button
+			class="carousel-control-next"
+			type="button"
+			(click)="arrowRight()"
+			*ngIf="showNavigationArrows"
+			aria-labelledby="button-span-next"
+		>
 			<span class="carousel-control-next-icon" aria-hidden="true"></span>
-			<span class="visually-hidden" i18n="@@ngb.carousel.next">Next</span>
+			<span class="visually-hidden" i18n="@@ngb.carousel.next" id="button-span-next">Next</span>
```
maxokorokov, line 132: This will not be unique in case there are several carousels on the page

maxokorokov (changes requested):
Just one comment about ids

src/carousel/carousel.ts
ValentinNelu: indeed, I am gonna use the same logic as in accordion for the generated id

ValentinNelu: I made the change to add unique ids for the spans. Should we use that ids (that are generated for each carousel) also as id on the carousel itself as we do for the accordion with the input?

demo/src/app/components/components.e2e-spec.ts
```diff
@@ -26,6 +27,30 @@ const COMPONENTS = [
...
+	nav: ['aria-allowed-attr', 'aria-required-children'],
+	offcanvas: [],
+	pagination: [],
+	popover: [],
+	progressbar: ['color-contrast'],
+	rating: ['aria-input-field-name'],
+	scrollspy: ['color-contrast'],
+	table: ['select-name'],
+	timepicker: ['color-contrast'], //because of the alert
+	toast: [],
```
maxokorokov, line 47: I think you should add contrast to toasts as well

maxokorokov (approved):
Thanks a lot!

PR #4605, opened 2023-11-09
fix(typeahead): use inline style for live element and localize message

### LINKED ISSUE #4560: Typeahead: Invisible "ngb-live" element is not translatable and may be visible
### Description:

When using the typeahead component, a new element appears in the DOM indicating the number of results:

```html
<div id="ngb-live" aria-live="polite" aria-atomic="true" class="visually-hidden">1 result available</div>
```

I have two problems with this:
1. The printed message is in English, I need a way to translate it.
2. My application is encapsulated in a shadow DOM but this new element is injected into the outside body. I have no control over the CSS outside the shadow DOM and therefore the element could be visible to the user. I need to be able to specify a container (within the shadow DOM) for this element.

[https://stackblitz.com/edit/angular-te4ccq?file=src%2Findex.html](https://stackblitz.com/edit/angular-te4ccq?file=src%2Findex.html)

### Versions 

Angular: v16.1
ng-bootstrap: v15.1
Bootstrap: v5.2.3

jeremychoisy:
Use inline style instead of the bootstrap class for the live element to support the usage of the component in the shadow DOM and localize the message itself. Fixes #4560


src/util/accessibility/live.ts
```diff
@@ -16,7 +16,7 @@ function getLiveElement(document: any, lazyCreate = false): HTMLElement | null {
 		element.setAttribute('aria-live', 'polite');
 		element.setAttribute('aria-atomic', 'true');
 
-		element.classList.add('visually-hidden');
+		element.style.display = 'none';
```
maxokorokov, line 19: Hmm, I think there is a reason `.visually-hidden` and the likes of it exist (ex. `.cdk-visually-hidden` in material). 

I don't think you should use `display: none` and `visiblity: hidden` if you want the the content to be read by the screen reader.

You should either inline all the styles necessary, or we should allow providing a `Live` service at a different level.

WDYT?

src/typeahead/typeahead.ts
```diff
@@ -397,6 +397,18 @@ export class NgbTypeahead implements ControlValueAccessor, OnInit, OnChanges, On
 		this._nativeElement.value = toString(value);
 	}
 
+	private getAnnounceLocalizedMessage(count: number): string {
```
maxokorokov, line 400: Maybe it would be better to introduce an `@Input()` for these kind of translations as well?


Like `@Input() liveAnnouncerMessage: (count: number) => string;`

With default being the function you're suggesting I guess.

maxokorokov: Also in this project private methods start with `_` → something like `_getLiveAnnouncerLocalizedMessage`

maxokorokov (commented):
Hey @jeremychoisy,

thanks for the PR.

I think there should be 2 PRs (or at least commits) for this:

- feat → allow customizing live announcer message (this PR)
- fix → to fix the shadow DOM issue (separate PR)

I've left a couple of comments in the code as well + the build is failing.

src/typeahead/typeahead.ts
jeremychoisy: Regarding the `@Input`, I'm not really sure tbh, I'm not super familiar with the way `@angular/localize` works but from what I understood, people using this library will generate/maintain translation files based on what is flagged with `i18n`/`$localize`. Do we really want/need to provide an additional way to override the translations on top of this already existing mechanism ? Provided keys for this use case seem to cover every possibilities, WDYT ?

src/util/accessibility/live.ts
jeremychoisy: You are 100% right, updated with the inline style taken from Bootstrap [visually-hidden](https://github.com/twbs/bootstrap/blob/8a965cadf80ed05d16dfe68a35f244cc5698ea8f/scss/mixins/_visually-hidden.scss#L8), noticed that the [cdk-visually-hidden class](https://github.com/angular/components/blob/59c8bb9a14c61c3932d6701994366f05ef719a91/src/cdk/a11y/_index.scss#L4C21-L4C21) might be a little more robust, let me know if you'd like to use this one instead.

Regarding inline style VS providing the `Live` service at a different level, was not really into the service providing based solution because imo it seemed a little bit too complex compared to the inline style one. But seeing the result, I'll give a try to see how it goes because maintaining this inline styling and not benefiting from what could be done on Bootstrap side is kind of sad.

Issue #1811, opened 2017-09-02
DatePicker Modes


say25:
### Bug description:
DatePicker doesn't have mode. Ability to use datepicker for selecting a date without a day (09/17 or 10/93) was awesome in ui-angular/bootstrap. Is this coming later? Purposefully abandoned?

![img_0066](https://user-images.githubusercontent.com/7467062/29993601-c29aba7c-8f89-11e7-9724-c192cfd09854.PNG)

### Version of Angular, ng-bootstrap, and Bootstrap:

Angular: 4.3.6

ng-bootstrap: 1.0.0-beta3

Bootstrap: 4.0.0-beta

maxokorokov:
I would say it was purposefully abandoned. If anybody wants to work on this - please feel free, though let's discuss design first before opening PRs

denniseffing:
Is there any update on this? This feature would be much appreciated and your competitor is actively working on it (PR merge will be pretty soon I guess): https://github.com/valor-software/ngx-bootstrap/pull/3354

crisdenverlabe:
Any update on this?

Angular 2 - How to show only month and year in ngxbootstrap/datepicker
https://stackoverflow.com/q/49165643/8134483?sem=2

EliseoPlunker:
it's only work with ngb-dropdown. I make a stackblitz https://stackblitz.com/edit/angular-zs1rxp
Feel free to use, improve or change

grosch:
Why would you purposefully abandon this?  Picking a month is a VERY common occurrence.

mboughaba:
Hi any news on this one? I also think it is a common use case

Cheers

val-samonte:
Please consider this feature, as this will benefit some of the users' experience. The behavior to select year and month in 2-3 clicks is great compared on creating 2 separate dropdowns for each

Mathaeusxs:
+1 
It would be rly good to have this feature.

rniculae:
As @mboughaba said, is a common use case

jmijmi:
+1

davidseek:
+1

ppppersy:
+1 really need this

jcarlos20:
> ngb-dropdown

This really helps me. Thank you so much. Lets go for a beer!

lapongua:
> ngb-dropdown

$event.stopPropagation(); not working form as in the stackblitz

lapongua:
[...]

PR #4613, opened 2023-11-12
Monthpicker

bastienmoulia:
WIP for #https://github.com/ng-bootstrap/ng-bootstrap/issues/1811


maxokorokov:
Hey, @bastienmoulia, this looks... big, haha.
I really do appreciate the effort to fix this though.

I don't think we should duplicate the datepikcer's code to make a monthpicker. It's already old and with quite a bit of technical debt in it. And we have new primitives to make it simpler as well (ex. signals). I don't think we should introduce new things like `NgbMonth`, `I18n`, `Calendar` (especially Calendar!).

I was playing with this on my side as well, I don't think it will make it for the next major, but I should share it soon as well.

On my side I was focusing on:

- making the datepicker "backend" signal-based
- making the layout responsive
- introduce `@Input() zoom: 'days' | 'months | 'years'` to change the view
- introduce `@Input() mode: 'days' | 'months'` to decide what should be selected

I'd also like to see how far what we have is from `Temporal` proposal, but this is probably for another day.

bastienmoulia:
It's a bit big I now. I made a monthpicker for a personal project so I would like to share how it can be done. But I'm agree with you, it's not very optimized.

vision10:
So are we going to see a rewrite of the datepicker @maxokorokov ?
I would very much like to see a mode changer to be able to select months 
And if possible, to reorganize the code in subfolders to be more manageable (picker, calendarType, adapters, ...)

Issue #3873, opened 2020-10-19
Some properties in the NgbInputDatepickerConfig are not working


lon360:
### Bug description:

`NgbInputDatepickerConfig` has a property of `restoreFocus`.

However, because the value is not set in the `NgbInputDatepicker` constructor, it is not running as a config.

https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-input.ts#L291

Similarly, the `minDate` / `maxDate` is not set, so the validator will not work.

Being able to set it to global reduces the amount of code, so this would be a big help.

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-ivy-cxryag

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: `10.1.6`

ng-bootstrap: `7.0.0`

Bootstrap: `^4.5.3`

PR #4620, opened 2023-11-14
docs(datepicker): fix config example

### LINKED ISSUE #3848: Global configuration of datepickers minDate and MaxDate example is misleading
### Bug description:
Hi, Maybe I'm wrong, but it seems that this example is misleading about minDate and maxDate

https://ng-bootstrap.github.io/#/components/datepicker/examples#config

When trying to customize minDate and MaxDate this config is not working even in StackBlitz

// customize default values of datepickers used by this component tree
    config.minDate = {year: 1900, month: 1, day: 1};
    config.maxDate = {year: 2099, month: 12, day: 31};

I end up using NgbDatepickerConfig instead of NgbInputDatepickerConfig

https://stackblitz.com/run?file=app/datepicker-config.ts

Angular: 10.0.0

ng-bootstrap:  v7.0.0

Bootstrap: 4.5.0

bastienmoulia:
Fix #3848 and #3873 but maybe the documentation is OK and it's the `NgbInputDatepickerConfig` which do not works well.

### DIFF
--- a/demo/src/app/components/datepicker/demos/config/datepicker-config.ts
+++ b/demo/src/app/components/datepicker/demos/config/datepicker-config.ts
@@ -2,6 +2,7 @@ import { Component } from '@angular/core';
 import {
 	NgbCalendar,
 	NgbDate,
+	NgbDatepickerConfig,
 	NgbDatepickerModule,
 	NgbDateStruct,
 	NgbInputDatepickerConfig,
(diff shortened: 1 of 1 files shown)


maxokorokov (commented):
I don't think the issue is with the demo. I don't think it should work this way, there should be only `NgbInputDatepickerConfig`.


- input config extends normal config, so it has all the same stuff
- we provide it here → https://github.com/ng-bootstrap/ng-bootstrap/blob/0ee19666c4678632119f28f264fd2cbaa49719d9/src/datepicker/datepicker-input.ts#L65
- only inputdatepikcer-specific inputs are initalized for some reason → https://github.com/ng-bootstrap/ng-bootstrap/blob/0ee19666c4678632119f28f264fd2cbaa49719d9/src/datepicker/datepicker-input.ts#L104 these are not → https://github.com/ng-bootstrap/ng-bootstrap/blob/0ee19666c4678632119f28f264fd2cbaa49719d9/src/datepicker/datepicker-input.ts#L176
- when we open datepicker popup, we copy inputs here → https://github.com/ng-bootstrap/ng-bootstrap/blob/0ee19666c4678632119f28f264fd2cbaa49719d9/src/datepicker/datepicker-input.ts#L532-L554


It all doesn't seem right...

Maybe we just need to initialize ALL inputs with the config values?

PR #4621, opened 2023-11-14
feat(progressbar): add stacked input

### LINKED ISSUE #905: Feature (Progressbar): Stacked Progressbar

bastienmoulia:
fix #905

### DIFF
--- a/demo/src/app/components/progressbar/demos/stacked/progressbar-stacked.html
+++ b/demo/src/app/components/progressbar/demos/stacked/progressbar-stacked.html
@@ -1,0 +1,5 @@
+<div class="progress-stacked">
+	<ngb-progressbar type="danger" [value]="20" [stacked]="true"></ngb-progressbar>
+	<ngb-progressbar type="warning" [value]="50" [stacked]="true"></ngb-progressbar>
+	<ngb-progressbar type="success" [value]="30" [stacked]="true"></ngb-progressbar>
+</div>
--- a/demo/src/app/components/progressbar/demos/stacked/progressbar-stacked.html
+++ b/demo/src/app/components/progressbar/demos/stacked/progressbar-stacked.html
@@ -1,5 +1,5 @@
-<div class="progress-stacked">
-	<ngb-progressbar type="danger" [value]="20" [stacked]="true"></ngb-progressbar>
-	<ngb-progressbar type="warning" [value]="50" [stacked]="true"></ngb-progressbar>
-	<ngb-progressbar type="success" [value]="30" [stacked]="true"></ngb-progressbar>
-</div>
+<ngb-progressbar-stacked>
+	<ngb-progressbar type="danger" [value]="20"></ngb-progressbar>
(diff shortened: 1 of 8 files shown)


maxokorokov (commented):
Hey @bastienmoulia,

what do you think about this API?


```html
<ngb-progressbar-stack>
  <ngb-progressbar type="danger" [value]="20"></ngb-progressbar>
  <ngb-progressbar type="warning" [value]="50"></ngb-progressbar>
  <ngb-progressbar type="success" [value]="30"></ngb-progressbar>
</ngb-progressbar-stack>
```

It will avoid having `@Input()` completely that you have to repeat. Just optionally inject the stack in the progressbar. If it's there →  we're in the stack, if not → it's a standalone progressbar.

Do you think it would work ?

maxokorokov:
I missed this migration completely oO

Also it might be good to move attributes from the inner `div` → https://getbootstrap.com/docs/5.3/migration/#progress-bars

bastienmoulia:
> Also, if you're working on this, it might be good to move attributes from the inner `div` to match the new markup → https://getbootstrap.com/docs/5.3/migration/#progress-bars

It was already done in https://github.com/ng-bootstrap/ng-bootstrap/pull/4461

bastienmoulia:
> Hey @bastienmoulia,
> 
> what do you think about this API?
> 
> ```
> <ngb-progressbar-stack>
>   <ngb-progressbar type="danger" [value]="20"></ngb-progressbar>
>   <ngb-progressbar type="warning" [value]="50"></ngb-progressbar>
>   <ngb-progressbar type="success" [value]="30"></ngb-progressbar>
> </ngb-progressbar-stack>
> ```
> 
> It will avoid having `@Input()` completely that you have to repeat. Just optionally inject the stack in the progressbar. If it's there → we're in the stack, if not → it's a standalone progressbar.
> 
> Do you think it would work ?

It's working !

maxokorokov (changes requested):
Hey, @bastienmoulia, LGTM!

Just 2 comments:
- you should export it from the `src/index.ts` as well
- I was also thinking whether it should be a directive or not, but probably component is better as `NgbProgressbar` is a component.

My thinking is:
- if you query with `<ngb-progressbar-stacked #s>`, `s` will be either an element (if directive) or component instance (if component)
- it doesn't have any API really, so won't matter much
- `<ngb-progressbar-stacked>` being a directive is confusing
- if we make it a directive with an attribute(`[ngbProgressbarStacked]`, then writing `<div ngbProgressBarStacked><ngb-progressbar></ngb-progressbar></div>` seems strange to me.


I guess it should stay as component in the end, unless you have other thoughts...

bastienmoulia:
In the long term, I think the 2 components should be deprecated for directives. Like what was done for accordion.

maxokorokov (approved):
LGTM, thanks!

PR #4645, opened 2023-12-06
remove dark classes in docs

bastienmoulia:
Because some buttons or text are not visible in dark mode


maxokorokov:
Hmm, I think this one is also merged already... https://github.com/ng-bootstrap/ng-bootstrap/commit/7957a79a99b88211e52048d2d083889bdf66cb3e

PR #4670, opened 2024-02-01
feat(modal): add role option

### LINKED ISSUE #4101: missing configuration for attribute role at NgbModal
### feature request:
Using NgbModal for inserting a dialog offers no option to configure the value of the **role** attribute and is default "dialog".
It would be important [considering WCAG](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role) to be able to set the value to "alertdialog" in certain cases.

Angular: 11.2.9

ng-bootstrap: 9.1.0

Bootstrap: 4.6.0

danilopiazza:

 - [X] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) and [DEVELOPER.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/DEVELOPER.md) guide.

This PR should close #4101.

### DIFF
--- a/e2e-app/src/app/app.routes.ts
+++ b/e2e-app/src/app/app.routes.ts
@@ -11,6 +11,7 @@ import { DropdownPositionComponent } from './dropdown/position/dropdown-position
 import { ModalAutoCloseComponent } from './modal/autoclose/modal-autoclose.component';
 import { ModalFocusComponent } from './modal/focus/modal-focus.component';
 import { ModalNestingComponent } from './modal/nesting/modal-nesting.component';
+import { ModalRoleComponent } from './modal/role/modal-role.component';
 import { ModalStackComponent } from './modal/stack/modal-stack.component';
 import { ModalStackConfirmationComponent } from './modal/stack-confirmation/modal-stack-confirmation.component';
 import { PopoverAutocloseComponent } from './popover/autoclose/popover-autoclose.component';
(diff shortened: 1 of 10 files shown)


danilopiazza:
I have a few remaining questions:

- API documentation: which version should I document in the `@since` tag for the new property?
- Demo: what should I showcase in the demo? Should I print the `role` attribute from the modal window, just to see if it is woking? It looks like I would be duplicating the end-to-end tests.
- Should the new `role` option be updatable, like `ariaDescribedBy` and `ariaLabelledBy`?
- At this time, the only possible values for `role` are `alertdialog` and `dialog`. Should the property be strictly typed, allowing these two values only, or should it be extensible, allowing any string value (like `fullscreen` and `size`)?

quentinderoubaix:
> I have a few remaining questions:
> 
> * API documentation: which version should I document in the `@since` tag for the new property?
> * Demo: what should I showcase in the demo? Should I print the `role` attribute from the modal window, just to see if it is woking? It looks like I would be duplicating the end-to-end tests.
> * Should the new `role` option be updatable, like `ariaDescribedBy` and `ariaLabelledBy`?
> * At this time, the only possible values for `role` are `alertdialog` and `dialog`. Should the property be strictly typed, allowing these two values only, or should it be extensible, allowing any string value (like `fullscreen` and `size`)?

Thanks for the contribution @danilopiazza !

Have rebased your PR and for your remaining questions:
* @since tag will be 18.0.0
* no need to showcase the role in the demo
* I prefer the property to be strictly typed to help users avoid typos. if there is a real need in the future to extend the type, we can always change it later

PR #4696, opened 2024-05-13
fix(highlight): Change highlight to track by index

kryVDH:
Change track statement in `highlight.ts` to use `$index` instead of `part`. 

The current statement causes performance issues and even browser crashes when running a large number of strings with the same characters repeated in a typeahead.


maxokorokov:
Argh, sorry, I have seen #4707 first and merged it already :/
Thanks for the PR anyway!

PR #4698, opened 2024-05-14
fix(datepicker): config not applied

### LINKED ISSUE #3848: Global configuration of datepickers minDate and MaxDate example is misleading

### LINKED ISSUE #4697: Datepicker does not inherit config if startDate is set before ngOnInit is called.
### Bug description:
If `NgbDatepicker.navigateTo()` is called (either manually, or by binding the `[startDate]` input) before the datepicker's `ngOnInit()` is called, then the datepicker will not inherit values from `NgbDatepickerConfig`.

Likely related to issue #3848.

[https://stackblitz.com/edit/angular-pznaci](https://stackblitz.com/edit/angular-pznaci)

I put the versions we're using, but from what I can tell, this bug has likely existed for quite some time.

Angular: 14

ng-bootstrap: 13

Bootstrap: 5

zxanderh:
Note that the proposed fix for #4697 will not resolve the issue when `NgbDatepicker.navigateTo()` is called manually, only when it is called by setting the `[startDate]` input. It seems reasonable to expect users not to call functions on components before they're initialized, so I believe this fix to be sufficient. That said, I'm more than happy to hear any feedback.

Fixes #4697
Fixes #3848

### DIFF
--- a/src/datepicker/datepicker-input.spec.ts
+++ b/src/datepicker/datepicker-input.spec.ts
@@ -26,11 +26,18 @@ function expectSameValues(inputDatepicker: NgbInputDatepicker, config: NgbInputD
 	);
 }
 
+function getWeekdays(element: HTMLElement): string[] {
+	return (<HTMLElement[]>Array.from(element.querySelectorAll('.ngb-dp-weekday')) || []).map((el) =>
+		el.textContent!.trim(),
+	);
+}
+
 function customizeConfig(config: NgbInputDatepickerConfig) {
 	config.autoClose = 'outside';
 	config.container = 'body';
-	config.positionTarget = 'positionTarget';
+	config.positionTarget = '#positionTarget';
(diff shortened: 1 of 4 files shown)


maxokorokov (approved):
LGTM, thanks

PR #4707, opened 2024-05-28
fix: track by index in NgbHilight

### LINKED ISSUE #4705: NgbHighlight cause NG0955 warning
### Bug description:

The NgbHighlight directive, used in the typeahead component, causes the following warning to be logged (when using Angular 18):

> 'NG0955: The provided track expression resulted in duplicated keys for a given collection. Adjust the tracking expression such that it uniquely identifies all the items in the collection.

This is because it splits the text of a typeahead result in parts and iterates through the parts using a `@for (part of parts; track part)`. So if we search with `'n'` in a result such as `'no name'`, then two parts will be identical.

Angular: 18.0.0

ng-bootstrap: 16.0.0

Bootstrap: 5.3.3

jnizet:
fix #4705


maxokorokov (approved):
LGTM, thanks

PR #4712, opened 2024-05-31
fix(datepicker): track weekdays by $index

bastienmoulia:
Because weekdays can be the same now, tracking by the $index remove the warning.


maxokorokov (approved):
LGTM, thanks!
Indeed, I've seen lots of warnings in tests because of this.

Issue #4709, opened 2024-05-31
Zoneless change detection support


maxokorokov:
This is an aggregate issue to track zoneless support with `provideExperimentalZonelessChangeDetection()` in apps using ng-bootstrap.

Please open separate issues for specific use cases, we'll reference them here.
And keep in mind this is still considered "experimental".

maxokorokov:
Demo site uses zoneless now → #4728

sysmat:
is this lib zonless ready?

ct-dixit:
is this lib zonless ready?

immotus:
> is this lib zonless ready?

Yes, it is. Tested with Angular 19, zoneless, with partial hydration. Also, apparently their demo website uses zoneless, too.

sysmat:
when I use ng 19 & "@ng-bootstrap/ng-bootstrap": "^18.0.0 & ngb-toast I get error

```
RuntimeError: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'bg-success text-bg-success'. Expression location: _ToastsContainerComponent component It seems like the view has been created after its parent and its children have been dirty checked. Has it been created in a change detection hook? Find more at https://angular.dev/errors/NG0100
    at throwErrorIfNoChangesMode (core.mjs:11268:9)
    at bindingUpdated (core.mjs:19869:9)
    at checkStylingMap (core.mjs:25579:30)
    at ɵɵclassMap (core.mjs:25523:3)
    at ToastsContainerComponent_For_1_Template 
    at executeTemplate (core.mjs:12051:5)
    at refreshView (core.mjs:14322:7)
    at detectChangesInView (core.mjs:14531:5)
    at detectChangesInViewIfAttached (core.mjs:14493:3)
    at detectChangesInEmbeddedViews (core.mjs:14453:7)
```

sysmat:
it seams if I use modal and in modal call toast I get this error `RuntimeError: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'undefined'. Current value: 'bg-success text-bg-success'. Expression location: _ToastsContainerComponent component It seems like the view has been`

hakimio:
@sysmat 
[...]

PR #4714, opened 2024-05-31
Refactor/avoid ng zone

jnizet:
This replaces a few usages of NgZone.onStable, which is not zoneless-compatible, with calls to afterRender or afterNextRender.
I'm not absolutely sure they behave exactly the same way, but the tests pass, and that's the recommended way of avoiding this method.
 
### DIFF
--- a/src/dropdown/dropdown.ts
+++ b/src/dropdown/dropdown.ts
@@ -1,5 +1,9 @@
 import {
 	AfterContentInit,
+	afterNextRender,
+	afterRender,
+	AfterRenderPhase,
+	AfterRenderRef,
 	ChangeDetectorRef,
 	ContentChild,
 	ContentChildren,
@@ -8,6 +12,7 @@ import {
 	EventEmitter,
 	forwardRef,
 	inject,
+	Injector,
 	Input,
 	NgZone,
 	OnChanges,
@@ -18,7 +23,7 @@ import {
 	SimpleChanges,
 } from '@angular/core';
 import { DOCUMENT } from '@angular/common';
-import { fromEvent, Subject, Subscription } from 'rxjs';
+import { fromEvent, Subject } from 'rxjs';
 import { take } from 'rxjs/operators';
 
 import { ngbPositioning, Placement, PlacementArray } from '../util/positioning';
@@ -165,11 +170,12 @@ export class NgbDropdown implements OnInit, AfterContentInit, OnChanges, OnDestr
 	private _changeDetector = inject(ChangeDetectorRef);
 	private _config = inject(NgbDropdownConfig);
 	private _document = inject(DOCUMENT);
+	private _injector = inject(Injector);
 	private _ngZone = inject(NgZone);
 	private _nativeElement = inject(ElementRef).nativeElement as HTMLElement;
 
 	private _destroyCloseHandlers$ = new Subject<void>();
-	private _zoneSubscription: Subscription;
+	private _afterRenderRef: AfterRenderRef | undefined;
 	private _bodyContainer: HTMLElement | null = null;
 
 	private _positioning = ngbPositioning();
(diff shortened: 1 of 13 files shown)


jnizet:
I'm fine with that (hopefully the next one would not be too long after).
I was just trying to be ready for the future with this, but as you rightly said in the issue about zoneless, let's remember it's only experimental for now.

BTW, experimenting with zoneless and ng-bootstrap allowed finding an issue with the framework, that has is being fixed: https://github.com/angular/angular/issues/56290

src/dropdown/dropdown.ts
```diff
@@ -350,7 +364,7 @@ export class NgbDropdown implements OnInit, AfterContentInit, OnChanges, OnDestr
 			this._open = false;
 			this._resetContainer();
 			this._positioning.destroy();
-			this._zoneSubscription?.unsubscribe();
+			this._afterRenderRef?.destroy();
```
JeanMeche, line 367: This probably isn't necessary anymore as AfterRender should autodestroy with the DestroyRef. (If the injector is indeed a NodeInjector)

src/tooltip/tooltip.ts
```diff
@@ -249,8 +253,13 @@ export class NgbTooltip implements OnInit, OnDestroy, OnChanges {
 				Promise.resolve().then(() => {
 					// This update is required for correct arrow placement
 					this._positioning.update();
-					this._zoneSubscription = this._ngZone.onStable.subscribe(() => this._positioning.update());
 				});
+				this._afterRenderRef = afterRender(
+					() => {
+						this._positioning.update();
+					},
+					{ phase: AfterRenderPhase.MixedReadWrite, injector: this._injector },
```
JeanMeche, line 261: `MixedReadWrite` is the default value, any reason for specifying it here ? 
Also haven't looked much into it, but its usage isn't recommended as it perf sensitive.

src/dropdown/dropdown.ts
jnizet: @JeanMeche but it will be destroyed only when the directive is destroyed, and this is the close() method, which just closes the dropdown. The directive is still alive after this call, but the afterRender "effect" that was created when open() was called before (and that will be re-crezted the next time open() is called again) should be destroyed on close. Am I missing something?

jnizet: Thank you for the review BTW :-)

src/tooltip/tooltip.ts
jnizet: I felt like specifying it allowed to make it clear that choosing that value was a conscious choice and not an oversight. I'm not entirely sure what update() does, but I guessed it had to read and write to the DOM, hence this option. 
Given that it's just a single tooltip positioning and that ther's nothing else being read or written to the DOM at the same time, I don't think it matters much, TBH.

maxokorokov:
Part of #4709

src/offcanvas/offcanvas-backdrop.ts
```diff
@@ -32,6 +34,7 @@ import { OffcanvasDismissReasons } from './offcanvas-dismiss-reasons';
 export class NgbOffcanvasBackdrop implements OnInit {
 	private _nativeElement = inject(ElementRef).nativeElement as HTMLElement;
 	private _zone = inject(NgZone);
+	private _injector = inject(Injector);
```
maxokorokov, line 37: I was thinking, if we don't really need it anywhere, can we maybe just move it to the constructor?


I think these are equivalent, right?

```ts

private _injector = inject(Injector);

ngOnInit() {
  afterNextRender(() => {...}, {phase: ..., injector: this._injector})
}
```

```ts
constructor() {
  afterNextRender(() => {...}, {phase})
}
```

src/util/popup.ts
```diff
@@ -50,8 +52,19 @@ export class PopupService<T> {
 		}
 
 		const { nativeElement } = this._windowRef.location;
-		const transition$ = this._ngZone.onStable.pipe(
-			take(1),
+
+		const nextRenderSubject = new Subject<void>();
```
maxokorokov, line 56: yeah, this one is not pretty :)

src/offcanvas/offcanvas-backdrop.ts
jnizet: Thanks for the review. I'm on holidays, so if you want to move this forward, don't hesitate to do the suggested refactorings. I won't be able to do them. Not before at least a week.

maxokorokov: Nice, have a good one!
I'll probably do it after or it will fix itself when moving to signals anyway.

Issue #4729, opened 2024-06-17
firstDayOfWeek is not the left most day of the week


wratho:
### Bug description:
In the online exampled and in my own application the `firstDayOfWeek` is the 2nd column. 
![image](https://github.com/ng-bootstrap/ng-bootstrap/assets/28812961/b1f80020-8c7c-4054-b7bb-7c5cb780ec51)
This is the first example in the [datepicker overview](https://ng-bootstrap.github.io/#/components/datepicker/overview) and as you can see the firstDayOfWeek of 1 has S[unday] in the left most column
In my app i have a firstDayOfWeek = 7 and S[aturday] is the left most column
![image](https://github.com/ng-bootstrap/ng-bootstrap/assets/28812961/bef91e64-dd2b-4d57-8745-b64a9e73a0fe)

### Link to minimally-working StackBlitz that reproduces the issue:

https://stackblitz.com/edit/angular-xiclhg-k5arou?file=src%2Fmain.ts

### Versions of Angular, ng-bootstrap and Bootstrap:

Angular: 18.0.3

ng-bootstrap: 17.0.0

Bootstrap:

wratho:
I've also noticed that the `weekday` config setting doesn't appear to work? I set it to "long" and nothing changes. Every value is kept to the single letter values.

maxokorokov:
Could you provide more data please, because your StackBlitz works fine for me (see screenshot)... maybe OS, time zone, date/time, basically any relevant info you might think of to reproduce the issue.

![Screenshot 2024-06-18 at 08 50 12](https://github.com/ng-bootstrap/ng-bootstrap/assets/8074436/cd7507b9-4846-4a6f-a8d9-9e6cdc1706a9)

HaiFish:
The issue happens here too, timezone here is -03:00, if i change timezone to London, the issue is not present anymore

cetdelosaltos:
I have the same issue and I'm in -4:00GMT

samvloeberghs:
I can confirm that this seems to be an issue for people in "negative" timezones (GMT-0n:00) and working for "positive" timezones ( GMT+0n:00 )

Core121:
Looking to release an angular 18 update for my website, but this bug is preventing me from releasing knowing it'll be shipped with it. Any idea as to when we can see a fix available?

edisonaugusthy:
Any Updates on this..?

phillipwildhirt:
[...]

PR #4730, opened 2024-06-24
fix(datepicker): generate weekday labels in correct order

sorin-davidoi:
`timeZone` defaults to the user's timezone, which means that the week names might be returned in an unexpected order:

```
> TZ=Europe/Copenhagen node --eval "console.log([...Array(7).keys()].map((day) => Intl.DateTimeFormat('en', { weekday: 'narrow' }).format(new Date(Date.UTC(2021, 5, day - 1)))))"
[
  'S', 'M', 'T',
  'W', 'T', 'F',
  'S'
]

> TZ=America/New_York  node --eval "console.log([...Array(7).keys()].map((day) => Intl.DateTimeFormat('en', { weekday: 'narrow' }).format(new Date(Date.UTC(2021, 5, day - 1)))))"
[
  'S', 'S', 'M',
  'T', 'W', 'T',
  'F'
]
```

This would have been caught if the unit tests were run in different timezones (presumably they are run with `TZ=UTC` in CI?), as they are currently either passing or failing, depending on the timezone they are run in:

```
> TZ=UTC npm run test
...
TOTAL: 1579 SUCCESS
...

> TZ=America/New_York npm run test
...
TOTAL: 11 FAILED, 1568 SUCCESS
...
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#timezone

Closes https://github.com/ng-bootstrap/ng-bootstrap/issues/4729.

### DIFF
--- a/src/datepicker/datepicker-i18n.ts
+++ b/src/datepicker/datepicker-i18n.ts
@@ -116,7 +116,9 @@ export class NgbDatepickerI18nDefault extends NgbDatepickerI18n {
 		width: Exclude<Intl.DateTimeFormatOptions['weekday'], undefined> = 'narrow',
 	): string {
 		const weekdaysStartingOnSunday = [...Array(7).keys()].map((day) =>
-			Intl.DateTimeFormat(this._locale, { weekday: width }).format(new Date(Date.UTC(2021, 5, day - 1))),
+			Intl.DateTimeFormat(this._locale, { weekday: width, timeZone: 'UTC' }).format(
+				new Date(Date.UTC(2021, 5, day - 1)),
+			),
 		);
 		const weekdays = weekdaysStartingOnSunday.map((day, index) => weekdaysStartingOnSunday[(index + 1) % 7]);
 		return weekdays[weekday - 1] || '';
--- a/src/datepicker/datepicker-i18n.ts
+++ b/src/datepicker/datepicker-i18n.ts
@@ -105,10 +105,10 @@ export class NgbDatepickerI18nDefault extends NgbDatepickerI18n {
 	private _locale = inject(LOCALE_ID);
 
 	private _monthsShort = [...Array(12).keys()].map((month) =>
-		Intl.DateTimeFormat(this._locale, { month: 'short' }).format(new Date(2000, month)),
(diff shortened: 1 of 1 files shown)


sorin-davidoi:
> This would have been caught if the unit tests were run in different timezones (presumably they are run with TZ=UTC in CI?), as they are currently either passing or failing, depending on the timezone they are run in.

Depending on how much resources can be used in CI, there are a few things we could do:
 - Run the tests (or a subset of them) in every timezone (they can be fetched on most Linux systems via `timedatectl list-timezones`)
 - Pick a somewhat representative list of timezones (one hour increments from UTC-12 to UTC+14, plus a few edge cases like Asia/Kathmandu?) and run the tests (or a subset of them) in those

With the usage of a [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy) it should be possible to keep the running times reasonable. If there's agreement on what to do I could take a stab at changing the CI to increase the test coverage.

bastienmoulia:
Hi Sorin, your are right about the test, we should mock some other timezone.

Also can you add your fix in the other place where `Intl.DateTimeFormat` is used?

samvloeberghs:
when is 17.1.0 milestone expected?
This is not a small bug, any chance we can expect this sooner in a patch release? Thank you for considering..

src/datepicker/datepicker-i18n.ts
```diff
@@ -105,18 +105,20 @@ export class NgbDatepickerI18nDefault extends NgbDatepickerI18n {
 	private _locale = inject(LOCALE_ID);
 
 	private _monthsShort = [...Array(12).keys()].map((month) =>
-		Intl.DateTimeFormat(this._locale, { month: 'short' }).format(new Date(2000, month)),
+		Intl.DateTimeFormat(this._locale, { month: 'short', timeZone: 'UTC' }).format(new Date(2000, month, 15)),
```
leifjones, line 108: What is the significance of this date? (Similar question for the 2021 date below.)

maxokorokov:
Thanks

samvloeberghs:
Thx for the release @maxokorokov.
Very appreciative of your time 🙏🏼

PR #4738, opened 2024-06-27
build: migrate to Yarn 4

maxokorokov:


bastienmoulia:
Hi, is it possible to merge this before working on Angular 19? Because there is some issue with Yarn 1.

PR #4778, opened 2024-11-21
fix: set the value of typeahead's control to null instead of undefined

### LINKED ISSUE #4777: non-editable typeahead sets form control to undefined instead of null
### Bug description:

When using a non-editable typeahead, if we start editing and then do not select any value, the typeahead sets the form control's value to `undefined` instead of `null`.

This is a problem because

- `undefined` is used in Angular form values when the control is disabled
- setting a form value with an undefined property leads to the error `NG01002: Must supply a value for form control with name: 'state'`. 

This means that having a form containing a non-editable typeahead, getting the FormGroup's value value, and setting it back to the FormGroup causes an error. 

https://stackblitz.com/edit/angular-zpctfu?file=src%2Fapp%2Ftypeahead-format.html,src%2Fapp%2Ftypeahead-format.ts

Start editing the typeahead and do not select any suggested state. Then press the "Reset to current value" button. Check the error in the console.

Angular: 19.0.0

ng-bootstrap: 17.0.1

Bootstrap: 5.3.2

jnizet:
fix #4777

BREAKING CHANGE:
When typing into a non-editable typeahead and not selecting any value, the associated form control value is set to null instead of undefined.

### DIFF
--- a/src/typeahead/typeahead.spec.ts
+++ b/src/typeahead/typeahead.spec.ts
@@ -633,7 +633,7 @@ describe('ngb-typeahead', () => {
 			fixture.detectChanges();
 			expect(getNativeInput(compiled)).toHaveCssClass('ng-invalid');
 			expect(getNativeInput(compiled)).not.toHaveCssClass('ng-valid');
-			expect(fixture.componentInstance.model).toBeUndefined();
+			expect(fixture.componentInstance.model).toBeNull();
 
 			const event = createKeyDownEvent('Enter');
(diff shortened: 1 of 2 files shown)


jnizet:
@quentinderoubaix could you please consider this PR before releasing the next version: it's a small change, but it's breaking, and I was hoping it to land in the previous major of ng-bootstrap already.

quentinderoubaix (approved):
lgtm

PR #4793, opened 2025-01-14
fix(pagination): put aria-current on link element instead of li for a11y

MarkoOleksiyenko:
Focusable pagination element should be marked with `aria-current` for the SR to read it properly (NVDA works with the attribute set on li but JAWS needs it to be on <a> element). 

reference of the `aria-current` and recommended usage: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#example


maxokorokov (approved):
LGTM, thanks!

For reference:

- https://a11ysupport.io/tests/tech__aria__aria-current#assertion-aria-aria-current_attribute-convey_value_page-html-a(href)_element
- https://a11ysupport.io/tests/html/aria/aria-current.html

PR #4799, opened 2025-01-31
fix(datepicker): add missing attributes aria-disabled and aria-selected

quentinderoubaix:


MarkoOleksiyenko (approved):
lgtm, thanks @quentinderoubaix !

PR #4806, opened 2025-03-01
feat: Add a color mode handler script into service

misotoldy:
- Add a new service to manage the color mode (light, dark, auto) of the application, according to demo and bootstrap.
- Exposed provideColorMode() and ColorModeModule to easily include it in application (standalone or module based application).
```
bootstrapApplication(AppComponent, {
	providers: [
		provideColorMode(),
		...
	]
});
```

```
@NgModule({
     imports: [
         ColorModeModule,
     ],
 })
```

### DIFF
--- a/src/color-mode/color-mode.module.ts
+++ b/src/color-mode/color-mode.module.ts
@@ -1,0 +1,46 @@
+import { ModuleWithProviders, NgModule } from '@angular/core';
+
+import { provideColorMode } from './color-mode.provider';
+import { ColorModeConfig } from './color-mode.service';
+
+/**
+ * The `ColorModeModule` is an Angular module that provides color mode configuration and services.
+ *
+ * This module can be imported into your Angular application to enable color mode functionality.
+ *
+ * @example
+ * // Importing the module in your application
+ * import { ColorModeModule } from './color-mode/color-mode.module';
+ *
+ * @NgModule({
+ *   imports: [
+ *     ColorModeModule,
+ *   ],
+ * })
+ *
+ * or with configurations
+ *
+ * @NgModule({
+ *   imports: [
+ *     ColorModeModule.forRoot({
+ *       	defaultColorMode: 'auto',
+ *				saveToLocalStorage: true,
+ *				localStorageKey: 'bs_preferred_color_mode',
+ *     }),
+ *   ],
+ * })
+ * export class AppModule {}
+ *
+ * @module
+ */
+@NgModule({
+	providers: provideColorMode(),
+})
+export class ColorModeModule {
+	static forRoot(options?: Partial<ColorModeConfig>): ModuleWithProviders<ColorModeModule> {
+		return {
+			ngModule: ColorModeModule,
+			providers: provideColorMode(options),
+		};
+	}
+}
--- a/src/color-mode/color-mode.service.ts
+++ b/src/color-mode/color-mode.service.ts
@@ -1,0 +1,95 @@
+import { DOCUMENT } from '@angular/common';
+import { computed, inject, Injectable, InjectionToken, signal } from '@angular/core';
+
+export type ColorMode = 'light' | 'dark' | 'auto';
+
+export type ColorModeConfig = {
+	defaultColorMode: ColorMode;
+	saveToLocalStorage: boolean;
+	localStorageKey: string;
+};
+
+export const ColorModeConfigDefaults: ColorModeConfig = {
+	defaultColorMode: 'auto',
+	saveToLocalStorage: true,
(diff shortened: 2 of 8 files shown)


src/color-mode/color-mode.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	private setColorModeLS(colorMode: ColorMode) {
+		localStorage.setItem(this.settings.localStorageKey, colorMode);
+	}
+
+	/**
+	 * Initialize the color mode. Service should be initialized during the APP Initialization
+	 */
+	initialize() {
+		const colorMode = this.getColorModeLS() ?? this.defaultColorMode;
```
jnizet, line 68: There is a preference to save or not in local storage, but it's ignored here

src/color-mode/color-mode.service.ts
```diff
@@ -0,0 +1,95 @@
...
+			this.setColorMode();
+		});
+	}
+
+	/**
+	 * Set the color mode to an 'html' element and set the local storage with the preference of the color mode
+	 * @param mode
+	 */
+	setColorMode(colorMode?: ColorMode): void {
+		if (colorMode) {
```
jnizet, line 82: Setting the color scheme on the document and saving it to local storage could all be done in an effect.

src/color-mode/color-mode.service.ts
```diff
@@ -0,0 +1,95 @@
...
+			this.systemPreffersColorScheme.set(e.matches ? 'dark' : 'light');
+			this.setColorMode();
+		});
+	}
+
+	/**
+	 * Set the color mode to an 'html' element and set the local storage with the preference of the color mode
+	 * @param mode
+	 */
+	setColorMode(colorMode?: ColorMode): void {
```
jnizet, line 81: Why is the argument optional?

src/color-mode/color-mode.module.ts
```diff
@@ -0,0 +1,46 @@
...
+ *   ],
+ * })
+ * export class AppModule {}
+ *
+ * @module
+ */
+@NgModule({
+	providers: provideColorMode(),
+})
+export class ColorModeModule {
```
jnizet, line 39: We shouldn't use modules anymore. The provide function can be used whether or not modules are used in the application.

src/color-mode/color-mode.service.ts
```diff
@@ -0,0 +1,95 @@
...
+		if (colorMode) {
+			this.colorMode.set(colorMode);
+			if (this.settings.saveToLocalStorage) {
+				this.setColorModeLS(colorMode);
+			}
+		}
+
+		const bsDomeTheme = this.currentColorMode() === 'auto' ? this.systemPreffersColorScheme() : this.currentColorMode();
```
jnizet, line 89: bsDome: typo?

src/color-mode/color-mode.module.ts
```diff
@@ -0,0 +1,46 @@
...
+ *     ColorModeModule,
+ *   ],
+ * })
+ *
+ * or with configurations
+ *
+ * @NgModule({
+ *   imports: [
+ *     ColorModeModule.forRoot({
+ *       	defaultColorMode: 'auto',
```
jnizet, line 26: nit: indentation uses a mix of spaces and tabs

src/color-mode/color-mode.module.ts
```diff
@@ -0,0 +1,46 @@
...
+ *				localStorageKey: 'bs_preferred_color_mode',
+ *     }),
+ *   ],
+ * })
+ * export class AppModule {}
+ *
+ * @module
+ */
+@NgModule({
+	providers: provideColorMode(),
```
jnizet, line 37: This shouldn't be there, but only in forRoot (but this module shouldn't exist anyway)

src/color-mode/color-mode.service.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+@Injectable({
+	providedIn: 'root',
+})
+export class NgbColorModeService {
+	private document = inject(DOCUMENT);
+	private settings = inject(NG_COLORMODE_CONFIG);
+
+	private systemPreffersColorScheme = signal<ColorMode>(
+		window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light',
```
jnizet, line 49: This will fail on the server (if SSR is used).

jnizet (commented):
I think this feature needs more thought and design. 

IMHO, to be useful, it should be more flexible (should auto always be supported for example) and be SSR-ready. 

But for SSR to work fine, the server would need to know what the preferred mode is, and it can't do that if the preference is stored in local storage.

PR #4831, opened 2025-05-31
refactor: bump peer deps and leftover angular 20 migration

quentinderoubaix:


quentinderoubaix:
@jnizet SauceLabs build is failing since we are still only running those tests on Safari 15 & 16 (instead of 17 & 18, following Angular browser support policies) and Safari 15 tests are failing without `downlevelIteration` set to `true`

I'm thinking to ignore this failure for the RC, but will make sure they are up and running before the release. wdyt ?

jnizet:
@quentinderoubaix I don't understand why tests are still being run on Safari 15 and 16, given the changes you made in this PR. Aren't these changes done precisely to make the tests run on Safari 17 and 18?

PR #4862, opened 2025-11-10
feat: Added granular entry points to improve tree-shaking and reduce bundle size.

### LINKED ISSUE #4686: Huge Bundle Size when import every module
### Bug description:
dist bundle includes every module when a single module in included.

```
var NGB_MODULES = [NgbAccordionModule, NgbAlertModule, NgbCarouselModule, NgbCollapseModule, NgbDatepickerModule, NgbDropdownModule, NgbModalModule, NgbNavModule, NgbOffcanvasModule, NgbPaginationModule, NgbPopoverModule, NgbProgressbarModule, NgbRatingModule, NgbScrollSpyModule, NgbTimepickerModule, NgbToastModule, NgbTooltipModule, NgbTypeaheadModule];
```

is present in the main-[hash].js when importing a single module in a standalone component.

Minimal reproduction
- create a new project
- [optional] use optimization: false in the production configuration (if you want to see the name of the bundles)
- run `ng add @ng-bootstrap/ng-bootstrap`
- remove everything from the app component (ts,html)

app.component.ts

```
@Component({
  selector: 'app-root',
  standalone: true,
  imports: [
    NgbToast,
    NgbToastHeader
  ],
  templateUrl: './app.component.html'
})
export class AppComponent {
}
```

app.component.html

```
<ngb-toast></ngb-toast>
```

- run command `ng build your-project`
- open the dist folder and check the main bundle

Angular: 17.2
ng-bootstrap: 16.0.0
Bootstrap: 5.3.2

JEANJACO:
## Description
Added granular entry points to improve tree-shaking and reduce bundle size.

### Changes:
- Added new entry points for every src folder (except test)
- Created a new `environment` folder with its entry point
- Created a new `ngbConfig` folder with its entry points
- Updated `tsconfig.json` with new path mappings

### Example usage:
```typescript
// Before
import { NgbModal } from '@ng-bootstrap/ng-bootstrap';
// After (using new entry points for better tree-shaking)
import { NgbModal } from '@ng-bootstrap/ng-bootstrap/modal';
```

### Related Issue
Closes #4686

cc @quentinderoubaix @Alex3917 @alexnoise79 @PieterjanDeClippel @jnizet

### Note for maintainers
I'm currently using my fork in production. Would it be possible to include this in an upcoming release so I can switch back to the official package?

## Checklist
 - [X] read and followed the [CONTRIBUTING.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/CONTRIBUTING.md) and [DEVELOPER.md](https://github.com/ng-bootstrap/ng-bootstrap/blob/master/DEVELOPER.md) guide.


quentinderoubaix:
hi @JEANJACO thanks for the PR !

will review in details end of this week when I am back from holidays.

I had a very first look (viewing a 342-files PR on mobile is kinda rough 😅) and would prefer if we kept the structure src/{component}/{files} instead of src/{component}/src/{files}

Todo later, after this is merged we might want to update all demo examples and use the component entry points.

PieterjanDeClippel:
@JEANJACO I don't think this will work yet. Since the source files for the sub-entrypoints are still within the root `src` folder, they'll still be packaged with the main entrypoint.

AFAIK You have to move each sub-entrypoint one level up, so that eg the accordion folder is next to the src folder.

I did try to have nothing within the main entrypoint, but that doesn't work either, so you can just expose a VERSION const there

JEANJACO:
> @JEANJACO I don't think this will work yet. Since the source files for the sub-entrypoints are still within the root `src` folder, they'll still be packaged with the main entrypoint.
> 
> AFAIK You have to move each sub-entrypoint one level up, so that eg the accordion folder is next to the src folder.
> 
> I did try to have nothing within the main entrypoint, but that doesn't work either, so you can just expose a VERSION const there

@PieterjanDeClippel I see your point about the package size. However, there's a requirement to maintain backward compatibility with the current import style (`@ng-bootstrap/ng-bootstrap`), which is why everything still needs to be accessible from the main entrypoint.

In practice though, users importing from sub-entrypoints (`@ng-bootstrap/ng-bootstrap/accordion`) will benefit from tree-shaking and won't have the unused modules in their final bundle. The only downside is a slightly larger node_modules folder (what i did on demo app changes -> no more import from @ng-bootstrap/ng-bootstrap)

The alternative approach (moving folders outside src/) would break backward compatibility for users still importing from the main entrypoint. Unless we're planning a major version bump with breaking changes?

PR #4882, opened 2025-12-12
docs: document missing default inputs

### LINKED ISSUE #4860: Review input default values docs
#4848 showed that we are missing some proper `@defaultValue` in some component inputs and thus info is missing from the doc.

Goal of this issue would be to fix for the Accordion Item at least, then take some time to review other component inputs.

AntonChesnokov:
- add @defaultValue tags to inputs so API docs show them
- teach misc/api-doc.ts + spec to pick up @defaultValue from setters or initializers

Fixes #4860 

While inputs like `@Input() dismissible = this._config.dismissible` technically have their default value defined in the config service, the `@defaultValue` JSDoc tag is essential for:
- User-facing documentation;
- Consistency;
- Clearer API docs

If needed, I can add the tag only for places where automatic value detection doesn't work.

### DIFF
--- a/src/accordion/accordion.directive.ts
+++ b/src/accordion/accordion.directive.ts
@@ -215,7 +215,7 @@ export class NgbAccordionItem implements AfterContentInit {
 	 *
 	 * This property can also be set up on the parent [`NgbAccordion` directive](#/components/accordion/api#NgbAccordionDirective).
 	 *
-	 * @defaultValue `true - initialized from the parent NgbAccordion directive`
+	 * @defaultValue `true` - initialized from the parent NgbAccordion directive
 	 */
 	@Input() set destroyOnHide(destroyOnHide: boolean) {
 		this._destroyOnHide = destroyOnHide;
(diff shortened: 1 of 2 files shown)


AntonChesnokov:
The API docs table still shows "-" when a default value isn't provided. I kept that behavior as is, but I can switch it to display "undefined" if you prefer.
An example is `@Input() positionTarget = this._config.positionTarget;` in `NgbInputDatepicker`

src/accordion/accordion.directive.ts
```diff
@@ -390,17 +396,23 @@ export class NgbAccordionDirective {
...
 	 * If `true`, accordion will be animated.
+	 *
+	 * @defaultValue `false`
 	 */
 	@Input() animation = this._config.animation;
 
 	/**
 	 * If `true`, only one item at the time can stay open.
+	 *
+	 * @defaultValue `false`
```
quentinderoubaix: Not needed.

AntonChesnokov:
@quentinderoubaix I see CI failed because of `misc/api-doc.ts` formatting.
I've just fixed that. Could you please restart CI workflow?

Issue #4879, opened 2025-12-12
Zonless unit and e2e tests


quentinderoubaix:
20.0.0 has been released but we could not pass all unit and e2e tests to zoneless in time.

This issue tracks the migration of all those tests to zoneless.  
Any issue found in a component (not the test) will need to be included in ng-bootstrap 20 as a patch.

## Unit tests

- [ ] accordion
- [ ] alert
- [ ] carousel
- [ ] collapse
- [ ] datepicker
- [ ] dropdown
- [ ] modal
- [ ] nav
- [ ] offcanvas
- [ ] pagination
- [ ] popover
- [ ] progressbar
- [ ] rating
- [ ] scrollspy
- [ ] table
- [ ] timepicker
- [ ] toast
- [ ] tooltip
- [ ] typeahead
- [ ] utils

## E2e tests

- [ ] datepicker
- [ ] dropdown
- [ ] modal
- [ ] nav
- [ ] offcanvas
- [ ] popover
- [ ] timepicker
- [ ] tooltip
- [ ] typeahead

PR #4884, opened 2025-12-14
test(dropdown): make the dropdown test zoneless

jnizet:

This allowed finding a bug: when `open()` is called programmatically on a dropdown, it changes the value of its `_open` input (yikes), but since this isn't a signal and since markForCheck wasn't called, the template wasn't updated in zoneless.

On top of #4883
ref #4879

### DIFF
--- a/src/dropdown/dropdown.spec.ts
+++ b/src/dropdown/dropdown.spec.ts
@@ -2,42 +2,51 @@ import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { createGenericTestComponent } from '../test/common';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-import { Component, provideZoneChangeDetection } from '@angular/core';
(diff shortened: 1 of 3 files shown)


quentinderoubaix:
@jnizet flakiness is usually a pain, and if this is indeed coming from the polyfill, the more unit tests we move to zoneless, the worse it will get...

so I am ok to have two different angular test configs until we migrate them all to zoneless

src/dropdown/dropdown.spec.ts
```diff
@@ -67,10 +73,10 @@ expect.extend({
 
 describe('ngb-dropdown', () => {
 	beforeEach(() => {
-		TestBed.configureTestingModule({ providers: [provideZoneChangeDetection()] });
+		TestBed.configureTestingModule({});
```
quentinderoubaix: The entire `beforeEach` can be removed.

quentinderoubaix (approved):
@jnizet lgtm, just a minor comment

PR #4885, opened 2025-12-15
test(nav): make the nav test zoneless

jnizet:

On top of #4884
ref #4879


quentinderoubaix (commented):
just one minor comment @jnizet
lgtm otherwise

quentinderoubaix:
@jnizet noticed the flakiness too, we might need to investigate deeply at one point

PR #4886, opened 2025-12-15
test: split zone and zoneless configs

quentinderoubaix:


package.json
```diff
@@ -26,7 +26,7 @@
     "e2e-app:lint": "ng lint e2e-app",
     "ngb:static": "ts-node --project misc/tsconfig.json misc/copy-static-files.ts",
     "ngb:lint": "ng lint ng-bootstrap",
-    "ngb:test": "ng test -c other-browsers && ng test ng-bootstrap --coverage",
+    "ngb:test": "ng test -c other-browsers && ng test -c zone-other-browsers && ng test -c zone && ng test ng-bootstrap --coverage",
```
jnizet: We should probably add a `--watch=false` option on each command here, otherwise, when running locally, only the first command is being run.

And it's useful to test locally. BTW, the developer doc recommends running `yarn ci` locally before submitting a PR.

jnizet (changes requested):

package.json
quentinderoubaix: indeed, I tend to run `CI=true ng test` or `CI=true yarn ci` locally, as it's the actual ci behavior (and fwks like vitest use that environment variable to log less and change some configs) but this is not documented anywhere

updated the PR to add `--watch=false` in that command, simpler for everyone

jnizet (approved):

PR #4888, opened 2025-12-16
test(progressbar): refactor unit test to validate zoneless

quentinderoubaix:

ref #4879

### DIFF
--- a/src/progressbar/progressbar.spec.ts
+++ b/src/progressbar/progressbar.spec.ts
@@ -1,49 +1,23 @@
 import { TestBed, ComponentFixture } from '@angular/core/testing';
 import { createGenericTestComponent } from '../test/common';
 
-import { Component, provideZoneChangeDetection } from '@angular/core';
+import { ChangeDetectionStrategy, Component, ComponentRef, signal } from '@angular/core';
 
 import { NgbProgressbar, NgbProgressbarStacked } from './progressbar';
 import { NgbProgressbarConfig } from './progressbar-config';
 import { beforeEach, describe, expect, it } from 'vitest';
-
-const createTestComponent = (html: string) =>
-	createGenericTestComponent(html, TestComponent) as ComponentFixture<TestComponent>;
-
-function getAriaLabel(nativeEl): string {
-	return getProgress(nativeEl).getAttribute('aria-label') || '';
-}
-
-function getBarWidth(nativeEl): string {
-	return getProgressbar(nativeEl).style.width;
-}
-
-function getBarHeight(nativeEl): string {
-	return getProgress(nativeEl).style.height;
-}
-
-function getBarValue(nativeEl): number {
-	return parseInt(getProgress(nativeEl).getAttribute('aria-valuenow')!, 10);
-}
-
-function getProgress(nativeEl: Element): HTMLElement {
-	return nativeEl.querySelector('.progress') as HTMLElement;
-}
-
-function getProgressbar(nativeEl: Element): HTMLElement {
-	return nativeEl.querySelector('.progress-bar') as HTMLElement;
-}
+import { page } from 'vitest/browser';
 
 describe('ngb-progressbar', () => {
-	beforeEach(() => {
-		TestBed.configureTestingModule({ providers: [provideZoneChangeDetection()] });
-	});
-
 	describe('business logic', () => {
+		let fixture: ComponentFixture<NgbProgressbar>;
+		let progressRef: ComponentRef<NgbProgressbar>;
 		let progressCmp: NgbProgressbar;
 
 		beforeEach(() => {
-			progressCmp = TestBed.createComponent(NgbProgressbar).componentInstance;
+			fixture = TestBed.createComponent(NgbProgressbar);
+			progressRef = fixture.componentRef;
+			progressCmp = fixture.componentRef.instance;
 		});
 
 		it('should initialize inputs with default values', () => {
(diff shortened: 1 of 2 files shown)


src/progressbar/progressbar.spec.ts
```diff
@@ -55,265 +29,309 @@ describe('ngb-progressbar', () => {
...
+
 			expect(progressCmp.getValue()).toBe(120);
 		});
 	});
 
 	describe('UI logic', () => {
-		it('accepts a value and respond to value changes', () => {
-			const html = '<ngb-progressbar [value]="value"></ngb-progressbar>';
-			const fixture = createTestComponent(html);
+		class PaginationTester {
```
jnizet: Should be named to `ProgressbarTester`

src/progressbar/progressbar.spec.ts
```diff
@@ -55,265 +29,309 @@ describe('ngb-progressbar', () => {
...
+			private readonly fixture: ComponentFixture<TestComponent>;
+			readonly componentInstance: TestComponent;
+			readonly progress = page.getByCss('.progress');
+			readonly progressBar = page.getByCss('.progress-bar');
 
-			expect(getBarWidth(fixture.nativeElement)).toBe('10%');
+			constructor(html: string) {
+				this.fixture = createGenericTestComponent(html, TestComponent);
```
jnizet: Should pass `false` to avoid initial detectChanges.

src/progressbar/progressbar.spec.ts
```diff
@@ -55,265 +29,309 @@ describe('ngb-progressbar', () => {
...
+			const tester = new PaginationTester('<ngb-progressbar [value]="value()" [max]="max()"/>');
 
-			expect(getBarWidth(fixture.nativeElement)).toBe('20%');
+			await tester.expectBarWidthToBe('20%');
 
-			fixture.componentInstance.max = 200;
-			fixture.detectChanges();
-			expect(getBarWidth(fixture.nativeElement)).toBe('5%');
+			tester.componentInstance.max.set(200);
+			await tester.whenStable();
```
jnizet, line 198: Shouldn't be necessary, should it?

src/progressbar/progressbar.spec.ts
```diff
@@ -55,265 +29,309 @@ describe('ngb-progressbar', () => {
...
+			const tester = new PaginationTester('<ngb-progressbar [value]="value()" [type]="type()"/>');
 
-			expect(getProgressbar(fixture.nativeElement)).toHaveCssClass('text-bg-warning');
+			await expect.element(tester.progressBar).toHaveClass('text-bg-warning');
 
-			fixture.componentInstance.type = 'info';
-			fixture.detectChanges();
-			expect(getProgressbar(fixture.nativeElement)).toHaveCssClass('text-bg-info');
+			tester.componentInstance.type.set('info');
+			await tester.whenStable();
```
jnizet, line 215: Same here: shouldn't be necessary.

src/progressbar/progressbar.spec.ts
```diff
@@ -55,265 +29,309 @@ describe('ngb-progressbar', () => {
...
 
-			fixture.componentInstance.type = 'dark';
-			fixture.detectChanges();
-			expect(getProgressbar(fixture.nativeElement)).toHaveCssClass('text-bg-dark');
+			await expect.element(tester.progressBar).toHaveClass('text-bg-info');
+
+			tester.componentInstance.type.set('dark');
+			await tester.whenStable();
```
jnizet, line 220: Same here. 

Have you added them to make the tests faster (to avoid the timeout between retried assertions)? That's fine with me: the whenStable don't hurt.

Given that there are plenty of other ones, I won't make this remark again, but let you choose whether we should leave them there or remove them.

jnizet (changes requested):
Some nits, but LGTM.

src/progressbar/progressbar.spec.ts
quentinderoubaix: I think initial detectChanges is fine, as it's also the behavior of Angular in a real app.
Makes units faster and is the recommended approach in Angular's unit test documentation:
https://angular.dev/guide/testing/components-basics

What we want is to make sure we **never** call `detectChanges` outside of this initial call, just after component creation.

src/progressbar/progressbar.spec.ts
quentinderoubaix: Yes, I let them here for speed. I do not have a very strong opinion on keeping them or removing them, up to the dev doing the rework.

As in the comment below, focusing on cleaning up to avoid wrong `detectChanges` calls and refactoring to use the **Tester** pattern with locators is still the main priority.

src/progressbar/progressbar.spec.ts
jnizet: That documentation is out of date, and pre-dates zoneless.

A unit test generated by the CLI doesn't ude detectChanges (see https://github.com/cexbrayat/angular-cli-diff/blob/21.0.3/ponyracer/src/app/app.spec.ts for example).

Not a big deal, but I find the rule "never call detectChanges" simpler to follow.

quentinderoubaix: Thanks for the info, I indeed thought that doc was up to date...

I also really like the idea that at the end of the entire refactor, we could search in our codebase `fixture.detectChanges` and not find anything...

So I am proposing instead this static function:
```typescript
static async create(html: string) {
  const tester = new ProgressbarTester(html);
  await tester.whenStable();
  return tester;
}
```
and in the constructor we would have `createGenericTestComponent(html, TestComponent, false)`

Sounds ok for you @jnizet ?

jnizet: Fine by me.

Issue #4909, opened 2026-06-05
Angular 22 and ng-bootstrap 21


quentinderoubaix:
This issue tracks work to support Angular 22 and release ng-bootstrap 21.

- [x] bump dependencies / run migrations
- [x] make sure ng-bootstrap custom form components also support form signal patterns
- [x] release RC
- [x] update demo site / docs
- [x] release 21.0.0

Current plan is RC end of next week and 21.0.0 the week after.

quentinderoubaix:
21.0.0-rc.0 is now available, you may check [its release notes](https://github.com/ng-bootstrap/ng-bootstrap/releases/tag/21.0.0-rc.0)

Folks that want to use Signal Forms with ng-bootstrap custom form controls (rating, datepickerinput...) can use Angular's [compatForm utility](https://angular.dev/api/forms/signals/compat/compatForm)

quentinderoubaix:
21.0.0 is out!

PR #4911, opened 2026-06-12
chore: bump to Angular v22

quentinderoubaix:
ref #4909

Ran `ng update` schematics, updated tsconfig files when necessary.

Next steps:

- [x] removing unnecessary `changeDetection: ChangeDetectionStategy:OnPush`
- [x] use the new `@Service`
- [ ] review places where Eager is set and validate no impact on putting to OnPush instead => another PR
- [ ] fix the comments `// TODO mig v22` left => another PR


tomupson:
@quentinderoubaix Regarding commit https://github.com/ng-bootstrap/ng-bootstrap/pull/4911/changes/9cf40043af33c596ed9ba595e6e607c3a4cd519c, I think it's worth pointing out that the new `@Service()` decorator doesn't support constructor-based injection, so it may be a breaking change for some users who have not yet moved their codebases over to the `inject()` function

See https://github.com/angular/angular/commit/8f3d0b9d97424e058eb7bce57d80833fb68dec4a

"2. \`@Service\` doesn't allow constructor-based injection, only the \`inject\` function."


I'd say it's the way forward, so I'm personally not suggesting reverting it, but I would consider adding a breaking change notice on the changelog

quentinderoubaix:
Hi @tomupson, thx for raising this.

Actually, it's the opposite!  
Users can still inject ng-bootstrap services declared using the new `@Service` api with constructor-based patterns.  
But implementations of Services using `@Service` **cannot** use `inject` in their implementation.

So for a new Angular app, this is fine:
```ts
import { Component, computed, inject, Service, signal } from '@angular/core';

@Service()
export class MySecondService {
  readonly user = signal('World');
}

@Service()
export class MyService {
  private readonly mySecondService = inject(MySecondService);
  readonly title = computed(() => `Hello ${this.mySecondService.user()}!`);
}

@Component({
  selector: 'app-root',
  template: ` {{ myService.title( ) }}`,
})
export class App {
  constructor(public readonly myService: MyService) {} // ==> no error, works as expected
}
```

while this is not:
```ts
import { Component, computed, inject, Service, signal } from '@angular/core';

@Service()
export class MySecondService {
  readonly user = signal('World');
}

@Service()
export class MyService {
  constructor(private readonly mySecondService: MySecondService) {} // ==> error at compilation
  readonly title = computed(() => `Hello ${this.mySecondService.user()}!`);
}

@Component({
  selector: 'app-root',
  template: ` {{ myService.title( ) }}`,
})
export class App {
  constructor(public readonly myService: MyService) {}
}
```

tomupson:
Apologies, my misunderstanding

